PR #7, opened 2018-08-06
Restart grant wait for synchronous branch changes

towoe:
In case the signals `instr_gnt_i` and `branch_i` are asserted at the
same time, the branch request invalidates the current grant. Wait for
the next grant in order to be in sync with the data output of the fifo.

### DIFF
--- a/zeroriscy_prefetch_buffer.sv
+++ b/zeroriscy_prefetch_buffer.sv
@@ -151,7 +151,7 @@ module zeroriscy_prefetch_buffer
           addr_valid   = 1'b1;
         end
 
-        if(instr_gnt_i)
+        if(instr_gnt_i & ~branch_i)
           NS = WAIT_RVALID;
         else
           NS = WAIT_GNT;


wallento:
ping, any news about this? This was a quite painful bug for us and others will most probably face it too..

daviderossi1982:
hello stefan, now I see what you were referring to at orconf, I'll synchronize with the team and check the pull request. It would be also good to open an issue in case of bugs like this.

Cheers,
Davide

davideschiavone:
Hi sorry, I missed this pull request somehow.

Can you describe me with an example (if possible) what problem you see? I would like to reproduce it before merging it to the master :)

Thanks a lot for your precious contribution,
Best
Davide

wallento:
Thanks, @davideschiavone, unfortunately we cannot share the original code as we were migrating from another core and it was already pretty large. We can try to re-create a test case that re-produces the error, as @towoe wrote it came down to the core using the wrong instruction on a very basic branch. @towoe, can you describe a bit what led to the situation, thanks.

towoe:
Hi, it is also not easy to reproduce this error, but I still have the waveform where one can see how the PC is set to an incorrect value.

waveform

And the objdump of the section:
objdump

After the PC is at 0x370 (lw), it should increase to 0x374, but the instruction from address 0x2f4 is fetched. If one decodes the data from address 0x370, 0xfff78793, as it would be a branch instruction, an offset of -124 would be the change of the PC, resulting in 0x2f4. So I would assume the branch instruction from 0x34c is still in the decoder, but the calculation of the PC is done with the data from 0x370.

This was the only time I observed the signals `instr_gnt_i` and `branch_i` to be active at the same time.
I hope this helps to understand what is happening.

felixmiller:
We observe this behavior with both the zeroriscy and the ri5cy core and therefore opened an issue here: https://github.com/pulp-platform/riscv/issues/50

Best regards
Felix

davideschiavone:
Thanks a lot, I am trying to reproduce it

wallento:
This can be closed in favor of changing core2axi as discussed in https://github.com/pulp-platform/riscv/issues/50

wallento:
It is also in a PR for better documentation now: https://github.com/pulp-platform/zero-riscy/pull/12/commits/52ffb372a30ed3ba63625475bb8888171433e37b

See here: https://zero-riscy.readthedocs.io/en/latest/instruction_fetch.html#protocol

PR #9, opened 2018-11-13
Convert documentation to restructured text

wallento:
This makes it a bit easier to contribute to the documentation, because it can be handled by source control. I know it doesn't look 100% like the original, but I wanted to minimize the extra customization low for now. We can tweak it if you like.

I hope this is useful to you.

There is a preview here: https://zero-riscy.readthedocs.io
PDF: https://readthedocs.org/projects/zero-riscy/downloads/pdf/latest/
I can transfer this readthedocs project to you once merged.

The revision string is extracted based on the tag versioning. I propose to tag the current HEAD to 0.2.


davideschiavone:
Thanks a lot :) definitively a good contribution :+1:

PR #11, opened 2018-11-21
Doc: Inline documentation of waveforms

wallento:
So far I used wavedrom for rendering, but that required external
tooling. After patching the extension, sphinx can now render inline
waveforms!


wallento:
part of #12 too

PR #22, opened 2019-05-03
Cleanup includes and defines

imphil:
- Move ibex_tracer_defines.sv and ibex_defines.sv out of the 'include'
  directory, since these files are not actually included.
- Remove ibex_config.sv, it's mostly unused code. The remaining defines,
  SYNTHESIS, ASIC_SYNTHESIS, TRACE_EXECUTION, and CHECK_MISALIGNED should
  be set through command-line flags to the simulation/synthesis tools.

Initial version by @NilsGraf.


ibex.core
```diff
@@ -28,7 +28,6 @@ filesets:
 #      - rtl/ibex_register_file.sv # ASIC
       - rtl/ibex_register_file_ff.sv # FPGA
       - rtl/ibex_core.sv
-      - rtl/include/ibex_config.sv: { is_include_file: true }
```
NilsGraf: Don't you still need the ibex_config.sv file here? But not as an include.

imphil: ibex_config.sv is completely gone now

NilsGraf:
Awesome!

PR #26, opened 2019-05-07
Some more ibex cleanup

imphil:
Switch from parameters to enums where possible, and remove unused ones.


imphil:
Thanks @vogelpi! LGTM.

PR #28, opened 2019-05-07
Code cleanup

vogelpi:
Some more cleanup

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -175,10 +175,10 @@ module ibex_id_stage #(
   // Immediate decoding and sign extension
   logic [31:0] imm_i_type;
   logic [31:0] imm_s_type;
-  logic [31:0] imm_sb_type;
+  logic [31:0] imm_b_type;
   logic [31:0] imm_u_type;
-  logic [31:0] imm_uj_type;
-  logic [31:0] imm_z_type;
+  logic [31:0] imm_j_type;
+  logic [31:0] zimm_rs1_type;
 
   logic [31:0] imm_a;       // contains the immediate for operand b
   logic [31:0] imm_b;       // contains the immediate for operand b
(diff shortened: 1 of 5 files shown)


asb:
The only one that gave me pause was the removal of `imm_iz` which in theory is needed for the CSR instructions, but I see the sign-extended bits are just discarded (which makes sense).

While we're at this - any ideas for a better name for `imm_z_type`? I suppose `imm_izrs1_type` is overly specific as the same logic could be used if instructions other than the I-types had zero-extended integers in the rs1 field.

vogelpi:
I removed `imm_iz` primarily because it was not used anywhere in the design.

Maybe we could just rename `imm_z_type` to `imm_rs1_type`?

asb:
Or `zimm_rs1_type` so it's clear that it's zero-extended unlike the other surviving immediate formats which are all sign-extended.

vogelpi:
Yes, that is even better. I'll change it to `zimm_rs1_type` then.

rtl/ibex_id_stage.sv
```diff
@@ -245,21 +239,15 @@ module ibex_id_stage #(
...
-  assign imm_uj_type = { {12 {instr[31]}}, instr[19:12], instr[20], instr[30:21], 1'b0 };
+  assign imm_i_type = { {20 {instr[31]}}, instr[31:20] };
+  assign imm_s_type = { {20 {instr[31]}}, instr[31:25], instr[11:7] };
+  assign imm_b_type = { {19 {instr[31]}}, instr[31], instr[7], instr[30:25], instr[11:8], 1'b0 };
+  assign imm_u_type = { instr[31:12], 12'b0 };
+  assign imm_j_type = { {12 {instr[31]}}, instr[19:12], instr[20], instr[30:21], 1'b0 };
 
   // immediate for CSR manipulatin (zero extended)
-  assign imm_z_type = { 27'b0, instr[`REG_S1] };
+  assign zimm_rs1_type = { 27'b0, instr[`REG_S1] };
```
imphil: We should get rid of those defines as well, I've opened https://github.com/lowRISC/ibex/issues/30 to track that. (Not blocking this merge.)

PR #32, opened 2019-05-10
Code cleanup

vogelpi:
This pull request aligns the coding style of ibex to our guidelines: all I/Os use _i/_o suffices, configurable parameters use UpperCamelCase, and the types of all parameters are specified.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -498,7 +498,7 @@ module ibex_controller (
   assign operand_a_fw_mux_sel_o = data_misaligned_i ? SEL_MISALIGNED : SEL_REGFILE;
 
   // update registers
-  always_ff @(posedge clk_i, negedge rst_ni) begin : UPDATE_REGS
+  always_ff @(posedge clk_i or negedge rst_ni) begin : UPDATE_REGS
     if (!rst_ni) begin
       ctrl_fsm_cs <= RESET;
       //jump_done_q <= 1'b0;
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -24,8 +24,8 @@
  * Main CPU controller of the processor
  */
 module ibex_controller (
-    input  logic                      clk,
-    input  logic                      rst_n,
+    input  logic                      clk_i,
+    input  logic                      rst_ni,
 
     input  logic                      fetch_enable_i,        // Start the decoding
     output logic                      ctrl_busy_o,           // Core is busy processing instructions
@@ -126,7 +126,7 @@ module ibex_controller (
   // synopsys translate_off
   // make sure we are called later so that we do not generate messages for
   // glitches
-  always_ff @(negedge clk) begin
+  always_ff @(negedge clk_i) begin
     // print warning in case of decoding errors
     if (is_decoding_o && illegal_insn_i) begin
       $display("%t: Illegal instruction (core %0d) at PC 0x%h: 0x%h", $time, ibex_core.core_id_i,
@@ -498,8 +498,8 @@ module ibex_controller (
   assign operand_a_fw_mux_sel_o = data_misaligned_i ? SEL_MISALIGNED : SEL_REGFILE;
 
   // update registers
-  always_ff @(posedge clk, negedge rst_n) begin : UPDATE_REGS
-    if (!rst_n) begin
+  always_ff @(posedge clk_i, negedge rst_ni) begin : UPDATE_REGS
+    if (!rst_ni) begin
       ctrl_fsm_cs <= RESET;
       //jump_done_q <= 1'b0;
       debug_mode_q   <= 1'b0;
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -498,16 +498,13 @@ module ibex_controller (
   assign operand_a_fw_mux_sel_o = data_misaligned_i ? SEL_MISALIGNED : SEL_REGFILE;
 
   // update registers
-  always_ff @(posedge clk_i or negedge rst_ni) begin : UPDATE_REGS
+  always_ff @(posedge clk_i or negedge rst_ni) begin : update_regs
     if (!rst_ni) begin
-      ctrl_fsm_cs <= RESET;
-      //jump_done_q <= 1'b0;
-      debug_mode_q   <= 1'b0;
+      ctrl_fsm_cs  <= RESET;
+      debug_mode_q <= 1'b0;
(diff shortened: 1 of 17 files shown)


rtl/ibex_core.sv
```diff
@@ -25,11 +25,11 @@
...
-    parameter N_EXT_PERF_COUNTERS  = 1,
-    parameter bit RV32E            = 0,
-    parameter bit RV32M            = 1,
-    parameter DM_HALT_ADDRESS      = 32'h1A110800,
-    parameter DM_EXCEPTION_ADDRESS = 32'h1A110808
+    parameter int unsigned NumExtPerfCntrs = 1,
+    parameter bit Rv32e                    = 0,
+    parameter bit Rv32m                    = 1,
+    parameter int unsigned DmHaltAddr      = 32'h1A110800,
+    parameter int unsigned DmExcAddr       = 32'h1A110808
```
imphil: I must say I'm not a huge fan of the new naming, but it's in the style guide, so let's do at least some damage control.

- `Rv32e`/`Rv32m` should stay `RV32E`/`RV32M`. They are well-known names in the RISC-V world, the new capitalization is pretty strange.
- I'd prefer less abbreviated names at least for `DmExcAddr` -> `DmExceptionAddr`. `NumExtPerfCntrs` is also much harder to read than `NumExtPerfCounters`

rtl/ibex_core.sv
```diff
@@ -463,64 +465,66 @@ module ibex_core #(
...
+      .branch_i                ( perf_branch         ),
+      .branch_taken_i          ( perf_tbranch        ),
+      .mem_load_i              ( perf_load           ),
+      .mem_store_i             ( perf_store          ),
+
+      .ext_counters_i          ( ext_perf_counters_i )
   );
 
+  assign perf_load  = data_req_o & data_gnt_i & (~data_we_o);
+  assign perf_store = data_req_o & data_gnt_i & data_we_o;
```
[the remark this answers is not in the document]
vogelpi: Alrighty, I did the same for the various `csr_*` signals assigned next to these.

rtl/ibex_fetch_fifo.sv
```diff
@@ -45,7 +45,7 @@ module ibex_fetch_fifo (
                                            // available immediately as it comes directly out of FFs
 );
 
-  localparam DEPTH = 3; // must be 3 or greater
+  localparam int unsigned DEPTH = 3; // must be 3 or greater
 
   // index 0 is used for output
   logic [DEPTH-1:0] [31:0]  addr_n,    addr_int,    addr_Q;
```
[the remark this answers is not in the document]
vogelpi: Yep.

rtl/ibex_core.sv
vogelpi: Regarding `RV32*`/`Rv32*`, lets first wait on the outcome of the discussion. As for the others, I do not really care, but I can change it back.

rtl/ibex_controller.sv
```diff
@@ -498,8 +498,8 @@ module ibex_controller (
   assign operand_a_fw_mux_sel_o = data_misaligned_i ? SEL_MISALIGNED : SEL_REGFILE;
 
   // update registers
-  always_ff @(posedge clk, negedge rst_n) begin : UPDATE_REGS
```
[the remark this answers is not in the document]
vogelpi: We do not have any rules regarding this. I will create a new issue.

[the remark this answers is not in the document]
sjgitty: This reminds me: should we release the coding style guide here as well?

rtl/ibex_controller.sv
```diff
@@ -498,8 +498,8 @@ module ibex_controller (
...
-    if (!rst_n) begin
+  always_ff @(posedge clk_i, negedge rst_ni) begin : UPDATE_REGS
```
sjgitty: we also don't have a rule on "," or "or" for these always_ff statements. Do we care about consistency? Worth putting in the style guide?

rtl/ibex_core.sv
```diff
@@ -383,10 +385,10 @@ module ibex_core #(
   ibex_ex_block #(
       //change the localparam MULT_TYPE to 0 or 1
```
sjgitty: It is odd to have this comment here, since the localparam is inside the _ex_block. I recommend remove.

rtl/ibex_core.sv
```diff
@@ -383,10 +385,10 @@ module ibex_core #(
...
       //if you want a SLOW or FAST multiplier
-      .RV32M(RV32M)
+      .Rv32m ( Rv32m )
   ) ex_block_i (
-      .clk                        ( clk                   ),
-      .rst_n                      ( rst_ni                ),
+      .clk_i                      ( clk                   ),
```
sjgitty: ditto

rtl/ibex_controller.sv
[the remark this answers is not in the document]
asb: We should share the style guide we're using, but it should probably be under lowRISC/style-guides or something like that.

rtl/ibex_controller.sv
vogelpi: The style guide does not explicitly specify it, but all the examples in the guide use the `or` instead of `,`. So I will change it to that. I would say we care about consistency and therefore, it would make sense to explicitly mention our choice in the style guide.

rtl/ibex_core.sv
vogelpi: Yep, I will remove it.

rtl/ibex_core.sv
vogelpi: Good catch!

rtl/ibex_controller.sv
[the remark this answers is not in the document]
vogelpi: I changed all labels to use `lower_snake_case` as suggested.

rtl/ibex_core.sv
vogelpi: Sorry @sjgitty , I do not get the issue here. Could you please elaborate a bit?

sjgitty: Oops, sorry, I had thought I found an inconsistency above, then "ditto"d it here. Then I realized it was as intended, and forgot to remove this "ditto". Please ignore.

rtl/ibex_core.sv
vogelpi: I changed `Rv32*` back to `RV32*` for now.

imphil:
Great, thanks!

Issue #24, opened 2019-05-03
Fix PDF documentation build


imphil:
```cmd
$ LANG=C make latexpdf
Running Sphinx v2.0.1
building [mo]: targets for 0 po files that are out of date
building [latex]: all documents
updating environment: 12 added, 0 changed, 0 removed
reading sources... [100%] register_file                                                                                                                                                      
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
processing ibex.tex... index introduction integration instruction_fetch load_store_unit register_file cs_registers interrupts performance_counters exception_interrupts debug copyright 
resolving references...
done
writing... failed

Exception occurred:
  File "/home/philipp/src/ibex/doc/venv/lib64/python3.7/site-packages/cairosvg/parser.py", line 414, in __init__
    'No tag with id="{}" found.'.format(element_id))
TypeError: No tag with id="vvv-2-2" found.
The full traceback has been saved in /tmp/sphinx-err-u9zes32v.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:20: latexpdf] Error 1
```

The SVG image in question actually is generated through wavedrom, as this part of the stacktrace shows:

```
  File "/home/philipp/src/ibex/doc/venv/lib64/python3.7/site-packages/sphinxcontrib/wavedrom.py", line 187, in visit_wavedrom
    imgname = render_wavedrom(self, node, outpath, bname, format)
  File "/home/philipp/src/ibex/doc/venv/lib64/python3.7/site-packages/sphinxcontrib/wavedrom.py", line 162, in render_wavedrom
    cairosvg.svg2pdf(svgout.tostring(), write_to=fpath)
```

The SVG (which is then converted to PDF and causes the error) contains a couple of tags with `xlink:href="#vvv-2-2"` statements, but no tag with the `id="vvv-2-2"` attribute. So the error message of cairosvg is correct, and something is wrong in the generation of the wavedrom SVG (for reference: the same problem happens with `vvv-2`, but `vvv-3` is handled correctly in the SVG)

The following wavedrom source code is trying to be rendered. It's in [doc/instruction_fetch.rst](https://github.com/lowRISC/ibex/blob/master/doc/instruction_fetch.rst):

```
{"signal":
  [
    {"name": "clk", "wave": "p......"},
[...]

PR #36, opened 2019-05-14
Fix wavedrom versions

wallento:
Ensure to pickup the correct versions for wavedrom.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,5 @@
 setuptools_scm
 sphinx>=1.8
 sphinx_rtd_theme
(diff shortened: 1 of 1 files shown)


wallento:
This fixes the wavedrom bug, that was solved by the major rewrite. This will be part of the next release which is ETA in a couple of days. Until then, manually pick up the release candidate. Fixes #24

wallento:
PDF still cannot build, but this is unrelated to this issue: It seems that sphinx emits incorrect latex code for images inside a note/warning. A simple workaround is to move the wavedrom image out of the warning box, but I can also have a look into the sphinx generation part.

imphil:
Thanks Stefan!

PR #42, opened 2019-05-16
Typo fix in muldiv: Reminder->Remainder

wallento:
Simple drive-by I assume: I am pretty sure this is about the remainder of the operation. But not entirely ;)


imphil:
Oh no, you took that one away from a Hackoktoberfest contributor :smile:  Thanks!

PR #48, opened 2019-05-22
Compressed hints

### LINKED ISSUE #46: Decode HINT as NOP
The core fetched an compressed instruction like below.

From a downstream bug report:

core   0: 0xffffffff8000f2ec (0x00009056) add     zero, zero, s5

0x9056 is correctly decoded as HINT instruction by spike, but ibex decode it as illegal C.EBREAK instruction:

https://github.com/lowRISC/stwg-base/blob/master/hw/vendor/lowrisc_ibex/rtl/ibex_compressed_decoder.sv#L215

```
              if (instr_i[11:7] == 5'b0) begin
                // c.ebreak -> ebreak
                instr_o = {32'h00_10_00_73};
                if (instr_i[6:2] != 5'b0) illegal_instr_o = 1'b1;
```

This code first checks if instr[11:7] is zero to determine if it's c.ebreak instruction, then check instr[6:2] to see if it's illegal. However, when instr[6:2] == 0 and instr_i[11:7] != 0, it should be treated as HINT instruction rather than illegal instruction.

imphil:
Fix handling of HINTs in compressed instructions. 

Fixes #46 (and a bit more)

@taoliug can you re-test with these modifications?
@asb does my reading of the spec sound about right?

### DIFF
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -214,16 +214,18 @@ module ibex_compressed_decoder (
                 instr_o = {12'b0, instr_i[11:7], 3'b0, 5'b0, {OPCODE_JALR}};
               end
             end else begin
-              // c.add -> add rd, rd, rs2
-              instr_o = {7'b0, instr_i[6:2], instr_i[11:7], 3'b0, instr_i[11:7], {OPCODE_OP}};
-
-              if (instr_i[11:7] == 5'b0) begin
-                // c.ebreak -> ebreak
-                instr_o = {32'h00_10_00_73};
-                if (instr_i[6:2] != 5'b0) illegal_instr_o = 1'b1;
-              end else if (instr_i[6:2] == 5'b0) begin
-                // c.jalr -> jalr x1, rs1, 0
-                instr_o = {12'b0, instr_i[11:7], 3'b000, 5'b00001, {OPCODE_JALR}};
+              if (instr_i[6:2] != 5'b0) begin
+                // c.add -> add rd, rd, rs2
+                // (c.add hints are translated into an add hint)
+                instr_o = {7'b0, instr_i[6:2], instr_i[11:7], 3'b0, instr_i[11:7], {OPCODE_OP}};
+              end else begin
+                if (instr_i[11:7] == 5'b0) begin
+                  // c.ebreak -> ebreak
+                  instr_o = {32'h00_10_00_73};
+                end else begin
+                  // c.jalr -> jalr x1, rs1, 0
+                  instr_o = {12'b0, instr_i[11:7], 3'b000, 5'b00001, {OPCODE_JALR}};
+                end
               end
             end
           end
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -206,12 +206,14 @@ module ibex_compressed_decoder (
 
           3'b100: begin
             if (instr_i[12] == 1'b0) begin
-              // c.mv -> add rd/rs1, x0, rs2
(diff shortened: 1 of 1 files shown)


rtl/ibex_compressed_decoder.sv
```diff
@@ -214,16 +214,17 @@ module ibex_compressed_decoder (
...
-                // c.jalr -> jalr x1, rs1, 0
-                instr_o = {12'b0, instr_i[11:7], 3'b000, 5'b00001, {OPCODE_JALR}};
+              if (instr_i[6:2] != 5'b0) begin
+                // c.add -> add rd, rd, rs2
+                instr_o = {7'b0, instr_i[6:2], instr_i[11:7], 3'b0, instr_i[11:7], {OPCODE_OP}};
+              end else begin
+                if (instr_i[11:7] == 5'b0) begin
+                  // c.ebreak -> ebreak
+                  instr_o = {32'h00_10_00_73};
+                end else if (instr_i[6:2] == 5'b0) begin
```
asb: Isn't this check redundant due to line 217? From a readability perspective, I don't think the extra check helps. Maybe a comment on line 220 or just after `// instr_i[6:2] == 5'b0`?

rtl/ibex_compressed_decoder.sv
```diff
@@ -214,16 +214,17 @@ module ibex_compressed_decoder (
...
-
-              if (instr_i[11:7] == 5'b0) begin
-                // c.ebreak -> ebreak
-                instr_o = {32'h00_10_00_73};
-                if (instr_i[6:2] != 5'b0) illegal_instr_o = 1'b1;
-              end else if (instr_i[6:2] == 5'b0) begin
-                // c.jalr -> jalr x1, rs1, 0
-                instr_o = {12'b0, instr_i[11:7], 3'b000, 5'b00001, {OPCODE_JALR}};
+              if (instr_i[6:2] != 5'b0) begin
+                // c.add -> add rd, rd, rs2
```
asb: I think it would be nice to indicate that this is written to handle both legal c.add and the associated HINT encodings. Maybe another line: `c.add hint -> add x0, x0, rs2`?

That would at least serve as a note to a future reader that this was indeed written with the intent of handling the c.add HINT encodings properly.

rtl/ibex_compressed_decoder.sv
```diff
@@ -206,12 +206,13 @@ module ibex_compressed_decoder (
 
           3'b100: begin
             if (instr_i[12] == 1'b0) begin
-              // c.mv -> add rd/rs1, x0, rs2
-              instr_o = {7'b0, instr_i[6:2], 5'b0, 3'b0, instr_i[11:7], {OPCODE_OP}};
-
-              if (instr_i[6:2] == 5'b0) begin
+              if (instr_i[6:2] != 5'b0) begin
+                // c.mv/c.add -> add rd/rs1, x0, rs2
```
asb: Given instr_i[12] is 0, don't we always have a c.mv here? The c.add reference seems confusing

asb: As with my comment on the last commit, it would be nice to explicitly indicate the intent to capture the c.mv hint encodings here too.

rtl/ibex_compressed_decoder.sv
imphil: Yes, this should have been just an else.

rtl/ibex_compressed_decoder.sv
imphil: Yes, the c.add comment is wrong here. (Got confused by reading the spec.)

imphil:
Thanks for your review @asb. I've updated the PR to address your comments.

PR #49, opened 2019-05-24
Code cleanup

vogelpi:
This PR implements three main changes:
- Enforce line wrapping after 100 characters. This is required by our style guide and avoids lint warnings.
- Simplify default assignments of `'X` and `'0` to enum types.
- Make state signal names for the prefetch buffer FSM lowercase.

Successfully tested in Verilator and Vivado.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -163,10 +163,10 @@ module ibex_controller (
 
     csr_save_cause_o       = 1'b0;
 
-    exc_cause_o            = exc_cause_e'({$bits(exc_cause_e){1'b0}});
+    exc_cause_o            = EXC_CAUSE_INSN_ADDR_MISA; // = 6'h00
     exc_pc_mux_o           = EXC_PC_IRQ;
 
-    csr_cause_o            = exc_cause_e'({$bits(exc_cause_e){1'b0}});
+    csr_cause_o            = EXC_CAUSE_INSN_ADDR_MISA; // = 6'h00
 
     pc_mux_o               = PC_BOOT;
     pc_set_o               = 1'b0;
@@ -513,7 +513,7 @@ module ibex_controller (
 
       default: begin
         instr_req_o = 1'b0;
-        ctrl_fsm_ns = ctrl_fsm_e'({$bits(ctrl_fsm_e){1'bX}});
+        ctrl_fsm_ns = ctrl_fsm_e'(1'bX);
       end
     endcase
   end
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -27,34 +27,35 @@ module ibex_controller (
     input  logic                      clk_i,
     input  logic                      rst_ni,
 
-    input  logic                      fetch_enable_i,        // Start the decoding
-    output logic                      ctrl_busy_o,           // Core is busy processing instructions
-    output logic                      first_fetch_o,         // Core is at the FIRST FETCH stage
-    output logic                      is_decoding_o,         // Core is in decoding state
+    input  logic                      fetch_enable_i,        // start decoding
+    output logic                      ctrl_busy_o,           // core is busy processing instrs
+    output logic                      first_fetch_o,         // core is at the FIRST FETCH stage
+    output logic                      is_decoding_o,         // core is in decoding state
 
     // decoder related signals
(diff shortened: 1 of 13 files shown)


rtl/ibex_controller.sv
```diff
@@ -160,10 +163,10 @@ module ibex_controller (
 
     csr_save_cause_o       = 1'b0;
 
-    exc_cause_o            = exc_cause_e'({$bits(exc_cause_e){1'b0}});
+    exc_cause_o            = EXC_CAUSE_INSN_ADDR_MISA; // = 6'h00
     exc_pc_mux_o           = EXC_PC_IRQ;
 
-    csr_cause_o            = exc_cause_e'({$bits(exc_cause_e){1'b0}});
+    csr_cause_o            = EXC_CAUSE_INSN_ADDR_MISA; // = 6'h00
```
imphil: These two changes look strange. We only assign these values because we want to assign 0, even though 0 has "special meaning" in this case, as demonstrated by the constant name? I we don't have a valid default value, can we assign x or simply nothing?

rtl/ibex_multdiv_fast.sv
```diff
@@ -122,7 +122,8 @@ module ibex_multdiv_fast (
   // 1. The 2 MSBs of the multiplicants are always equal, and
   // 2. The 16 MSBs of the addend (accum[33:18]) are always equal.
   // Thus, it is safe to ignore mac_res_ext[34].
-  assign mac_res_signed = $signed({sign_a, mult_op_a})*$signed({sign_b, mult_op_b}) + $signed(accum);
+  assign mac_res_signed =
+      $signed({sign_a, mult_op_a})*$signed({sign_b, mult_op_b}) + $signed(accum);
```
imphil: nit: add spaces around `*`

rtl/ibex_tracer.sv
```diff
@@ -331,8 +331,8 @@ module ibex_tracer #(
     instr_trace_t trace;
     mem_acc_t     mem_acc;
     // special case for WFI because we don't wait for unstalling there
-    if ( (id_valid_i || mret_insn_i || ecall_insn_i || pipe_flush_i || ebrk_insn_i || dret_insn_i ||
-          csr_status_i || ex_data_req_i) && is_decoding_i) begin
+    if ( (id_valid_i || mret_insn_i || ecall_insn_i || pipe_flush_i || ebrk_insn_i ||
+        dret_insn_i || csr_status_i || ex_data_req_i) && is_decoding_i) begin
```
imphil: Indentation:
```verilog
    if ((id_valid_i || mret_insn_i || ecall_insn_i || pipe_flush_i || ebrk_insn_i ||
         dret_insn_i || csr_status_i || ex_data_req_i) && is_decoding_i) begin
```

vogelpi: Sorry but the indentation is 4 spaces for line wraps, but I will remove the space between the parentheses.

rtl/ibex_multdiv_fast.sv
vogelpi: Yep, I'll do that.

rtl/ibex_tracer.sv
imphil: It's a typical issue for readability > code *guidelines*

rtl/ibex_controller.sv
vogelpi: Yes, that is the case: We assign 0 by default which in this case has a special meaning. I think we should not assign X because we use X only to propagate invalid input/state. We only do so in the default statement of `unique case` constructs which should not be reached under normal circumstances.

Here, we assign a default to a signal that is not assigned in all legal cases. So we cannot assign X but we must assign something. Otherwise we create latches. Thus, assigning 0 or the value which resolves to 0 is the cleanest thing to do.

rtl/ibex_tracer.sv
vogelpi: :-)

rtl/ibex_controller.sv
imphil: I don't have any great solutions either, unfortunately, so let's go with your choice.

vogelpi: Alright, thanks for your feedback BTW!

Issue #15, opened 2019-01-19
PCMR bits swapped in documentation


brabect1:
Looking at the RTL code, the function of Global Enable and Saturation is swapped in the documentation. Saturation is really controlled by bit 1 and enable by bit 0.

Please indicate which one is correct. RTL or doc?

https://github.com/pulp-platform/zero-riscy/blob/c6eeb34f57ff1cec354a74e025b889c26138146b/doc/performance_counters.rst#L16-L26

https://github.com/pulp-platform/zero-riscy/blob/b1b89f55fb1dbbf6f35ac5f92c5c1d12b2f41d45/zeroriscy_cs_registers.sv#L401-L406

vogelpi:
Hi @brabect1 ,
thanks for creating this issue. You are right. Bit 0 is the global enable, and Bit 1 is the saturation control bit. I changed the documentation. The fix will be upstream in my next PR.
Best regards,
Pirmin

PR #53, opened 2019-05-30
Perf counters & CSRs

### LINKED ISSUE #16: Bit clear operation (`csrrc` and `csrrci` ops) wrong for performance counter registers
Looking at RTL, clear operations are implemented bad. Applies to all CSRs related to performance counters (PCMR, PCER, PCCRx).

Below is the sample code for PCMR, but the same pattern applies to the other CSRs. Rather than inverting the register value, do invert the mask (i.e. `csr_wdata_i`).

https://github.com/pulp-platform/zero-riscy/blob/b1b89f55fb1dbbf6f35ac5f92c5c1d12b2f41d45/zeroriscy_cs_registers.sv#L431

### LINKED ISSUE #27: Standard-compliant performance counters
We need to look into performance counters and see how we can implement them in a standard-compliant way (or ensure that we document where we have good reasons to deviate).

While at it, check #17, #16 and  #15 to see if they are valid and fix/close as needed.

vogelpi:
This PR brings several changes related to CSRs and performance counters that bring the CSRs closer to the specs. The most important changes are:

- Fixes to the clear operation and and documentation: resolves #16 , #15 
- Replace legacy/custom performance counters by standard compliant ones, resolves #27 
- Add access checks for CSR operations and cause illegal instruction exceptions on violations
- Modify illegal clear and set CSR ops to be read-only as mandated by the specs 
- Update outdated documentation on CSRs and performance counters

The changes have been successfully tested using Verilator and Xilinx Vivado.


doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
-.. figure:: images/pcer.png
-   :name: pcer
-
-   Events and PCCR, PCMR and PCER on the ASIC.
+Ibex implements performance counters according to the RISC-V Privileged Specification, draft version 1.11 (see Hardware Performance Monitor, Section 3.1.11).
+The performance counters are placed inside the Control and Status Registers (CSRs) and can be accessed with the ``CSRRW(I)`` and ``CSRRS/C(I)`` instructions.
+
+Ibex implements the clock cycle counter ``mcycle(h)``, the retired instruction counter ``minstret(h)``, as well as the 29 event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)``. 
+
+All counters have a width of 64 bits. The lower 32 bits can be accessed through the base register, whereas the upper 32 bits are accessed through the ``h``-register.
```
imphil: Maybe make the detailed discussion of the counters a separate subsection?

doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
+| mhpmcounter7(h)  | 0xB07 (0xB87)  | Number of unconditional jumps (j, jal, jr, jalr)        |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter8(h)  | 0xB08 (0xB88)  | Number of branches (conditional)                        |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter9(h)  | 0xB09 (0xB89)  | Number of taken branches (conditional)                  |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter10(h) | 0xB0A (0xB8A)  | Number of compressed instructions retiredz              |
++------------------+----------------+---------------------------------------------------------+
+
+By default, all these counters are enabled after reset.
```
imphil: Could we make this a second subsection?

doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
+They can be individually enabled or disabled by overwriting the corresponding bit in the ``mcountinhibit`` CSR at address ``0x320`` with 0 or 1, respectively.
+To enable/disable ``mcycle(h)``, bit 0 must be written. For ``minstret(h)``, it is bit 2. For event counter ``mhpmcounterX(h)``, it is bit X.
```
imphil: Could you add a reference to the RISC-V Privileged Spec here and mention that we implement exactly that?

doc/cs_registers.rst
```diff
@@ -3,32 +3,63 @@
...
+|  0xB83  | MHPMCOUNTER3H  | R/W    | Upper 32 bits of ``mhmpcounter3``             |
++---------+----------------+--------+-----------------------------------------------+
+|     .           .             .                    .                              |
++---------+----------------+--------+-----------------------------------------------+
+|  0xB9F  | MHPMCOUNTER31H | R/W    | Upper 32 bits of ``mhmpcounter31``            |
++---------+----------------+--------+-----------------------------------------------+
+|  0xF14  | MHARTID        | R      | Hardware Thread ID                            |
++---------+----------------+--------+-----------------------------------------------+
+
+* Write accesses are ignored but allowed as mandated by the specifications in the sense that they do not cause illegal instruction exceptions.
```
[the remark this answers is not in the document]
asb: WARL is the term the RISC-V spec uses - write any values, read legal values

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
+  // update enable signals
+  always_comb begin : mcountinhibit_update
+    if (mcountinhibit_we == 1'b1) begin
+      mcountinhibit_n = csr_wdata_int;
+    end else begin
+      mcountinhibit_n = mcountinhibit_q;
+    end
+    // bit 1 must always be 0
+    mcountinhibit_n[1] = 1'b0;
   end
```
imphil: What prevents you from putting the code in `mcountinhibit_update` into the case statement above and remove this combinatorial block here?

```verilog

      CSR_MCOUNTINHIBIT: begin
        if (csr_we_int) begin
          mcountinhibit_n = csr_wdata_int;
          mcountinhibit_n[1] = 1'b0;
        end
      end
```

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
+                                                                          // (unconditional)
+  assign mhpmcounter_incr[8] = ~mcountinhibit[8] & branch_i;              // num of branches
+                                                                          // (conditional)
+  assign mhpmcounter_incr[9] = ~mcountinhibit[9] & branch_taken_i;        // num of taken
+                                                                          // branches
+                                                                          // (conditional)
+  assign mhpmcounter_incr[10] = ~mcountinhibit[10]                        // num of compressed
+      & id_valid_i & is_decoding_i & is_compressed_i;                     // instr
+
+  assign mhpmcounter_incr[31:11] = ~mcountinhibit[31:11];
```
imphil: To simplify this block of code and avoid future errors by forgetting to take the mcountinhibit array into account, I'd remove the `~mcountinhibit[x] & ` part and do this check when actually incrementing the counter in `mhpmcounter_update`.

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
+
+  // event selector (hardwired, 0 means no event)
+  always_comb begin: mhpmevent_active
+    for (int i=0; i<32; i++) begin : gen_mhpmevent_active
+      mhpmevent[i][i] = 1'b1;
```
imphil: mhpmevent[e] is not a bitfield, it's giving the "event number" for a given counter. So we should have something like ` mhpmevent[i] = i`

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
+  assign mhpmcounter_incr[9] = ~mcountinhibit[9] & branch_taken_i;        // num of taken
+                                                                          // branches
+                                                                          // (conditional)
+  assign mhpmcounter_incr[10] = ~mcountinhibit[10]                        // num of compressed
+      & id_valid_i & is_decoding_i & is_compressed_i;                     // instr
+
+  assign mhpmcounter_incr[31:11] = ~mcountinhibit[31:11];
+
+  // event selector (hardwired, 0 means no event)
+  always_comb begin: mhpmevent_active
```
imphil: The code below is static at compile time, you can make this just a for loop without the `always_comb` block.

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
+      // lower half
+      if (mhpmcounter_we[i] == 1'b1) begin // write
+        mhpmcounter_n[i][31:0] = csr_wdata_int;
+      end else if (mhpmcounter_incr[i] == 1'b1) begin // increment
+        mhpmcounter_n[i][31:0] = mhpmcounter_q[i][31:0] + 32'h1;
+      end else begin // hold
+        mhpmcounter_n[i][31:0] = mhpmcounter_q[i][31:0];
       end
     end
   end
```
imphil: How about this?


```verilog

  // update
  always_comb begin : mhpmcounter_update
    for (int i=0; i<31; i++) begin : gen_mhpmcounter_update
      mhpmcounter_n = mhpmcounter_q;

      if (mhpmcounter_incr[i] & ~mcountinhibit[i]) begin
        mhpmcounter_n[i] = mhpmcounter_n[i] + 64'h1;
      end

      if (mhpmcounter_we[i]) begin
        mhpmcounter_n[i][31:0] = csr_wdata_int;
      end else if (mhpmcounterh_we[i]) begin
        mhpmcounter_n[i][63:32] = csr_wdata_int;
      end
    end
  end
```

rtl/ibex_id_stage.sv
```diff
@@ -403,9 +405,25 @@ module ibex_id_stage #(
...
+  ///////////////////////
+  // CSR operand check //
+  ///////////////////////
+  always_comb begin : csr_operand_check
+    csr_op_ex_o = csr_op;
+
+    // CSRRSI/CSRRCI must not write 0 to CSRs (uimm[4:0]=='0)
+    // CSRRS/CSRRC must not write from x0 to CSRs (rs1=='0)
+    if (((csr_op == CSR_OP_SET) || (csr_op == CSR_OP_CLEAR)) &&
+         (instr[`REG_S1] == '0)) begin
```
imphil: No need for parentheses around each condition, the precedence of `==` is sufficiently high.

```
    if ((csr_op == CSR_OP_SET || csr_op == CSR_OP_CLEAR) &&
        instr[`REG_S1] == '0) begin
```

imphil:
Even more ideas reading the SiFive docs: It might be a good idea to make the width of the counter configurable, the E24 uses 40 bit for example; would be an easy-to-introduce parameter with large gains in area efficiency for many use-cases.

doc/cs_registers.rst
```diff
@@ -3,32 +3,63 @@
...
-| 01    | 11  | 00  | 0xxxxx | 0x780- | PCCRs   | R/W    | Performance Counter Counter       |
-|       |     |     |        | 0x79F  |         |        | Registers                         |
-+-------+-----+-----+--------+--------+---------+--------+-----------------------------------+
-| 01    | 11  | 10  | 100000 | 0x7A0  | PCER    | R/W    | Performance Counter Enable        |
-+-------+-----+-----+--------+--------+---------+--------+-----------------------------------+
-| 01    | 11  | 10  | 100001 | 0x7A1  | PCMR    | R/W    | Performance Counter Mode          |
-+-------+-----+-----+--------+--------+---------+--------+-----------------------------------+
-| 11    | 11  | 00  | 010100 | 0xF14  | MHARTID | R      | Hardware Thread ID                |
-+-------+-----+-----+--------+--------+---------+--------+-----------------------------------+
+Ibex does not implement all Control and Status Registers (CSRs) specified in the RISC-V privileged specifications, but is limited to the registers that were needed for the PULP system. The reason for this is that we wanted to keep the footprint of the core as low as possible and avoid any overhead that we do not explicitly need.
```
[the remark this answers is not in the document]
vogelpi: Done

doc/cs_registers.rst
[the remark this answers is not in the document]
vogelpi: Done

doc/cs_registers.rst
```diff
@@ -3,32 +3,63 @@
...
+
+The CSRs listed below are currently implemented with the following access permissions.
+
++---------+----------------+--------+-----------------------------------------------+
+| Address | Name           | Access | Description                                   |
++=========+================+========+===============================================+
+|  0x300  | MSTATUS        | R/W    | Machine Status                                |
```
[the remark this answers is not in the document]
vogelpi: Done

doc/performance_counters.rst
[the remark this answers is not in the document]
vogelpi: Done

doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
+Reads to all these registers are non-destructive.
+
+Of the 29 event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)``, only the lower 8 counters ``mhpmcounter3(h)`` - ``mhpmcounter10(h)`` are enabled.
+The association of events with counters is hardwired as listed in the following table.
+The remaining counters are disabled.
+
++------------------+----------------+---------------------------------------------------------+
+| Counter          | CSR Address    | Event Description                                       |
++==================+================+=========================================================+
+| mcycle(h)        | 0xB00 (0xB80)  | Number of cycles                                        |
```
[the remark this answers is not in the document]
vogelpi: Done

doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
++------------------+----------------+---------------------------------------------------------+
+| minstret(h)      | 0xB02 (0xB82)  | Number of instructions retired                          |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter3(h)  | 0xB03 (0xB83)  | Number of cycles waiting for data memory                |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter4(h)  | 0xB04 (0xB84)  | Cycles waiting for instruction fetches, i.e., number of |
+|                  |                | instructions wasted due to non-ideal caching            |
++------------------+----------------+---------------------------------------------------------+
+| mhpmcounter5(h)  | 0xB05 (0xB85)  | Number of data memory loads. Misaligned accesses are    |
+|                  |                | counted twice                                           |
```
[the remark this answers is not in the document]
vogelpi: Done

doc/performance_counters.rst
```diff
@@ -3,144 +3,48 @@
...
-:numref:`pcer` shows how events are first masked with the PCER register and then ORed together to increase the one performance counter PCCR.
-
-.. figure:: images/pcer.png
-   :name: pcer
-
-   Events and PCCR, PCMR and PCER on the ASIC.
+Ibex implements performance counters according to the RISC-V Privileged Specification, draft version 1.11 (see Hardware Performance Monitor, Section 3.1.11).
+The performance counters are placed inside the Control and Status Registers (CSRs) and can be accessed with the ``CSRRW(I)`` and ``CSRRS/C(I)`` instructions.
+
+Ibex implements the clock cycle counter ``mcycle(h)``, the retired instruction counter ``minstret(h)``, as well as the 29 event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)``.
```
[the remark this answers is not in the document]
vogelpi: Done

doc/performance_counters.rst
vogelpi: Done

doc/performance_counters.rst
vogelpi: Done

doc/performance_counters.rst
vogelpi: Done

rtl/ibex_core.sv
```diff
@@ -470,8 +472,11 @@ module ibex_core #(
   assign perf_load  = data_req_o & data_gnt_i & (~data_we_o);
   assign perf_store = data_req_o & data_gnt_i & data_we_o;
 
+  // Whenever the ID stage gets a new instr and the previously seen instr was valid, it has
+  // been executed & retired.
```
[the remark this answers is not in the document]
vogelpi: Done

rtl/ibex_id_stage.sv
vogelpi: Done

rtl/ibex_cs_registers.sv
```diff
@@ -15,18 +15,17 @@
...
+//                 RISC-V draft priviledged instruction set spec (v1.11)      //
 //                                                                            //
 ////////////////////////////////////////////////////////////////////////////////
 
 /**
  * Control and Status Registers
  *
- * Control and Status Registers (CSRs) loosely following the RiscV draft
- * priviledged instruction set spec (v1.9)
+ * Control and Status Registers (CSRs) loosely following the RISC-V draft
```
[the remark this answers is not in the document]
vogelpi: Done

rtl/ibex_cs_registers.sv
```diff
@@ -151,60 +146,74 @@ module ibex_cs_registers #(
...
   /////////////
   // CSR reg //
   /////////////
 
+  logic [$bits(csr_num_e)-1:0] csr_addr;
+  assign csr_addr           = {csr_addr_i};
+  assign mhpmcounter_idx    = csr_addr[4:0];
+
+  assign illegal_csr_priv   = 1'b0; // we only support M-mode
+  assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && (csr_we_int == 1'b1);
```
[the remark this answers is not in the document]
vogelpi: Done

rtl/ibex_cs_registers.sv
vogelpi: Done

rtl/ibex_cs_registers.sv
```diff
@@ -151,60 +146,74 @@ module ibex_cs_registers #(
...
-  logic [4:0]                    pccr_index;
-  logic                          pccr_all_sel;
-  logic                          is_pccr;
-  logic                          is_pcer;
-  logic                          is_pcmr;
+  // Hardware performance monitor signals
+  logic [31:0]       mcountinhibit_n, mcountinhibit_q, mcountinhibit;
+  logic [31:0]       mcountinhibit_force;
+  logic              mcountinhibit_we;
+  logic [31:0][63:0] mhpmcounter_n, mhpmcounter_q;
```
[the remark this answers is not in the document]
vogelpi: Done

rtl/ibex_cs_registers.sv
```diff
@@ -151,60 +146,74 @@ module ibex_cs_registers #(
...
+  logic [31:0]       mhpmcounter_we;
+  logic [31:0]       mhpmcounterh_we;
+  logic [31:0]       mhpmcounter_incr;
+  logic [31:0][31:0] mhpmevent;
```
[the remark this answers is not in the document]
vogelpi: Done

rtl/ibex_cs_registers.sv
vogelpi: Actually, it can be a bit field (as discussed). Thus I am leaving it as is.

rtl/ibex_cs_registers.sv
vogelpi: Done

rtl/ibex_cs_registers.sv
vogelpi: Seems more elegant than my implementation. Thanks for the suggestion.

vogelpi:
Thanks a lot @imphil for your comments and the help. I implemented your feedback. Two important changes I made:

1. The `ASIC_SYNTHESIS` flag is now gone. I have added a module parameter to define the number of counters (in addition to `mcycle` and `minstret`) that should be enabled. By default it is 8. The module then defines what these 8 counters should count.

2. I added a second parameter to define the width of these counters. `mcycle` and `minstret` are always 64 bit wide though.

Now, I am updating the documentation + I will add some info on the event association.

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +460,128 @@ module ibex_cs_registers #(
...
-  assign PCCR_inc[0] = (|(PCCR_in & PCER_q)) & PCMR_q[0];
-
-  always_comb begin
-    PCCR_n[0]   = PCCR_q[0];
-
-    if ((PCCR_inc_q[0] == 1'b1) && ((PCCR_q[0] != 32'hFFFFFFFF) || (PCMR_q[1] == 1'b0))) begin
-      PCCR_n[0] = PCCR_q[0] + 32'h1;
+  // event selector (hardwired, 0 means no event)
+  for (genvar i=0; i<3+NumMHPMCounters; i++) begin : gen_mhpmevent_active
+    if (i==1) begin
```
imphil: nit: `if (i == 1)`

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +460,128 @@ module ibex_cs_registers #(
...
+      assign mhpmevent[i] = '0; // not existing, reserved
+    end else begin
+      assign mhpmevent[i][i] = 1'b1;
     end
+  end
+  for (genvar i=3+NumMHPMCounters; i<32; i++) begin : gen_mhpmevent_inactive
+    assign mhpmevent[i] = '0;
+  end
```
imphil: How about putting these two generate blocks (gen_mhpmevent_active and gen_mhpmevent_inactive) in a always_comb block? This makes it easier to check that the signal is written only once.

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +460,128 @@ module ibex_cs_registers #(
...
 
-      PCCR_n[c]   = PCCR_q[c];
+  // update
+  always_comb begin : mhpmcounter_update
+    mhpmcounter_n = mhpmcounter_q;
 
-      if ((PCCR_inc_q[c] == 1'b1) && ((PCCR_q[c] != 32'hFFFFFFFF) || (PCMR_q[1] == 1'b0))) begin
-        PCCR_n[c] = PCCR_q[c] + 32'h1;
-      end
+    for (int i=0; i<31; i++) begin : gen_mhpmcounter_update
```
imphil: 31 -> 32?

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +460,128 @@ module ibex_cs_registers #(
...
-      endcase
-    end
+  // mask, controls effective counter width
+  for (genvar i=0; i<3; i++) begin : gen_mask_fixed
+    // mcycle, mtime, minstret are always 64 bit wide
+    assign mhpmcounter_mask[i] = {64{1'b1}};
+  end
+  for (genvar i=3; i<32; i++) begin : gen_mask_configurable
+    // mhpmcounters have a configurable width
+    assign mhpmcounter_mask[i] = {{{64-WidthMHPMCounters}{1'b0}}, {WidthMHPMCounters{1'b1}}};
```
imphil: As before, I'd use one block to set mhpmcounter_mask.

Reasoning: You want to set each signal only in exactly one block to quickly check for signals which have multiple drivers.

rtl/ibex_cs_registers.sv
```diff
@@ -214,11 +226,49 @@ module ibex_cs_registers #(
...
+          if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCOUNTINHIBIT
+              (csr_addr[4:0] == 5'b00001) ||
+              (csr_addr[4:0] == 5'b00010)) begin
+            illegal_csr = csr_access_i;
+          end
+
+        end else if (csr_addr_i == CSR_MCOUNTER_MASK) begin
+          csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][31: 0];
+          // check access to non-existent CSRs
+          if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCYCLE
```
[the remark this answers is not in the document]
vogelpi: Yes, these cases are already covered in the code blocks above using an explicit case. I put it here for completeness. I prefer to have the explicit case for CSR_MCYCLE(H) and CSR_MINSTRET(H), and only use the default for the event counters and selectors (to not list these 3*29 registers). Theoretically, we would just need to check for `5b00001` for CSR_MCOUNTER(H)_MASK and `5b00001` and `5b00010` for CSR_MCOUNTER_SETUP_MASK.

But in my opinion, it is just cleaner to list all of them here. I adjusted the comments to make clear that we also check already covered CSRs besides non-existent ones here.

rtl/ibex_cs_registers.sv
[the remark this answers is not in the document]
vogelpi: Sorry, I fixed it but it got lost when I was fixing a rebase conflict.

rtl/ibex_cs_registers.sv
```diff
@@ -362,197 +457,131 @@ module ibex_cs_registers #(
...
-  for (genvar i = 0; i < NumExtCounters; i++) begin : gen_ext_counters
-    assign PCCR_in[N_PERF_COUNTERS - NumExtCounters + i] = ext_counters_i[i];
+  // update enable signals
+  always_comb begin : mcountinhibit_update
+    if (mcountinhibit_we == 1'b1) begin
+      mcountinhibit_n = csr_wdata_int;
+    end else begin
+      mcountinhibit_n = mcountinhibit_q;
+    end
+    // bit 1 must always be 0
```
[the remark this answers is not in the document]
vogelpi: Sorry, but the spec does not say that this is for mtime. This is just our assumption. The only thing that the spec says is that there are some mtime and mtimecmp registers that are memory mapped, and that bit 1 in mcountinhibit is fixed to 0.

rtl/ibex_cs_registers.sv
vogelpi: Actually, I did not see your comment. There were too many and GitHub decided to hide some and only show them when hitting an unfold button.

vogelpi:
I implemented the remaining requests. There are two things left:
- a ref to why mcountinhibit[1] must be 0 (the spec just says it must be 0 and nothing more, this is already in the comment)
- the check for accessing mcycle(h) in the block handling the accesses to the event counters and event selectors (I adapted the comment).
Both I would leave as is.

rtl/ibex_cs_registers.sv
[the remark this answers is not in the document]
imphil: Not insisting on that comment, but it says: "Because the time counter can be shared between multiple cores, it cannot be inhibited with the mcountinhibit mechanism." (Section 3.1.13 of the privileged spec.)

PR #74, opened 2019-06-09
Fix linting errors

vogelpi:
This PR fixes 3 linting errors.

Both Verilator and Vivado builds successfully tested.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -186,8 +186,10 @@ module ibex_cs_registers #(
   /////////////
 
   logic [$bits(csr_num_e)-1:0] csr_addr;
+  logic [4:0]                  unused_csr_addr;
   assign csr_addr           = {csr_addr_i};
   assign mhpmcounter_idx    = csr_addr[4:0];
+  assign unused_csr_addr    = csr_addr[9:5];
 
   assign illegal_csr_priv   = 1'b0; // we only support M-mode
   assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_we_int;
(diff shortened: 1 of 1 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -541,7 +543,7 @@ module ibex_cs_registers #(
 
     // deactivate
     mhpmevent[1] = '0; // not existing, reserved
-    for (int i=3+MHPMCounterNum; i<32; i++) begin : gen_mhpmevent_inactive
+    for (int unsigned i=3+MHPMCounterNum; i<32; i++) begin : gen_mhpmevent_inactive
```
imphil: Why do we need the unsigned here? (It should be default, and yosys for example doesn't support the explicit "unsigned", so there's value in not having it.)

vogelpi: Thanks for your comment. I am not sure I get the point about the "default".
The `unsigned` is needed because by our guidelines we use `int unsigned` parameters. Otherwise the linting tool generates an error because of the unsigned parameter is assigned to a signed variable. In contrast, `genvar` by default generates a variable that can take natural positive numbers (i.e. unsigned).
Alternatively, we could write

 ```for (int i=3+$(signed)MHPMCounterNum; i<32; i++) begin```

This works with Verilator and Vivado as well. But I am not sure about Yosys. Or we could use an localparam that is just `int`. But this seems a bit overkill to me. What do you think?

imphil: You're right, "int" is signed by default (I thought it was unsigned by default). So please ignore my comment.

PR #61, opened 2019-06-05
RISC-V Formal Interface

towoe:
Interface for formal verification

PR #76, opened 2019-06-17
Fix RVFI signal name

towoe:
This updates the name of a signal which was renamed in the process of https://github.com/lowRISC/ibex/pull/61


imphil:
Thanks Tobi!

Issue #62, opened 2019-06-05
Switch ibex_tracer to use RVFI


imphil:
Currently ibex_tracer is instantiated within the ibex_core.sv file depending on a define. It would be cleaner if we do the following:

- Modify ibex_tracer.sv take the RVFI signals as input. Requires #61 to be merged first.
- Remove the instantiation of ibex_tracer from ibex_core.sv
- Show people how to instantiate ibex_core and ibex_tracer in their top-level design; either in documentation, or by creating ibex_core_tracing.sv which instantiates ibex_core and ibex_tracer and passes through all signals except for the tracing ones.

PR #77, opened 2019-06-18
Tracer

towoe:
First draft for https://github.com/lowRISC/ibex/issues/62


imphil:
Thanks, I'll have a look next week.

towoe:
Hey @imphil, I can definitely change this. My first assumption was that the tracer module should be the same as before, but if we a free to change it that this is fine by me :)
The restructuring is fine? Having another ibex_core_tracer which has the ibex_core and the tracer modules together with another .core file?

imphil:
Yes, the restructuring is fine and shows people how to integrate the tracer in their own design

towoe:
Found something which I will change right now.

towoe:
Sorry @imphil for the premature review request, found some unneeded signals.
In total I remove the `tracer_*` signals and only the RVFI signals are used. Did some changes to the tracer itself, but I guess this could be rewritten in total, maybe allow for compatibility with verilator (but not sure).

doc/tracer.rst
```diff
@@ -0,0 +1,8 @@
+.. _tracer:
+
+Tracer
+======
+
+The module ``ibex_tracer`` can be used to create a log of the executed instructions.
+It is used by ``ibex_core_tracer`` which forwards the signals added by :ref:`rvfi` as an input for the tracer.
+Currently it is not possible to use this module with verilator.
```
imphil: nit: verilator -> Verilator

How about a RST note?

```rst
.. note::

   ``ibex_tracer`` is not compatible with Verilator.
```

examples/sim/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+export MODEL_TECH=/path/to/modelsim/bin
+```
+
+## Usage
+
+In the `examples` directory of `ibex` run the following command:
+
+```
+fusesoc --cores-root=.. run --target=sim lowrisc:ibex:top_tracer_sim
```
imphil: We typically recommend running fusesoc from the toplevel, which would make the comment this:
```
fusesoc --cores-root=. run --target=sim lowrisc:ibex:top_tracer_sim
```

examples/sim/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+The trace output can be found in the directory `examples/build/lowrisc_ibex_top_tracer_sim_0.1/sim-modelsim/`
```
imphil: If you run fusesoc from the toplevel the build directory changes as well.

examples/sim/rtl/prim_clock_gating.sv
```diff
@@ -0,0 +1,16 @@
...
+// Dummy clock gating module
+
+module prim_clock_gating (
+    input        clk_i,
+    input        en_i,
+    input        test_en_i,
+    output logic clk_o
+);
+
+  assign clk_o = en_i ? clk_i : 0;
```
imphil: Please use the code from `./dv/uvm/tb/prim_clock_gating.sv` for this module, your example breaks with the latch-based register file and confuses people.

examples/sim/tb/ibex_tracer_tb.sv
```diff
@@ -0,0 +1,123 @@
...
+    #10ns instr_rdata = 32'h00000013;
+    #10ns instr_rdata = 32'h00000013;
+    #10ns instr_rdata = 32'h00000013;
+    #10ns instr_rdata = 32'h00000013;
+    #10ns instr_rdata = 32'h00000013;
+    #10ns instr_rdata = 32'h00000013;
+  end
+
+  ibex_core_tracer #(
+    .RV32M(0)
```
imphil: Can we not set any parameter here, i.e. leave the default configuration?

examples/sim/top_tracer_sim.core
```diff
@@ -0,0 +1,34 @@
...
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:ibex:top_tracer_sim:0.1"
+description: "Ibex tracer example for modelsim"
+filesets:
+  files_tb:
+    depend:
+      - lowrisc:ibex:tracer
+      - vlog_tb_utils
```
imphil: Where does that come from?

examples/sim/tb/ibex_tracer_tb.sv
```diff
@@ -0,0 +1,123 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Sample testbench for Ibex tracer
+// The `nop` instruction is the only input
+
+`timescale 1ns/1ns
```
imphil: You already set the timescale in the *.core file, you can remove it here.

rtl/ibex_core.sv
```diff
@@ -24,6 +24,11 @@
 `ifdef RISCV_FORMAL
   `define RVFI
 `endif
+// Some signals needed for tracing are already defined by
+// RISC-V Formal Interface, so use those as well
+`ifdef TRACE_EXECUTION
+  `define RVFI
+`endif
```
imphil: The TRACE_EXECUTION define is not used in ibex_core or any of its children. We can therefore remove it from this file and put this code into the file using ibex_core.sv

rtl/ibex_core_tracer.sv
```diff
@@ -0,0 +1,129 @@
...
+
+`ifdef RISCV_FORMAL
+  `define RVFI
+`endif
+`ifdef TRACE_EXECUTION
+  `define RVFI
+`endif
+
+/**
+ * Top level module of the ibex RISC-V core
```
imphil: ... with tracing enabled.

rtl/ibex_core_tracer.sv
```diff
@@ -0,0 +1,129 @@
...
+
+  ibex_core #(
+    .MHPMCounterNum(MHPMCounterNum),
+    .MHPMCounterWidth(MHPMCounterWidth),
+    .RV32E(RV32E),
+    .RV32M(RV32M),
+    .DmHaltAddr(DmHaltAddr),
+    .DmExceptionAddr(DmExceptionAddr)
+  ) u_ibex_core (
+    .*
```
imphil: Please replace this wildcard with a explicit port list (https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#language-features-1).

You can use the short form of writing a signal name only once, e.g.

```
  ) u_ibex_core (
    .clk_i,
    .rst_ni,
   ...
  );

File: rtl/ibex_core_tracer.sv [hunk 11]
```diff
@@ -0,0 +1,129 @@
...
+  );
+
+
+`ifndef VERILATOR
+  ibex_tracer ibex_tracer_i (
+      .clk_i            ( clk_i          ), // always-on clk for tracer
```
imphil: You can remove that comment, nothing gates the clock in this module.

rtl/ibex_id_stage.sv
```diff
@@ -28,6 +28,9 @@
 `ifdef RISCV_FORMAL
   `define RVFI
 `endif
+`ifdef TRACE_EXECUTION
+  `define RVFI
+`endif
```
imphil: Remove these lines (see comment in ibex_core.sv).

rtl/ibex_if_stage.sv
```diff
@@ -23,6 +23,9 @@
 `ifdef RISCV_FORMAL
   `define RVFI
 `endif
+`ifdef TRACE_EXECUTION
+  `define RVFI
+`endif
```
imphil: Remove these lines (see comment in ibex_core.sv).

towoe:
@imphil I am not really sure I get https://github.com/lowRISC/ibex/pull/77#discussion_r301082651
The `TRACE_EXECUTION` is not really useful, I get that, so you would suggest to just define `RVFI` in the `ibex_tracer.core`?

towoe:
Thanks @imphil for the helpful review. And thanks for finding the unused tracer signals in the id stage.
I removed vlog_tb_utils (yes, this comes from Olof), but for reference, I used it as it was quite convenient as the VCD file was created at the same time as the trace log and is was easy to reload it with gtkwave.

imphil:
> The TRACE_EXECUTION is not really useful, I get that, so you would suggest to just define RVFI in the ibex_tracer.core?

almost; I'd define it in the ibex_tracer.sv file. The reasoning is: RVFI enables the tracing interface, so if someone (in this case, ibex_tracer.sv) wants to use it, they must define it. The downside is of course that now the compile order matters (ibex_tracer.sv must be compiled first). I would just live with that requirement.

rtl/ibex_tracer.sv
```diff
@@ -19,6 +19,9 @@
 
 `ifndef VERILATOR
 
+// ibex_tracer relies on the signals from the RISC-V Formal Interface
+`define RVFI
```
imphil: You need to define this one in ibex_core_tracer.sv, not in ibex_tracer.sv.

rtl/ibex_core_tracer.sv
```diff
@@ -0,0 +1,181 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+
+`ifdef RISCV_FORMAL
+  `define RVFI
+`endif
```
imphil: You can remove that, formal isn't to be used with this tracer wrapper.

imphil:
Thanks @towoe for your work on this! Looks like our investment in cleanups and more generic infrastructure (like the RVFI port) is paying off.

tjaychen:
sorry, i'm a bit late to this.  But there's something a bit confusing to me.
I see [here](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_core_tracer.sv#L7) that RVFI is directly defined in the tracer, but [here](https://github.com/lowRISC/ibex/blob/master/ibex_tracer.core#L28) it's defined in the tracer.core file

My understanding of this update is that we should just instantiate ibex_tracer_core.sv in the top level, as that wraps both the core and the tracer.  If that's the case, should we not leave it to the core file to set the define bool instead of putting it directly in the file?

tjaychen:
@towoe 
@imphil

towoe:
@tjaychen the idea was to define it in the sv file https://github.com/lowRISC/ibex/pull/77#issuecomment-509540775 but I encountered a problem that RVFI was not defined in the other files in the modelsim run so I put it in the core file for now.

imphil:
@tjaychen Defines are compile-order dependent, so we need to compile ibex_core_tracer.sv first if we set the define there.

In our core file (https://github.com/lowRISC/ibex/blob/master/ibex_tracer.core#L8) we depend on lowrisc:ibex:ibex, so that code is compiled first (since it is a dependency), and the ibex_core_tracer.sv comes afterwards.

We could work around that in various ways, but the easiest solution is just to set the define globally and avoid the compile-order dependency.

tjaychen:
Sorry just to clarify, I thought the way to do it globally would be to put
it in the core file?

PR #78, opened 2019-06-18
Ibex example Arty A7-100T

towoe:
Simple example of how Ibex can be used to run software on a FPGA target.

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -3,7 +3,21 @@
 Examples
 ========
 
+To make use of Ibex it has to be integrated as described in :ref:`core-integration`.
+
 Tracer Simulation
 -----------------
 
 How to run a simple testbench to test the tracer is described in ``examples/sim/README.md``.
+
+FPGA
+----
+
+A minimal example for the `Arty A7 <https://reference.digilentinc.com/reference/programmable-logic/arty-a7/start>`_ FPGA Development board is provided.
+In this example Ibex is directly linked to a SRAM memory instance.
+Four LEDs from the board are connected to the data bus and are updated each time when a word is written.
+The memory is separated into a instruction and data section.
+The instructions memory is initialized at synthesis time by reading the output from the software build.
+The software writes to the data section the complementary lower for bits of a word every second resulting in blinking LEDs.
+
+Find the description of how to build and program the Arty board in ``examples/fpga/artya7-100/README.md``.
--- a/examples/fpga/artya7-100/README.md
+++ b/examples/fpga/artya7-100/README.md
@@ -1,0 +1,48 @@
+# Ibex RISC-V Core SoC Example
+
+Please see [examples](https://ibex-core.readthedocs.io/en/latest/examples.html "Ibex User Manual") for a description of this example.
+
+## Requirements
+
+### Tools
+
+  - RV32 compiler
+  - srecord
+  - `fusesoc` and its dependencies
+  - Xilinx Vivado
+
+## Build
+
+### Software
+
+First the software must be built. Go into `examples/sw/led` and call:
+
+```
+make CC=/path/to/RISC-V-compiler
+```
+
+The path to the RV32 compiler `/path/to/RISC-V-compiler` depends on the environment.
+For example, it can be `riscv32-unknown-elf-gcc` if the binary is available through the `PATH` environment or `/opt/riscv/bin/riscv-none-embed-gcc` if a specific path is used.
+
+This should produce a `led.vmem` file which is used in the synthesises to update the SRAM storage.
+
+### Hardware
+
+Run the following command at the top level to build the hardware.
+
+```
+fusesoc --cores-root=. build lowrisc:ibex:top_artya7_100
+```
+
+This will create a directory `build` which contains the output files, including
+the bitstream.
+
+## Program
+
+After the board is connected to the computer it can be programmed with:
+
+```
+fusesoc --cores-root=. pgm lowrisc:ibex:top_artya7_100
+```
+
+LED1/LED3 and LED0/LED2 should alternately be on after the FPGA programming is finished.
--- a/examples/fpga/artya7-100/rtl/clkgen_xil7series.sv
+++ b/examples/fpga/artya7-100/rtl/clkgen_xil7series.sv
@@ -1,0 +1,79 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module clkgen_xil7series (
+    input IO_CLK,
+    input IO_RST_N,
+    output clk_sys,
+    output rst_sys_n
+);
+  logic locked_pll;
+  logic io_clk_buf;
+  logic clk_50_buf;
+  logic clk_50_unbuf;
+  logic clk_fb_buf;
+  logic clk_fb_unbuf;
+
+  // input buffer
+  IBUF io_clk_ibuf(
+    .I (IO_CLK),
+    .O (io_clk_buf)
+  );
+
+  PLLE2_ADV #(
+    .BANDWIDTH            ("OPTIMIZED"),
+    .COMPENSATION         ("ZHOLD"),
+    .STARTUP_WAIT         ("FALSE"),
+    .DIVCLK_DIVIDE        (1),
+    .CLKFBOUT_MULT        (12),
+    .CLKFBOUT_PHASE       (0.000),
+    .CLKOUT0_DIVIDE       (24),
+    .CLKOUT0_PHASE        (0.000),
+    .CLKOUT0_DUTY_CYCLE   (0.500)
+  ) pll (
+    .CLKFBOUT            (clk_fb_unbuf),
+    .CLKOUT0             (clk_50_unbuf),
+    .CLKOUT1             (),
+    .CLKOUT2             (),
+    .CLKOUT3             (),
+    .CLKOUT4             (),
+    .CLKOUT5             (),
+     // Input clock control
+    .CLKFBIN             (clk_fb_buf),
+    .CLKIN1              (io_clk_buf),
+    .CLKIN2              (1'b0),
+     // Tied to always select the primary input clock
+    .CLKINSEL            (1'b1),
+    // Ports for dynamic reconfiguration
+    .DADDR               (7'h0),
+    .DCLK                (1'b0),
+    .DEN                 (1'b0),
+    .DI                  (16'h0),
+    .DO                  (),
+    .DRDY                (),
+    .DWE                 (1'b0),
+    // Other control and status signals
+    .LOCKED              (locked_pll),
+    .PWRDWN              (1'b0),
+    // Do not reset PLL on external reset, otherwise ILA disconnects at a reset
+    .RST                 (1'b0));
+
+  // output buffering
+  BUFG clk_fb_bufg (
+    .I (clk_fb_unbuf),
+    .O (clk_fb_buf)
+  );
+
+  BUFG clk_50_bufg (
+    .I (clk_50_unbuf),
+    .O (clk_50_buf)
+  );
+
+  // outputs
+  // clock
+  assign clk_sys = clk_50_buf;
+
+  // reset
+  assign rst_sys_n = locked_pll & IO_RST_N;
+endmodule
--- a/examples/fpga/artya7-100/rtl/top_artya7_100.sv
+++ b/examples/fpga/artya7-100/rtl/top_artya7_100.sv
@@ -1,0 +1,153 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module top_artya7_100 (
+    input               IO_CLK,
+    input               IO_RST_N,
+    output [3:0]        LED
+);
+
+  parameter int          MEM_SIZE  = 64 * 1024; // 64 kB
+  parameter logic [31:0] MEM_START = 32'h00000000;
+  parameter logic [31:0] MEM_MASK  = MEM_SIZE-1;
+
+  logic clk_sys, rst_sys_n;
+
+  // Instruction connection to SRAM
+  logic        instr_req;
+  logic        instr_gnt;
+  logic        instr_rvalid;
+  logic [31:0] instr_addr;
+  logic [31:0] instr_rdata;
+
+  // Data connection to SRAM
+  logic        data_req;
+  logic        data_gnt;
+  logic        data_rvalid;
+  logic        data_we;
+  logic [31:0] data_addr;
+  logic [31:0] data_wdata;
+  logic [31:0] data_rdata;
+
+  // SRAM arbiter
+  logic [13:0] mem_addr_index;
+  logic        mem_req;
+  logic        mem_write;
+  logic [31:0] mem_wdata;
+  logic        mem_rvalid;
+  logic [31:0] mem_rdata;
+
+
+  ibex_core #(
+     .DmHaltAddr(32'h00000000),
+     .DmExceptionAddr(32'h00000000)
+  ) u_core (
+     .clk_i                 (clk_sys),
+     .rst_ni                (rst_sys_n),
+
+     .test_en_i             ('b0),
+
+     .core_id_i             (4'b0),
+     .cluster_id_i          (6'b0),
+     // First instruction executed is at 0x0 + 0x80
+     .boot_addr_i           (32'h00000000),
+
+     .instr_req_o           (instr_req),
+     .instr_gnt_i           (instr_gnt),
+     .instr_rvalid_i        (instr_rvalid),
+     .instr_addr_o          (instr_addr),
+     .instr_rdata_i         (instr_rdata),
+
+     .data_req_o            (data_req),
+     .data_gnt_i            (data_gnt),
+     .data_rvalid_i         (data_rvalid),
+     .data_we_o             (data_we),
+     // TODO: Byte access needs to be implemented
+     .data_be_o             (),
+     .data_addr_o           (data_addr),
+     .data_wdata_o          (data_wdata),
+     .data_rdata_i          (data_rdata),
+     .data_err_i            ('b0),
+
+     .irq_i                 ('b0),
+     .irq_id_i              ('b0),
+     .irq_ack_o             (),
+     .irq_id_o              (),
+
+     .debug_req_i           ('b0),
+
+     .fetch_enable_i        ('b1)
+  );
+
+  // Connect Ibex to SRAM
+  always_comb begin
+    mem_req        = 1'b0;
+    mem_addr_index = 14'b0;
+    mem_write      = 1'b0;
+    mem_wdata      = 32'b0;
+    if (instr_req) begin
+      mem_req        = (instr_addr & ~MEM_MASK) == MEM_START;
+      mem_addr_index = instr_addr[15:2];
+    end else if (data_req) begin
+      mem_req        = (data_addr & ~MEM_MASK) == MEM_START;
+      mem_write      = data_we;
+      mem_addr_index = data_addr[15:2];
+      mem_wdata      = data_wdata;
+    end
+  end
+
+  // SRAM block for instruction and data storage
+  ram_1p #(
+    .Width(32),
+    .Depth(MEM_SIZE / 4)
+  ) u_ram (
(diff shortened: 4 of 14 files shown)


examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module top (
```
imphil: Can we find a more descriptive name than `top`? Once we start adding more examples, having multiple files named `top` in our repository makes it harder to communicate and to find them in an editor/IDE.

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+
+  /*
+  *  SRAM block for instruction and data storage
+  *
+  *  The address range of the core is move to the
+  *  storage address space.
+  */
+  ram_2p #(
+    .Width(32),
+    .Depth(4 * 1024 / 4) // 4 kB
```
imphil: Is that the maximum amount of blockram we can get on this board? Otherwise I'd try to go for 64kB.

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+  ) u_ram (
+    .clk_a_i     (clk_sys),
+    .clk_b_i     (clk_sys),
+
+    .a_req_i     (instr_req),
+    .a_write_i   ('b0),
+    .a_addr_i    ((instr_addr - 32'h00001000)/4), // Translate addr to storage
```
imphil: Let's do this like a normal interconnect would do it:
- Gate the request signal with the address mask for the RAM area
- The RAM byte address is then only the LSB parts of the address.
- Please assign the result of these operations to an intermediate signal instead of doing the logic inside the ram_2p instantiation. This way you can have signal names which "speak for themselves", e.g. "instr_addr_index" (with index being a typical name for word addresses; without this suffix one would assume byte addresses)

towoe:
Thanks a lot @imphil for the comments! I will incorporate those soon.

examples/fpga/artya7-100/rtl/top.sv
towoe: do we want something like `top_artya7_100` or maybe `top_alpine_artya7_100` to have a little more Ibex in there?

examples/fpga/artya7-100/rtl/top.sv
towoe: I was aiming more for the minimum, but I see that it might be preferable to go for a bigger default.

examples/fpga/artya7-100/rtl/top.sv
imphil: I'd go for something like top_artya7_100, and use the same name in the core file (=> "lowrisc:ibex:top_artya7_100")

examples/README.md
```diff
@@ -0,0 +1,46 @@
...
+This will create a directory `build` which contains the output files, including
+the bitstream.
+
+## Program
+
+Open the Vivado project:
+`build/lowrisc_ibex_top_artya7-100_0.1/synth-vivado/lowrisc_ibex_top_artya7-100_0.1.xpr`
+
+And use the Hardware Manager to program the Arty board:
+`build/lowrisc_ibex_top_artya7-100_0.1/synth-vivado/lowrisc_ibex_top_artya7-100_0.1.bit`
```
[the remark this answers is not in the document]
towoe: done

examples/fpga/artya7-100/rtl/clkgen_xil7series.sv
```diff
@@ -0,0 +1,92 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module clkgen_xil7series (
+    input IO_CLK,
+    input IO_RST_N,
+    output clk_sys,
+    output clk_48MHz,
```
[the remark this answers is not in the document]
towoe: done

examples/fpga/artya7-100/rtl/clkgen_xil7series.sv
```diff
@@ -0,0 +1,92 @@
...
+  );
+
+  BUFG clk_48_bufg (
+    .I (clk_48_unbuf),
+    .O (clk_48_buf)
+  );
+
+  // outputs
+  // clock
+  assign clk_sys = clk_50_buf; // TODO: choose 50 MHz clock as sysclock for now
```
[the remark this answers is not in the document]
towoe: I was able to get something around 57 MHz (1200/21). Could you maybe test if for example 60 MHz is not possible?

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+
+     .fetch_enable_i        ('b1)
+  );
+
+  /*
+  *  SRAM block for instruction and data storage
+  *
+  *  The address range of the core is move to the
+  *  storage address space.
+  */
```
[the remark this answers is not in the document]
towoe: done

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+    .b_req_i     (data_req),
+    .b_write_i   (data_we),
+    .b_addr_i    ((data_addr - 32'h00001000)/4), // Translate addr to storage
+    .b_wdata_i   (data_wdata),
+    .b_rdata_o   (data_rdata)
+  );
+
+  /*
+  * Connect the led output the the lower four bits of a written data word
+  */
```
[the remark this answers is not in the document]
towoe: done

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+  logic [3:0] leds;
+  always_ff @(posedge clk_sys or negedge rst_sys_n) begin
+    if (!rst_sys_n) begin
+      leds <= 4'b0;
+    end else if (data_we) begin
+      leds <= data_wdata[3:0];
+    end else begin
+      leds <= leds;
```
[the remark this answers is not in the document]
towoe: done

examples/fpga/artya7-100/rtl/top.sv
```diff
@@ -0,0 +1,122 @@
...
+  );
+
+  /*
+  * Connect the led output the the lower four bits of a written data word
+  */
+  logic [3:0] leds;
+  always_ff @(posedge clk_sys or negedge rst_sys_n) begin
+    if (!rst_sys_n) begin
+      leds <= 4'b0;
+    end else if (data_we) begin
```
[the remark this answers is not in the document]
towoe: done

examples/sw/led/Makefile
```diff
@@ -0,0 +1,63 @@
...
+endif
+ARCH = rv32imc
+# ARCH = rv32im # to disable compressed instructions
+SRCS = $(PROGRAM).c
+
+CC = /tools/riscv/bin/riscv32-unknown-elf-gcc
+
+OBJCOPY ?= $(subst gcc,objcopy,$(wordlist 1,1,$(CC)))
+OBJDUMP ?= $(subst gcc,objdump,$(wordlist 1,1,$(CC)))
+REGTOOL = ../../util/regtool.py
```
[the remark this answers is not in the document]
towoe: removed

examples/sw/led/led.c
```diff
@@ -0,0 +1,47 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stdint.h>
+#define CLK_FIXED_FREQ_HZ (50ULL * 1000 * 1000)
```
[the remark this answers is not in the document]
towoe: checked, thanks for the pointer!

examples/fpga/artya7-100/rtl/top.sv
towoe: used top_artya7_100

examples/fpga/artya7-100/rtl/top.sv
towoe: I used 64kB with 48kB for the "rom" and 16kB for data

examples/fpga/artya7-100/rtl/top.sv
towoe: Thanks for the guidance, could you check if this is ok now?

examples/fpga/artya7-100/rtl/top.sv
albmoriconi: I can see how both choices (maximal vs minimal amount) have their own merit; however, wouldn't clocking multiple (unused) BRAMs add to the device power consumption?

imphil: That's certainly possible, but low power consumption is a not a goal for this simple demo example, helping people to get started with Ibex.

examples/fpga/artya7-100/rtl/top_artya7_100.sv
```diff
@@ -0,0 +1,130 @@
...
+     .instr_gnt_i           ('b1),
+     .instr_rvalid_i        ('b1),
+     .instr_addr_o          (instr_addr),
+     .instr_rdata_i         (instr_rdata),
+
+     .data_req_o            (data_req),
+     .data_gnt_i            ('b1),
+     .data_rvalid_i         ('b1),
+     .data_we_o             (data_we),
+     .data_be_o             (data_be),
```
albmoriconi: The data byte-write enable is connected to the core out, but not used. Could this be a problem if another application is used?

towoe: For simplicity I am ignoring the byte enable, as the SRAM is only addressed on a word level basis.
What kind of problem are you thinking about?

towoe: And following the comment from @imphil the interface to the SRAM is going to be changed.

albmoriconi: I guess it would be used in case of a misaligned write; sure @imphil can shed some light.

towoe: Yes, you are right, misaligned accesses use the data enable signal. And ignoring this is not correct if one accesses the data storage on a byte granularity and or not word aligned. But the purpose of this example is to have a simple starting point and not a complete system, as the direct connection of the LEDs to the write data signal is also an example of.

towoe:
>     * Can we go for a unified address space for data and instruction, starting from 0? (So essentially remove the ROM/RAM masking.)
Changed to 0
>     * Can we go from a 2p memory to a 1p memory? The core won't issue a data and instruction operation at the same time anyways, so we can go simpler here (in line with having that a minimal example). You need to have a small arbitration block between the two interfaces (data/instruction) in front of the SRAM (essentially a mini-bus). (Please ping me on Zulip if that paragraph doesn't make sense to you.)
Using 1p memory now
>     * Optional: 57 MHz looks rather strange. So I'd go back to 50 MHz if we cannot go higher to a number that looks "normal". (But I leave that up to you, changing things around all the time isn't the greatest use of time.)
I only put in 57 MHz to show the highest one possible for now, but I agree, 50 MHz is nicer, so used this now again.

jrrk:
> The core won't issue a data and instruction operation at the same time anyways, so we can go simpler here (in line with having that a minimal example). 

@imphil This seems to be a convenient but surprising limitation. Why go for two buses at all at the top level of the Ibex if the two buses are never used simultaneously ?

examples/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
imphil: As in the tracer example: please use fusesoc from the source root, and put this gitignore there as well.

doc/examples.rst
```diff
@@ -0,0 +1,12 @@
...
+Examples
+========
+
+To make use of Ibex it has to be integrated as described in :ref:`core-integration`.
+Here is an example of how this could look like.
+
+FPGA
+----
+
+A minimal example for the `Arty A7 <https://reference.digilentinc.com/reference/programmable-logic/arty-a7/start>`_ FPGA Development board is provided. Find the description of how to use this in ``examples/fpga/artya7-100/README.md``.
```
imphil: I'd actually prefer having the documentation in the manual and link to it from the README, i.e. the other way around.

examples/fpga/artya7-100/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This is a simple example of how Ibex can be utilized.
+The core is not connected to a bus, but to the SRAM directly, which is initialized
+by the `led.vmem` file from the software section.
+
+## Requirements
+
+### Tools
+
+  - RISC-V compiler
```
imphil: A RV32 compiler

examples/fpga/artya7-100/README.md
```diff
@@ -0,0 +1,46 @@
...
+  - Xilinx Vivado
+
+## Build
+
+### Software
+
+First the software must be built. Go into `sw/led` and call:
+
+```
+make CC=/path/to/RISC-V-compiler
```
imphil: Can you make this `/path/to/riscv32-unknown-elf-gcc` to make it easier to find the right compiler for people?

examples/fpga/artya7-100/README.md
```diff
@@ -0,0 +1,46 @@
...
+This will create a directory `build` which contains the output files, including
+the bitstream.
+
+## Program
+
+Open the Vivado project:
+`build/lowrisc_ibex_top_artya7_100_0.1/synth-vivado/lowrisc_ibex_top_artya7_100_0.1.xpr`
+
+And use the Hardware Manager to program the Arty board:
+`build/lowrisc_ibex_top_artya7_100_0.1/synth-vivado/lowrisc_ibex_top_artya7_100_0.1.bit`
```
imphil: `fusesoc pgm` should do the trick as well

examples/fpga/artya7-100/rtl/top_artya7_100.sv
```diff
@@ -0,0 +1,154 @@
...
+    output [3:0]        LED
+);
+
+  parameter int          MEM_SIZE  = 64 * 1024; // 64 kB
+  parameter logic [31:0] MEM_START = 32'h00000000;
+  parameter logic [31:0] MEM_MASK  = MEM_SIZE-1;
+
+  logic clk_sys, rst_sys_n;
+
+  // Instruction connection so SRAM
```
imphil: so -> to

examples/fpga/artya7-100/rtl/top_artya7_100.sv
```diff
@@ -0,0 +1,154 @@
...
+  logic        instr_req;
+  logic        instr_gnt;
+  logic        instr_rvalid;
+  logic [31:0] instr_addr;
+  logic [31:0] instr_rdata;
+
+  // Data connection so SRAM
```
imphil: so -> to

examples/fpga/artya7-100/rtl/top_artya7_100.sv
```diff
@@ -0,0 +1,154 @@
...
+  logic        mem_req;
+  logic        mem_write;
+  logic [31:0] mem_wdata;
+  logic        mem_rvalid;
+  logic [31:0] mem_rdata;
+
+
+  ibex_core #(
+     .RV32E(0),
+     .RV32M(1),
```
imphil: Don't set these two parameters, it's good to test the default configuration (whatever that might be if we change things)

examples/fpga/artya7-100/rtl/top_artya7_100.sv
```diff
@@ -0,0 +1,154 @@
...
+  end
+  assign LED = leds;
+
+  // Clock and reset
+  clkgen_xil7series
+    clkgen(
+      .IO_CLK       (IO_CLK),
+      .IO_RST_N     (IO_RST_N),
+      .clk_sys      (clk_sys),
+      .rst_sys_n    (rst_sys_n)
```
imphil: If you prefer, you can simplify this to:

```

  clkgen_xil7series
    clkgen(
      .IO_CLK
      .IO_RST_N 
      .clk_sys
      .rst_sys_n
  );
```

examples/sw/led/Makefile
```diff
@@ -0,0 +1,59 @@
...
+$(PROGRAM).elf: $(OBJS) $(LINKER_SCRIPT)
+	$(CC) $(CFLAGS) -T $(LINKER_SCRIPT) $(OBJS) -o $@ $(LIBS)
+
+%.dis: %.elf
+	$(OBJDUMP) -SD $^ > $@
+
+# Note: this target requires the srecord package to be installed.
+# XXX: This could be replaced by objcopy once
+# https://sourceware.org/bugzilla/show_bug.cgi?id=19921
+# is merged.
```
imphil: This bug is already merged now, but I'm not sure if the implementation has been cleaned up enough to be usable. And it's not really available in a standard toolchain yet, so slighly reword this to read "is merged" -> "is widely available"

examples/fpga/artya7-100/README.md
asb: That can also cause confusion unfortunately - multilib builds will use riscv64-unknown-elf-gcc but may include libs for ilp32.

examples/fpga/artya7-100/rtl/top_artya7_100.sv
towoe: done

examples/fpga/artya7-100/rtl/top_artya7_100.sv
towoe: done

examples/fpga/artya7-100/rtl/top_artya7_100.sv
towoe: done

examples/fpga/artya7-100/rtl/top_artya7_100.sv
towoe: Yes, this looks better. Done.

examples/sw/led/Makefile
towoe: Good to know. I guess this would also simplify `link.ld` in regards to the alignment of the sections.

examples/fpga/artya7-100/README.md
towoe: So `/path/to/riscvXX-unknown-elf-gcc`?

towoe:
I will make the changes to the documentation after the tracer PR is merged as this simplifies the rebase.

examples/fpga/artya7-100/README.md
imphil: Well, even the *unknown* part is not really standard (unfortunately), with the rather commonly used [Eclipse prebuilts](https://gnu-mcu-eclipse.github.io/toolchain/riscv/install/) defaulting to *none* here. I'd go for either riscv32-unknown-elf-gcc (which is what you get with a SiFive toolchain source build) or just leave it as you had it.

doc/examples.rst
towoe: So the description should move to doc, but the README will still hold the build instruction?

examples/fpga/artya7-100/README.md
towoe: I just checked the Eclipse site, looks like they now use `riscv-none-embed-gcc`.
I will just put in some examples of what this path could look like. I would assume that the chances are high that if someone finds the way to Ibex Arty example, they are able to set the path to the right compiler correctly.

examples/fpga/artya7-100/README.md
towoe: Thanks for that, much better to program this way.

imphil:
Looks like this is ready to be merged now?

One more thing: This currently doesn't support byte reads/writes. That's fine for now, but we should at least add a TODO in the code and file an issue for it.

towoe:
I will add a TODO regarding the byte access.
My last push was already rebased on https://github.com/lowRISC/ibex/pull/77 so the commits are appearing here as well now, so we should wait until tracer is merged so I can rebase on the master branch for better attribution.

imphil:
This should be ready after a rebase on top of master now that the tracer has gone in.

imphil:
Thanks Tobias!

PR #82, opened 2019-06-20
Align data and instruction address output

vogelpi:
This PR adds two commits to ensure the data and instruction address output by the core is always word aligned.

The core handles unaligned accesses by issuing two separate, aligned accesses. However, without these commits, the core still outputs unaligned addresses for the two accesses and relies on the memory to ignore the address LSBs.

Tested with Verilator and Vivado.

### DIFF
--- a/rtl/ibex_prefetch_buffer.sv
+++ b/rtl/ibex_prefetch_buffer.sv
@@ -54,6 +54,7 @@ module ibex_prefetch_buffer (
   pf_fsm_e pf_fsm_cs, pf_fsm_ns;
 
   logic [31:0] instr_addr_q, fetch_addr;
+  logic [31:0] instr_addr, instr_addr_w_aligned;
   logic        addr_valid;
 
   logic        fifo_valid;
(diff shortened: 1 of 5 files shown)


rtl/ibex_prefetch_buffer.sv
```diff
@@ -217,9 +218,15 @@ module ibex_prefetch_buffer (
...
-        instr_addr_q <= instr_addr_o;
+        instr_addr_q <= instr_addr_int;
       end
     end
   end
 
+  /////////////////
+  // Output Addr //
+  /////////////////
+  assign instr_addr_aligned = {instr_addr_int[31:2], 2'b00};
```
imphil: `aligned` is pretty generic (word or half-word aligned, for example?). How about `instr_addr_wordaligned` or `instr_addr_aligned_32b`?

rtl/ibex_prefetch_buffer.sv
```diff
@@ -54,6 +54,7 @@ module ibex_prefetch_buffer (
   pf_fsm_e pf_fsm_cs, pf_fsm_ns;
 
   logic [31:0] instr_addr_q, fetch_addr;
+  logic [31:0] instr_addr_int, instr_addr_aligned;
```
imphil: I'm not a huge fan of the `_int` naming, since the aligned signal is actually a version of the int signal; therefore I'd expect the aligned signal have the same base name as the int name with a suffix (such as aligned). 

How about `instr_addr` and `instr_addr_aligned`? (I know the `data_addr_int` has already been there, how about renaming for consistency? And see below for a discussion on the `aligned` suffix.)

rtl/ibex_prefetch_buffer.sv
vogelpi: Done.

rtl/ibex_prefetch_buffer.sv
vogelpi: Done

vogelpi:
Thanks for your feedback @imphil . I implemented the requested changes, updated the documentation accordingly, and added the requested assertions.

PR #83, opened 2019-06-21
Fix perf counters

### LINKED ISSUE #80: Number of performance counters
the parameter MHPMCounterNum does not seem to influence the number of performance counters. I synthesized the core once with MHPMCounterNum=1 and then with MHPMCounterNum=8 and it results in almost exactly the same area.

by looking at the code I also don't see how this parameter is changing the number of physically present counters/flipflops. I'm pretty sure there are 32 64b registers and 32 64b adders no matter what the parameters are.

I think MHPMCounterWidth has the same problem. it probably only counts to the desired width, but the registers [63:MHPMCounteWidth] cannot be optimized away.

the problem with the width can be easily fixed by changing:
```
csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][63:32];
```
to
```
csr_rdata_int[MHPMCounterWidth-1:0] = mhpmcounter_q[mhpmcounter_idx][MHPMCounterWidth-1:32];
csr_rdata_int[32-1:MHPMCounterWidth-32] = '0;
```

if mhpmcounter_q[63:MHPMCounterWidth] is never accessed, the synthesizer can optimize it away. otherwise these registers are there but cannot really be used.

vogelpi:
This PR solves two problems in the performance counters:
1. Fix increment of performance counters
2. Make sure performance counters are not accidentally optimized away during synthesis. Only those are optimized away, which are disabled.

It resolves #80.

The changes have been tested using Verilator and Vivado.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -579,7 +579,7 @@ module ibex_cs_registers #(
 
       // increment
       if (mhpmcounter_incr[i] & ~mcountinhibit[i]) begin
-        mhpmcounter_n[i] = mhpmcounter_mask[i] & (mhpmcounter_n[i] + 64'h1);
+        mhpmcounter_n[i] = mhpmcounter_mask[i] & (mhpmcounter_q[i] + 64'h1);
       end
 
       // write
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -186,10 +186,8 @@ module ibex_cs_registers #(
   /////////////
 
   logic [$bits(csr_num_e)-1:0] csr_addr;
-  logic [4:0]                  unused_csr_addr;
   assign csr_addr           = {csr_addr_i};
   assign mhpmcounter_idx    = csr_addr[4:0];
-  assign unused_csr_addr    = csr_addr[9:5];
 
   assign illegal_csr_priv   = 1'b0; // we only support M-mode
   assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_we_int;
@@ -242,7 +240,7 @@ module ibex_cs_registers #(
       CSR_MINSTRETH:     csr_rdata_int = mhpmcounter_q[2][63:32];
 
       default: begin
-        if (csr_addr_i == CSR_MCOUNTER_SETUP_MASK) begin
+        if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTER_SETUP) begin
           csr_rdata_int = mhpmevent[mhpmcounter_idx];
           // check access to non-existent or already covered CSRs
           if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCOUNTINHIBIT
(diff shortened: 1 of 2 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -515,35 +513,41 @@ module ibex_cs_registers #(
...
     // activate all
     for (int i=0; i<32; i++) begin : gen_mhpmevent_active
+      mhpmevent[i]    =   '0;
       mhpmevent[i][i] = 1'b1;
     end
 
     // deactivate
     mhpmevent[1] = '0; // not existing, reserved
-    for (int unsigned i=3+MHPMCounterNum; i<32; i++) begin : gen_mhpmevent_inactive
+    for (int i=3+$signed(MHPMCounterNum); i<32; i++) begin : gen_mhpmevent_inactive
```
imphil: Why is this change? How is int unsigned not working? I'm really not a fan of such ugly constructs for the sake of making lint happy. It's better in this case to change the lint waivers, and not give up code readability.

rtl/ibex_defines.sv
```diff
@@ -240,15 +240,18 @@ typedef enum logic[11:0] {
...
+  CSR_MINSTRET           = 12'hB02,
+  CSR_MINSTRETH          = 12'hB82
 } csr_num_e;
 
+typedef enum logic [11:0] {
+  CSR_OFF_MCOUNTER_SETUP = 12'h320, // mcounter_setup @ 12'h323 - 12'h33F
+  CSR_OFF_MCOUNTER       = 12'hB00, // mcounter       @ 12'hB03 - 12'hB1F
+  CSR_OFF_MCOUNTERH      = 12'hB80, // mcounterh      @ 12'hB83 - 12'hB9F
+  CSR_MASK_MCOUNTER      = 12'hFE0
+} csr_mcounter_e;
```
imphil: How about making these just a normal parameter? It's not really an enumeration of values, but more a grouping of related (but not identical) things. This would also make the code using these signals easier as you don't need to add the curly braces around each name (e.g. you can write `if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTERH) begin` instead of `if ((csr_addr & {CSR_MASK_MCOUNTER}) == {CSR_OFF_MCOUNTERH}) begin`).

rtl/ibex_cs_registers.sv
vogelpi: Thanks for your feedback. I made this change based on your feedback for a previous PR #74 and your point that `int unsigned` is not supported by yosys. Thinking about it and a comment from Nils made me conclude that doing the type cast is more clean than switching to unsigned loop variables.

rtl/ibex_defines.sv
vogelpi: Thanks for your feedback. I implemented the requested change.

rtl/ibex_cs_registers.sv
imphil: This code is a almost trivial, textbook-style for loop. Having a strange looking cast in there really hurts readability, so it needs to go away. 

Options we have:
- use `int unsigned`. Yosys is currently not a requirement, it has larger problems in the parser than this one.
- Make `MHPMCounterNum` `int` instead of `int unsigned`
- Waive this lint warning

I'll leave it up to you to choose an option.

vogelpi: Ok, I went for `int unsigned`.

PR #85, opened 2019-06-24
Disable performance counters by default

imphil:
Performance counters are an optional feature. Disable them by default to
avoid users having them enabled unknowingly and paying the (rather
large) area price for it.

### DIFF
--- a/doc/performance_counters.rst
+++ b/doc/performance_counters.rst
@@ -7,6 +7,8 @@ Ibex implements performance counters according to the RISC-V Privileged Specific
 The performance counters are placed inside the Control and Status Registers (CSRs) and can be accessed with the ``CSRRW(I)`` and ``CSRRS/C(I)`` instructions.
 
 Ibex implements the clock cycle counter ``mcycle(h)``, the retired instruction counter ``minstret(h)``, as well as the 29 event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)`` and the corresponding event selector CSRs ``mhpmevent3`` - ``mhpmevent31``, and the ``mcountinhibit`` CSR to individually enable/disable the counters.
+``mcycle(h)`` and ``minstret(h)`` are always available and 64 bit wide.
+The ``mhpmcounter`` performance counters are optional (unavailable by default) and parametrizable in width.
 
 Event Selector
 --------------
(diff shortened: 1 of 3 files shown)


imphil:
Good catch @vogelpi! I've updated the documentation accordingly.

On the motivation for this change: I was reminded again in Zurich how important defaults are. Since Ibex is meant to be a "small RV32I" core we should configure it by default in a way that lives up to this promise.

doc/performance_counters.rst
```diff
@@ -54,16 +56,22 @@ In particular, to enable/disable ``mcycle(h)``, bit 0 must be written. For ``min
...
-The ``mcycle(h)`` and ``minstret(h)`` counters are 64 bit wide.
-The bit width of the event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)`` can be controlled via the ``WidthMHPMCounters`` parameter.
+The event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)`` are parametrizable.
+Their width can be parametrized between 32 and 64 bit through the ``WidthMHPMCounters`` parameter, which defaults to 40 bit wide counters.
 
 The effective number of available event counters ``mhpmcounterX(h)`` can be controlled via the ``NumMHPMCounters`` parameter.
-By default, only the first 8 counters ``mhpmcounter3(h)`` - ``mhpmcounter10(h)`` are available.
-The association of events with these counters is hardwired as listed in the following table.
-The remaining counters are disabled and tied to 0.
+By default, no counters are enabled.
```
vogelpi: I would clearly differentiate between available counters and enabled/disabled counters here. Enabled means the counter is counting whereas available means the counter can be enabled/disabled. In the spec, an enabled counter means a counter that is effectively counting up. Enabling/disabling counters it is done in software by writes to the `mcountinhibit` and `mevent3` - `mevent31` CSRs.

doc/performance_counters.rst
```diff
@@ -54,16 +56,22 @@ In particular, to enable/disable ``mcycle(h)``, bit 0 must be written. For ``min
...
---------------
+Parametrization at synthesis time
+---------------------------------
+
+The ``mcycle(h)`` and ``minstret(h)`` counters are unconditionally enabled and 64 bit wide.
 
-The ``mcycle(h)`` and ``minstret(h)`` counters are 64 bit wide.
-The bit width of the event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)`` can be controlled via the ``WidthMHPMCounters`` parameter.
+The event counters ``mhpmcounter3(h)`` - ``mhpmcounter31(h)`` are parametrizable.
+Their width can be parametrized between 32 and 64 bit through the ``WidthMHPMCounters`` parameter, which defaults to 40 bit wide counters.
```
vogelpi: Actually, also widths smaller than 32 bit are supported in the RTL.

imphil: Ah ok, then `integration.rst` had it wrong (fixed that), and we probably had this limitation in a previous RTL version. (I should double-check more and rely less on my memory ...)

doc/performance_counters.rst
imphil: good point, fixed.

asb:
> I agree that performance counters are optional and disabling them saves roughly 15% slices when targeting FPGAs. However, if you really want to disable them by default, you must at least adapt the documentation accordingly. And even then, one can discuss if it will not confuse people working on software if even the performance counters connected in the RTL (at the first glance) always read as 0.

Just to chime in on this one. I think `mhpmcounter*` being disabled/not present by default is sensible. Disabling mcycle and minstret by default would be more surprising and cause more confusion for the software world. Software should be able to determine that counters aren't present by observing mhpmevent is 0.

PR #92, opened 2019-06-25
Expand word widths to keep Verilator lint checks happy

jrrk:
Verilator bombs out by default unless all the widths match (and -Wno-width is not specified). This patch expands widths where necessary to overcome this problem. However it may not be the most elegant solution.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -97,15 +97,15 @@ module ibex_cs_registers #(
       (0     <<  0)  // A - Atomic Instructions extension
     | (1     <<  2)  // C - Compressed extension
     | (0     <<  3)  // D - Double precision floating-point extension
-    | (RV32E <<  4)  // E - RV32E base ISA
+    | {27'b0,RV32E,4'b0}  // E - RV32E base ISA
     | (0     <<  5)  // F - Single precision floating-point extension
     | (1     <<  8)  // I - RV32I/64I/128I base ISA
-    | (RV32M << 12)  // M - Integer Multiply/Divide extension
+    | {19'b0,RV32M,12'b0}  // M - Integer Multiply/Divide extension
     | (0     << 13)  // N - User level interrupts supported
     | (0     << 18)  // S - Supervisor mode implemented
     | (0     << 20)  // U - User mode implemented
     | (0     << 23)  // X - Non-standard extensions present
-    | (MXL   << 30); // M-XLEN
+    | {MXL,30'b0}; // M-XLEN
 
   `define MSTATUS_UIE_BITS        0
   `define MSTATUS_SIE_BITS        1
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -93,20 +93,37 @@ module ibex_cs_registers #(
 
   // misa
   localparam logic [1:0] MXL = 2'd1; // M-XLEN: XLEN in M-Mode for RV32
+  localparam logic MISA_X = 0,
+  MISA_W = 0,
+  MISA_V = 0,
+  MISA_U = 0,
+  MISA_T = 0,
+  MISA_S = 0,
+  MISA_R = 0,
+  MISA_Q = 0,
+  MISA_P = 0,
+  MISA_O = 0,
+  MISA_N = 0,
+  MISA_L = 0,
+  MISA_K = 0,
+  MISA_J = 0,
+  MISA_I = 0,
+  MISA_H = 0,
+  MISA_G = 0,
+  MISA_F = 0,
+  MISA_D = 0,
+  MISA_C = 1,
+  MISA_B = 0,
+  MISA_A = 0;
   localparam logic [31:0] MISA_VALUE =
-      (0     <<  0)  // A - Atomic Instructions extension
-    | (1     <<  2)  // C - Compressed extension
-    | (0     <<  3)  // D - Double precision floating-point extension
-    | {27'b0,RV32E,4'b0}  // E - RV32E base ISA
-    | (0     <<  5)  // F - Single precision floating-point extension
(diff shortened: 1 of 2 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -97,15 +97,15 @@ module ibex_cs_registers #(
...
     | (0     <<  5)  // F - Single precision floating-point extension
     | (1     <<  8)  // I - RV32I/64I/128I base ISA
-    | (RV32M << 12)  // M - Integer Multiply/Divide extension
+    | {19'b0,RV32M,12'b0}  // M - Integer Multiply/Divide extension
     | (0     << 13)  // N - User level interrupts supported
     | (0     << 18)  // S - Supervisor mode implemented
     | (0     << 20)  // U - User mode implemented
     | (0     << 23)  // X - Non-standard extensions present
-    | (MXL   << 30); // M-XLEN
+    | {MXL,30'b0}; // M-XLEN
```
imphil: The Verilator warnings here could to be a bug in Verilator lint. The SV spec is rather hard to read in this part, but that's my understanding:

According to the SV spec (SV2018, page 283, "Table 11-21—Bit lengths resulting from self-determined expressions") the whole expression is as large as its largest component. 

image

The first part of the expression `(0 << 0)` is 32 bit wide (a shifted integer), hence all the other parts are 32 bit wide as well.

What I don't really get is how this works together with the shift of the bit and logic values you modified. Looking at the table here:
image
The shift bit width is given by the left operand, which is `bit` (for `RV32*`) or `logic [1:0]` (for `MXL`). This would make the shift expression 1 or 2 bit wide, making the subexpression actually 0. That cannot really be right, and it's also not what we're seeing in other tools.
So far for the confusion part.

AscentLint doesn't complain about this code.

So I'm proposing to do the following:
- Please look at the SV spec to check if you agree with my reading of the spec.
- If you feel this could be a bug in Verilator, please file a bug report and confirm with Wilson.
- Waive this lint warning in Verilator (preferred), or work around it. 
   I'm not a big fan of the workaround you propose, as it destroys the symmetry we have in the code. How about `((RV32E + 0) <<  4)` instead to force `RV32E` from `bit` to `int`? If you do so, please add a comment as to why this was done and add a link to the Verilator bug report so we can remove the workaround once it has been fixed upstream.

rtl/ibex_register_file_ff.sv
```diff
@@ -60,7 +60,7 @@ module ibex_register_file #(
 
   always_comb begin : we_a_decoder
     for (int i = 1; i < NUM_WORDS; i++) begin
-      we_a_dec[i] = (waddr_a_i == i) ?  we_a_i : 1'b0;
+      we_a_dec[i] = ({27'b0,waddr_a_i} == i) ?  we_a_i : 1'b0;
```
imphil: `waddr_a_i` is `DataWidth` bit, you cannot hardcode it to 28 here. Can you make `i` of `logic [DataWidth-1:0]` or cast `i` to be `DataWidth` bit wide.

imphil:
Thanks for this PR Jonathan! Getting Ibex Verilator-lint clean is an important goal, thanks for looking into it. However, we need to be careful to fix lint warnings only if they are really a problem in our code, and not only to work around tool issues. I've added some comments, could you please have a look at them?

rtl/ibex_register_file_ff.sv
jrrk: In the checked in copy waddr_a_i is [4:0]. It would make no sense to relate it to data width.

imphil: Uh sorry, I mixed up waddr and wdata. The suggestion holds nonetheless: either use a width cast, or simply do `waddr_a_i + 0` to cast to int

rtl/ibex_cs_registers.sv
jrrk: I don't want to get into a discussion about what looks nicer. To my mind one concatenation
of the entire 32-bit field would be the nicest approach. As a hacker I am not really in a position to launch into a polemic on how things should be done.

jrrk: The workaround you suggest would just move the warning message from the | operation to the + operation, and is surely not an increase in legibility

imphil: > I don't want to get into a discussion about what looks nicer.

It's not about nicer. It's about consistency (for this piece of code, which consistently uses shifts right now), and about (not) working around possible Verilator lint bugs.

> To my mind one concatenation of the entire 32-bit field would be the nicest approach.

Why not do that then? If it gets us around the Verilator lint issue, and creates consistent code for the assignment, then that sounds like a good solution.

rtl/ibex_register_file_ff.sv
```diff
@@ -52,15 +52,15 @@ module ibex_register_file #(
...
 
   logic [NUM_WORDS-1:0][DataWidth-1:0] rf_reg;
   logic [NUM_WORDS-1:1][DataWidth-1:0] rf_reg_tmp;
   logic [NUM_WORDS-1:1]                we_a_dec;
 
   always_comb begin : we_a_decoder
-    for (int i = 1; i < NUM_WORDS; i++) begin
-      we_a_dec[i] = (waddr_a_i == i) ?  we_a_i : 1'b0;
+    for (logic [ADDR_WIDTH:0] i = 1; i < NUM_WORDS; i++) begin
+      we_a_dec[i] = {1'b0,waddr_a_i} == i ?  we_a_i : 1'b0;
```
imphil: Doesn't `ADDR_WIDTH` depend on the RV32* parameter, while `waddr_a_i` is always 5 bit? Please double-check with RV32E enabled. Could casting `i` work, e.g. `i`, e.g. `waddr_a_i == 5'(i)` (untested)?

imphil:
We took #101 instead. Thanks @jrrk!

PR #101, opened 2019-06-27
Avoid WIDTH lint warnings in Verilator

imphil:
Verilator displays the following lint warnings:

```
%Warning-WIDTH: ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_cs_registers.sv:100: Operator SHIFTL expects 32 bits on the LHS, but LHS's VARREF 'RV32E' generates 1 bits.
                ... Use "/* verilator lint_off WIDTH */" and lint_on around source to disable this message.
%Warning-WIDTH: ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_cs_registers.sv:103: Operator SHIFTL expects 32 bits on the LHS, but LHS's VARREF 'RV32M' generates 1 bits.
%Warning-WIDTH: ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_cs_registers.sv:108: Operator SHIFTL expects 32 bits on the LHS, but LHS's VARREF 'MXL' generates 2 bits.
%Warning-WIDTH: ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_register_file_ff.sv:63: Operator EQ expects 32 bits on the LHS, but LHS's VARREF 'waddr_a_i' generates 5 bits.
```

It's not quite clear (to me) from reading the SV spec if this is a bug
in Verilator lint, or if this is actually a code bug.

Alternative proposal for #92

### DIFF
--- a/rtl/ibex_register_file_ff.sv
+++ b/rtl/ibex_register_file_ff.sv
@@ -60,7 +60,7 @@ module ibex_register_file #(
 
   always_comb begin : we_a_decoder
     for (int i = 1; i < NUM_WORDS; i++) begin
-      we_a_dec[i] = (waddr_a_i == i) ?  we_a_i : 1'b0;
+      we_a_dec[i] = (waddr_a_i == 5'(i)) ?  we_a_i : 1'b0;
     end
   end
 
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -94,18 +94,18 @@ module ibex_cs_registers #(
   // misa
   localparam logic [1:0] MXL = 2'd1; // M-XLEN: XLEN in M-Mode for RV32
   localparam logic [31:0] MISA_VALUE =
-      (0     <<  0)  // A - Atomic Instructions extension
-    | (1     <<  2)  // C - Compressed extension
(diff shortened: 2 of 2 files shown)


asb:
For what it's worth, I quite like this fix.

imphil:
@jrrk would you be OK with this fix?

imphil:
Thanks @jrrk. We have this syntax actually in a couple other places, I'm pretty sure it goes through Vivado. I don't have access to DC at the moment; if we run into trouble there we need to revisit.

rtl/ibex_register_file_ff.sv
```diff
@@ -60,7 +60,7 @@ module ibex_register_file #(
 
   always_comb begin : we_a_decoder
     for (int i = 1; i < NUM_WORDS; i++) begin
-      we_a_dec[i] = (waddr_a_i == i) ?  we_a_i : 1'b0;
+      we_a_dec[i] = (waddr_a_i == 5'(i)) ?  we_a_i : 1'b0;
```
sjgitty: Are all tools fine with this notation? I don't know of any that wouldn't be, but haven't seen it widely used.

Issue #97, opened 2019-06-26
Implement the mscratch CSR


imphil:
The title says it all.

Issue #70, opened 2019-06-07
Implement trap handling according to latest RISC-V spec


davideschiavone:
Hi guys (@vogelpi , @imphil )

according to the Priv ISA Spec of RISC-V 1.10, you should handle the exceptions by always jumping to mtvec rather than vectoring them as it was in the past

Look at the section
3.1.12 Machine Trap-Vector Base-Address Register (mtvec)
of this document please: https://riscv.org/specifications/privileged-isa/

I have already done it in RISCY.
So now in the controller everytime you get an exception you set exc_pc_mux_o to exceptions:
https://github.com/pulp-platform/riscv/blob/master/rtl/riscv_controller.sv#L731

and in the IF stage you always jump to the same address regardless the exception cause:
https://github.com/pulp-platform/riscv/blob/master/rtl/riscv_if_stage.sv#L141

asb:
Thanks for the note Davide - I believe @vogelpi is already working on this. We should have set up a tracking issue in this repo for that work. This issue can now fulfill that role!

taoliug:
Does this apply to interrupt handling as well?

vogelpi:
Thanks @davideschiavone , I am working on this as mentioned by @asb . Thanks for the pointers to RI5CY. This will be helpful.

@taoliug : For interrupts we have two options. Either we do it as for the exceptions (direct mode). Or we use vectored mode (similar to what we do today). The plan is to discuss this also next week in Zurich.

davideschiavone:
For what concerns interrupts, RI5CY is vectorized by default, so there is one field in the mtvec hardwired at VECTORIZED interrupt FYI.

taoliug:
The base address of the vectorized interrupt should come from MTVEC, right now it's hardwired to 0? Here's what's said in the spec:

> When MODE=Vectored,all synchronous exceptions into machine mode cause the pc to be set to the address in the BASE field, whereas interrupts cause the pc to be set to the address in the BASE field plus four times interrupt cause number.

vogelpi:
The changes discussed here were implemented and merged in PR #104 .
Regarding `mtvec`: It can be modified by changing the boot address at runtime.

However, this will change in the future. We will at some point make `mtvec` writable by software. This will allow to change it independently of the boot address. This work is tracked in issue #118

PR #104, opened 2019-06-27
Spec compliant trap handling

vogelpi:
This PR mainly modifies the trap handling infrastructure to comply with the RISC-V privileged spec, draft versions 1.11. To this end, additional CSRs were added.

All exceptions now jump to the base address of the trap-vector table. Interrupts are handled in vectored mode and jump to the base address + 4*irq_id. The exception cause is stored in `mcause` and `mtval` has been added to hold additional information (illegal instruction for illegal instruction exception, address of faulting transaction part for LSU errors).

To implement these changes the AGU mux for misaligned loads/stores has been reworked.
In addition, the `mscratch` CSR has been implemented.

This PR resolves: #97 and #70.

### DIFF
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -79,10 +79,10 @@ Reset Value: ``0x0000_1800``
 | 3     | RW  | **Interrupt Enable (MIE):** If set to 1'b1, interrupts are globally enabled.    |
 +-------+-----+---------------------------------------------------------------------------------+
 
-When an exception is encountered, MPIE will be set to MIE.
-When the MRET instruction is executed, the value of MPIE will be stored back to IE.
+When an exception is encountered, ``mstatus``.MPIE will be set to ``mstatus``.MIE.
+When the MRET instruction is executed, the value of MPIE will be stored back to ``mstatus``.MIE.
 
-If you want to enable interrupt handling in your exception handler, set MIE to 1'b1 inside your handler code.
+If you want to enable interrupt handling in your exception handler, set ``mstatus``.MIE to 1'b1 inside your handler code.
 
 
 Machine Trap-Vector Base Address (mtvec)
@@ -92,7 +92,7 @@ CSR Address: ``0x305``
 
 When an exception is encountered, the core jumps to the corresponding handler using the content of ``mtvec`` as base address.
 It is a read-only register which contains the boot address.
-Its LSBs (mode field) are set to 2'b01 to indicate vectored interrupt handling.
+``mtvec``.MODE is set to 2'b01 to indicate vectored interrupt handling.
 
 
 Machine Exception PC (mepc)
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -22,9 +22,11 @@ Ibex implements all the Control and Status Registers (CSRs) listed in the follow
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x33F  | ``mhpmevent31``    | WARL   | Machine performance-monitoring event selector |
 +---------+--------------------+--------+-----------------------------------------------+
-|  0x341  | ``mepc``           | RW     | Machine Exception Program Counter             |
+|  0x340  | ``mscratch``       | RW     | Machine Scratch Register                      |
 +---------+--------------------+--------+-----------------------------------------------+
-|  0x342  | ``mcause``         | RW     | Machine Trap Cause                            |
+|  0x341  | ``mepc``           | WARL   | Machine Exception Program Counter             |
++---------+--------------------+--------+-----------------------------------------------+
+|  0x342  | ``mcause``         | WLRL   | Machine Cause Register                        |
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x343  | ``mtval``          | WARL   | Machine Trap Value Register                   |
 +---------+--------------------+--------+-----------------------------------------------+
@@ -78,7 +80,7 @@ Reset Value: ``0x0000_1800``
 +-------+-----+---------------------------------------------------------------------------------+
 
 When an exception is encountered, MPIE will be set to MIE.
-When the ``mret`` instruction is executed, the value of MPIE will be stored back to IE.
+When the MRET instruction is executed, the value of MPIE will be stored back to IE.
 
 If you want to enable interrupt handling in your exception handler, set MIE to 1'b1 inside your handler code.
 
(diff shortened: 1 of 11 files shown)


rtl/ibex_controller.sv
```diff
@@ -45,8 +45,8 @@ module ibex_controller (
     input  logic                      csr_status_i,          // decoder has CSR status instr
 
     // from IF/ID pipeline
-    input  logic                      instr_valid_i,         // instruction coming from IF/ID stage
-                                                             // is valid
+    input  logic [31:0]               instr_i,               // instr from IF/ID stage, for mtval
```
asb: This must (according to the spec) hold the compressed form of the instruction if it was compressed. I might be tracing the logic incorrectly, but looking at ibex_compressed_decoder.sv and ibex_if_stage.sv I can't convince myself that's always true. Can you double check?

vogelpi: Thanks @asb ! When I read the spec initially I also had the same understanding. During the implementation, I just did not find anymore the term compressed in the mtval section and thus took the decompressed. But the spec clearly says the actual faulting instruction. I have changed it now.

doc/cs_registers.rst
```diff
@@ -76,7 +80,7 @@ Reset Value: ``0x0000_1800``
 +-------+-----+---------------------------------------------------------------------------------+
 
 When an exception is encountered, MPIE will be set to MIE.
-When the ``mret`` instruction is executed, the value of MPIE will be stored back to IE.
+When the MRET instruction is executed, the value of MPIE will be stored back to IE.
```
imphil: Should that read `stored back to MIE`?

Beyond that, MIE vs mie is confusing in the RISC-V spec: MIE is a field in mstatus, while mie is a register. To avoid confusion, can we use mstatus.MPIE and mstatus.MIE here (I never understood why the spec uses all-caps for fields, but lowercase for registers, but hey ...)

doc/cs_registers.rst
```diff
@@ -88,6 +92,7 @@ CSR Address: ``0x305``
 
 When an exception is encountered, the core jumps to the corresponding handler using the content of ``mtvec`` as base address.
 It is a read-only register which contains the boot address.
+Its LSBs (mode field) are set to 2'b01 to indicate vectored interrupt handling.
```
imphil: Its LSBs (mode field) -> mtvec.MODE

doc/cs_registers.rst
```diff
@@ -117,6 +122,23 @@ Reset Value: ``0x0000_0000``
...
+--------------------------
+
+CSR Address: ``0x343``
+
+Reset Value: ``0x0000_0000``
+
+When an exception is encountered, this register can hold exception-specific information to assist software in handling the trap.
+In the case of errors in the load-store unit ``mtval`` holds the address of the transaction causing the error.
+If this transaction is misaligned, ``mtval`` holds the address of the missing transaction part.
+In the case of illegal instruction exceptions, ``mtval`` holds the faulting instruction.
```
imphil: How about making this an unordered list?

doc/exception_interrupts.rst
```diff
@@ -3,28 +3,17 @@
 Exceptions and Interrupts
 =========================
 
-Ibex currently implements a free-form vectored trap handler mechanism for interrupts and exceptions.
-The base address of the interrupt vector table is given by the boot address (must be aligned to 256 bytes, i.e., its least significant byte must be 0x00).
+Ibex currently implements trap handling for interrupts and exceptions according to the RISC-V Privileged Specification, draft version 1.11.
```
imphil: It's actually ratified now

doc/exception_interrupts.rst
```diff
@@ -3,28 +3,17 @@
...
+
+All exceptions cause the core to jump to the base address of the vector table in the ``mtvec`` CSR.
+Interrupts are handled in vectored mode, i.e., the core jumps to the base address plus four times the interrupt cause number.
```
imphil: i.e., -> i.e. (no comma)

rtl/ibex_if_stage.sv
```diff
@@ -50,12 +50,11 @@ module ibex_if_stage #(
     output logic                      instr_valid_id_o,      // instr in IF/ID is valid
     output logic [31:0]               instr_rdata_id_o,      // read instr is sampled and sent
                                                              // to ID stage for decoding
+    output logic [15:0]               instr_rdata_c_id_o,    // compressed instr sent to ID stage
+                                                             // for mtval, meaningful only
+                                                             // if is_compressed_id_o = 1'b1
     output logic                      is_compressed_id_o,    // compressed decoder thinks this is
```
imphil: Could you rename this signal to be in line with the naming in the controller (instr_is_compressed_i)?

rtl/ibex_if_stage.sv
```diff
@@ -100,6 +103,9 @@ module ibex_if_stage #(
 
   logic       [31:0] exc_pc;
 
+  // trap-vector base address, mode set to vectored
```
imphil: mode -> MODE

(I guess we could make mtvec a struct, but I'm not sure if it's worth the effort.)

doc/cs_registers.rst
vogelpi: Oh, that's a good one. It should be MIE indeed. I fixed it also in another place and switched to the csr_num.FIELD_NAME syntax.

doc/cs_registers.rst
vogelpi: Done.

doc/exception_interrupts.rst
vogelpi: Thanks, fixed that and also in other places in the doc.

doc/exception_interrupts.rst
vogelpi: The comma is not required in BE, but recommended in AE.

rtl/ibex_if_stage.sv
vogelpi: Changed.

rtl/ibex_if_stage.sv
vogelpi: Done.

doc/cs_registers.rst
vogelpi: Done.

vogelpi:
Thanks for your feedback @imphil ! I changed it and do the merge now.

PR #110, opened 2019-07-01
Fix RVFI rd write enable

towoe:
The original query was (rvfi_rd_we_id == 1'b0), so this fixes the incorrect transformation.


imphil:
Thanks Tobias!

PR #120, opened 2019-07-03
Cleanup ID stage and decoder

### LINKED ISSUE #103: Diagnose/remove combinatorial loop
Verilator reports this loop:

```
%Warning-UNOPTFLAT: ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_core.sv:129: Signal unoptimizable: Feedback to clock or circular logic: ibex_core.is_decoding
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_core.sv:129:      Example path: ibex_core.is_decoding
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_controller.sv:527:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_id_stage.sv:158:      Example path: ibex_core.id_stage_i.deassert_we
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_decoder.sv:550:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_id_stage.sv:174:      Example path: ibex_core.id_stage_i.jump_in_id
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_id_stage.sv:602:      Example path: ALWAYS
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_id_stage.sv:172:      Example path: ibex_core.id_stage_i.jump_set
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_controller.sv:151:      Example path: ALWAYS
                    ../src/lowrisc_ibex_ibex_0.1/rtl/ibex_core.sv:129:      Example path: ibex_core.is_decoding
```

We need to diagnose this loop in more detail, and hopefully remove it. Otherwise add an explanation and a lint waiver.

(My gut feeling is that we need to apply deassert_we later to signals that need it.)

vogelpi:
This PR contains a major cleanup of the ID stage and the decoder. Control logic (e.g. for stalls) that depends not on the decoder exclusively, is moved out of the decoder. Similarly, logic inside the ID stage that is strictly related to instruction decoding (immediate & register file address generation and checking) is moved into the decoder. 

As a result, the ID stage now instantiates the modules of the controller, decoder and registers, and contains the the main muxes, the write-back (WB) FSM to handle multicycle instructions and minimal glue logic, only.
This will ease future work on any of these parts. 

Finally, the WB FSM inside the ID stage is cleaned up to remove circular dependencies between FSM and decoder. This resolves #103 .

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -106,7 +106,7 @@ module ibex_controller (
     input  logic                      stall_multdiv_i,
     input  logic                      stall_jump_i,
     input  logic                      stall_branch_i,
-    input  logic                      instr_multicyle_i,     // multicycle instructions active
+    input  logic                      instr_multicycle_i,    // multicycle instructions active
 
     output logic                      id_valid_o,
 
@@ -268,7 +268,7 @@ module ibex_controller (
 
         /*
          * TODO: What should happen on
-         * instr_valid_i && (instr_multicyle_i || branch_in_id_i)?
+         * instr_valid_i && (instr_multicycle_i || branch_in_id_i)?
          * Let the instruction finish executing before serving debug or
          * interrupt requests?
          */
@@ -289,7 +289,7 @@ module ibex_controller (
           end
 
           default: begin
-            exc_kill_o    = irq_req_ctrl_i & ~instr_multicyle_i & ~branch_in_id_i;
+            exc_kill_o    = irq_req_ctrl_i & ~instr_multicycle_i & ~branch_in_id_i;
 
             if (instr_valid_i) begin
               // analyze the current instruction in the ID stage
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -44,16 +44,19 @@ module ibex_controller (
     input  logic                      ebrk_insn_i,           // decoder has EBREAK instr
     input  logic                      csr_status_i,          // decoder has CSR status instr
 
-    // from IF/ID pipeline
+    // from IF-ID pipeline stage
     input  logic                      instr_valid_i,         // instr from IF-ID reg is valid
     input  logic [31:0]               instr_i,               // instr from IF-ID reg, for mtval
     input  logic [15:0]               instr_compressed_i,    // instr from IF-ID reg, for mtval
     input  logic                      instr_is_compressed_i, // instr from IF-ID reg is compressed
 
-    // from prefetcher
-    output logic                      instr_req_o,           // start fetching instructions
+    // to IF-ID pipeline stage
+    output logic                      instr_valid_clear_o,   // kill instr in IF-ID reg
+    output logic                      id_ready_o,            // ID stage is ready for new instr
+    output logic                      halt_if_o,             // request halt of IF stage
 
     // to prefetcher
+    output logic                      instr_req_o,           // start fetching instructions
     output logic                      pc_set_o,              // jump to address set by pc_mux
     output ibex_defines::pc_sel_e     pc_mux_o,              // IF stage fetch address selector
                                                              // (boot, normal, exception...)
@@ -68,8 +71,6 @@ module ibex_controller (
     input  logic                      branch_in_id_i,        // branch in id
     input  logic                      branch_set_i,          // branch taken set signal
     input  logic                      jump_set_i,            // jump taken set signal
-
-    input  logic                      instr_multicyle_i,     // multicycle instructions active
 
     // External Interrupt Req Signals, used to wake up from wfi even if the interrupt is not taken
     input  logic                      irq_i,
@@ -101,16 +102,19 @@ module ibex_controller (
     output logic [31:0]               csr_mtval_o,
 
     // stall signals
-    output logic                      halt_if_o,
-    output logic                      halt_id_o,
+    input  logic                      stall_lsu_i,
+    input  logic                      stall_multdiv_i,
+    input  logic                      stall_jump_i,
+    input  logic                      stall_branch_i,
+    input  logic                      instr_multicyle_i,     // multicycle instructions active
 
-    input  logic                      id_ready_i,             // ID stage is ready
+    output logic                      id_valid_o,
 
     // Performance Counters
-    output logic                      perf_jump_o,            // we are executing a jump
-                                                              // instruction (j, jr, jal, jalr)
-    output logic                      perf_tbranch_o          // we are executing a taken branch
-                                                              // instruction
+    output logic                      perf_jump_o,           // we are executing a jump
+                                                             // instruction (j, jr, jal, jalr)
+    output logic                      perf_tbranch_o         // we are executing a taken branch
+                                                             // instruction
 );
   import ibex_defines::*;
 
@@ -127,6 +131,8 @@ module ibex_controller (
   logic debug_mode_q, debug_mode_n;
   logic load_err_q, load_err_n;
   logic store_err_q, store_err_n;
+
+  logic halt_id;
 
 `ifndef SYNTHESIS
   // synopsys translate_off
@@ -174,7 +180,7 @@ module ibex_controller (
     first_fetch_o          = 1'b0;
 
     halt_if_o              = 1'b0;
-    halt_id_o              = 1'b0;
+    halt_id                = 1'b0;
     irq_ack_o              = 1'b0;
     irq_id_o               = irq_id_ctrl_i;
     irq_enable_int         = m_IE_i;
@@ -213,7 +219,7 @@ module ibex_controller (
         ctrl_busy_o   = 1'b0;
         instr_req_o   = 1'b0;
         halt_if_o     = 1'b1;
-        halt_id_o     = 1'b1;
+        halt_id       = 1'b1;
         ctrl_fsm_ns   = SLEEP;
       end
 
@@ -224,7 +230,7 @@ module ibex_controller (
         ctrl_busy_o   = 1'b0;
         instr_req_o   = 1'b0;
         halt_if_o     = 1'b1;
-        halt_id_o     = 1'b1;
+        halt_id       = 1'b1;
 
         // normal execution flow
         // in debug mode or single step mode we leave immediately (wfi=nop)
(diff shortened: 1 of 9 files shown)


rtl/ibex_controller.sv
```diff
@@ -514,8 +514,9 @@ module ibex_controller (
   // Stall control //
   ///////////////////
 
-  // deassert WE when the core is not decoding instructions
-  // or in case of illegal instruction
+  // deassert write enable when the core is not decoding instructions, i.e., current instruction
+  // in ID stage done, but waiting for next instruction from IF stage, or in case of illegal
+  // instruction
   assign deassert_we_o = ~is_decoding_o | illegal_insn_i;
```
imphil: The need for the lengthy comment shows IMHO that this signal should be renamed. It's currently named after what is will be used for, and not what it is. Using this signal will deassert some write enables somewhere, that's pretty generic. To actually see what's going on you need a comment here and another comment in `ibex_id_stage.sv`. 

So how about renaming this signal to what it actually signals, like `id_is_stalled_o`, `id_idle_o` or something similar?

Once you have done this renaming you can remove most of the comment here and from `ibex_id_stage.sv`.

rtl/ibex_id_stage.sv
```diff
@@ -569,15 +566,26 @@ module ibex_id_stage #(
       .m_IE_i               ( m_irq_enable_i     )
   );
 
+  //////////////
+  // ID-EX/WB //
+  //////////////
+  // Do not forward decoder output to EX or WB if:
+  // - current instr is already done, ID waiting for IF stage
+  // - current instr is illegal
```
imphil: Please see my comment in `ibex_controller.sv` about this comment and the `deassert_we` signal name.

rtl/ibex_decoder.sv
```diff
@@ -506,10 +500,25 @@ module ibex_decoder #(
...
-    // make sure invalid compressed instruction causes an exception
+    // make sure illegal compressed instructions cause illegal instruction exceptions
     if (illegal_c_insn_i) begin
       illegal_insn_o = 1'b1;
     end
+
+    // make sure illegal instructions detected in the decoder do not propagate from decoder
+    // into register file, LSU, EX, WB, CSRs
+    // NOTE: instructions can also be detected to be illegal inside the CSRs (upon accesses with
+    // insufficient privileges), in ID stage (when accessing Reg 16 or higher in RV32E config)
```
imphil: Add a last sentence to the NOTE: "These cases are not handled here."

rtl/ibex_controller.sv
```diff
@@ -44,16 +44,19 @@ module ibex_controller (
...
-    // from IF/ID pipeline
+    // from IF-ID pipeline
     input  logic                      instr_valid_i,         // instr from IF-ID reg is valid
     input  logic [31:0]               instr_i,               // instr from IF-ID reg, for mtval
     input  logic [15:0]               instr_compressed_i,    // instr from IF-ID reg, for mtval
     input  logic                      instr_is_compressed_i, // instr from IF-ID reg is compressed
 
-    // from prefetcher
-    output logic                      instr_req_o,           // start fetching instructions
+    // to IF-ID pipeline
```
imphil: s/pipeline/pipeline stage/

rtl/ibex_controller.sv
```diff
@@ -519,6 +525,15 @@ module ibex_controller (
...
   assign deassert_we_o = ~is_decoding_o | illegal_insn_i;
 
+  // signal to IF stage that ID stage is ready for next instruction
+  assign id_ready_o = ~stall_lsu_i & ~stall_multdiv_i & ~stall_jump_i & ~stall_branch_i;
+
+  // kill instruction in IF-ID reg for instructions that are done
+  assign instr_valid_clear_o = id_ready_o | halt_id;
+
+  // signal that ID stage has valid output
+  assign id_valid_o = id_ready_o & ~halt_id;
```
imphil: It's unexpected for a valid/ready interface to have dependencies on each other. 

In this case, it's probably fine, as valid and ready to in different directions (ready goes to the previous stage, valid to the next stage). Can we reflect these different directions more clearly in the signal name? (e.g. `id_in_ready`, `id_out_valid`).

We could probably go even further and rename `~halt_id` to `id_in_valid`, and then we have a complete valid/ready interface.

rtl/ibex_ex_block.sv
```diff
@@ -145,21 +140,12 @@ module ibex_ex_block #(
...
-      multdiv_en:
-        ex_ready_o = multdiv_ready;
-      lsu_en_i:
-        ex_ready_o = lsu_ready_ex_i;
-      default:
-        //1 Cycle case
-        ex_ready_o = 1'b1;
-    endcase
-  end
+  // ALU output valid in same cycle, multiplier/divider may require multiple cycles
```
imphil: output *is* valid

rtl/ibex_id_stage.sv
```diff
@@ -680,7 +681,7 @@ module ibex_id_stage #(
       end
 
       WAIT_MULTICYCLE: begin
-        if (ex_ready_i) begin
+        if ( (data_req_id & lsu_valid_i) | (~data_req_id & ex_valid_i) ) begin
```
imphil: style nit: `if ( (` -> `if ((`

rtl/ibex_id_stage.sv
```diff
@@ -335,22 +334,14 @@ module ibex_id_stage #(
...
-    end else begin
-      regfile_wdata_mux = regfile_wdata_ex_i;
-    end
+  // Register file write data mux
+  always_comb begin : regfile_wdata_mux
+    unique case (regfile_wdata_sel)
+      RF_WD_EX:  regfile_wdata_id = regfile_wdata_ex_i;
+      RF_WD_LSU: regfile_wdata_id = regfile_wdata_lsu_i;
+      RF_WD_CSR: regfile_wdata_id = csr_rdata_i;
+      default:   regfile_wdata_id = regfile_wdata_ex_i;
```
imphil: The default can't happen, can it? Can you remove the default statement, or output x here to avoid masking errors when extending the RF_WD_* enum in the future?

rtl/ibex_defines.sv
```diff
@@ -153,6 +153,12 @@ typedef enum logic [2:0] {
   IMM_B_INCR_ADDR
 } imm_b_sel_e;
 
+// Regfile write data selection
+typedef enum logic [1:0] {
```
imphil: Would it make sense to not use an explicit type or int in this enum since we actually don't need any specific width here (just like in a state machine, the synthesis tool should figure it out)

rtl/ibex_id_stage.sv
```diff
@@ -602,85 +605,88 @@ module ibex_id_stage #(
...
       end
 
       WAIT_MULTICYCLE: begin
-        if ( (data_req_id & lsu_valid_i) | (~data_req_id & ex_valid_i) ) begin
-          regfile_we        = regfile_we_id;
-          id_wb_fsm_ns      = IDLE;
-          stall_lsu         = 1'b0;
-          stall_multdiv     = 1'b0;
-          select_data_rf    = data_req_id ? RF_LSU : RF_EX;
+        if ( (data_req_dec & lsu_valid_i) | (~data_req_dec & ex_valid_i) ) begin
```
imphil: style nit: `if ((`

rtl/ibex_decoder.sv
```diff
@@ -84,9 +100,60 @@ module ibex_decoder #(
...
+  opcode_e     opcode;
+
+  assign instr = instr_rdata_i;
+
+  //////////////////////////////////////
+  // Register and immediate selection //
+  //////////////////////////////////////
+
+  // immediate extraction and sign extension
+  assign imm_i_type_o = { {20 {instr[31]}}, instr[31:20] };
```
imphil: style nit: I find this space between 20 and { confusing, I'd write it as `assign imm_i_type_o = { {20{instr[31]}}, instr[31:20] };`

rtl/ibex_decoder.sv
vogelpi: Done

rtl/ibex_controller.sv
vogelpi: Done

rtl/ibex_id_stage.sv
vogelpi: Done

rtl/ibex_id_stage.sv
vogelpi: Done

rtl/ibex_decoder.sv
vogelpi: Done

rtl/ibex_defines.sv
vogelpi: I am not sure I get the point. Do you mean to just write ``typedef enum logic {`` here?
I don't know if this would work. Inside Ibex, we currently always specify the required width for enum types (also for FSM state types) unless the number of values is just 2, in which case a 1-bit `logic` is enough.

imphil: I'd use `typedef enum {` (no type at all), but feel free to leave as-is.

rtl/ibex_id_stage.sv
vogelpi: We can't remove the default (required by our guidelines) but we could output X. And I think it is a good idea to do so. I changed also the other muxes accordingly.

rtl/ibex_defines.sv
vogelpi: Ah I see. Actually, our guidelines demand the specification of a type and for synthesizable enums it must be `logic`.

rtl/ibex_controller.sv
vogelpi: As discussed, this will go a future PR that cleans up the controller and its FSM. The change is already done there.

rtl/ibex_id_stage.sv
vogelpi: As discussed, this will go a future PR that cleans up the controller and its FSM. The change is already done there.

rtl/ibex_controller.sv
vogelpi: Yes I totally agree. I will also include this in the same future PR that cleans up the controller and its FSM.

PR #122, opened 2019-07-09
Fix LSU + lint errors

vogelpi:
I prepared this PR to get urgent stuff fixed before the bigger changes for the controller and interrupts/exceptions.

This PR contains two commits:
- The first one is a cleanup of the LSU that also fixes a critical error. Previously, misaligned loads/stores were broken as the byte enable generation and read data alignment for the second transaction part were not working. These operations rely on the updated input address having the same alignment as the original address, whereas the LSU provided the aligned output address to the AGU.
- The second one fixes recently introduced linting errors.

Tested using Verilator and Vivado.

### DIFF
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -30,7 +30,7 @@ module ibex_load_store_unit (
     input  logic         clk_i,
     input  logic         rst_ni,
 
-    // output to data memory
+    // data interface
     output logic         data_req_o,
     input  logic         data_gnt_i,
     input  logic         data_rvalid_i,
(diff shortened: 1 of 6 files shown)


rtl/ibex_load_store_unit.sv
```diff
@@ -74,44 +74,51 @@ module ibex_load_store_unit (
...
 
   logic [3:0]   data_be;
   logic [31:0]  data_wdata;
 
-  logic         misaligned_st;   // high if we are currently performing the second part
-                                 // of a misaligned store
-  logic         data_misaligned, data_misaligned_q;
+  logic [31:0]  data_rdata_ext;
+
+  logic [31:0]  rdata_w_ext; // sign extension for words, actually only misaligned assembly
```
imphil: What does "actually only misaligned assembly" mean?

rtl/ibex_load_store_unit.sv
```diff
@@ -74,44 +74,51 @@ module ibex_load_store_unit (
...
-  // registers for data_rdata alignment and sign extension
-  logic [1:0]   data_type_q;
-  logic [1:0]   rdata_offset_q;
-  logic         data_sign_ext_q;
-  logic         data_we_q;
+  logic [31:0]  rdata_q, rdata_n;
+  logic [1:0]   rdata_offset_q, rdata_offset_n;
+  logic [1:0]   data_type_q, data_type_n;
+  logic         data_sign_ext_q, data_sign_ext_n;
+  logic         data_we_q, data_we_n;
```
imphil: Shouldn't this all be _d (register input) instead of _n (active low)?

rtl/ibex_load_store_unit.sv
```diff
@@ -291,188 +314,180 @@ module ibex_load_store_unit (
...
-  // to know what kind of error to signal, we need to know the type of the transaction to which
-  // the outsanding rvalid belongs.
-  assign load_err_o    = data_err_i & data_rvalid_i & ~data_we_q;
-  assign store_err_o   = data_err_i & data_rvalid_i &  data_we_q;
+        default: begin
+          data_misaligned = 1'bX;
+        end
+      endcase
+    end
+  end
```
imphil: I would write this combination of case and if as a simple assign:

```verilog
assign data_misaligned = 
  (data_type_ex_i == 2'b00 && data_addr[1:0] != 2'b00) || // misaligned word access
  (data_type_ex_i == 2'b01 && data_addr[1:0] == 2'b11); // misaligned half-word access
```

Symmetry question: Doesn't `(data_type_ex_i == 2'b01 && data_addr[1:0] != 2'b01)` also work for the half-word case?

rtl/ibex_load_store_unit.sv
```diff
@@ -291,188 +314,180 @@ module ibex_load_store_unit (
...
 
     data_req_o        = 1'b0;
 
-    lsu_update_addr_o = 1'b0;
-
     data_valid_o      = 1'b0;
     data_misaligned_o = 1'b0;
+    data_misaligned_n = data_misaligned_q;
 
     unique case(ls_fsm_cs)
```
imphil: drive-by style nit: `unique case (` (space)

rtl/ibex_load_store_unit.sv
```diff
@@ -74,44 +74,51 @@ module ibex_load_store_unit (
...
-  logic         data_misaligned, data_misaligned_q;
+  logic [31:0]  data_rdata_ext;
+
+  logic [31:0]  rdata_w_ext; // sign extension for words, actually only misaligned assembly
+  logic [31:0]  rdata_h_ext; // sign extension for half words
+  logic [31:0]  rdata_b_ext; // sign extension for bytes
+
+  logic         data_misaligned;
+  logic         data_misaligned_q, data_misaligned_n; // high if we are currently performing the
+                                                      // second part of a misaligned access
```
imphil: Now we have `data_misaligned_n`, `data_misaligned_q`, `data_misaligned`, and `data_misaligned_o`? That's rather confusing. Can we find better names?

rtl/ibex_load_store_unit.sv
```diff
@@ -291,188 +314,180 @@ module ibex_load_store_unit (
...
-        2'b10,
-        2'b11:;
+  // output to register file
+  assign data_rdata_ex_o = data_rvalid_i ? data_rdata_ext : rdata_q;
 
-        default: data_misaligned = 1'bX;
-      endcase // case (data_type_ex_i)
-    end
-  end
+  // output data address must be aligned to word
```
imphil: must be aligned to word -> must be word-aligned

rtl/ibex_load_store_unit.sv
vogelpi: That is an old comment. It means that the signal is used for read data alignment for misaligned loads. I change it.

rtl/ibex_load_store_unit.sv
vogelpi: Yes, you are right. Thanks.

rtl/ibex_load_store_unit.sv
vogelpi: Thanks for the suggestion. That's much more elegant. I will change it.

Regarding your question: Two loads/stores are only required when the requested word or half-word touches two consecutive word-aligned addresses. This means for misaligned word accesses it is any offset != 2b'00. For half-word accesses it means only if the offset is 2'b11.

rtl/ibex_load_store_unit.sv
vogelpi: Fixed.

rtl/ibex_load_store_unit.sv
vogelpi: Fixed.

rtl/ibex_load_store_unit.sv
vogelpi: Good point! I replaced these signals. To request the address update from ID/EX, the LSU now uses `addr_incr_req_o`. To find out whether a misaligned access needs to be split the LSU uses `split_misaligned_access`. To indicate that the LSU is currently handling a misaligned access, it uses `misa_handling_d/q`.

rtl/ibex_load_store_unit.sv
```diff
@@ -42,76 +42,81 @@ module ibex_load_store_unit (
...
-  logic         data_misaligned, data_misaligned_q;
+  logic [31:0]  data_rdata_ext;
+
+  logic [31:0]  rdata_w_ext; // word realignment for misaligned loads
+  logic [31:0]  rdata_h_ext; // sign extension for half words
+  logic [31:0]  rdata_b_ext; // sign extension for bytes
+
+  logic         split_misaligned_access;
+  logic         misa_handling_q, misa_handling_d; // high after receiving the grant for the
+                                                  // first part of a misaligned access
```
imphil: `misa` led me to believe this has something to do with the misa CSR. `handle_misaligned_d/q` maybe (hard to find a reasonably short and non-confusing name, I know.)

rtl/ibex_load_store_unit.sv
```diff
@@ -291,188 +312,152 @@ module ibex_load_store_unit (
...
+  /////////////
 
-  // to know what kind of error to signal, we need to know the type of the transaction to which
-  // the outsanding rvalid belongs.
-  assign load_err_o    = data_err_i & data_rvalid_i & ~data_we_q;
-  assign store_err_o   = data_err_i & data_rvalid_i &  data_we_q;
+  // check for misaligned accesses that need to be split into two word-aligned accesses
+  assign split_misaligned_access = ~data_req_ex_i ? 1'b0 :
+      ((data_type_ex_i == 2'b00) && (data_addr[1:0] != 2'b00)) || // misaligned word access
+      ((data_type_ex_i == 2'b01) && (data_addr[1:0] == 2'b11));   // misaligned half-word access
```
imphil: On the dependency on `~data_req_ex_i`:
- The first user of `split_misaligned_access` is in the `IDLE` state, where you already check for `data_req_ex_i`, so the condition can be removed here.
- The second user of `split_misaligned_access` is in the `WAIT_GNT_MIS` state, where `data_req_ex_i` is not checked. However, it seems that we only get into this state if `split_misaligned_access == 1'b1`, so you could simply set `misa_handling_d = 1'b1` in the `WAIT_GNT_MIS` state? I'm not entirely sure about the timing of this signal right now, so I might be missing something. If so, it still might make sense to check the `data_req_ex_i` in the FSM instead of this assignment.

rtl/ibex_load_store_unit.sv
vogelpi: OK, I changed it to `handle_misaligned_d/q`.

rtl/ibex_load_store_unit.sv
vogelpi: You are totally right! I changed it as suggested.

Issue #105, opened 2019-06-27
Instructions accessing x16-x31 should be considered illegal for RV32E


asb:
The specification doesn't require that all unused instruction encodings generate illegal instruction traps. However, we prefer to do so for Ibex as this increases the chances of catching errors early and provides some forward-compatibility for trap+emulate of future standard or non-standard extensions.

Right now, an illegal instruction exception won't be signalled on an RV32E system for instructions accessing x16-x31. There is some relevant commented out logic in ibex_id_stage.sv: https://github.com/lowRISC/ibex/blob/90796e3800207c67f5f0349b2216ef8bdf623ab2/rtl/ibex_id_stage.sv#L274-L278 Logic may also be needed in ibex_compressed_decoder.sv for the CI and CSS formats (ultimately, after #104 we need to ensure the mcause holds the compressed version of the instruction).

vogelpi:
I have implemented this functionality. But since it is based on PR #120 which first needs to be reviewed and merged. I did not yet create a PR for this.

One thing worth to mention is that the check previously commented out can generate false alarms as the bits used to encode the register file addresses for some instructions can also be used to encode other arguments and/or immediates for other instructions. Thus, the check must also evaluate if the current instruction uses the register file output or not.

As for the compresses instructions, the controller always writes the actual faulting instruction to `mtval`. Independently whether the instruction is detected as being illegal in the decoder or the compressed decoder. This means there is no check of the register file address required in the compressed decoder. I added a comment to the compressed decoder regarding this.

vogelpi:
I am closing this issue. The discussed feature got added with #126 .

PR #126, opened 2019-07-09
Check register address for RV32E

vogelpi:
This PR adds register address checks to the decoder for RV32E configurations. In such a configurations, registers x16-x31 are not available. If accessed, the core now generates an illegal instruction exception as discussed in #105 .

Without this PR, the core just ignores the MSB of register file addresses which is problematic and masks the actual problem.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -100,8 +100,11 @@ module ibex_decoder #(
 
   import ibex_defines::*;
 
+  logic        illegal_insn;
   logic        illegal_reg_rv32e;
   logic        csr_illegal;
+  logic        regfile_we;
+
   logic [31:0] instr;
 
   csr_op_e     csr_op;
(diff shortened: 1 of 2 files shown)


rtl/ibex_decoder.sv
```diff
@@ -564,33 +568,27 @@ module ibex_decoder #(
...
     end
 
     // make sure illegal instructions detected in the decoder do not propagate from decoder
     // into register file, LSU, EX, WB, CSRs
     // NOTE: instructions can also be detected to be illegal inside the CSRs (upon accesses with
     // insufficient privileges), in ID stage (when accessing Reg 16 or higher in RV32E config),
     // these cases are not handled here
-    if (illegal_insn_o) begin
-      regfile_we_o    = 1'b0;
+    if (illegal_insn) begin
```
imphil: How about `if (illegal_insn | illegal_reg_rv32e | illegal_c_insn_i)` here and remove the `illegal_reg_rv32e` from the assigns below (line 604 + 607) and the two if statement above (`if (illegal_c_insn_i) begin`)

vogelpi: Thanks for your fast review @imphil . I really appreciate it!

The problem here is that this would lead to a zero-latency loop:

- We only want to check the register write address if the instruction wants to write the register file (`regfile_we`). Otherwise the `rd` field of the instruction can have a different meaning.
- If we detect a wrong register file write address, we want to deassert the write enable to not change the content of the register file.

But I agree, these statements are a bit convoluted. I am open for suggestions.

vogelpi: For the record: we had a look together and concluded that there is not other and better solution to this.

PR #139, opened 2019-07-11
Update opcode enum

towoe:
Rename fence and sort

### DIFF
--- a/rtl/ibex_defines.sv
+++ b/rtl/ibex_defines.sv
@@ -31,7 +31,7 @@ package ibex_defines;
 typedef enum logic [6:0] {
   OPCODE_LOAD     = 7'h03,
   OPCODE_MISC_MEM = 7'h0f,
-  OPCODE_OPIMM    = 7'h13,
+  OPCODE_OP_IMM   = 7'h13,
   OPCODE_AUIPC    = 7'h17,
   OPCODE_STORE    = 7'h23,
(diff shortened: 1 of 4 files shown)


asb:
If we wanted to complete the alignment with the RISC-V spec, we'd rename OPIMM to OP_IMM as well. I'd be in favour of this. Looks like it would be a straight-forward and non-disruptive replacement, but worth seeing if @vogelpi and @imphil feel negatively about the churn.

rtl/ibex_defines.sv
```diff
@@ -29,17 +29,17 @@ package ibex_defines;
...
-  OPCODE_STORE  = 7'h23,
-  OPCODE_OP     = 7'h33,
-  OPCODE_LUI    = 7'h37,
-  OPCODE_BRANCH = 7'h63,
-  OPCODE_JALR   = 7'h67,
-  OPCODE_JAL    = 7'h6f,
-  OPCODE_SYSTEM = 7'h73
+  OPCODE_LOAD     = 7'h03,
+  OPCODE_MISC_MEM = 7'h0f,
+  OPCODE_OPIMM    = 7'h13,
```
vogelpi: Please also rename OPCODE_OPIMM to correspond to the standard.

towoe: Done

towoe:
All names in `opcode_e` reflect now the spec.

PR #150, opened 2019-07-12
Generate `instr_ret/compressed` inside ID stage

vogelpi:
This commit moves the generation of these signals into the ID stage to generate it with more accurate timing. More precisely, an instruction is now considered as retired once it has been completely executed and not once the next instruction is pushed into the IF-ID pipeline register.

Also, it no longer depends on whether the instruction was valid or illegal. For RVFI and the tracer, also illegal instructions must be considered.

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -459,7 +459,9 @@ module ibex_core #(
       // Performance Counters
       .perf_jump_o                  ( perf_jump              ),
       .perf_branch_o                ( perf_branch            ),
-      .perf_tbranch_o               ( perf_tbranch           )
+      .perf_tbranch_o               ( perf_tbranch           ),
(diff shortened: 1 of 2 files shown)


towoe:
Thanks @vogelpi for the PR. To me it makes sense to do it this way, but I noticed that the RVFI part needs to be changed as well.
This is because right now it waits until the new instruction is indeed ready as it makes in this way sure that the new program counter is correct. I think this can be changed as `instr_ret_o` also considers the branch information.
What I would suggest is that I will work on this and we leave the PR open for now.

towoe:
@vogelpi I think it makes sense to just merge it and I will do a separate PR for the RVFI and tracer changes.

vogelpi:
Yes, thanks for reviewing, testing and approving. I'll merge now.

PR #157, opened 2019-07-15
Fix dpc CSR not updated

### LINKED ISSUE #153: dpc is not updated properly if debug_req_i is a one cycle pulse
As you can see from below waveform, debug_req_i is modeled as a pulse interrupt which is only asserted for one cycle. ibex can enter debug mode and start fetching instructions from DmHaltAddr. However, dpc is not updated properly (keep at default value 0). I cannot find in the RISC-V debug spec that whether the halt request should be a level or pulse signal, can you help to clarify?  

screencapture-remotedesktop-corp-google-access-session-6fd70a76-592d-4962-9324-fb0b435c9f3d-2019-07-14-16_45_08

vogelpi:
This PR lets the core also check `debug_req_i` and `debug_single_step_i` while entering debug mode to avoid entering debug mode without setting up the debug CSRs properly (which anyway depend on these signals), if these signals go low during the procedure.

This resolves #153 .

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -370,8 +370,6 @@ module ibex_controller (
           debug_cause_o = DBG_CAUSE_STEP;
         end else if (debug_req_i) begin
           debug_cause_o = DBG_CAUSE_HALTREQ;
-        end else if (ebrk_insn_i) begin
-          debug_cause_o = DBG_CAUSE_EBREAK;
         end
 
         // enter debug mode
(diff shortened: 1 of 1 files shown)


vogelpi:
@taoliug , would you mind to check locally if the problem still persists before we merge?

rtl/ibex_controller.sv
```diff
@@ -358,24 +358,24 @@ module ibex_controller (
       DBG_TAKEN_IF: begin
         // enter debug mode and save PC in IF to dpc
         // jump to debug exception handler in debug memory
-        pc_mux_o         = PC_EXC;
-        pc_set_o         = 1'b1;
-        exc_pc_mux_o     = EXC_PC_DBD;
+        if (debug_single_step_i || debug_req_i)
```
taoliug: Missing begin here

PR #169, opened 2019-07-17
ID/EX stage: do not write to register file upon load errors

### LINKED ISSUE #162: Load instructions that cause an error still write to registers
Hi,
There seems to be a bug in Ibex when performing a read from memory which returns an error.

I would expect that when the memory slave sets data_err high, the core would not write the data that was read in into a register, but it appears to still be written to the register regardless.

What i observe is that the data is still written into the register.

I'm using the following instructions:
```
addi r2, r0, 1
lw r2, r2[2]
addi r3, r2, 1
```
When using the load word instruction, the register is overwritten with 0x0, but when using load halfword or load byte, it seems to actually read what is returned by memory and store it in the register.

My memory slave is configured such that accesses outside of the range 0x80000000 - 0x80007fff will set data_err high.

I've performed a fix that seemed to work locally by changing the line
https://github.com/lowRISC/ibex/blob/b7ef64b62da74afe52bb40165e17a7e87539d47a/rtl/ibex_id_stage.sv#L283
to
```
assign regfile_we = (illegal_csr_insn_i || !instr_executing || lsu_load_err_i) ? 1'b0          :
```

This fix seems to work with other failing loads as well, but I'm not sure that it conforms to your style guide or whether it should be done here or at some other point in the code.

vogelpi:
This commit fixes the write back FSM to not store values returned from
memory to the register file when the LSU is reporting a load error.

This bug was reported by @ivanmgribeiro. This commit resolves #162.

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -622,7 +622,7 @@ module ibex_id_stage #(
         if ((data_req_dec & lsu_valid_i) | (~data_req_dec & ex_valid_i)) begin
           id_wb_fsm_ns            = IDLE;
           instr_multicycle_done_n = 1'b1;
-          regfile_we_wb           = regfile_we_dec;
+          regfile_we_wb           = regfile_we_dec & ~lsu_load_err_i;
           instr_ret_o             = 1'b1;
         end else begin
           stall_lsu               = data_req_dec;


rtl/ibex_id_stage.sv
```diff
@@ -622,7 +622,7 @@ module ibex_id_stage #(
         if ((data_req_dec & lsu_valid_i) | (~data_req_dec & ex_valid_i)) begin
           id_wb_fsm_ns            = IDLE;
           instr_multicycle_done_n = 1'b1;
-          regfile_we_wb           = regfile_we_dec;
+          regfile_we_wb           = lsu_load_err_i ? 1'b0 : regfile_we_dec;
```
imphil: I would simply go for `regfile_we_dec & ~lsu_load_err_i`, but leaving that up to you.

vogelpi: Thanks @imphil . Your suggestion is more concise. I changed it.

PR #174, opened 2019-07-18
Rename packages

imphil:
Rename _defines.sv to _pkg.sv. Patch best viewed without whitespace changes.


imphil:
Good catch, I've updated the Bender file as well.

PR #176, opened 2019-07-19
Decoder: remove reg-reg load, make LWU illegal

### LINKED ISSUE #25: Remove reg-reg loads
reg-reg load (i.e., offset taken from a register and not from an immediate) should not be part of the Ibex ISA.

The code at line 261 here:
https://github.com/lowRISC/ibex/blob/master/rtl/ibex_decoder.sv

is a piece of code coming from RISCY and it should be removed.

vogelpi:
The custom reg-reg instruction was added in the original design but is
no longer needed. This commit removes it. Also, load instructions with
`instr[14:12] == 3'b110` are now decoded as illegal.

This resolves #25.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -299,12 +299,10 @@ module ibex_decoder #(
 
         // store size
         unique case (instr[13:12])
-          2'b00: data_type_o = 2'b10; // SB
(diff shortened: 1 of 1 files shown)


vogelpi:
Thanks @imphil !

PR #177, opened 2019-07-19
Rename cores for fusesoc, mainly tracer

imphil:
Painful, but needed for symmetry and to reduce confusion: rename the cores in the fusesoc core files.

This PR requires RVFI to be defined globally if the ibex_core_tracing.sv is used to avoid potential design bugs caused by differing tool behavior. It's better to be explicit here.

If RVFI is not set and the ibex_core_tracing.sv file is used a compile error will be thrown as the file includes invalid syntax. I'm not yet sure that this is the right way to go, but it seems that there's no universally understood way to convince a Verilog parser to issue a compile-time fatal error.

### DIFF
--- a/ibex_core_tracing.core
+++ b/ibex_core_tracing.core
@@ -1,0 +1,28 @@
+CAPI=2:
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:ibex:ibex_core_tracing:0.1"
+description: "Ibex CPU core with tracing enabled"
+filesets:
+  files_rtl:
+    depend:
+      - lowrisc:ibex:ibex_core
+      - lowrisc:ibex:ibex_tracer
+    files:
+      - rtl/ibex_core_tracing.sv
+    file_type: systemVerilogSource
+
+parameters:
+  # The tracer uses the RISC-V Formal Interface (RVFI) to collect trace signals.
+  RVFI:
+    datatype: bool
+    paramtype: vlogdefine
+    default: true
+
+targets:
+  default:
+    filesets:
+      - files_rtl
+    parameters:
+      - RVFI=true
--- a/rtl/ibex_core_tracing.sv
+++ b/rtl/ibex_core_tracing.sv
@@ -3,13 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 
-// ibex_tracer relies on the signals from the RISC-V Formal Interface
-`define RVFI
-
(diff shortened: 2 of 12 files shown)


tjaychen:
@imphil 
I know this is probably related to the issue that was discussed.  But in this current form, what's the suggestion of integrating into the top level? Should we instantiate ibex_core_tracing?

imphil:
@towoe can you run the tracer example with modelsim on your end to see if it still works? I did a couple blind renames without the ability to test it.

ibex_core_tracing.core
```diff
@@ -0,0 +1,29 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:ibex:ibex_core_tracing:0.1"
+description: "Ibex CPU core with tracing enabled"
+filesets:
+  files_rtl:
+    depend:
+      - lowrisc:ibex:ibex_core
+      - lowrisc:ibex:ibex_tracer
+  files_rtl:
```
towoe: yaml error as this is a duplicate, please delete line.

rtl/ibex_core_tracing.sv
```diff
@@ -55,40 +52,18 @@ module ibex_core_tracer #(
     // Debug Interface
     input  logic        debug_req_i,
 
-    // RISC-V Formal Interface
```
towoe: The RVFI signals should still be an output of `ibex_core_tracing`?

ibex_core_tracing.core
imphil: done.

rtl/ibex_core_tracing.sv
imphil: You could argue either way, but I'd say: 
- Use ibex_core_tracing instead of ibex_core when you want to do tracing. It's use of RVFI is an implementation detail and shouldn't be exposed to the top. 
- Use ibex_core with RVFI defined if you want to run RISC-V Formal; don't use ibex_core_tracing in this case.

imphil:
Can you have another look Tobi?

rtl/ibex_core_tracing.sv
towoe: Thanks, I now get the intention behind this.
Then we need to define those signals locally if they are no longer available as an output?

imphil: Very much so. Looks like this PR is killing me ... Updated.

towoe: Thanks @imphil

imphil:
Thanks for testing it!

PR #192, opened 2019-07-25
Change sim timescale option - Fixes #181

### LINKED ISSUE #181: Compile issue while running ibex_core_tracer example
Hi, 

I am running into compile issues while trying to run the `ibex_core_tracer` example using `fusesoc`.  I guess the issue is in the way `timescale` value is specified on the command-line. 

Updating the timescale option from `timescale=1ns/1ns` to `timescale 1ns/1ns` should fix it.

raulbehl:
- Updated the timescale option to not include "=" in between the
    timescale directive and the value passed

### DIFF
--- a/examples/sim/top_tracing_sim.core
+++ b/examples/sim/top_tracing_sim.core
@@ -22,4 +22,4 @@ targets:
       - ibex_tracing_tb
     tools:
       modelsim:
-        vlog_options: [-timescale=1ns/1ns]
+        vlog_options: [-timescale 1ns/1ns]


towoe:
Thanks for creating this PR!
Could you perhaps reword the commit message header to be more meaningful. Something like: "Change sim timescale option (fixes #181)"

vogelpi:
Hi @raulbehl,

thanks for preparing the fix and submitting the PR.

Please change the commit message to a more meaningful name as suggested by @towoe .

It is fine to have "fixing #181" in the message body but we usually do not put it in the title as it can cause problems when importing Ibex inside another repository (depending on the setup the commit might then automatically close issues in this other repo).

Best regards,
Pirmin

raulbehl:
Hi @vogelpi, 

Sure, it makes sense. I usually follow the approach of indicating the issue number in the commit header followed by a small description in the commit body. Will make the needed changes.

towoe:
Hi @raulbehl,

thanks for updating this PR. Unfortunately we now have three commits here instead of one, so we now have to get rid of those extra two commits.
We can do this by rebasing.
Looking at your fork this could look like this:
```
git rebase -i HEAD~3
```
You can then drop the commits b5566760ffcf56a3428d3fc3656d9a777e4f00a3 and 6a03aa710544f7edf794184112ad4d39f99c1909
I used `HEAD~3` to start the rebase beginning at your commits.
In general a more robust approach is to create a branch for working. Then the rebase is normally the same: `git rebase -i master`.

raulbehl:
Thank you for the comment. I should have used a branch in the first place. My bad!

raulbehl:
Removed the extra two commits with the last push. Let me know if it looks good. 

Thank you!

imphil:
Thanks for your contribution @raulbehl!

raulbehl:
@imphil Thank you for the merge

PR #193, opened 2019-07-25
Change mtvec implementation and boot address

ivanmgribeiro:
This PR aims to change the implementation of the mtvec CSR in order to allow it to be written to and read from by software.
This also allows for the boot address to be set to the input boot address instead of adding an offset to it, and it no longer needs to be 256-byte aligned

### DIFF
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -129,9 +129,14 @@ Machine Trap-Vector Base Address (mtvec)
 
 CSR Address: ``0x305``
 
+Reset Value: ``0x0000_0001``
+
 When an exception is encountered, the core jumps to the corresponding handler using the content of ``mtvec`` as base address.
-It is a WARL register which contains the boot address.
-It contains a hard-wired value, so will remain unchanged after any writes.
+``mtvec`` is a WARL register.
+Its value can be changed using the instructions added by the Zicsr extension.
+The value written to and read from the register will always be 256-byte aligned, and only vectored mode is supported.
+This means the bottom 8 bits of the write input are ignored, and instead ``0x01`` is written in the bottom 8 bits of the register.
+``mtvec``.BASE is initialized to the 256-byte aligned boot address on booting the core.
 ``mtvec``.MODE is set to 2'b01 to indicate vectored interrupt handling.
 
 
--- a/doc/exception_interrupts.rst
+++ b/doc/exception_interrupts.rst
@@ -8,10 +8,10 @@ Ibex implements trap handling for interrupts and exceptions according to the RIS
 All exceptions cause the core to jump to the base address of the vector table in the ``mtvec`` CSR.
 Interrupts are handled in vectored mode, i.e., the core jumps to the base address plus four times the interrupt ID.
 
-The base address of the vector table is given by the boot address (must be aligned to 256 bytes, i.e., its least significant byte must be 0x00).
-The most significant 3 bytes of the boot address given to the core are used for the first instruction fetch of the core and as the basis of the interrupt vector table.
+The base address of the vector table is initialized on booting the core.
+The most significant 3 bytes of the boot address given to the core are used for the first instruction fetch of the core and to initialize the base of the interrupt vector table when the core is booted.
 The core starts fetching at the address made by concatenating the most significant 3 bytes of the boot address and the reset value (0x80) as the least significant byte.
-The boot address can be changed after the first instruction was fetched to change the interrupt vector table address.
+It is possible to change the base address after booting using the instructions added by the Zicsr extension.
 It is assumed that the boot address is supplied via a register to avoid long paths to the instruction fetch unit.
 
 
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -124,6 +124,7 @@ module ibex_core #(
   pc_sel_e     pc_mux_id;              // Mux selector for next PC
   exc_pc_sel_e exc_pc_mux_id;          // Mux selector for exception PC
   exc_cause_e  exc_cause;              // Exception cause
+  logic        csr_mtvec_init;
 
   logic        lsu_load_err;
   logic        lsu_store_err;
@@ -325,6 +326,7 @@ module ibex_core #(
       .pc_mux_i                 ( pc_mux_id              ),
       .exc_pc_mux_i             ( exc_pc_mux_id          ),
       .exc_cause                ( exc_cause              ),
+      .csr_mtvec_init_o         ( csr_mtvec_init         ),
 
       // jump targets
       .jump_target_ex_i         ( jump_target_ex         ),
@@ -561,8 +563,9 @@ module ibex_core #(
       .core_id_i               ( core_id_i              ),
       .cluster_id_i            ( cluster_id_i           ),
 
-      // mtvec reset value
-      .mtvec_resetval_i        ( boot_addr_i            ),
+      // mtvec initialization
+      .csr_mtvec_resetval_i    ( boot_addr_i            ),
+      .csr_mtvec_init_i        ( csr_mtvec_init         ),
 
       // Interface to CSRs (SRAM like)
       .csr_access_i            ( csr_access             ),
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -332,7 +332,7 @@ module ibex_core #(
       // CSRs
       .csr_mepc_i               ( csr_mepc               ), // exception return address
       .csr_depc_i               ( csr_depc               ), // debug return address
-      .csr_mtvec_o              ( csr_mtvec              ), // trap-vector base address
+      .csr_mtvec_i              ( csr_mtvec              ), // trap-vector base address
 
       // pipeline stalls
       .id_in_ready_i            ( id_in_ready            ),
@@ -595,7 +595,7 @@ module ibex_core #(
       .csr_save_id_i           ( csr_save_id            ),
       .csr_restore_mret_i      ( csr_restore_mret_id    ),
       .csr_save_cause_i        ( csr_save_cause         ),
-      .csr_mtvec_i             ( csr_mtvec              ),
+      .csr_mtvec_o             ( csr_mtvec              ),
       .csr_mcause_i            ( exc_cause              ),
       .csr_mtval_i             ( csr_mtval              ),
       .illegal_csr_insn_o      ( illegal_csr_insn_id    ),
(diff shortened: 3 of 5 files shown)


ivanmgribeiro:
I'm not able to test this locally, so please let me know if there's any bugs that come up.

vogelpi:
Hi @ivanmgribeiro,

thanks for your PR.
What do you mean exactly by you can't this test locally?

ivanmgribeiro:
My current testing setup doesn't work well with the current RVFI signals that ibex is producing, so I've had to make local workarounds for these issues.
I've managed to get a setup that works but isn't ideal.

For reference, I am using [TestRIG](https://github.com/CTSRD-CHERI/TestRIG/) with Sail as the model implementation for testing. TestRIG generates a random instruction stream, passes it to two implementations and compares the outputs.

vogelpi:
Thanks for getting back Ivan. I will have a look at your PR and give feedback.

vogelpi:
> My current testing setup doesn't work well with the current RVFI signals that ibex is producing, so I've had to make local workarounds for these issues.
> I've managed to get a setup that works but isn't ideal.

Just as a remark: we cannot really test the RVFI interface here as we do not have access to the required tools. If there are issues with the RVFI signals and how they are generated, and if you can fix and test this, this would be very much appreciated!

rtl/ibex_cs_registers.sv
```diff
@@ -572,6 +580,8 @@ module ibex_cs_registers #(
       mepc_q         <= '0;
       mcause_q       <= '0;
       mtval_q        <= '0;
+      // set mtvec to 1 to signify vectored mode
+      mtvec_q        <= '1;
```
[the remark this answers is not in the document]
ivanmgribeiro: Just to clarify, do you think mtvec should indeed be set to ```32h0000_0001``` on reset, or should the base be set to another value (perhaps passed in as a parameter)?

ivanmgribeiro:
@vogelpi Just to clarify, do you think mtvec should indeed be set to ```32h0000_0001``` on reset, or should the base be set to another value (perhaps passed in as a parameter)?

imphil:
> do you think mtvec should indeed be set to 32h0000_0001 on reset

No, the idea is to keep the reset/default behavior as it is today, as @vogelpi outlined in https://github.com/lowRISC/ibex/pull/193#pullrequestreview-267058207.

ivanmgribeiro:
The requested changes should be implemented now. Please let me know if there's anything I've missed out or anything else you would like me to change.
Also, thanks for the great feedback!

rtl/ibex_if_stage.sv
```diff
@@ -113,14 +111,11 @@ module ibex_if_stage #(
...
 
-  // trap-vector base address, mtvec.MODE set to vectored
-  assign csr_mtvec_o = {boot_addr_i[31:8], 6'b0, 2'b01};
-
   // exception PC selection mux
   always_comb begin : exc_pc_mux
     unique case (exc_pc_mux_i)
-      EXC_PC_EXC:     exc_pc = { boot_addr_i[31:8], 8'h00                    };
-      EXC_PC_IRQ:     exc_pc = { boot_addr_i[31:8], 1'b0, irq_id[4:0], 2'b00 };
+      EXC_PC_EXC:     exc_pc = { csr_mtvec_i[31:2], 2'b0 };
```
vogelpi: `csr_mtvec_i[31:8], 8'h00`

rtl/ibex_if_stage.sv
```diff
@@ -113,14 +111,11 @@ module ibex_if_stage #(
...
+      EXC_PC_IRQ:     exc_pc = { csr_mtvec_i[31:8], 1'b0, irq_id[4:0], 2'b00 } );
```
vogelpi: Remove the `)` at the end.

rtl/ibex_cs_registers.sv
```diff
@@ -232,6 +236,9 @@ module ibex_cs_registers #(
     csr_rdata_int = '0;
     illegal_csr   = 1'b0;
 
+    // we always want the current value of mtvec to be accessible so the IF stage can jump to it
+    csr_mtvec_o = mtvec_q;
```
vogelpi: `csr_mtvec_o` is always driven by the same signal `mtvec_q`. It does not depend on the read logic described in this section at all.

Please move this statement down to about line 519 marked by a comment `//directly output some registers`.

rtl/ibex_cs_registers.sv
```diff
@@ -386,6 +394,11 @@ module ibex_cs_registers #(
       // mtval: trap value
       CSR_MTVAL: if (csr_we_int) mtval_d = csr_wdata_int;
 
+      // mtvec
+      // the 2 LSBs of mtvec should always be 01 since we use only vectored mode
```
vogelpi: mtvec.MODE set to vectored

rtl/ibex_cs_registers.sv
```diff
@@ -572,6 +585,7 @@ module ibex_cs_registers #(
       mepc_q         <= '0;
       mcause_q       <= '0;
       mtval_q        <= '0;
+      mtvec_q        <= {mtvec_resetval_i[31:8], 6'b0, 2'b01};
```
vogelpi: This is not good practice and also not safe. Your reset value should not depend on a signal that is run-time dependent (`mtvec_resetval_i` depends on `boot_addr_i` which can change anytime). The synthesis tool will thus create a mux here which uses the reset signal as control input. Further, if the boot address is changed after the reset is deasserted but before the core is booted. It will not have a valid `mtvec` CSR. 

What you need to do is the following:
1. Reset `mtvec_q` to `32'b01`.
2. Use a mux for `mtvec_d` that is either:

   - {csr_wdata_int[31:8], 6'b0, 2'b01} when the mtvec CSR is written
   - {boot_addr_i[31:8], 6'b0, 2'b01} when the core is booting. For this, you need to generate e.g. a signal `csr_mtvec_init_o` in `if_stage.sv` when `pc_mux_i == PC_BOOT` and `pc_set_i == 1'b1`.
   - `mtvec_q` otherwise.

ivanmgribeiro:
Hi,
I've committed some changes to address your suggestions. I'm not sure if I correctly interpreted what you meant by having a mux for mtvec_d which picks between those 3 options, please let me know if these changes are correct or not.
I'll work on adjusting the documentation next.

rtl/ibex_if_stage.sv
```diff
@@ -111,11 +113,14 @@ module ibex_if_stage #(
   assign irq_id         = {exc_cause};
   assign unused_irq_bit = irq_id[5];   // MSB distinguishes interrupts from exceptions
 
+  // tell CS register file to initialize mtvec on boot
+  assign csr_mtvec_init_o = (pc_mux_i == PC_BOOT);
```
vogelpi: This is not sufficient, `pc_mux_i` can be set to `PC_BOOT` also if the core is not booting (see `ibex_controller.sv` where the signal is coming from). As suggested in the previous review, also `pc_set_i` must be set.

rtl/ibex_core.sv
```diff
@@ -559,8 +561,9 @@ module ibex_core #(
       .core_id_i               ( core_id_i              ),
       .cluster_id_i            ( cluster_id_i           ),
 
-      // mtvec reset value
-      .mtvec_resetval_i        ( boot_addr_i            ),
+      // mtvec initialization
+      .csr_mtvec_resetval_i    ( boot_addr_i            ),
```
vogelpi: I would rename the signal `csr_mtvec_resetval_i` to `boot_addr_i` because it is simply the boot address.

ivanmgribeiro:
For documentation purposes, would you say the reset value is then ```0x0000_0001``` or ```{boot_addr[31:8], 8'h01}```? Or would it be best to instead explain in the documentation what the reset and initialization logic does?

vogelpi:
> For documentation purposes, would you say the reset value is then `0x0000_0001` or `{boot_addr[31:8], 8'h01}`? Or would it be best to instead explain in the documentation what the reset and initialization logic does?

Hi Ivan,
the documentation should say that the reset value of `mtvec` is `0x0000_0001` but also that the BASE field of mtvec (i.e., the trap-handler base address) is initialized to the boot address when the core is booting. Also, the documentation should state that the value in the BASE field needs to be aligned to 256 Bytes.
Best regards,
Pirmin

rtl/ibex_if_stage.sv
ivanmgribeiro: I missed that in the last review, sorry. Should be resolved now.

doc/cs_registers.rst
```diff
@@ -129,9 +129,14 @@ Machine Trap-Vector Base Address (mtvec)
...
+
 When an exception is encountered, the core jumps to the corresponding handler using the content of ``mtvec`` as base address.
-It is a WARL register which contains the boot address.
-It contains a hard-wired value, so will remain unchanged after any writes.
+``mtvec`` is a WARL register.
+Its value can be changed using the instructions added by the Zicsr extension.
+The value written to and read from the register will always be 256-byte aligned, and only vectored mode is supported.
+This means the bottom 8 bits of the write input are ignored, and instead ``0x01`` is written in the bottom 8 bits of the register.
+``mtvec``.BASE is initialized to the 256-byte aligned boot address on booting the core.
 ``mtvec``.MODE is set to 2'b01 to indicate vectored interrupt handling.
```
vogelpi: I think this information would be better conveyed using a table (like for mcause).

doc/exception_interrupts.rst
```diff
@@ -8,10 +8,10 @@ Ibex implements trap handling for interrupts and exceptions according to the RIS
...
 Interrupts are handled in vectored mode, i.e., the core jumps to the base address plus four times the interrupt ID.
 
-The base address of the vector table is given by the boot address (must be aligned to 256 bytes, i.e., its least significant byte must be 0x00).
-The most significant 3 bytes of the boot address given to the core are used for the first instruction fetch of the core and as the basis of the interrupt vector table.
+The base address of the vector table is initialized on booting the core.
+The most significant 3 bytes of the boot address given to the core are used for the first instruction fetch of the core and to initialize the base of the interrupt vector table when the core is booted.
 The core starts fetching at the address made by concatenating the most significant 3 bytes of the boot address and the reset value (0x80) as the least significant byte.
-The boot address can be changed after the first instruction was fetched to change the interrupt vector table address.
+It is possible to change the base address after booting using the instructions added by the Zicsr extension.
```
vogelpi: It is possible to change the base address after booting through the ``mtvec`` CSR.

vogelpi:
@ivanmgribeiro , I implemented the suggested changes, tested it and created a new PR into your own branch here: https://github.com/ivanmgribeiro/ibex/pull/2

Please let me know if something is unclear to you.

vogelpi:
Thanks for you feedback @ivanmgribeiro , I created a new PR #226 which supersedes this one. I am thus closing it.

Thanks for taking the initiative and for your contribution to Ibex!

PR #194, opened 2019-07-25
src_files.yml: correct path of source files

vogelpi:
This PR contains a single commit to update the path of source files in the `src_files.yml` file used in the PULP world.

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -1,23 +1,23 @@
 ibex:
   incdirs: [
+    rtl,
   ]
   files: [
-    ibex_pkg.sv,
-    ibex_alu.sv,
-    ibex_compressed_decoder.sv,
-    ibex_controller.sv,
-    ibex_cs_registers.sv,
-    ibex_debug_unit.sv,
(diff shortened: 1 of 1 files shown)


towoe:
Thanks for updating this Pirmin. LGTM but I am curious what the format is supposed to be. Could you maybe point me to the tool which will consume this file? I am interested why there is a field for include directories and additionally an entry for each include file.
Should `incdirs` be also set for `ibex_vip_rtl`?

vogelpi:
Hi Tobias, thanks for approving these changes. The tool consuming this file is here: https://github.com/pulp-platform/IPApproX

This tool generates the tcl scripts to setup the include directories and load the source files for many different design tools including Xilinx Vivdao. I think in Vivado, the package files need to be added as source file as well as being accessible through the include directories.

I think `ibex_vip_rtl` does not need to have a separate `incdirs` as it does not have additional include files in a different place. It will inherit the `incdirs` from `ibex`.

Best regards,
Pirmin

Issue #216, opened 2019-08-06
instr_multicycle_done is not asserted, continuous register writes


udinator:
While executing the code shown below:
![Screenshot from 2019-08-05 18-14-11](https://user-images.githubusercontent.com/16736281/62504340-e2de7b00-b7ac-11e9-8b26-0cc4090ccec3.png)
We see this incorrect behavior in the core:
![Screenshot from 2019-08-05 18-13-52](https://user-images.githubusercontent.com/16736281/62504356-ee31a680-b7ac-11e9-8d28-748683269f7e.png)

As shown, the `addi` instruction at PC `0x80000088` appears to correctly load register `x13` with the value `0xa5a5a5a5`, but at the very next instruction (PC: `0x8000008c`), we can see from the trace log that `x13` now has the value `0xa5a5de17`. 
![Screenshot from 2019-08-05 18-12-33](https://user-images.githubusercontent.com/16736281/62504456-76b04700-b7ad-11e9-94ab-524900644a48.png)

From the waveform, `rf_reg[13]` is written correctly by the `addi` after adding the immediate value `0x5a5` (`alu_operand_b_i`), and becomes `0xa5a5a5a5`. After this, the testbench does not assert the `instr_gnt_i` for a number of cycles, leading the core to stay in the DECODE phase while waiting for the next instruction fetch. However, we see that `rf_reg[13]` keeps getting written with the incremented values of `alu_result` in every consecutive cycle that the core waits, as `instr_multicycle_done_q` in the ID stage does not go high after the `addi` finishes. This keeps `regfile_we` at 1, causing the continuous writes.

vogelpi:
Hi @udinator ,
thanks for reporting the issue and tracking down the bug! `instr_multicycle_done_q` is only used for multicycle instructions and currently prevents `instr_executing` from being deasserted for normal instructions stalled due to the instruction fetch stage. I will prepare fix for it.
Best,
Pirmin

PR #222, opened 2019-08-06
Fix IF stalls preventing deassertion of regfile write enable

vogelpi:
This PR makes sure the signal `instr_multicycle_done_q` is only evaluated when the decoder currently has a multicycle instruction (load/store, mult/div, branch, jump). If this signal is evaluated
for single-cycle instructions, it can prevent the reg file write enable from being de-asserted in the case of stalls due to the IF stage.

@udinator: Could you please verify that this actually fixes the bug you reported in #216 before we merge it?

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -163,6 +163,7 @@ module ibex_id_stage #(
   logic        jump_set;
 
   logic        instr_executing;
+  logic        instr_multicycle;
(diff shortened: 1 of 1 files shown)


udinator:
@vogelpi This fix seems to have fixed the issue I reported yesterday. Thank you!

vogelpi:
Thanks for testing and for your feedback @udinator !

Issue #118, opened 2019-07-02
Writable mtvec


imphil:
`mtvec` is currently read-only and hardwired to the boot address. We need it writable.

ivanmgribeiro:
I've created a pull request (https://github.com/lowRISC/ibex/pull/193) to try to implement this.

PR #226, opened 2019-08-08
Make BASE field of `mtvec` CSR writeable

vogelpi:
This commit makes the BASE field of the `mtvec` CSR writable to
allow changing the trap-vector base address at runtime without
changing the boot address input. Similar to the boot address,
the trap-vector base address must always be aligned to 256 bytes.
At bootup, the trap-vector base address is initialized to the boot
address.

Most of this work has been done by @ivanmgribeiro.
This commit resolves lowrisc/ibex#118.

This supersedes PR #193 .


asb:
Could you please add a reference to #193 to the commit message, just so it's easier to find Ivan's original contribution?

vogelpi:
> Could you please add a reference to #193 to the commit message, just so it's easier to find Ivan's original contribution?

Yes, sure.

rtl/ibex_if_stage.sv
```diff
@@ -43,7 +43,7 @@ module ibex_if_stage #(
     input  logic                  instr_rvalid_i,
     input  logic [31:0]           instr_rdata_i,
 
-    // Output of IF Pipeline stage
+    // output of ID stage
```
asb: Even if bundled in this same PR (which I'm fine with), I think these non-functional comment fixes would be best in a separate commit. When looking through commit history, it's confusing when you see a mix of functional changes and non-functional unrelated improvements in a single commit.

imphil:
Thanks @ivanmgribeiro for your first larger contribution to Ibex!

rtl/ibex_if_stage.sv
vogelpi: I get your point @asb. I sometimes tend to do too much cleanup combined with other changes and mix it into the same commit. I will try to reduce/avoid this in the future.

PR #227, opened 2019-08-08
CSRs: add `irq_` prefix to `Interrupts_t` members

vogelpi:
This PR adds a prefix to the members of the `Interrupts_t` struct to avoid linting errors in AMS mode (timer is a key word).


vogelpi:
Thanks @tomroberts-lowrisc  for reviewing!

PR #232, opened 2019-08-09
Update ibex testbench

udinator:
- Added a new register file interface, to enable monitoring for any register state updates
- Added a new `core_ibex_csr_test`, to be run with the privileged CSR test generation in `riscv-dv`, that uses this regfile interface to monitor for writes to x3 at the end of test to determine success or failure
- Extended the `uvm_default_report_server` to create a custom reporting class that reports whether simulation has passed or failed (will report for all risc-v simulations, and is just a measure of whether the DV UVM simulation has completed without errors/failures, and as such is useful for self-checking tests)
- Modified interrupt driver to now assert interrupts for 2 cycles, which ensures that Ibex will actually respond to all asserted interrupts, as they are treated as level sensitive

### DIFF
--- a/dv/uvm/riscv_dv_extension/csr_description.yaml
+++ b/dv/uvm/riscv_dv_extension/csr_description.yaml
@@ -61,27 +61,28 @@
       msb: 25
       lsb: 0
 
+# TODO(udij) - clarify expected write behavior
 # MHARTID
-- csr: mhartid
-  description: >
-    Contains integer ID of hardware thread running code
-  address: 0xF14
-  privilege_mode: M
-  rv32:
-    - field_name: cluster_id
-      description: >
-        ID of the cluster
-      type: R
-      reset_val: 0
-      msb: 10
-      lsb: 5
-    - field_name: core_id
-      description: >
-        ID of the core within cluster
-      type: R
-      reset_val: 0
-      msb: 3
-      lsb: 0
+#- csr: mhartid
+#  description: >
+#    Contains integer ID of hardware thread running code
+#  address: 0xF14
+#  privilege_mode: M
+#  rv32:
+#    - field_name: cluster_id
+#      description: >
+#        ID of the cluster
+#      type: R
+#      reset_val: 0
+#      msb: 10
+#      lsb: 5
+#    - field_name: core_id
+#      description: >
+#        ID of the core within cluster
+#      type: R
+#      reset_val: 0
+#      msb: 3
+#      lsb: 0
 
 # MSTATUS
 - csr: mstatus
--- a/dv/uvm/tests/core_ibex_csr_test.sv
+++ b/dv/uvm/tests/core_ibex_csr_test.sv
@@ -1,0 +1,33 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class core_ibex_csr_test extends core_ibex_base_test;
+
+  `uvm_component_utils(core_ibex_csr_test)
+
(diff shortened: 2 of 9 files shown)


dv/uvm/env/core_ibex_reg_file_if.sv
```diff
@@ -0,0 +1,21 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Interface to probe ibex register file ports
+interface core_ibex_reg_file_if #(
```
taoliug: As discussed, please use a store instruction to write out the pass/fail signature rather than monitoring register file update

dv/uvm/tests/core_ibex_csr_test.sv
```diff
@@ -0,0 +1,34 @@
...
+
+  `uvm_component_utils(core_ibex_csr_test)
+
+  function new(string name="", uvm_component parent=null);
+    super.new(name, parent);
+  endfunction
+
+  virtual task wait_for_test_done();
+    fork
+    begin
```
taoliug: Use enum instead of hard coded value

dv/uvm/tests/core_ibex_csr_test.sv
```diff
@@ -0,0 +1,34 @@
...
+      wait(regfile_vif.we_a && regfile_vif.waddr_a === 3);
+      @(negedge regfile_vif.clk);
+      if (regfile_vif.wdata_a === 1) begin
+        `uvm_info(`gfn, "CSR test completed successfully", UVM_LOW)
+      end else if (regfile_vif.wdata_a === 2) begin
+        `uvm_error(`gfn, "CSR TEST FAILED")
+      end
+      wait(dut_vif.ecall === 1'b1);
```
taoliug: Do you still need to wait for ecall? I guess the test should be trapped by an infinite loop. If you use ecall to terminate the test, you need to have a trap handler for it as well.

dv/uvm/riscv_dv_extension/csr_description.yaml
```diff
@@ -62,26 +62,26 @@
...
-      msb: 10
-      lsb: 5
-    - field_name: core_id
-      description: >
-        ID of the core within cluster
-      type: R
-      reset_val: 0
-      msb: 3
-      lsb: 0
+#- csr: mhartid
```
taoliug: Please add a TODO here. After get clarified with the designer, you can add a comment for it.

dv/uvm/tests/core_ibex_csr_test.sv
```diff
@@ -0,0 +1,33 @@
...
+  `uvm_component_utils(core_ibex_csr_test)
+
+  function new(string name="", uvm_component parent=null);
+    super.new(name, parent);
+  endfunction
+
+  virtual task wait_for_test_done();
+    fork
+    begin
+      wait(dmem_vif.request && dmem_vif.we && dmem_vif.grant &&
```
taoliug: I think it's OK for now. Later we may need to change to a monitor or scoreboard to get the transactions from the dmem agent monitor directly. It's better to limit the interface protocol details in the agent itself.

dv/uvm/riscv_dv_extension/csr_description.yaml
udinator: Done

dv/uvm/tests/core_ibex_csr_test.sv
udinator: Sure, I'll keep that as TODO. Depending on how I handle EOT for interrupt tests, this change might be necessary then.

PR #233, opened 2019-08-09
Update ibex simulation flow

udinator:
Integrated ibex simulation flow with `riscv-dv` privileged CSR test
- Added options for CSR test generation script to `Makefile`
- Added `csr_description.yaml`, containing YAML descriptions of all implemented CSRs, as per `docs/cs_registers.rst` for use in privileged CSR testing
- Modified license of `testlist.yaml` to match standard Ibex licenses
- Modified `sim.py` to deal with the optional `no_post_compare` license
- Updated  `ibex_log_to_trace_csv.py` to add functionality to parse UVM sim logs for the new report messages

### DIFF
--- a/dv/uvm/Makefile
+++ b/dv/uvm/Makefile
@@ -2,35 +2,38 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-DV_DIR     := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-GEN_DIR    := $(realpath ${DV_DIR}/../../vendor/google_riscv-dv)
-TOOLCHAIN  := ${RISCV_TOOLCHAIN}
-OUT        := "${DV_DIR}/out"
+DV_DIR     		:= $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+GEN_DIR    		:= $(realpath ${DV_DIR}/../../vendor/google_riscv-dv)
+TOOLCHAIN  		:= ${RISCV_TOOLCHAIN}
+OUT        		:= "${DV_DIR}/out"
 # Run time options for the instruction generator
-GEN_OPTS   :=
+GEN_OPTS   		:=
 # Run time options for ibex RTL simulation
-SIM_OPTS   :=
+SIM_OPTS   		:=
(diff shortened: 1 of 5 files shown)


taoliug:
Please use commit --amend going forward.

asb:
And for this PR, merge and fixup commits make the history more confusing. Happy to offer got advice if you're not sure how to rebase.

dv/uvm/riscv_dv_extension/csr_description.yaml
```diff
@@ -0,0 +1,279 @@
...
+  privilege_mode: M
+  rv32:
+    - field_name: base
+      description: >
+        trap vector base address (256 byte aligned)
+      type: WARL
+      reset_val: 0x800000
+      msb: 31
+      lsb: 8
+    - field_name: base_zero_end
```
taoliug: I prefer to partition the field the same as spec. Is this creating issue to your test?

dv/uvm/Makefile
```diff
@@ -57,8 +57,12 @@ COMMON_OPTS=--seed=${SEED} \
 	python3 ./run.py \
      --o=${OUT}/instr_gen ${GEN_OPTS} \
      --steps=gen \
+		 --csr_yaml=${DV_DIR}/riscv_dv_extension/csr_description.yaml \
```
taoliug: alignment, here can below

dv/uvm/riscv_dv_extension/csr_description.yaml
udinator: Yes, somewhat, as the Ibex spec requires `mtvec[7:2]` to be set to `6'b0`, and if this is partitioned the same way as the spec, the gen_csr script does not have a way to tell which sections of fields should be hardwired or not, so decided to make this a separate field.
I could add a `mask` value for each CSR field to solve this, which would convey to the script if any bits are hardwired or not, as well as providing the script a r/w mask, saving the computation of having to calculate the mask from the script itself. What do you think?

taoliug: Adding a bitmask sounds like a good idea, you can do it in a later PR, as it requires changes in the test as well.

dv/uvm/Makefile
udinator: Done.

PR #243, opened 2019-08-15
Add basic debug test and modify sim flow

udinator:
- Modified sim.py to correctly process any sim_opts and compare_opts passed specified in the testlist
- Removed shared utility functions from sim.py (duplicates of functions in lib.py)
- Changed location of vseq.stop() call in testbench to prevent debug or interrupt signal assertions after ecall has been detected

### DIFF
--- a/dv/uvm/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/riscv_dv_extension/testlist.yaml
@@ -102,10 +102,30 @@
     +no_ebreak=0
   rtl_test: core_ibex_base_test
 
-- test: riscv_ebreak_debug_mode_test
+- test: riscv_debug_basic_test
+  description: >
+    Randomly assert debug_req_i, random instruction sequence in debug_rom section
+  iterations: 10
+  gen_test: riscv_instr_base_test
+  gen_opts: >
+    +no_ebreak=1
+    +instr_cnt=6000
+    +no_branch_jump=1
+    +no_csr_instr=1
+    +no_fence=1
+    +num_of_sub_program=0
+  rtl_test: core_ibex_base_test
+  sim_opts: >
+    +enable_debug_seq=1
+  compare_opts:
+    compare_final_value_only: 1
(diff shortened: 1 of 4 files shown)


dv/uvm/riscv_dv_extension/testlist.yaml
```diff
@@ -102,7 +102,27 @@
...
+    +no_fence=1
+    +num_of_sub_program=0
+  rtl_test: core_ibex_base_test
+  sim_opts: >
+    +enable_debug_seq=1
+  compare_opts: >
+    compare_final_value_only=1
+    verbose=1
+
+# TODO(udij) - this test is later in testlist - ignore for now
```
taoliug: If it's ignored, should the iteration be set to 0 for now?

udinator: Sure, good point.

dv/uvm/sim.py
```diff
@@ -222,7 +189,17 @@ def compare(test_list, iss, output_dir, verbose):
         else:
           print("Unsupported ISS" % iss)
           sys.exit(1)
-        compare_trace_csv(rtl_csv, iss_csv, "ibex", iss, report)
+        # construct trace script call from list of strings and call run_cmd()
+        compare_cmd = ("python3 %s/instr_trace_compare.py " % gen_dir) + \
```
taoliug: What's the reason for this change? I'd prefer just do a functional call rather than running a bash command.

udinator: I changed this to match the rest of the various simulation scripts, and just using a functional call makes argument parsing more wordy, as well as the fact that the argparser  in instr_trace_compare.py will not be needed with that approach

taoliug: I think each python script should be able to run independently(argparser is needed in this case), and it can also be included in other python program.  I think we have already avoided running another python script with a bash command. The run_cmd is used for non-python programs.

Issue #6, opened 2018-04-25
why mstatus accessed  by CSR instruction will flush prefetch fifo


pearson78:
Please refer the line 306 in “zeroriscy_controller.sv”, what purpose is for the csr_status_i (mstatus accessed  by CSR instruction) will enter the FLUSH state?

              else if (~branch_set_i & ~jump_set_i & (mret_insn_i | ecall_insn_i | pipe_flush_i | ebrk_insn_i | illegal_insn_i | csr_status_i))
              begin
                ctrl_fsm_ns = FLUSH;
                halt_if_o   = 1'b1;
                halt_id_o   = 1'b1;
              end

vogelpi:
Hi @pearson78 ,
thanks for reporting this issue. I know it's a very long time but we had absolutely no clue why the pipeline flush was there. It was not documented anywhere and we only recently found out what it is needed for.

The reason for the pipeline flush being inserted when the `mstatus` CSR is accessed is to not miss pending interrupts when quickly enabling and disabling them right away. Imagine interrupts are globally disabled and before entering a longer phase that should not be interrupted, the core quickly wants to check if there are any pending interrupts.

1. To this end, it uses a CSR instruction to switch the MIE bit in `mstatus` to 1.
2. Only in the next cycle, the controller then sees if there are any interrupts pending. But since the core is already executing the CSR instruction to again set the MIE bit in `mstatus` to 0 and since the controller always gives priority to the instruction currently being executed (an interrupt should not abort an instruction and mess up the core state), the controller does not see the pending interrupt.
3. In the next cycle, interrupts are again disabled.

To actually give the controller a chance to see a pending interrupt, a NOP between 1. and 2. is needed. This is what the pipeline flush has been introduced for.

Best regards,
Pirmin

PR #244, opened 2019-08-15
Rework CSR-related pipeline flushes

vogelpi:
This commit clarifies why CSR-related pipeline flushes are needed (e.g. when enabling interrupts), only introduces them when doing the critical modifications (write/set bits in `mstatus` and `mie` CSRs
for enabling interrupts, reads and clears are uncritical for these CSRs), and makes sure the controller is actually able to start handling interrupts while doing a CSR-related pipeline flush.

This resolves lowrisc/ibex#6.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -82,7 +82,7 @@ module ibex_decoder #(
     // CSRs
     output logic                 csr_access_o,          // access to CSR
     output ibex_pkg::csr_op_e    csr_op_o,              // operation to perform on CSR
-    output logic                 csr_status_o,          // access to xstatus CSR
+    output logic                 csr_pipe_flush_o,      // CSR-related pipeline flush
 
     // LSU
     output logic                 data_req_o,            // start transaction to data memory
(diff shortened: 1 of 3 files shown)


rtl/ibex_decoder.sv
```diff
@@ -159,6 +159,31 @@ module ibex_decoder #(
     end
   end
 
+  /////////////////////////////////
+  // CSR-related pipline flushes //
+  /////////////////////////////////
+  always_comb begin : csr_pipeline_flushes
+    csr_pipe_flush_o = 1'b0;
+
+    // a pipeline flush is needed to let the controller react after modifying certain CSRs:
```
imphil: Can you make this line and the lines below full sentences?

a pipeline -> A pipeline ...

(etc.)

rtl/ibex_decoder.sv
```diff
@@ -159,6 +159,31 @@ module ibex_decoder #(
...
+    // - when enabling interrupts, the controller only sees pending IRQs during the next cycle
+    //   if during that cycle the core disables interrupts again, it does not see pending IRQs
+    // - when modifying debug CSRs (not sure whether this is really needed)
```
imphil: Add a "TODO: Check if this is really needed" to help finding this comment more easily.

rtl/ibex_decoder.sv
```diff
@@ -159,6 +159,31 @@ module ibex_decoder #(
...
 
+  /////////////////////////////////
+  // CSR-related pipline flushes //
+  /////////////////////////////////
+  always_comb begin : csr_pipeline_flushes
+    csr_pipe_flush_o = 1'b0;
+
+    // a pipeline flush is needed to let the controller react after modifying certain CSRs:
+    // - when enabling interrupts, the controller only sees pending IRQs during the next cycle
+    //   if during that cycle the core disables interrupts again, it does not see pending IRQs
```
imphil: Can you be more specific about what it means to "see pending IRQs"? Does that mean the ISR is called, or does that mean the interrupt is visible in the interrupt pending registers?

vogelpi:
Thanks for your feedback @imphil . I will now integrate that and then merge the PR.

rtl/ibex_decoder.sv
vogelpi: Done.

rtl/ibex_decoder.sv
vogelpi: Done.

rtl/ibex_decoder.sv
vogelpi: Done.

PR #245, opened 2019-08-15
Adapt line numbers in Verilator lint waiver

vogelpi:
This PR contains a single commit to adapt some line numbers in the Verilator lint waiver file. These
got out of sync with the RTL recently.


vogelpi:
Thanks for approving @imphil !

PR #247, opened 2019-08-19
CI: Use Azure Pipelines to run lint and some DV

imphil:
This sets up Azure Pipelines to run the following two tasks:

- Run Verilator lint on the SystemVerilog code.
- Run the RISC-V compliance test suite for RV32IMC


imphil:
This isn't the most efficient way to do CI with re-building Verilator each time, but good enough for now.

imphil:
Also out of scope for this PR: improved test reporting. (Azure Pipelines can actually report passed/failed tests individually if you pass it a JUnit-compatible test report, or similar.)

PR #253, opened 2019-08-21
Consolidate some debug generation options, and make the signature_add…

udinator:
…r handshake optional


dv/uvm/tests/core_ibex_base_test.sv
```diff
@@ -15,6 +15,10 @@ class core_ibex_base_test extends uvm_test;
   bit                                             enable_debug_seq;
   irq_seq                                         irq_seq_h;
   int unsigned                                    timeout_in_cycles = 2000000;
+  // If no signature_addr handshake functionality is desired between the
```
taoliug: do you expect to use this mode for ibex?

udinator: Probably not, I just wanted to include the functionality to show how it would be used

PR #271, opened 2019-08-29
Update ibex_register_file_latch.sv

pbing:
Fix latch to transparent on low

### DIFF
--- a/rtl/ibex_register_file_latch.sv
+++ b/rtl/ibex_register_file_latch.sv
@@ -122,7 +122,7 @@ module ibex_register_file #(
     mem[0] = '0;
 
     for (int k = 1; k < NUM_WORDS; k++) begin : latch_wdata_word_iter
-      if (mem_clocks[k]) begin
+      if (!mem_clocks[k]) begin
         mem[k] = wdata_a_q;
       end
     end


vogelpi:
Hi @pbing,
thanks for creating this pull request. Could you please explain the motivation for this change?
Best,
Pirmin

pbing:
Because to match the comment in line 137.

vogelpi:
Thanks for getting back. I see your point. I think we should rather change the comment then. What do you think?

The reason I am asking is that we can't test it here using Verilator but the implementation of the latch-based register file has been like this for ages, it is used by many people and it is also the same for RI5CY. So I assume it is actually right and just the comment is wrong.

pbing:
Looks like the comment was right (figure 1)
https://www.edn.com/design/integrated-circuit-design/4368424/Latches-and-timing-closure-a-mixed-bag

Am Do., 29. Aug. 2019 um 11:41 Uhr schrieb Pirmin Vogel <
notifications@github.com>:

vogelpi:
Thanks for providing the reference.

In my opinion, the time borrowing feature is not relevant in the latch-based register file of Ibex/RI5CY for the following reasons:

- There is no subsequent latch-to-latch path from which a long combinational path could borrow time from.
- Path 1 in Figure 2 contains no logic in the case of the register file. Register (1) is `wdata_a_q` and the latch (2) is `mem[k]`. They are directly connected without logic in between. 
- The critical path of Ibex is not in the register file. It starts at the register file output and goes through the ALU into the prefetch buffer. Changing the latches to be transparent on low, would not help there.

I more see a problem with this PR as it only changes the latches to be transparent on low but does not drive the latch clock high when the latches are not written. If I understand it correctly, this change would constantly overwrite all registers with whatever data is in `wdata_a_q`. Please check the timing diagram in Figure 6 [here](https://ibex-core.readthedocs.io/en/latest/register_file.html) for illustration. Did you actually test this change?

Best regards,
Pirmin

pbing:
Thank you for providing the link to the documentation.

I did not simulate the design but only synthesized it. Your are right that time-borring is not relevant in this case because the path from wdata_a_q/Q to mem/D is back-to-back. On the other hand I got now only a half period for setup time from mem/QN to mhpmcounter_q/D. The resulting margin was only 22 ps for TSMC CL018 and fclk = 25 MHz.

Using the original latch with transparent on high the slack of the above path was 14.6 ns.

I also think that the comment should be changed. You can close this pull request.

vogelpi:
Thanks for getting back @pbing. I will prepare a commit to fix the comments.

PR #273, opened 2019-08-29
Update comments in latch-based register file

vogelpi:
This commit updates the comments inside the latch-based register file. Some of them were outdated or just wrong.


vogelpi:
This PR superseedes lowRISC/ibex#271.

PR #276, opened 2019-08-30
[RTL] - Remove timing loop in LSU

tomroberts-lowrisc:
- External signals data_rvalid_i and data_err_i were factored
  into the external data_req_o signal
- To improve timing, these signals are decoupled
- The second part of an unaligned transaction will now be issued
  even if the first received an error response
- The state machine will service the abandoned requests
- pmp_err_q fixed to only update at specific times

### DIFF
--- a/doc/load_store_unit.rst
+++ b/doc/load_store_unit.rst
@@ -48,6 +48,10 @@ The LSU is able to handle misaligned memory accesses, meaning accesses that are 
 However, it does so by performing two separate word-aligned accesses.
 This means that at least two cycles are needed for misaligned loads and stores.
 
+If an error response is received for the first transaction, the second transaction will still be issued.
+The second transaction will then follow the normal bus protocol, but its response/data will be ignored.
+If a new load/store request is received while waiting for an abandoned second part to complete, it will not be serviced until the state machine returns to IDLE.
+
 .. _lsu-protocol:
 
 Protocol
@@ -60,6 +64,8 @@ The protocol that is used by the LSU to communicate with a memory works as follo
 2. After receiving a grant, the address may be changed in the next cycle by the LSU. In addition, the ``data_wdata_o``, ``data_we_o`` and ``data_be_o`` signals may be changed as it is assumed that the memory has already processed and stored that information.
 
 3. The memory answers with a ``data_rvalid_i`` set high for exactly one cycle to signal the response from the bus or the memory using ``data_err_i`` and ``data_rdata_i`` (during the very same cycle). This may happen one or more cycles after the grant has been received. If ``data_err_i`` is low, the request could successfully be handled at the destination and in the case of a load, ``data_rdata_i`` contains valid data. If ``data_err_i`` is high, an error occurred in the memory system and the core will raise an exception.
+
+4. When multiple granted requests are outstanding, it is assumed that the memory requests will be kept in-order and one ``data_rvalid_i`` will be signalled for each of them, in the order they were issued.
 
 :numref:`timing1`, :numref:`timing2` and :numref:`timing3` show example-timing diagrams of the protocol.
 
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -76,11 +76,13 @@ module ibex_load_store_unit (
   logic         split_misaligned_access;
   logic         handle_misaligned_q, handle_misaligned_d; // high after receiving grant for first
                                                           // part of a misaligned access
+  logic         pmp_err_d;
   logic         pmp_err_q;
   logic         data_or_pmp_err;
 
   typedef enum logic [2:0]  {
-    IDLE, WAIT_GNT_MIS, WAIT_RVALID_MIS, WAIT_GNT, WAIT_RVALID
+    IDLE, WAIT_GNT_MIS, WAIT_RVALID_MIS, WAIT_GNT, WAIT_RVALID,
+    WAIT_GNT_ERR, WAIT_RVALID_ERR, WAIT_RVALID_DONE
   } ls_fsm_e;
 
   ls_fsm_e ls_fsm_cs, ls_fsm_ns;
@@ -321,12 +323,14 @@ module ibex_load_store_unit (
     addr_incr_req_o     = 1'b0;
     handle_misaligned_d = handle_misaligned_q;
     data_or_pmp_err     = 1'b0;
+    pmp_err_d           = pmp_err_q;
 
     unique case (ls_fsm_cs)
 
       IDLE: begin
         if (data_req_ex_i) begin
           data_req_o = 1'b1;
+          pmp_err_d  = data_pmp_err_i;
           if (data_gnt_i) begin
             handle_misaligned_d = split_misaligned_access;
             ls_fsm_ns           = split_misaligned_access ? WAIT_RVALID_MIS : WAIT_RVALID;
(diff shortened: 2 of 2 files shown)


rtl/ibex_load_store_unit.sv
```diff
@@ -451,7 +503,8 @@ module ibex_load_store_unit (
   assign load_err_o    = data_or_pmp_err & ~data_we_q;
   assign store_err_o   = data_or_pmp_err &  data_we_q;
 
-  assign busy_o = (ls_fsm_cs == WAIT_RVALID) | (data_req_o == 1'b1);
+  assign busy_o = (ls_fsm_cs == WAIT_RVALID) | (ls_fsm_cs == WAIT_RVALID_ERR) |
+                  (ls_fsm_cs == WAIT_RVALID_DONE) | (data_req_o == 1'b1);
```
vogelpi: Should we just change this to `assign busy_o = (ls_fsm_cs != IDLE);`?

rtl/ibex_load_store_unit.sv
```diff
@@ -397,6 +411,44 @@ module ibex_load_store_unit (
...
+        if (data_rvalid_i) begin
+          // Update the pmp error for the second part
+          pmp_err_d = data_pmp_err_i;
+          // The first part cannot see a PMP error in this state
+          if (data_err_i) begin
+            // first part created an error, abort transaction and wait for second rvalid
+            data_valid_o        = 1'b1;
+            data_or_pmp_err     = 1'b1;
+            handle_misaligned_d = 1'b0;
+            ls_fsm_ns           = WAIT_RVALID_ERR;
```
vogelpi: I think this will lock up the whole core: in case you get a bus error in the second part of a misaligned transaction (signaled together with `rvalid_i`) there will be no other `rvalid_i` from the bus. The FSM will stay inside `WAIT_RVALID_ERR` forever.

doc/load_store_unit.rst
```diff
@@ -48,6 +48,10 @@ The LSU is able to handle misaligned memory accesses, meaning accesses that are
 However, it does so by performing two separate word-aligned accesses.
 This means that at least two cycles are needed for misaligned loads and stores.
 
+If an error response is received for the first transaction, the second transaction will still be issued.
+The second transaction will then follow the normal bus protocol, but it's response/data will be ignored.
```
imphil: it's -> its

rtl/ibex_load_store_unit.sv
```diff
@@ -370,7 +383,10 @@ module ibex_load_store_unit (
           end
         end else begin
           // first part rvalid is NOT received
-          ls_fsm_ns = WAIT_RVALID_MIS;
+          if (data_gnt_i) begin
+            // second grant is received
+            ls_fsm_ns = WAIT_RVALID_DONE;
+          end
         end
       end
```
vogelpi: The FSM was not very nice before I know, but I don't like that the super complex `WAIT_RVALID_MIS` state we have now. There are now three nested levels of if/else and so many statements, it is really hard to follow. You should at least add an empty line here and there to visually group what does (NOT) belong together. For example: what is below Line 357 does not depend on what is above. The same holds for Line 384.

doc/load_store_unit.rst
tomroberts-lowrisc: fixed

rtl/ibex_load_store_unit.sv
tomroberts-lowrisc: agreed

rtl/ibex_load_store_unit.sv
tomroberts-lowrisc: You're right, treating the two errors differently was over-complicating things. I have made the behaviour the same and simplified the WAIT_RVALID_MIS state a bit.

rtl/ibex_load_store_unit.sv
tomroberts-lowrisc: why wouldn't there be a second rvalid_i?

vogelpi: You are right. I thought the first `rvalid_i` would have been received already when jumping into `WAIT_RVALID_DONE`. Sorry!

rtl/ibex_load_store_unit.sv
vogelpi: Well done, it looks much better now!

rtl/ibex_load_store_unit.sv
```diff
@@ -345,32 +349,34 @@ module ibex_load_store_unit (
...
 
       WAIT_RVALID_MIS: begin
+        // push out second request, unless the first request was suppressed by a PMP error
+        data_req_o = 1'b1;
         // tell ID/EX stage to update the address
         addr_incr_req_o = 1'b1;
+
         // first part rvalid is received, or gets a pmp error
```
vogelpi: pmp -> PMP

rtl/ibex_load_store_unit.sv
```diff
@@ -397,6 +401,44 @@ module ibex_load_store_unit (
         end
       end
 
+      WAIT_GNT_ERR: begin
+        // Wait for the grant of the abandoned second access
+        data_req_o = 1'b1;
+        // tell ID/EX stage to update the address
+        addr_incr_req_o = 1'b1;
+        if (pmp_err_q) begin
+          // The second part was suppressed by a pmp error
```
vogelpi: pmp -> PMP

rtl/ibex_load_store_unit.sv
```diff
@@ -345,32 +349,34 @@ module ibex_load_store_unit (
...
+        data_req_o = 1'b1;
         // tell ID/EX stage to update the address
         addr_incr_req_o = 1'b1;
+
         // first part rvalid is received, or gets a pmp error
-        // pmp_err_i will hold stable until the address is updated, and
-        // therefore pmp_err_q is valid in both WAIT_GNT_MIS and WAIT_RVALID_MIS states
         if (data_rvalid_i || pmp_err_q) begin
+          // Update the PMP error for the second part
+          pmp_err_d = data_pmp_err_i;
```
vogelpi: Question: shouldn't we always update `pmp_err_d` with `data_pmp_err_i` in this state, in particular also when we do not get the `rvalid_i`, i.e., when we go to line 374? As far as I understand, we do a new PMP check in this state and would like to buffer the outcome of that check.

tomroberts-lowrisc: No we can't update it yet - if the first request did not get a PMP error (and is just sitting there waiting for rvalid) but the second request will get a PMP error, if we update the pmp_err_q flop here, then we will see pmp_err_q set on the following cycle and assume it relates to the first request and take the error without waiting for the first rvalid to come back.

vogelpi: I see, thanks for clarifying this to me.

arjanbink:
Thank you for the proposed update. I don't see anything wrong with it (and it would remove the combinational path on the data interface that I asked about; but of course not yet the similar path on the instruction side). 

It would help me however if you could describe the LSU protocol with respect to data_pmp_err_i signal (I see pmp_err_q being used together with data_err_i, data_gnt_i, as well as data_rvalid_i).

tomroberts-lowrisc:
> Thank you for the proposed update. I don't see anything wrong with it (and it would remove the combinational path on the data interface that I asked about; but of course not yet the similar path on the instruction side).
> 
> It would help me however if you could describe the LSU protocol with respect to data_pmp_err_i signal (I see pmp_err_q being used together with data_err_i, data_gnt_i, as well as data_rvalid_i).

I'll add some extra comments to describe the timing / functionality of the two types of error

vogelpi:
Thanks @arjanbink for your feedback and @tomroberts-lowrisc for adding the clarifying comment.

Shall we finally merge this PR?

arjanbink:
Okay with me; thank you for the quick follow-up

PR #277, opened 2019-08-30
CSR: Access checks on Debug CSRs

### LINKED ISSUE #275: Accessing debug CSRs while not in debug mode
The RISC-V Debug Spec v.0.13.2 (p.41) mandates that the core debug CSRs `dcsr`, `dpc`, `dscratch0` and `dscratch1` must not be accessible if not in debug mode. Currently, Ibex always allows accesses to these. This should be changed.

See also pulp-platform/riscv#136.

raulbehl:
- The RISC-V Debug Spec v.0.13.2 (p.41) mandates that the core
    debug CSRs dcsr, dpc, dscratch0 and dscratch1 must not be
    accessible if not in debug mode. Fixes #275

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -57,6 +57,7 @@ module ibex_cs_registers #(
     output logic [33:0]          csr_pmp_addr_o [PMPNumRegions],
 
     // debug
+    input  logic                 debug_mode_i,
     input  ibex_pkg::dbg_cause_e debug_cause_i,
     input  logic                 debug_csr_save_i,
     output logic [31:0]          csr_depc_o,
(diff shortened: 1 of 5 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -57,6 +57,7 @@ module ibex_cs_registers #(
     output logic [33:0]          csr_pmp_addr_o [PMPNumRegions],
 
     // debug
+    input  logic                 debug_req_i,
```
imphil: debug_req_i is only the debug request, a kind of an interrupt. You need to check for debug_mode instead (you can find this signal in the controller).

raulbehl:
CI failing due to the following issue:

Screenshot 2019-08-30 at 9 26 36 PM

imphil:
Thanks for your patch! 

CI fails because the waiver file in lint/verilator_waiver.vlt needs updating (it contains file numbers which changed in your patch). Let me know if you have trouble running lint, and I can help you out there.

rtl/ibex_cs_registers.sv
raulbehl: Thank you for pointing this out. I should have checked a bit more on the surrounding logic before using this signal. I've updated the patch now to use the suggested signal.

raulbehl:
Have pushed a fix for the lint checks and see them passing now. :) 
The fix also updates the logic to use debug mode signal.

PR #279, opened 2019-09-03
Replace cluster_id/core_id with hart_id (Fixes #29)

### LINKED ISSUE #29: Replace core_id and cluster_id parameters in favour of a single hartid parameter
Ibex should allow system integrators to directly specify the hartid using whatever scheme makes most sense for their design. Currently core_id and cluster_id must be specified - this makes sense for the PULP use case, but will require some gymnastics for users who have a different hartid generation scheme.

[EDIT: "gymnastics" is hyperbolic given they're just concatenated - but you get my point. An integrator may also want a hartid that is greater than 10 bits in length]

GregAC:
Uncertain if we actually want a HartIDWidth parameter or just hard-code it to 32 bit.

### DIFF
--- a/doc/integration.rst
+++ b/doc/integration.rst
@@ -28,8 +28,7 @@ Instantiation Template
       .test_en_i      (),
 
       // Configuration
-      .core_id_i      (),
-      .cluster_id_i   (),
+      .hart_id_i      (),
       .boot_addr_i    (),
 
       // Instruction memory interface
@@ -90,7 +89,6 @@ Parameters
 | ``DmExceptionAddr``   | int         | 0x1A110808 | Address to jump to when an exception occurs while in debug mode |
 +-----------------------+-------------+------------+-----------------------------------------------------------------+
 
-
 Interfaces
 ----------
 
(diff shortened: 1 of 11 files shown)


imphil:
I'd go for simplicity and simply hardcode the width to 32 bit. Everything else will be synthesized out anyway.

GregAC:
Yeah that was my feeling too, I had mostly done the change at that point though so thought I'd see what was preferred.

doc/integration.rst
```diff
@@ -103,10 +101,8 @@ Interfaces
 +-------------------------+-------------------------+-----+----------------------------------------+
 | ``test_en_i``           | 1                       | in  | Test input, enables clock              |
 +-------------------------+-------------------------+-----+----------------------------------------+
-| ``core_id_i``           | 4                       | in  | Core ID, usually static, can be read   |
+| ``hart_id_i``           | 32                      | in  | hart ID, usually static, can be read   |
```
imphil: tiny nit: Everything in this column starts with a capital letter, so you might want to do the same here.

rtl/ibex_core.sv
```diff
@@ -27,9 +27,8 @@ module ibex_core #(
 
     input  logic        test_en_i,     // enable all clock gates for testing
 
-    // Core ID, Cluster ID and boot address are considered more or less static
-    input  logic [ 3:0] core_id_i,
-    input  logic [ 5:0] cluster_id_i,
+    // Hart ID and boot address are considered more or less static
```
imphil: That's a pretty strange comment, how about removing it? (Or the other way around: what does "more or less static" mean? You can change it, but then everything breaks?)

GregAC: Well I was just replacing the existing comment to change the Core ID, Cluster ID to Hart ID.  I don't thing anything would break if the signal wasn't static for hart_id_i.  Uncertain about the boot address but it's probably ok

imphil: Yeah, I know the comment was there before. Feel free to leave it or remove it, it just sounds strange and half-useless.

GregAC: Well nothing will obviously go wrong if someone goes for a dynamically changing boot address, comment removed.

imphil:
Thanks! Please feel free to rebase-merge once CI is done.

asb:
I think for GitHub to automatically close #29 when this lands, you need to say something like "Fixes #29" in the commit message. Just referencing the issue number isn't enough.

GregAC:
Though that was just the title of the pull request?  Guess I'll need to push a new version of the commit with the commit message changed?  I'll just close the issue manually for now and remember next time

Issue #261, opened 2019-08-23
[ibex tracer] enhancement request - runtime knob to disable tracer


sriyerg:
At chip level DV, we "stub" the ibex cpu by forcing its clk_i to 0 and taking control of the TL interface by overdriving it with our TL transactor. The advantage of doing that is choosing to 'stub' the cpu becomes a runtime config knob / setting as opposed to being a build time, which reduces overhead in terms of requiring additional parallel builds. 

In chip level dv tests where the ibex core is stubbed, the ibex tracer still appears to be running. I would like to request that it be enhanced through a runtime plusarg setting (call it "enable_ibex_tracer" or something similar), that is set to 1 by default. Users can override it via plusarg to set it to 0. If set to 0, then the tracer should stay dormant (and it should not even create the log file). 

Thanks
Sri

raulbehl:
This should be fixed with #360 
With this implementation in place - the file would only be created if the core retires valid instructions. 

@imphil - Should be okay to mark this as complete? Or you reckon some further updates might be necessary? 
@sriyerg - Hope the above implementation meets your criteria. Basically, gating the clock to ibex would ensure that there is no trace generated hence no trace log as well.

imphil:
@raulbehl thanks for reminding me. Yes, this should be fixed. Please re-open if there are additional issues @sriyerg .

Issue #197, opened 2019-07-30
Ibex tracer - compressed instructions


udinator:
Hi,
Would it be possible to have the Ibex tracer indicate that an instruction is compressed, and if so, log the compressed instruction name? This would bring the log file closer in similarity to Spike log files, and would make debugging regression failures a bit easier to do.
Not sure if this complies with RVFI spec though.
Thanks!
Udi

imphil:
Sure, that's possible. We have the `rvfi_insn` signal, which contains the original (compressed) instruction. (The non-standard `rvfi_insn_uncompressed` signal always contains the uncompressed instruction. This one is currently used by ibex_tracer.)

We have two options going forward:
- Either you open a PR which extends `ibex_tracer` to include the compressed instruction.
- Or you let me know how you'd like the logs to look and I'll implement it.

udinator:
Current ibex trace logs are below:
![Screenshot from 2019-07-31 10-37-14](https://user-images.githubusercontent.com/16736281/62234432-867bf580-b37f-11e9-96b6-ecf0d713ba7d.png)
![Screenshot from 2019-07-31 10-38-44](https://user-images.githubusercontent.com/16736281/62234439-8aa81300-b37f-11e9-9b39-0e7283f2f886.png)

I was just referring to denoting the instructions as compressed (such as c.addi, etc...) like Spike logs do:
![Screenshot from 2019-07-31 10-37-00](https://user-images.githubusercontent.com/16736281/62234481-9d224c80-b37f-11e9-9523-acf7c7fcad10.png)
![Screenshot from 2019-07-31 10-38-28](https://user-images.githubusercontent.com/16736281/62234490-a3182d80-b37f-11e9-88ef-8cb01cb9be9f.png)

I was just thinking it could be efficient for the ibex logs to show the compressed instruction name (c.li, c.addi, c.lui, etc...) to match what Spike shows and give a better idea of which instructions are compressed or not. Thanks!

PR #280, opened 2019-09-03
Add verilator support to Ibex tracer

raulbehl:
- Removed the tracer class in the ibex_tracer file and implemented class methods and properties as functions and global variables
- Updated the class new method with init function which implements the same functionality
- Updated usage of queues to systemverilog arrays of structs
- Updated $fwrite function to use verilator friendly string formatting
- Added verilator top-tb which performs the same sequence of instructions as the system-verilog top tracing TB
- Updated top_tracing_sim.core file to support verilator as well

### DIFF
--- a/examples/sim/tb/ibex_tracing_tb.cpp
+++ b/examples/sim/tb/ibex_tracing_tb.cpp
@@ -1,0 +1,115 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stdlib.h>
+#include <iostream>
+#include <utility>
+#include <string>
+#include "Vibex_tracing_tb.h"
+#include "verilated_vcd_c.h"
+
+vluint64_t sim_time = 0;
+
+// This function is needed to ensure verilator
+// knows the current simulation time
+double sc_time_stamp () {
+  return sim_time;
+}
+
+int main(int argc, char **argv) {
+  Verilated::commandArgs(argc, argv);
+  Verilated::traceEverOn(true);
+
+  Vibex_tracing_tb *top = new Vibex_tracing_tb;
+  VerilatedVcdC* tfp = new VerilatedVcdC;
+
+  top->trace (tfp, 99);
+  tfp->open ("ibex_trace.vcd");
+
+  top->clk   = 0;
+  top->rst_n = 0;
+
+  while (!Verilated::gotFinish()) {
+    if (sim_time == 160) {
+      top->rst_n = 1;
+    } else if (sim_time == 200) {
+      top->instr_rdata  = 0x81868106;
+      top->instr_rvalid = 0;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 210) {
+      top->instr_rvalid = 1;
+      top->instr_gnt    = 0;
+    } else if (sim_time == 220) {
+      top->instr_rvalid = 0;
+    } else if (sim_time == 250) {
+      top->instr_rdata  = 0x00400113;
+      top->instr_rvalid = 0;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 260) {
+      top->instr_rvalid = 1;
+      top->instr_gnt    = 0;
+    } else if (sim_time == 270) {
+      top->instr_rvalid = 0;
+      top->instr_rdata  = 0xff810113;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 280) {
+      top->instr_rvalid = 1;
+      top->instr_gnt    = 0;
+    } else if (sim_time == 290) {
+      top->instr_rvalid = 0;
+      top->instr_rdata  = 0x4000006f;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 300) {
+      top->instr_rvalid = 1;
+      top->instr_gnt    = 0;
+    } else if (sim_time == 310) {
+      top->instr_rvalid = 0;
+    } else if (sim_time == 320) {
+      top->instr_rdata  = 0x039597b3;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 330) {
+      top->instr_rvalid = 1;
+      top->instr_gnt    = 1;
+    } else if (sim_time == 340) {
+      top->instr_rdata  = 0x13410d13;
+    } else if (sim_time == 350) {
+      top->instr_rdata  = 0xe1070713;
+    } else if (sim_time == 360) {
+      top->instr_rdata  = 0xfff7c793;
+    } else if (sim_time == 370) {
+      top->instr_rdata  = 0x00000013;
+    } else if (sim_time == 380) {
+      top->instr_rdata  = 0x002d2c23;
+    } else if (sim_time == 390) {
+      top->instr_rdata  = 0x00000013;
+    } else if (sim_time == 400) {
+      top->instr_rdata  = 0x000d2083;
+      top->data_rdata   = 0x12345678;
+    } else if (sim_time == 410) {
+      top->instr_rdata  = 0x60008113;
+    } else if (sim_time == 420) {
+      top->instr_rdata  = 0x00000013;
+    } else if (sim_time == 430) {
+      top->instr_rdata  = 0x002d2023;
+    } else if (sim_time == 450) {
+      top->instr_rdata  = 0x0000000f;
+    } else if (sim_time == 460) {
+      top->instr_rdata  = 0x00000113;
+    } else if (sim_time == 490) {
+      top->instr_rdata  = 0x00000013;
+    } else if (sim_time == 500) {
+      top->instr_rdata  = 0x0000000f;
+    }
+    if (sim_time == 560) {
+      top->sim_done = 1;
+    }
+    top->clk = !top->clk;
+    top->eval();
+    tfp->dump (sim_time);
+    sim_time += 5;
+  }
+  tfp->close();
+  delete top;
+  exit(0);
+}
--- a/rtl/ibex_core_tracing.sv
+++ b/rtl/ibex_core_tracing.sv
@@ -65,6 +65,8 @@ module ibex_core_tracing #(
     Fatal error: RVFI needs to be defined globally.
   `endif
 
+  parameter int unsigned RegAddrWidth = RV32E ? 4 : 5;
+
   logic        rvfi_valid;
   logic [63:0] rvfi_order;
   logic [31:0] rvfi_insn;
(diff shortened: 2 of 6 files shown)


rtl/ibex_tracer.sv
```diff
@@ -43,373 +43,392 @@ module ibex_tracer #(
...
+        if (regs_read[i].addr != 0) begin
+          $fwrite(f, " %s:0x%08x", regAddrToStr(regs_read[i].addr), regs_read[i].value);
         end
+      end
 
-        if (mem_access.size() > 0) begin
-          mem_acc = mem_access.pop_front();
-
-          $fwrite(f, " PA:0x%08x", mem_acc.addr);
+      foreach(mem_access[i]) begin
```
imphil: nit: space after `foreach`

rtl/ibex_tracer.sv
```diff
@@ -43,373 +43,392 @@ module ibex_tracer #(
...
     end
   end
 
   // open/close output file for writing
   initial begin
+  `ifndef VERILATOR
     wait(rst_ni == 1'b1);
     wait(fetch_enable_i == 1'b1);
+  `endif
     $sformat(fn, "trace_core_%h_%h.log", cluster_id_i, core_id_i);
```
imphil: That part recently (#29) changed and needs adjustment to only use the hart_id.

rtl/ibex_tracer.sv
```diff
@@ -43,373 +43,392 @@ module ibex_tracer #(
...
-      cycles = 0;
+      cycles <= 0;
     end else begin
-      cycles = cycles + 1;
+      cycles <= cycles + 1;
     end
   end
 
   // open/close output file for writing
   initial begin
```
imphil: Instead of opening the file in an initial block, can you open it on demand whenever there is a first write being done? That would also solve #261. The easiest option is to probably wrap `$fwrite()`, and keep a flag indicating that the file has already been opened (or check the file handle if it is open).

rtl/ibex_tracer.sv
```diff
@@ -423,144 +442,142 @@ module ibex_tracer #(
 
   // log execution
   always @(posedge clk_i) begin
```
imphil: nit: Please replace this by always_ff

rtl/ibex_tracer.sv
```diff
@@ -423,144 +442,142 @@ module ibex_tracer #(
...
+    init();
     // special case for WFI because we don't wait for unstalling there
     if (valid_i) begin
-      trace = new ();
 
-      trace.simtime    = $time;
-      trace.cycles     = cycles;
-      trace.pc         = pc_i;
-      trace.instr      = instr_i;
+      simtime    = $time;
```
imphil: Can $time change while you're writing the log entry? If not, use $time directly.

rtl/ibex_tracer.sv
```diff
@@ -423,144 +442,142 @@ module ibex_tracer #(
...
+          INSTR_FENCE:      printMnemonic("fence");
+          default:          printMnemonic("INVALID");
         endcase // unique case (instr_i)
       end
 
       // replace register written back
-      foreach(trace.regs_write[i]) begin
-        if ((trace.regs_write[i].addr == ex_reg_addr_i)) begin
-          trace.regs_write[i].value = ex_reg_wdata_i;
+      foreach(regs_write[i]) begin
```
imphil: nit: space after foreach

imphil:
Now that #197 is merged, can you rebase this PR on top of master and fix the first round of review comments? I'll then follow up as needed, and I'm confident we can get this merged rather soon.

rtl/ibex_tracer.sv
raulbehl: Done here and at other places too.

rtl/ibex_tracer.sv
raulbehl: Taken care by rebase.

rtl/ibex_tracer.sv
raulbehl: Moved `$fopen()` inside `print_instr_trace()` function (which is the place for the first write to the file).

rtl/ibex_tracer.sv
raulbehl: Yep. `$time` doesn't change for one log entry. Removed `simtime` and using `$time` directly.

raulbehl:
Updated the file as per the review comments. I have renamed all the functions to follow the `lower_snake_case` naming convention. I have removed `printIuInstr()` as this is no longer needed and print can be taken care by the `printIInstr()` function itself. 

Plus a couple of function have been renamed to match the latest RISC-V ISA.  The function `printUInstr()` is renamed to `print_u_instr()`, `printUJInstr()` is renamed to `print_j_instr()` and `printSBInstr()` is renamed to `print_b_instr()`.

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
   import ibex_pkg::*;
   import ibex_tracer_pkg::*;
 
-  integer      f;
+  int          f;
   string       fn;
-  integer      cycles;
   logic [ 4:0] rd, rs1, rs2, rs3;
```
imphil: These variables are shadowed below by the functions using the same variable names. Please rename either one of them.

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
...
+    begin
+      if (addr < 10) begin
+        return $sformatf(" x%0d", addr);
+      end else begin
+        return $sformatf("x%0d", addr);
+      end
+    end
+  endfunction
+
+  function void print_instr_trace();
```
imphil: The naming of this function is confusing. It's named print_, but unlike all other print_ functions, it doesn't assign to str, but writes to file. Can you rename it to `dump_printbuffer_to_file()` or something similar?

raulbehl:
I have resolved most of the lint warnings except for the following:
```
%Warning-WIDTH: Use "/* verilator lint_off WIDTH */" and lint_on around source to disable this message.
%Warning-CASEX: ../src/lowrisc_ibex_ibex_tracer_0.1/rtl/ibex_tracer.sv:451: Suggest casez (with ?'s) in place of casex (with X's)
%Warning-CASEX: ../src/lowrisc_ibex_ibex_tracer_0.1/rtl/ibex_tracer.sv:483: Suggest casez (with ?'s) in place of casex (with X's)
```

Let me know if it should be okay to waive these or should we update the case statement as per the suggestion?

rtl/ibex_tracer.sv
raulbehl: Done.

rtl/ibex_tracer.sv
raulbehl: Removed the variables declared in the functions instead the global variables would be used.

rtl/ibex_tracer.sv
```diff
@@ -430,146 +435,143 @@ module ibex_tracer #(
...
+          default:          print_mnemonic("INVALID");
         endcase // unique case (instr_i)
       end
 
       // replace register written back
-      foreach(trace.regs_write[i]) begin
-        if ((trace.regs_write[i].addr == ex_reg_addr_i)) begin
-          trace.regs_write[i].value = ex_reg_wdata_i;
+      foreach (regs_write[i]) begin
+        if ((regs_write[i].addr == ex_reg_addr_i)) begin
```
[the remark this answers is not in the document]
raulbehl: Removed them.

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
...
-      begin
-        regs_write.push_back('{rd, 'x});
-        str = $sformatf("%-16s x%0d, 0x%0h", mnemonic, rd, {instr[31:12], 12'h000});
-      end
-    endfunction // printUInstr
+  function void print_mnemonic(input string mnemonic);
+    begin
+      str = mnemonic;
+    end
+  endfunction // print_mnemonic
```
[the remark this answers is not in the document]
raulbehl: Done. Have retained the ones for always_ff blocks.

rtl/ibex_core_tracing.sv
```diff
@@ -152,8 +152,6 @@ module ibex_core_tracing #(
     .core_sleep_o
   );
 
-
-`ifndef VERILATOR
   ibex_tracer u_ibex_tracer (
```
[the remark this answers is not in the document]
raulbehl: The parameter `RegAddrWidth` is set depending upon the value of `RV32E`. However, the implemented RVFI uses a hard-coded register address width (being 5-bits wide). This would need to be updated before the tracer is tried with E extension enabled. Let me know if I should file a bug?

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
...
-        end else begin
-          return $sformatf("x%0d", addr);
-        end
+  integer      cycles;
+  logic [31:0] pc;
+  logic [31:0] instr;
+  string       str;
+  reg_t        regs_read[1:0];
+  reg_t        regs_write[1:0];
+  mem_acc_t    mem_access[1:0];
```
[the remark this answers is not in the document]
raulbehl: Done.

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
...
+
+  function void init ();
```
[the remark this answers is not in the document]
raulbehl: Updated.

rtl/ibex_tracer.sv
```diff
@@ -40,388 +37,396 @@ module ibex_tracer #(
   import ibex_pkg::*;
   import ibex_tracer_pkg::*;
 
-  integer      f;
+  int          f;
```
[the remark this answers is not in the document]
raulbehl: Updated the variables to the following:

f -> file_handle
fn -> file_name

examples/sim/tb/ibex_tracing_tb.cpp
```diff
@@ -0,0 +1,111 @@
+#include <stdlib.h>
```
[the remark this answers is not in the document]
raulbehl: Done.

imphil:
@raulbehl In the process of reviewing this PR I found a couple things to clean up. I'll follow up with a proposal in roughly a week (I'm currently traveling).

imphil:
@raulbehl I'm sorry this review took longer than I usually aim for. When looking through your code, I found various ways to further improve things to better make use of the signals already available at the RVFI. What started out as review comments morphed into more and more changes, so I took the liberty to take over this PR and include my changes in #360. I invite you to look at this code and provide feedback, I hope we can converge to a great and sustainable solution.

Thanks for starting this work!

Issue #265, opened 2019-08-27
Combinational paths between RVALID_I/ERR_I and REQ_O


arjanbink:
Description: Combinational paths exist from data_rvalid_i and data_err_i to data_req_o and from instr_rvalid_i to instr_req_o. Such paths can potentially limit the achievable performance in systems that do not pipeline these signals as they would introduce a path from a slave to the master (i.e. Ibex) and then back to the slave. For details see the attached pdf.

Git tag: 242147239551be9fb8523fbc6f26f22b898d8d93 (august 26   2019)

[IBEX0.pdf](https://github.com/lowRISC/ibex/files/3546206/IBEX0.pdf)

imphil:
Thanks for the detailed report @arjanbink! I have seen this loop before but apparently I didn't file a bug for it, thanks for doing that! 

We also don't see this loop in Verilator lint at the moment because we don't connect the bus to ibex, that's something we should look into as well. (But probably not as part of this issue.)

arjanbink:
It is not a loop inside Ibex (but it could become a loop on system level)

vogelpi:
Thanks for your detailed report @arjanbink , the timing diagram is really helpful!

I think we should change this. It basically means that the LSU needs to push out the second request before receiving the first `RAVLID_I`. In the shown example, the first `RVALID_I` is received together with the second `GNT_I`. Handling this case seems not to be too difficult to me. But I think we should also support the case if the second `GNT_I` comes before the first `RVALID_I`. To properly handle errors in this case requires quite some changes in the LSU FSM though.

For the instruction interface, it is probably more difficult as the whole interface is still much more convoluted.

arjanbink:
I also noted that the second transfer in a misaligned load/store is suppressed in case the first transfer receives an error. In itself that is fine of course, but I don't think it is mandated by any specification, so I would think this part could be simplified by getting rid of the 'second transfer suppression in case of an error' (the performance impact would be a don't care as this should be an exceptional scenario)

davideschiavone:
@vogelpi , if the core does not wait the first valid before sending the second request, it means that we have potentially out of order memory transactions.
Imagine a load to an APB space that may take 3/4 cycles followed by a single cycle load to the scratch pad for instance.
So the second grant should not be received before the first valid (so the second request should wait the valid).

I agree with @arjanbink  that this may limit the performance, but I have the personal opinion that this should be handled by a wrapper that simply breaks the paths (penalties for cycles but gains in max frequency).

[...]

PR #282, opened 2019-09-05
[RTL] Fix I-side timing loop

tomroberts-lowrisc:
- See issue #265 (partially fixes)
- Remove path from instr_rvalid_i to instr_req_o
- Prefetch buffer unit can now issue up to two outstanding requests
- Structure moved from state machine to request queue
- Change fetch fifo to use an unaligned flag rather than updating
  the address each time

### DIFF
--- a/rtl/ibex_fetch_fifo.sv
+++ b/rtl/ibex_fetch_fifo.sv
@@ -6,64 +6,90 @@
 /**
  * Fetch Fifo for 32 bit memory interface
  *
- * input port: send address one cycle before the data
- * clear_i clears the FIFO for the following cycle. in_addr_i can be sent in
- * this cycle already.
+ * input port: send address and data to the FIFO
+ * clear_i clears the FIFO for the following cycle, including any new request
  */
 module ibex_fetch_fifo (
     input  logic        clk_i,
     input  logic        rst_ni,
 
     // control signals
-    input  logic        clear_i,          // clears the contents of the fifo
+    input  logic        clear_i,          // clears the contents of the FIFO
 
     // input port
+    input  logic        in_valid_i,
+    output logic        in_ready_o,
     input  logic [31:0] in_addr_i,
     input  logic [31:0] in_rdata_i,
     input  logic        in_err_i,
-    input  logic        in_valid_i,
-    output logic        in_ready_o,
-
 
     // output port
     output logic        out_valid_o,
     input  logic        out_ready_i,
-    output logic [31:0] out_rdata_o,
     output logic [31:0] out_addr_o,
-    output logic        out_err_o,
-
-    output logic        out_valid_stored_o // same as out_valid_o, except that if something is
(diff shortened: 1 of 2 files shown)


rtl/ibex_prefetch_buffer.sv
```diff
@@ -39,27 +39,30 @@ module ibex_prefetch_buffer (
     output logic        busy_o
 );
 
-  typedef enum logic [1:0] {
-    IDLE, WAIT_GNT, WAIT_RVALID, WAIT_ABORTED
-  } pf_fsm_e;
+  // Changes to the address flops would be required for > 2 outstanding requests
```
imphil: Does this mean that changing this parameter will actually break the design? Can you add an `assert` in this case? (Disabled with Verilator unfortunately at the moment.)

rtl/ibex_fetch_fifo.sv
```diff
@@ -30,40 +30,66 @@ module ibex_fetch_fifo (
     input  logic        out_ready_i,
     output logic [31:0] out_rdata_o,
     output logic [31:0] out_addr_o,
-    output logic        out_err_o,
-
-    output logic        out_valid_stored_o // same as out_valid_o, except that if something is
-                                           // incoming now it is not included. This signal is
-                                           // available immediately as it comes directly out of FFs
+    output logic        out_err_o
```
imphil: nit: can you sort the signals on the input and output ports the same way?

rtl/ibex_fetch_fifo.sv
```diff
@@ -30,40 +30,66 @@ module ibex_fetch_fifo (
...
+  // The FIFO contains word aligned memory fetches, but the instructions contained in each entry
+  // might be half-word aligned (due to compressed instructions)
+  // e.g.
+  //              | 31               16 | 15               0 |
+  // FIFO entry 0 | Instr 1 [15:0]      | Instr 0 [15:0]     |
+  // FIFO entry 1 | Instr 2 [15:0]      | Instr 1 [31:16]    |
+  //
+  // The FIFO also has a direct bypass path, so a complete instruction might be made up of data
+  // from the FIFO and new incoming data.
+  //
```
imphil: Trailing whitespace here. Please configure your editor to remove these when saving. You can also git to show you these whitespace problems in a diff.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -39,27 +39,30 @@ module ibex_prefetch_buffer (
...
+  localparam int unsigned NUM_REQS = 2;
 
-  pf_fsm_e pf_fsm_cs, pf_fsm_ns;
+  logic                valid_req;
+  logic                valid_req_n, valid_req_q;
```
imphil: _n -> _d

imphil:
Thanks for the update Tom! Please feel free to merge whenever you're ready.

PR #293, opened 2019-09-09
[I-side] - Fix issues found in tracing example

### LINKED ISSUE #288: Ibex tracing example needs update following the changes in prefetch buffer
The recent changes to I-side logic and prefetch buffer has changed how ibex asserts `instr_req_o`. Due to this the current example in the TB doesn't produce the intended results. The delays in the TB would need to be adjusted as per the new logic.

FYI: Here are the contents of the `trace_core_00000000.log`
```
                Time          Cycles PC       Instr                        Mnemonic
                 240               8 00000080 00008106                          c.mv x2, x1  x2=0x00000000  x1:0x00000000
                 250               9 00000082 00008186                          c.mv x3, x1  x3=0x00000000  x1:0x00000000
                 290              13 00000090 00400113                       addi x2, x0, 4  x2=0x00000004
                 310              15 00000094 ff810113                      addi x2, x2, -8  x2=0xfffffffc  x2:0x00000004
                 340              18 00000098 4000006f                         jal x0, 1024
                 390              23 00000498 039597b3                   mulh x15, x11, x25 x15=0x00000000 x25:0x00000000 x11:0x00000000
                 400              24 0000049c 13410d13                    addi x26, x2, 308 x26=0x00000130  x2:0xfffffffc
                 410              25 000004a0 e1070713                  addi x14, x14, -496 x14=0xfffffe10 x14:0x00000000
                 420              26 000004a4 fff7c793                    xori x15, x15, -1 x15=0xffffffff x15:0x00000000
```
The trace ends abruptly at `420ns` (since now the timing between `instr_req_o`, `instr_gnt_i` & `instr_valid_i` isn't in-sync anymore).

Also, it would be nice if we could come up with a new naming scheme for the dumped filename? The current one has way too many zeros getting appended!

tomroberts-lowrisc:
- Fixes #288
- Add missing grant qualification to stop spurious address updates
- Make RTL robust to spurious rvalid signalling

### DIFF
--- a/rtl/ibex_fetch_fifo.sv
+++ b/rtl/ibex_fetch_fifo.sv
@@ -76,7 +76,7 @@ module ibex_fetch_fifo (
   // Alignment is tracked with a flag, this records whether entry[0] of the FIFO has become unaligned.
   // The flag is set once any compressed instruction enters the FIFO and is only cleared once a
   // a compressed instruction realigns the FIFO, or the FIFO is cleared.
-  
+
                               // New incoming unaligned request (must be a branch) or already unaligned
   assign entry0_unaligned_d = ((((in_valid_i & in_addr_i[1]) | entry0_unaligned_q) &
                                 // cleared by a compressed unaligned instruction
--- a/rtl/ibex_if_stage.sv
+++ b/rtl/ibex_if_stage.sv
@@ -261,11 +261,6 @@ module ibex_if_stage #(
     @(posedge clk_i) (boot_addr_i[7:0] == 8'h00) ) else
       $error("The provided boot address is not aligned to 256 bytes");
 
-  // there should never be a grant when there is no request
-  assert property (
-    @(posedge clk_i) (instr_gnt_i) |-> (instr_req_o) ) else
-      $warning("There was a grant without a request");
-
   // assert that the address is word aligned when request is sent
   assert property (
     @(posedge clk_i) (instr_req_o) |-> (instr_addr_o[1:0] == 2'b00) ) else
--- a/rtl/ibex_prefetch_buffer.sv
+++ b/rtl/ibex_prefetch_buffer.sv
@@ -44,6 +44,7 @@ module ibex_prefetch_buffer (
 
   logic                valid_req;
   logic                valid_req_d, valid_req_q;
+  logic                hold_addr_d, hold_addr_q;
   logic                gnt_or_pmp_err, rvalid_or_pmp_err;
   logic [NUM_REQS-1:0] rdata_outstanding_n, rdata_outstanding_s, rdata_outstanding_q;
   logic [NUM_REQS-1:0] branch_abort_n, branch_abort_s, branch_abort_q;
@@ -100,8 +101,7 @@ module ibex_prefetch_buffer (
   //////////////
 
   // Make a new request any time there is space in the FIFO, and space in the request queue
-  assign valid_req = req_i & (fifo_ready | branch_i) &
-                     ~&rdata_outstanding_q;
+  assign valid_req = valid_req_q | (req_i & (fifo_ready | branch_i) & (~&rdata_outstanding_q));
 
   // If a request address triggers a PMP error, the external bus request is suppressed. We might
   // therefore never receive a grant for such a request. The grant is faked in this case to make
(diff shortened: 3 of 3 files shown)


tomroberts-lowrisc:
Yes good idea, I ran the tracer test example given by @raulbehl, but it would be could if @udinator could run the failing riscv_rand_instr_test

udinator:
Hi @vogelpi and @tomroberts-lowrisc, I reran a riscv_rand_instr_test, and Ibex is now able to run the test to completion, and passes the correctness log comparison check in our flow, but I am seeing that the second assertion in `ibex_if_stage.sv` starts to fail slightly less than halfway through the test (around 58000 cycles in), and the resultant simulation log looks like the below, though Ibex completely passes the test:
Screenshot from 2019-09-09 09-30-25

After digging into this failure, the below waveform shows the first time that this assertion fails:
Screenshot from 2019-09-09 11-49-36
At the location where the yellow cursor is, we see the request signal being asserted by Ibex for 3 cycles, but the testbench does not send a grant for this specific request for another 7-8 cycles, until after the request has been dropped by the core, thus triggering the assertion in question. From my discussion with Pirmin as well as the memory interface spec, it seems that grants should only be sent to the core while the request signal is high, which would make this stand-alone grant invalid, but Ibex still responds to the corresponding `rvalid` and `rdata`, which are sent around the `15,460,000ps` mark, and continues executing as normal.
I will modify the testbench driver protocol to only send grants to Ibex when requests are active, but would it be possible to clarify Ibex's expected behavior in the situation I described above?
Best,
Udi

vogelpi:
Hi Udi,
thanks for your report. Looking at the waveform, I see different points that need to be investigated, discussed and documented:

1. Is it allowed for the core to deassert `request` without getting a `grant`? The two of us earlier discussed that it makes sense to allow this. For example, if the core changes its mind about an outstanding request because of a branch. This is similar to the core changing the address as long as it does not get a grant. This has been confusing for many people in the past but it was allowed on the instruction interface. In the case of a PMP error, the same is also possible now on the data interface (the core asserts `data_req_o` for one cycle and then deasserts it if it does not get a grant). For our core interface it is currently not described if this is allowed or not and how the downstream should handle it. We need to take a decision here and document it. Just FYI: there are both interconnects that allow this behavior (TileLink) and such that do not (AXI).
-> I will create a separte issue for this.

2. Between the yellow marker in your wave diagram and the second time the core asserts `request` at around 15,420,000ps, the core does not change the address but quite a few cycles pass by. What's is happening there? This does not look good to me. Could you extract the waves and send to us for inspection?

3. There are `rvalid`s missing. After the yellow marker, I see 3 `request`/`grant` pairs + the `grant` without `request`, but just 1 associated `rvalid` (the first one after the marker belongs to the handshake before the marker). Probably the IF stage is getting out-of-sync because of the `grant` without `request`?

Best regards,
Pirmin

udinator:
Hi Pirmin,
1) I agree with your point about the spec saying that `instr_req_o` must stay high until `instr_gnt_i` is high for one cycle (I did not remember this either in our discussion).

2) Unfortunately, the original waveform file got overwritten, but I have recreated a very similar situation as in my previous post. I have sent the associated `waves.fsdb` and disassembled assembly code for further inspection via an email; the situation arises at around 17,410,000ps into the simulation, where `instr_gnt_i` is asserted when `instr_req_o` is low, and then there is quite a large delay until `instr_req_o` is asserted again, and the address does not change either. In this situation, shown in the new waveform snippet below, a `rem` instruction is being executed, and takes many cycles to complete (from the time the first `request` goes high just before 17,400,400ps until the next `request` goes high around 17,466,000ps. I recall in the previous scenario I posted above, a related instruction that utilizes the multdiv block was executing, causing the delay.
Screenshot from 2019-09-09 17-03-40

3) Regarding the waves image in my previous post, the `rvalid` shown near the right hand side of the waveform around 15,460,000ps is the one returning data associated with the stand-alone `grant`, as this `grant` actually is in response to the requested `response` at 15,370,000ps. This is due to the testbench inserting a random delay between receiving `response` and driving `grant`. In addition, our testbench inserts a delay of up to 20 clock cycles between detecting `response` and `grant` high during the same cycle, and driving `rvalid` with the associated read data. However, since the `grant` was standalone, the next time that the condition was true was at 15,420,000ps, which marks a separate transaction, which is why the `rvalid` corresponding to the standalone `grant` is so far after in the simulation.

tomroberts-lowrisc:
Thanks for looking into it Udi. I think the grant |-> req assertion should probably just be removed. That's not how a sensible interconnect should behave, and the core can handle it happening.

I think we need a separate issue on whether or not the request can be dropped (and the address changed) once asserted. It will need quite a few changes to fix, but I can see it being something that's required for the future (not just for AMBA but any system running N:1 clocking between core and interconnect).

tomroberts-lowrisc:
For now I think I can stop the ungranted request from dropping with an extra flop, and that gets us back to where we were before protocol-wise. Then we can look into the address changing as a separate issue.

tomroberts-lowrisc:
Updates made to stop requests being dropped. Hopefully @udinator can rerun the test again and confirm all is ok.
Thanks!

vogelpi:
I just discussed with Tom, the LSU is not affected by this behavior (in the case of PMP errors, I have changed my comment above). I have created a separate issue lowrisc/ibex#296 to discuss the address stability on the instruction interface.

udinator:
Hi @tomroberts-lowrisc, @vogelpi,
I reran the same test with the updates to the RTL, and I also added back in our testbench's randomized delay before asserting `grant` (which I'd removed in #295) to test the core's new behavior. For the most part, the waves look better now, and `request` is usually dropping after the `grant` is asserted after some random delay. However, I'm still seeing some interesting situations as shown in the waveform below:
Screenshot from 2019-09-10 10-23-25
Around 19,973,000ps, the core sends out two consecutive requests, followed by a third request a clock cycle after the first two have been sent. The first two are answered with grants from the testbench immediately, and `rvalid` is raised two cycles after the core has asserted `request`. However, the third request the core sends out at around 19,980,000ps drops as soon as the `rvalid` (which indicates `rdata` for only the first two requests) drops, and does not wait for its `grant` to be asserted, which occurs at around 20,000,000ps. Resultantly, the `rvalid` for this third request is sent by the testbench only after it detects both `request` and `grant` are high again, which occurs much later, around 20,070,000ps, thus the read data for this standalone `grant` is sent at the location of the yellow cursor. 
What is interesting is that we can see that around 20,053,000ps, the core sends out a `request` for the fourth time, but this one does wait until seeing a `grant` to deassert it.

tomroberts-lowrisc:
Thanks again @udinator - I think I see the issue, have pushed another update which should fix the dropped requests

udinator:
LGTM, I'm not seeing this issue anywhere in the tests. Thanks for all of the fixes @tomroberts-lowrisc !

vogelpi:
Cool, thanks for your feedback @udinator and fixing it @tomroberts-lowrisc , great work!

tomroberts-lowrisc:
Thanks again @udinator for testing this so quickly, and thanks @vogelpi for the review!

Issue #290, opened 2019-09-07
Incorrect irq_fast ID assignment


udinator:
According to the documentation, Ibex should assign all of the fast interrupts `irq_fast_i[14:0]` an ID from 16 to 30, so that `irq_fast_i[0]` has an ID of 16, `irq_fast_i[1]` has an ID of 17, and so on.  In the [Ibex CSR register file](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_cs_registers.sv#L572), the entire 15 bit array is output to the [Ibex controller module](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_controller.sv#L165), where `mfip_id` is assigned based on on the value of `irq_fast_i[14:0]`. However these assigned IDs are between 0 and 14. Later on in the controller, `exc_cause_o` is [assigned](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_controller.sv#L363) by appending 1'b1 to `mfip_id`. This is sent to the [Ibex IF stage](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_if_stage.sv#L109), where the 5 bit `irq_id`, which in this case is the ID assigned earlier in the controller, is multiplied by `0x4` and added to the address held in `mtvec` to get the trap address. However, the ID's of all fast interrupts are still from 0 to 14, which seems incorrect according to the RISC-V Privileged Spec Ch. 3.1.16. This means that there are now conflicts with all other interrupts that have ID's between 0 and 14, so Ibex will now jump to the wrong vectored address for all fast interrupts that are sent to the core. I think the `irq_fast` ID assignments in the CSR file should be from between 16 to 30 to fix this issue and comply with the spec.

udinator:
adding @tjaychen for visibility

vogelpi:
Hi @udinator ,
thanks for reporting this issue. You are right. The generation of `mfip_id` in the controller is wrong. For example, it should be
   ```if      (csr_mfip_i[14]) mfip_id = {1'b1, 4'd14}; // ID 30```
instead of
   ```if      (csr_mfip_i[14]) mfip_id = 5'd14;```

I will fix it accordingly and create a PR.

PR #294, opened 2019-09-09
Controller: Fix exception cause ID of fast interrupts

vogelpi:
This commit fixes the generation of the exception cause ID of fast interrupt. Without this commit, the core tries to handle fast interrupts using the handlers of other interrupts like external, software, and timer interrupts.

This commit resolves lowRISC/ibex#290 reported by @udinator.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -113,7 +113,7 @@ module ibex_controller (
   logic enter_debug_mode;
   logic handle_irq;
 
-  logic [4:0] mfip_id;
+  logic [3:0] mfip_id;
   logic       unused_csr_mtip;
 
 `ifndef SYNTHESIS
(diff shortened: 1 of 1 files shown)


rtl/ibex_controller.sv
```diff
@@ -360,7 +360,7 @@ module ibex_controller (
             exc_cause_o = EXC_CAUSE_IRQ_NM;
             nmi_mode_d  = 1'b1; // enter NMI mode
           end else if (csr_mfip_i != 15'b0) begin
-            exc_cause_o = exc_cause_e'({1'b1, mfip_id});
+            exc_cause_o = exc_cause_e'({2'b11, mfip_id});
```
imphil: Can you make it clearer what's going on here? With a comment (first bit is interrupt, second is "+16", and/or with clearer code.

E.g. `EXC_CAUSE_IRQ_FAST_0 | mfip_id`

imphil: Or, to avoid too much annoying lint workarounds, you can also do the mfip_id assignment at the beginning already with `EXC_CAUSE_IRQ_FAST_0 + 5'd14` etc. (possibly also rename mfip_id to mfip_exc_cause?)

vogelpi: Thanks @imphil, I added a clarifying comment.

PR #299, opened 2019-09-10
Add performance counter utils for verilator sims

GregAC:
These aren't used anywhere yet but have been tested with some verilator stuff I have locally.

They could live under dv/verilator/simutil_verilator but there's no reason it couldn't be used independently of that so put it under it's own directory.


dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -0,0 +1,57 @@
...
+    longest_name_length = std::accumulate(
+        &ibex_count_names[0], &ibex_count_names[ibex_pcount_num], 0,
+        [](unsigned int cur_longest, const std::string& pcount_name) {
+          if(pcount_name.length() > cur_longest)
+            return (unsigned int)pcount_name.length();
+          else
+            return cur_longest;
+        });
+
+    // Add 1 to always get at elast once space after the seperator
```
lenary:
```suggestion
    // Add 1 to always get at least once space after the separator
```

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,41 @@
+#ifndef IBEX_PCOUNTS_H_
+#define IBEX_PCOUNTS_H_
+
+#include <cstdint>
+#include <string>
+
+const unsigned int ibex_pcount_num = 11;
```
lenary: This is a very C way of doing this. It's potentially better to use a std::vector, so when someone adds a new counter, and has to update the list of performance counter names, they won't get issues by forgetting to update this variable?

imphil:
I'm waiting with a full review until #298 is in. In the mean time, you can
- Add license headers to all files
- Run clang-format on your changes. I'm adding a clang-format configuration in #303 which you can use.

GregAC:
Last fix was just correcting the path referred to in ibex_cs_registers.sv

GregAC:
The perils of code that isn't getting tested by CI, thought I'd tested it but I hadn't.  Vector have size() not length()

dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -0,0 +1,62 @@
...
+    "Loads",
+    "Stores",
+    "Jumps",
+    "Conditional Branches",
+    "Taken Conditional Branches",
+    "Compressed Instructions"};
+
+std::string ibex_pcount_string(uint64_t pcounts[], bool csv) {
+  char seperator = csv ? ',' : ':';
+  unsigned int longest_name_length;
```
asb: I'd probably drop the use of `unsigned` here. A number of coding styles are recommending against using `unsigned` unless strictly necessary (see [here](http://lists.llvm.org/pipermail/llvm-dev/2019-June/132890.html) for some background), and obviously you end up needing the cast for `pcount_name.length()` below.

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,45 @@
...
+ * printing/logging
+ *
+ * In Ibex the performance counters are held in the mhpmcounter flops in
+ * ibex_cs_registers.  In a verilator simulation the signal type of the
+ * mhpmcounter array should be compatible with the type of pcounts here and so
+ * can be passed in directly to this function.
+ *
+ * There are two options for string formatting, csv or pretty-print.  Both
+ * produce one counter name and value per line.  csv just seperates them with a
+ * comma and no further formatting.  pretty-print uses a colon and aligns the
```
asb: We don't tend to use double spaces after a period in the current comments and docs in this codebase. There are a few cases where where you're using print-print rather than pretty-print.

dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -0,0 +1,61 @@
...
+    "Conditional Branches",
+    "Taken Conditional Branches",
+    "Compressed Instructions"};
+
+std::string ibex_pcount_string(uint64_t pcounts[], bool csv) {
+  char seperator = csv ? ',' : ':';
+  std::string::size_type longest_name_length;
+
+  if (!csv) {
+    longest_name_length = std::accumulate(
```
asb: Don't you just want a loop like `for (const std::string &pcount_name : ibex_pcount_names) { /* logic here to update cur_longest */ }` to find the longest? (Or using auto if preferred - the codebase I normally work on limits use of auto quite a bit).

asb: Yes, I like the std algorithms in general, but in this case it obfuscates it whereas this is shorter+clearer.
```
for (const std::string &s : ibex_pcount_names) {
  cur_longest = std::max(cur_longest, s.length());
}
```

dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -0,0 +1,57 @@
...
+    "Taken Conditional Branches",
+    "Compressed Instructions"};
+
+std::string ibex_pcount_string(uint64_t pcounts[], bool csv) {
+  char seperator = csv ? ',' : ':';
+  std::string::size_type longest_name_length;
+
+  if (!csv) {
+    longest_name_length = 0;
+    for(const std::string& pcount_name : ibex_pcount_names) {
```
asb: Should be `for (const std::string &pcount_name : ibex_pcount_names) {`. Now Philipp has merged a `.clang-format` file you can get clang-format to sort this for you. The current Debian/Ubuntu clang-format package includes the git-clang-format helper so you can stage your work and use `git clang-format` to reformat changed hunks.

GregAC:
@asb @imphil happy with this now?

dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -0,0 +1,57 @@
...
+    pcount_ss << ibex_pcount_names[i] << seperator;
+
+    if (!csv) {
+      int padding = longest_name_length - ibex_pcount_names[i].length();
+
+      for (int j = 0; j < padding; ++j)
+        pcount_ss << ' ';
+    }
+
+    pcount_ss << pcounts[i] << '\n';
```
imphil: nit: \n -> std::endl

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+extern const std::vector<std::string> ibex_count_names;
+
+/**
+ * Given an array of performance counter values returns a formatted string for
+ * printing/logging
```
imphil: Can you shorten that to a single line to avoid confusing doxygen?

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,45 @@
...
+ * longercountername1, 43980
+ * ...
+ *
+ * pretty-print (csv == false):
+ * countername1:       1234
+ * longercountername1: 43980
+ * ...
+ *
+ * @param pcounts Array of performance counter values, must be of length
+ * ibex_pcount_num
```
imphil: nit:

```

 * @param pcounts Array of performance counter values, must be of length
 *                ibex_pcount_num
```

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,45 @@
...
+ * @param csv Choose csv or pretty-print formatting
+ * @return String of formatted performance counter values, newline at end
+ */
+std::string ibex_pcount_string(uint64_t pcounts[], bool csv);
```
imphil: You could split this into two functions, one for CSV and one for pretty printing. That would make the doxygen string easier to understand, and the implementation more focused. (You can share the lower half of the implementation between the two functions by putting it into a private function.)

dv/verilator/pcount/cpp/ibex_pcounts.h
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IBEX_PCOUNTS_H_
+#define IBEX_PCOUNTS_H_
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+extern const std::vector<std::string> ibex_count_names;
```
imphil: nit: ibex_count_names -> ibex_counter_names possibly?

GregAC:
Thanks for the comments @imphil decided not to do the refactor of ibex_pcount_string, felt it resulted in simpler code leaving it as one function.

PR #311, opened 2019-09-13
Simple system

GregAC:
Opening this PR now so I can get some feedback.  Currently missing documentation (beyond a few comments) which I intend to write as part of the PR but want to avoid re-writing it if there's lot of stuff to change.

It should sit on top of my pcount stuff so that's also dragged into this PR, will rebase this on top of ibex once that is merged in.


GregAC:
Oh and I think it'd be nice to get a printf as part of the common library.  Certainly easy enough to find some suitable source code but need to make sure it's something with a suitable license and properly include it in the repository (probably under a vendor/ directory somewhere).  So I'll probably handle that with another PR once we're happy with this.

examples/sw/simple_system/common/common.mk
```diff
@@ -0,0 +1,53 @@
+COMMON_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
GregAC: Add copyright header

examples/sw/simple_system/common/crt0.S
```diff
@@ -0,0 +1,101 @@
+#include "simple_system_regs.h"
```
GregAC: Add copyright header

examples/sw/simple_system/common/link.ld
```diff
@@ -0,0 +1,106 @@
+OUTPUT_ARCH(riscv)
```
GregAC: Add copyright header

GregAC:
Thanks for the feedback @vogelpi I have made your suggested changes.

I've also simplified the linker script, made some tweaks to the names used in the sim_output_manager.h/.cc and written a little bit of documentation.

I plan to write a README and/or add documentation to the Ibex docs about simple system once everyone's happy with what I have so far (that documentation will become part of this PR rather than a new one).

vogelpi:
Thanks for implementing the feedback @GregAC . Please let me know, when you would like me to re-review the PR.

imphil:
Just to warn you ahead of time: I had a quick look at this PR and I'll follow up with suggestions to make the current set of "simple systems" we have in our tree share more code. I hope to be able to do so later today or early tomorrow.

dv/common/dv_common.core
```diff
@@ -0,0 +1,22 @@
...
+description: "Collection of useful RTL for building simulations"
+filesets:
+  files_sim_sv:
+    files:
+      - rtl/prim_clock_gating.sv
+      - rtl/ram_1p.sv
+      - rtl/bus.sv
+      - rtl/simulator_ctrl.sv
+      - cpp/sim_output_manager.h:  { file_type: cppSource, is_include_file: true }
+      - cpp/sim_output_manager.cc: { file_type: cppSource }
```
imphil: You need to split this file to not include files which might not be needed in certain cases. So that would remove the C++ files at least.

dv/common/dv_common.core
```diff
@@ -0,0 +1,22 @@
+CAPI=2:
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:dv:common"
```
imphil: All code in this repository is in lowrisc:ibex:xxx

dv/common/rtl/simulator_ctrl.sv
```diff
@@ -0,0 +1,67 @@
...
+  localparam SIM_CTRL_ADDR = 1;
+
+  logic [7:0] ctrl_addr;
+
+  assign ctrl_addr = addr_i[9:2];
+
+  import "DPI-C" function void output_char(input int c,input string log_name);
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if(~rst_ni) begin
```
imphil: style nit (here and in other places): space after if, case, etc.

GregAC:
Thanks for the feedback @imphil 

On Topic 1 - Happy to move it though if we do merge the riscv compliance suite and this either fully or partially what ends up living in dv (as the riscv compliance suite definitely is dv)?  Depending on how it's done perhaps this is just a ibex_riscv_compliance.core and a README.md.  riscv_testutil.sv should probably continue to live somewhere in dv.  Would we want the shared code all in examples/ if the compliance suite was using it?  As then it's not just an example but part of our DV setup.

Thinking along similar lines should dv/verilator be moved to examples/verilator?  Though again it's part of the compliance suite so something more than a mere example, maybe we need a new common/verilator at the top-level of the repo or similar.

On Topic 2 - I did consider how to share things between the compliance system and simple system however I ran into a couple of difficulties

* I didn't like the idea of just instantiating the riscv_testutil in simple system (so it could be used directly for the compliance runs).  As it provides specialised functionality that only makes sense if you're doing a compliance suite run.  Also if we're aiming for a simple, fully-working, example system we don't want extra baggage hanging around

* The .cc files look very similar but there were a few differences that made trying to produce them out of a shared codebase potentially annoying (or just generate something more complex than just having the two separate files).  One of the main issues was each would have a different name for its verilog top level which is used at various points.  The recent changes to the dv/verilator stuff may have alleviated this.  There were also some extra things I wanted simple system to do (the output manager stuff, writing out the performance counters) that didn't really belong in the compliance suite

Maybe a way forward is to have a separate 'devices' module?  That way simple_system and riscv_compliance can share the same top-level verilog (with some parameterisation to define number of devices and number of hosts) and just have different versions of that module.  RAM would always be instantiated at the top-level rather than living in the 'devices' module.

Instead of a seperate module we could do it with an include and a define (if the verilog preprocessor supports this), place something like:

`include SYSTEM_DEVICES_SV

At an appropriate point in the module and a 

`include SYSTEM_DEVICES_MAP_SV

at another (to setup the cfg_device stuff)

We could use defines/params to conditionally include the testutil but then we're failing to keep the example system simple.

Finally we could autogen the top-level based on some config though we're maybe failing the 'simple' angle here too.

Topic 3 - Point taken about the over-engineering I was aiming to have something that would allow you to easily customise what happened to the output, though I guess in reality it's pretty much always going to go to a file so the direct $fwrite approach in verilog should be fine.

imphil:
If you're ok with it, let's discuss this in person when I'm back in Cambridge.

imphil:
Coming back to this (finally); sorry for the long delay.

I see your point about not combining the compliance test system and the simple system to keep it really simple [1].

So let's go ahead.
- Put the simple system somewhere in `examples` to be able to have a simple system for Verilator, FPGA, etc. Ideally align and share code with the other examples (at least think about it, can be TODO for later).
- Find a place for the common code.
    >  Would we want the shared code all in examples/ if the compliance suite was using it? 

    That's not beautiful, but I'd be OK accepting that for now. Or we can find a better directory for all these "not part of core ibex" files -- examples sounds slightly wrong, dv is slighly wrong as well. Any ideas?

That seems to be all? *(Now why did I take so long to get back to you on that?)*


[1] I hope we can remove the special code to read the compliance test signature once we have print character support in the simple system. We can then output the test signature through this mechanism and get rid of the compliance test system altogether. But that's for later.

GregAC:
Oh and I've added a README.md since the initial PR, added an exception handler (dumps some useful info and halts) and fiddled with the build system slightly (allow specifying assembly source files, and some tweaks to make it easier to build coremark/embench under the same framework)

imphil:
LGTM, thanks for getting this through @GregAC! I'm going ahead and merge that.

PR #318, opened 2019-09-16
Verilator simulation utility: Make it easier to use generated toplevel class

imphil:
Currently, we generate a toplevel class (I'll call it topname here), inheriting from the Verilator-generated toplevel class (V*), and from the VerilatedToplevel abstract class. This allows us to pass around pointers to VerilatedToplevel whenever we need any toplevel, e.g. to call `eval()` on.

```
                      topname
                   /          \
         VerilatedTopevel      Vtopname
           public eval()   public: clk_i, rst_ni, etc.
                           public eval(), etc.
```
We now want to access the members of Vtopname from a pointer to VerilatedToplevel. That's achievable by casting VerilatedToplevel* to topname*, and then simply accessing the inherited members:

```cpp
void dosomething(VerilatedToplevel* top) {
  dynamic_cast<topname*>(top)->clk_i = 1;

  // (or, using a convenience method in this PR)
  top->dut()->clk_i = 1;
}
```
(Edited for new version to not require cast any more here.)

Doing this cast requires type information for topname to be available, i.e. it requires including the corresponding header file in the compilation unit.

Previously, including `verilated_toplevel.h` wasn't sufficient: the `class topname` was only available when the `VERILATED_TOPLEVEL()` macro was called. 

With this PR, `#include verilated_toplevel.h` and adding `-DTOPLEVEL_NAME=topname` is all that is needed.

Thanks @lenary for help trying out various options to achieve this, and thanks to the C preprocessor for its unlimited supply of magic.

Side notes:
- ~`VM_TRACE` is always defined as 1 or 0 by Verilator, so we can use it directly in the `assert`.~ (That's true, but doesn't help, because `VERILATED_TOPLEVEL_NAME::trace()` doesn't exist without VM_TRACE.)
- The alternative approach we tried out earlier is available at https://github.com/imphil/ibex/commit/51e59dfe0ef1752b0bc56eecf7b5617264e3f2d6. It essentially does the same thing, but additionally changes the class hierarchy to have VerilatedToplevel at the top. I didn't follow through with that once since it's more intrusive, and I don't like the idea of a class of a given name (VerilatedToplevel) having different members available depending on how it's compiled.

### DIFF
--- a/dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
+++ b/dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
@@ -5,18 +5,42 @@
 #ifndef VERILATED_TOPLEVEL_H_
 #define VERILATED_TOPLEVEL_H_
 
+#ifndef TOPLEVEL_NAME
+#error "TOPLEVEL_NAME must be set to the name of the toplevel."
+#endif
+
 #include <verilated.h>
+
+#define STR(s) #s
+#define STR_AND_EXPAND(s) STR(s)
+
+// Include Verilator-generated toplevel
+#define VERILATED_TOPLEVEL_HEADER_STR2(s) STR(V##s)
+#define VERILATED_TOPLEVEL_HEADER_STR(s) VERILATED_TOPLEVEL_HEADER_STR2(s)
+
+#include VERILATED_TOPLEVEL_HEADER_STR(TOPLEVEL_NAME.h)
+
+// Name of the Verilated class
+#define VERILATED_TOPLEVEL_NAME3(s) V##s
+#define VERILATED_TOPLEVEL_NAME2(s) VERILATED_TOPLEVEL_NAME3(s)
+#define VERILATED_TOPLEVEL_NAME VERILATED_TOPLEVEL_NAME2(TOPLEVEL_NAME)
+
+// VM_TRACE is defined by Verilator and passed through the command line as 1 or
+// 0. The code below serves as safety net only.
+#ifndef VM_TRACE
+#define VM_TRACE 0
+#endif
 
 // VM_TRACE_FMT_FST must be set by the user when calling Verilator with
 // --trace-fst. VM_TRACE is set by Verilator itself.
 #if VM_TRACE == 1
-#  ifdef VM_TRACE_FMT_FST
-#    include "verilated_fst_c.h"
-#    define VM_TRACE_CLASS_NAME VerilatedFstC
-#  else
-#    include "verilated_vcd_c.h"
-#    define VM_TRACE_CLASS_NAME VerilatedVcdC
-#  endif
+#ifdef VM_TRACE_FMT_FST
+#include "verilated_fst_c.h"
+#define VM_TRACE_CLASS_NAME VerilatedFstC
+#else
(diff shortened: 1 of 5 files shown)


dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
```diff
@@ -97,27 +113,18 @@ class VerilatedToplevel {
...
+class TOPLEVEL_NAME : public VERILATED_TOPLEVEL_NAME, public VerilatedToplevel {
+ public:
+  TOPLEVEL_NAME(const char *name = "TOP")
+      : VERILATED_TOPLEVEL_NAME(name), VerilatedToplevel() {}
+  const char *name() const { return STR(TOPLEVEL_NAME); }
+  void eval() { VERILATED_TOPLEVEL_NAME::eval(); }
+  void final() { VERILATED_TOPLEVEL_NAME::final(); }
+  void trace(VerilatedTracer &tfp, int levels, int options = 0) {
+    assert(VM_TRACE && "Tracing not enabled.");
+    VERILATED_TOPLEVEL_NAME::trace(static_cast<VM_TRACE_CLASS_NAME *>(tfp),
```
lenary: I thought the point in the `ifdef` was to cut code size by not even including this call, if `VM_TRACE` was turned off.

dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
```diff
@@ -5,18 +5,34 @@
...
+#endif
+
 #include <verilated.h>
 
+#define STR(s) #s
+
+// Include Verilator-generated toplevel
+#define VERILATED_TOPLEVEL_HEADER_STR2(s) STR(V##s)
+#define VERILATED_TOPLEVEL_HEADER_STR(s) VERILATED_TOPLEVEL_HEADER_STR2(s)
+#include VERILATED_TOPLEVEL_HEADER_STR(TOPLEVEL_NAME.h)
```
lenary: Please include an empty line before the `#include` so it's obvious you're doing an include - it gets lost in the `#define`s here.

dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
imphil: We need to keep the `trace()` to fully implement `VerilatedToplevel`, the `#ifdef` code was probably just a leftover from when it wasn't obvious that `VM_TRACE` is actually an integer.

imphil:
Thanks for your comments @lenary! I've updated the code as discussed, including a change to `dut()` to return a reference instead of a pointer.

dv/verilator/simutil_verilator/cpp/verilated_toplevel.h
imphil: After the discussion, it was obvious that we need the `#if` to capture the case where `VERILATED_TOPLEVEL_NAME::trace()` doesn't exist. Added it back.

imphil:
Updated after discussion, also changed `TOPLEVEL_NAME::name()` to use `STR_AND_EXPAND` instead of `STR`, otherwise expansion wasn't done and `name()` was always `TOPLEVEL_NAME`.

PR #326, opened 2019-09-17
Lint: Fix parameter width

imphil:


imphil:
This uncovered a couple other things that need to be fixed, added corresponding commits to this PR. Let's see what CI has to say.

imphil:
(Fixed title to better reflect what's going on here. RV32M is checked anyways because it's enabled by default.)

PR #327, opened 2019-09-17
Ignore common editor files in Git

imphil:
Also removes the include/riscv_config.sv.bak line, which was added
in 346d14c5. We have no code that generates this file any more.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
-*.swp
-include/riscv_config.sv.bak
+# Build output
 build
+
+# Common editor/IDE config and temporary files
+.project
+.vscode/
+.sw[a-p]


asb:
Personally I prefer leaving people to add files specific to their personal workflow to .git/info/excludes or a global excludes file, and leaving .gitignore for files known to be generated by e.g. building the repo contents. Otherwise, what's the limit on what common editor files to add?

But of course plenty of projects do this, so happy for you to make this call if you prefer it. I think for vim you'd ideally have .sw[a-p] to allow for multiple swapfiles...

imphil:
Updated the VIM one. I agree with your point, but I'd also like to make it reasonably easy for people to get started without getting confused too much by git. If we can do so with a couple lines in a file that won't be in our way, that seems to be a reasonable trade-off to me.

PR #335, opened 2019-09-19
[RTL] Flop exc_req in controller to ease timing

### LINKED ISSUE #305: Potential timing issues around pc_set
The pc_set signal factors in both exceptions that have just occurred and branch conditions and directly feeds into the top-level instr_addr_o output (and associated signals).  This is a long path making the instruction memory interface outputs late.

The paper 'Slow and Steady Wins the Race? A Comparison of Ultra-Low-Power RISC-V Cores for Internet-of-Things Applications' (https://www.researchgate.net/profile/Francesco_Conti4/publication/321121782_Slow_and_steady_wins_the_race_A_comparison_of_ultra-low-power_RISC-V_cores_for_Internet-of-Things_applications/links/5a37aa3ea6fdccdd41fc99e1/Slow-and-steady-wins-the-race-A-comparison-of-ultra-low-power-RISC-V-cores-for-Internet-of-Things-applications.pdf) which looked at zero-riscy highlighted this path as the worst in their test system.

The simple fix (stall and set the pc cycle after branch decision/exception) adds a cycle latency to any branch.  A simple branch predictor may be able to reduce the impact of this.

GregAC:
Fixes #305

@udinator would you be able to run some tests on this change?  I've done some basic tests to ensure exceptions continue to function but if you had something more in-depth to try out that would be useful.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -104,7 +104,7 @@ module ibex_controller (
 
   logic stall;
   logic halt_if;
-  logic halt_id;
+  logic flush_id;
   logic illegal_dret;
   logic illegal_insn;
   logic exc_req;
@@ -115,6 +115,14 @@ module ibex_controller (
 
   logic [3:0] mfip_id;
   logic       unused_csr_mtip;
+
+  logic ecall_insn;
+  logic mret_insn;
+  logic dret_insn;
+  logic wfi_insn;
+  logic ebrk_insn;
+  logic csr_pipe_flush;
+  logic instr_fetch_err;
 
 `ifndef SYNTHESIS
   // synopsys translate_off
@@ -137,19 +145,28 @@ module ibex_controller (
   assign load_err_d  = load_err_i;
   assign store_err_d = store_err_i;
 
+  // Decoder doesn't take instr_valid into account, factor it in here.
+  assign ecall_insn      = ecall_insn_i      & instr_valid_i;
+  assign mret_insn       = mret_insn_i       & instr_valid_i;
+  assign dret_insn       = dret_insn_i       & instr_valid_i;
+  assign wfi_insn        = wfi_insn_i        & instr_valid_i;
+  assign ebrk_insn       = ebrk_insn_i       & instr_valid_i;
+  assign csr_pipe_flush  = csr_pipe_flush_i  & instr_valid_i;
+  assign instr_fetch_err = instr_fetch_err_i & instr_valid_i;
+
   // "Executing DRET outside of Debug Mode causes an illegal instruction exception."
   // [Debug Spec v0.13.2, p.41]
-  assign illegal_dret = dret_insn_i & ~debug_mode_q;
+  assign illegal_dret = dret_insn & ~debug_mode_q;
   assign illegal_insn = illegal_insn_i | illegal_dret;
 
   // exception requests
-  assign exc_req     = ecall_insn_i | ebrk_insn_i | illegal_insn | instr_fetch_err_i;
+  assign exc_req     = ecall_insn | ebrk_insn | illegal_insn | instr_fetch_err;
 
   // LSU exception requests
   assign exc_req_lsu = store_err_i | load_err_i;
 
   // special requests: special instructions, pipeline flushes, exceptions...
-  assign special_req = mret_insn_i | dret_insn_i | wfi_insn_i | csr_pipe_flush_i |
+  assign special_req = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
       exc_req | exc_req_lsu;
 
   ////////////////
@@ -210,7 +227,7 @@ module ibex_controller (
     first_fetch_o         = 1'b0;
 
     halt_if               = 1'b0;
-    halt_id               = 1'b0;
+    flush_id              = 1'b0;
 
     debug_csr_save_o      = 1'b0;
     debug_cause_o         = DBG_CAUSE_EBREAK;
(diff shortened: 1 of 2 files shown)


rtl/ibex_controller.sv
```diff
@@ -143,14 +144,14 @@ module ibex_controller (
   assign illegal_insn = illegal_insn_i | illegal_dret;
 
   // exception requests
-  assign exc_req     = ecall_insn_i | ebrk_insn_i | illegal_insn | instr_fetch_err_i;
+  assign exc_req_n = (ecall_insn_i | ebrk_insn_i | illegal_insn | instr_fetch_err_i) & (ctrl_fsm_cs != FLUSH);
```
vogelpi: You should add a comment here to clarify why the `& (ctrl_fsm_cs != FLUSH)` is needed. I think its basically about clearing the exception request such that it is no longer set when returning back to DECODE (or another state), but it is not obvious.

Question: would it be better (easier to understand) to leave `exc_req` as is, and drive `exc_req_n` inside the FSM with `exc_req` (default) or `1'b0` (in FLUSH state, and add the comment there)? This would also avoid the use of the FSM state out of the actual FSM, which I always try to avoid.

GregAC: Indeed it's to avoid exc_req being asserted when we're in the state past FLUSH (DECODE or one of the debug states).  I don't think it's strictly necessary (we only look at exc_req_q in FLUSH and FLUSH always transitions to something else) but leaving it out does rely on the state machine always behaving the way it currently does and I think this behaviour of exc_req_q is less surprising.  I can add the comment

With regard to putting in the FSM I see what you mean.  I think I prefer this style though as it's all in one place.  It's clear that exc_req_n is getting set on certain events unless we're in FLUSH.  In the state machine you'd seen a general assignment of exc_req_n = ecall_insn_i | ebrk_insn_i | illegal_insn | instr_fetch_err_i and it getting cleared in FLUSH.  You need to look into two places to understand what the signal is doing.

Now this is the case for all of the signals that are in the FSM but for most/all of them they have complex behaviour so separating them out is justified (the one-liner versions wouldn't be understandable).  As the exc_req_n behaviour is sufficiently simple I think it gains from being done this way.

vogelpi: Alright, thanks for sharing your thoughts on that matter. For me both works (as long as you add a comment).

udinator:
Hi @GregAC - I ran tests that use illegal instructions and `ebreak` to trigger exceptions, with your modifications to the controller, and these are all passing successfully!
With regards to the above discussion with @vogelpi about not setting the exception request when transitioning out of the `FLUSH` state into `DECODE` or something else, I'm seeing the following occur:
Screenshot from 2019-09-19 10-33-37
The instruction at `PC=0x8000068e` is illegal, so when it gets into the decode stage of the pipeline, it triggers an exception. As per what you said above, `exc_req_q` is high during `FLUSH` state, but then remains high afterwards in `DECODE` state, until the first instruction of the exception handler code (at `PC=0x80013000`) moves into the pipeline decode stage, as the illegal instruction still remains in the decode stage. This is something that should not happen based on your description of the intended behavior, correct?

GregAC:
@udinator thanks for the test.  Interesting observation, doesn't actually mean there's a functional issue but I think this is slightly surprising behaviour.  The root cause of the issue is the illegal instruction output of the decoder isn't being qualified with the valid.  Some new testing would be appreciated.

So this PR is now two commits:

1. Factor instr_valid into the illegal instruction and illegal dret signals
2. Flopping things to break the timing path.  Realised I also needed to flop illegal_insn (otherwise it's used directly in the FLUSH stage and the value of pc_set depends upon leaving a long path in)

I've been playing with yosys and I believe I've now got rid of all of the long paths going into pc_set.  yosys is what highlighted the need for illegal_insn to be flopped.

udinator:
@GregAC , the fixes you made to factor instr_valid into the illegal instruction signals seemed to work, the resultant waveforms for my illegal instruction test now look like below:
Screenshot from 2019-09-20 15-03-31
However, I'm noticing the same situation with `exc_req` being held high after the `FLUSH` state occurring for all other exception as well, an example using `ebreak` exceptions is below:
Screenshot from 2019-09-20 15-04-04
Of course, this is by no means a functional issue, just wanted to bring it to your attention, as the behavior of all other exceptions is still the same as before, maybe factoring instr_valid into ebreak_insn and ecall_insn would work also to normalize the new behavior?

GregAC:
Thanks @udinator.  I believe this latest version should cover everything now (or at least a good deal more than before!).

The first commit (Keep instr in ID valid for FLUSH state) is a bigger change than it used to be so I'll wait for input for @imphil and @vogelpi .  They're both away so may not get this merged til next week.

In the meantime another test would be appreciated.

udinator:
@GregAC, I tested the `exc_req_q` changes using illegal instruction, ebreak, and ecall exceptions, and everything looks good to me now! Thanks for all of the changes.

rtl/ibex_controller.sv
```diff
@@ -306,9 +323,11 @@ module ibex_controller (
 
           // get ready for special instructions, exceptions, pipeline flushes
           end else if (special_req) begin
+            // Halt IF but don't flush ID. This leaves a valid instruction in
+            // ID so controller can determing appropriate action in the
+            // flush state.
```
vogelpi: in the flush state -> in FLUSH

rtl/ibex_controller.sv
```diff
@@ -559,7 +578,7 @@ module ibex_controller (
 
   // kill instr in IF-ID pipeline reg that are done, or if a
   // multicycle instr causes an exception for example
-  assign instr_valid_clear_o = ~stall |  halt_id;
+  assign instr_valid_clear_o = ~(stall | halt_if) | flush_id;
```
vogelpi: I would add a comment here why you factor in `halt_if`. I guess you want to avoid that when halting IF (during an exception happening in a singlecycle instruction) without flushing ID, the instruction valid bit is cleared.

imphil:
@GregAC can you rebase this and fix the remaining small comments from Pirmin to get it over the finish line?

Issue #117, opened 2019-07-02
Support for Intel/Altera Quartus Lite/Standard (not Pro)


jrrk:
for (genvar ..) generate_statement() is not supported in Quartus

it likes 

generate
genvar g;
for(g ...) generate_statement() 
endgenerate

The suggestion is made to add ifdef QUARTUS around the four places that you do this

jrrk:
@jonwoodruff issue opened for comment

ivanmgribeiro:
I've done the proposed changes in this commit:
https://github.com/ivanmgribeiro/ibex/commit/b2e82dd57913ed472d3da2da0d86de8df2d988f0

These allow the core to compile in Quartus 16.1 if the QUARTUS macro is defined in the project settings.

imphil:
The currently used style (loop generate constructs, Chapter 27.4  of the SV spec) is recommended by our style guide at https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#generate-constructs. The proposed alternative style (generate regions) is not recommended. So if we cannot make Quartus accept this syntax, we need to change the style guide. Ifdefs are really only the last resort.

To double-check if a style guide change is required:

- Can you retry with the latest Quartus version?
- Is this documented as missing feature in Quartus somewhere? 
  According to https://www.intel.com/content/www/us/en/programmable/quartushelp/18.1/index.htm#hdl/vlog/vlog_list_sys_vlog.htm SV2009 should be supported (and loop generate constructs are SV2009). More details are hard to find in documentation, since unfortunately the documentation of language features ends at SV Spec Chapter 26, and generate blocks are discussed in Chapter 27 of the SV2009 spec.
- Did you check that Quartus doesn't need any special flag to choose the latest SystemVerilog version?

sjgitty:
I'm agreeing with @imphil re `ifdef. The proposed coding style in question
should be
supported, so let's pressure the tool vendors to explain why they are not
conformant
and fix as soon as possible.  Worst case is we allow code exceptions and/or
change the
coding style in the short term. If we need to temporarily change the code
to allow the
tools to catch up, we should tag the change with a comment that allows us
to find and
remove when the tools are fixed.

jrrk:
[...]

PR #336, opened 2019-09-19
[RTL] Quartus compatibility

MarekPikula:
I am working on porting PULPissimo project to support Altera/Intel Quartus and thus Intel FPGAs. Besides issues stated in #117 I had some weird problems with core not willing to enter debug mode. As it turns out Quartus isn't interpreting SystemVerilog in the same way as Vivado does (I'm currently documenting all the stuff I've stumbled upon [here](https://github.com/MarekPikula/quartus-sv-gotchas)) and makes some assumptions regarding `case` statement, which are not necessarily true.

In this commit I propose some changes, which result in correct RTL for Quartus (didn't test it in Vivado yet, but I guess it should work just fine) and IMHO make the code more readable in some places.

It's definitely not to be merged since I'd need to make some correct indentation and some cosmetic changes for a final version. I just wanted to make changes only in the places, which have indeed been changed, so that the diff is more readable. Purpose of this PR is to open discussion on some specific places in code, which don't synthesize correctly with Quartus.

Here's the list of changes with explanations:

## ibex_cs_registers.sv

Here is the place where Quartus just cut entire `default` section (line 319 in upstream), leaving `illegal_csr` high for some reason. I presume that it's caused by some partial cases being handled before (like `CSR_MCOUNTINHIBIT`, `CSR_MCYCLE`, …).

### CSR read logic 
Since we don't use `unique case inside` as pointed out in [style guides](https://github.com/lowRISC/style-guides) (this construct isn't supported in Quartus either way) I propose to use `casez` with wildcarded `CSR_OFF_MCOUNTER_SETUP`, `CSR_OFF_MCOUNTER` and `CSR_OFF_MCOUNTERH` with internal `unique case` for distinguishing between subaddresses in the same address range. This results in less code and IMHO more readable code, as a reader now clearly sees that these subaddresses are indded in this address range and that not all subaddresses are illegal as code suggests.

### CSR write logic

Similar thing is done for write section and I'd have suggestion to change assignment for `illegal_csr_write` from global assign without much meaning to an assignment inside this case statement for particular cases and some additional ones, which are not handled by global assign.

Another thing, which reduces clutter is making `if (csr_we_int == 1'b1)` condition outside case statement. That's a minor change not particularly connected with topic of this PR, but I'd say it makes the code more readable.

### Use of constant expression in case statement

As stated [here](https://github.com/MarekPikula/quartus-sv-gotchas/blob/master/Intel%20Quartus%20SystemVerilog%20gotchas.md#1252-constant-expression-in-case-statement) it simply doesn't synthesize correctly. For example for the code starting at line 499 in upstream the synthesizer basically replaced this case statement with `exception_pc = pc_if_i;` totally ignoring the second case. Correct me if I'm wrong, but I think that syntax in my patch is equivalent to constant expression case.

## ibex_id_stage.sv and ibex_multdiv_fast.sv 

The only problem was constant expression case.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -234,7 +234,7 @@ module ibex_cs_registers #(
     csr_rdata_int = '0;
     illegal_csr   = 1'b0;
 
-    unique case (csr_addr_i)
+    unique casez (csr_addr)
       // mhartid: unique hardware thread id
       CSR_MHARTID: csr_rdata_int = hart_id_i;
 
@@ -327,42 +327,33 @@ module ibex_cs_registers #(
       end
 
       // machine counter/timers
-      CSR_MCOUNTINHIBIT: csr_rdata_int = mcountinhibit;
-      CSR_MCYCLE:        csr_rdata_int = mhpmcounter_q[0][31: 0];
-      CSR_MCYCLEH:       csr_rdata_int = mhpmcounter_q[0][63:32];
-      CSR_MINSTRET:      csr_rdata_int = mhpmcounter_q[2][31: 0];
-      CSR_MINSTRETH:     csr_rdata_int = mhpmcounter_q[2][63:32];
+      CSR_Z_MCOUNTER_SETUP: begin
+        unique case (mhpmcounter_idx)
+          CSR_MCOUNTINHIBIT[4:0]: csr_rdata_int = mcountinhibit;
+          5'b00001, 5'b00010:     illegal_csr   = 1'b1;
+          default:                csr_rdata_int = mhpmevent[mhpmcounter_idx];
+        endcase
+      end
 
+      CSR_Z_MCOUNTER: begin
+        unique case (mhpmcounter_idx)
+          CSR_MCYCLE[4:0]:   csr_rdata_int = mhpmcounter_q[0][31:0];
+          5'b00001:          illegal_csr   = 1'b1;
+          CSR_MINSTRET[4:0]: csr_rdata_int = mhpmcounter_q[2][31:0];
+          default:           csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][31:0];
+        endcase
+      end
+
+      CSR_Z_MCOUNTERH: begin
+        unique case (mhpmcounter_idx)
+          CSR_MCYCLEH[4:0]:   csr_rdata_int = mhpmcounter_q[0][63:32];
+          5'b00001:           illegal_csr   = 1'b1;
+          CSR_MINSTRETH[4:0]: csr_rdata_int = mhpmcounter_q[2][63:32];
+          default:            csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][63:32];
+        endcase
+      end
       default: begin
-        if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTER_SETUP) begin
-          csr_rdata_int = mhpmevent[mhpmcounter_idx];
-          // check access to non-existent or already covered CSRs
-          if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCOUNTINHIBIT
-              (csr_addr[4:0] == 5'b00001) ||
-              (csr_addr[4:0] == 5'b00010)) begin
-            illegal_csr = 1'b1;
-          end
-
-        end else if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTER) begin
-          csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][31: 0];
-          // check access to non-existent or already covered CSRs
-          if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCYCLE
-              (csr_addr[4:0] == 5'b00001) ||
-              (csr_addr[4:0] == 5'b00010)) begin // CSR_MINSTRET
-            illegal_csr = 1'b1;
-          end
-
-        end else if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTERH) begin
-          csr_rdata_int = mhpmcounter_q[mhpmcounter_idx][63:32];
-          // check access to non-existent or already covered CSRs
-          if ((csr_addr[4:0] == 5'b00000) ||     // CSR_MCYCLEH
-              (csr_addr[4:0] == 5'b00001) ||
-              (csr_addr[4:0] == 5'b00010)) begin // CSR_MINSTRETH
-            illegal_csr = 1'b1;
-          end
-        end else begin
-          illegal_csr = 1'b1;
-        end
+        illegal_csr = 1'b1;
       end
     endcase
   end
@@ -393,7 +384,7 @@ module ibex_cs_registers #(
     mhpmcounterh_we  = '0;
 
     if (csr_we_int) begin
-      unique case (csr_addr_i)
+      unique casez (csr_addr)
         // mstatus: IE bit
         CSR_MSTATUS: begin
           mstatus_d = '{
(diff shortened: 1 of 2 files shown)


GregAC:
Thanks for PR and detailed explanations @MarekPikula good to know where Quartus is seeing issues.

Did you not encounter the issues with generated discussed in #117?

MarekPikula:
I did, but I didn't include them in this PR as they are discussed in #117.

MarekPikula:
@GregAC Any thoughts on my changes and propositions so far?

rtl/ibex_multdiv_fast.sv
```diff
@@ -92,14 +92,12 @@ module ibex_multdiv_fast (
...
 
-      unique case(1'b1)
-        mult_en_i:
-          mac_res_q <= mac_res_n;
-        div_en_i:
-          mac_res_q <= op_remainder_n;
-        default:
-          mac_res_q <= mac_res_q;
-       endcase
+      if (mult_en_i)
```
imphil: This needs begin/end blocks.

imphil:
Just for my understanding: All of this is for Quartus Lite/Standard, i.e. the "not developed any further but still heavily used" version of Quartus? This whole exercise feels like a game of whack-a-mole, where we, in a fragile way, try to work around miscompilations.

I'm fine in general with taking the changes you're proposing here, as they are (at least on first sight) in line with our coding standards.

What do you plan to do with the changes discussed in  #117 which require changes that go against our coding standard, since without those changes, you still cannot use Quartus?

Could you try pre-processing ibex with https://github.com/zachjs/sv2v to see if that gets you around some of the problems, including possibly the ones discussed in this PR?

MarekPikula:
Yeah, that's exactly this "not developed any further but still heavily used" version of Quartus.

Regarding changes from #117 they are minor enough that supporting them in my private branch will be possible. It's just a question of adding some `generate` blocks here and there.

I will try with sv2v next week, but to be honest it would be the best for me if it was possible to use upstream version from repo without some additional steps (apart from applying `generate` patch).

What do you think about proposed change in `illegal_csr_write` assignment? Since it's not really related to supporting Quartus maybe I should open a separate PR for that and after it's discussed go back to this PR?

imphil:
Yes, it would help to have cleanups which are not related to Quartus support in a separate PR with a descriptive commit message. Or at least split this PR into a series of commits which each tackle one issue which would make it easier for me to see where exactly the trade-offs are.

imphil:
I've opened https://github.com/lowRISC/style-guides/issues/8 to check if we can discourage `case (1'b1)` in the style guide to avoid falling into this miscompilation trap again.

MarekPikula:
I rebased to current `master`, so no changes from #363 are included there.

MarekPikula:
If you'll approve these changes I'll prepare clean version (with correct formatting) for final review.

MarekPikula:
I've just updated to recent master and moved `casez` constants to `ibex_pkg.sv`.

Today I'll experiment with constant expression case.

MarekPikula:
To my great surprise after reverting constant expression cases to state before the processor works just fine. Sorry for the hassle, during debugging the way Quartus synthesized block in CSR was the first thing that caught my attention and the real issue was the other case statement (now converted to casez).

Because of that I'll revert constant expression case and clean up the code and lowRISC/style-guides#8 can be closed.

imphil:
Thanks for trying and confirming that @MarekPikula!

MarekPikula:
I'd say that the code is ready for final review in this shape. I rebased to master and removed constant expressions cases.

rtl/ibex_cs_registers.sv
```diff
@@ -464,19 +455,25 @@ module ibex_cs_registers #(
...
+
+        CSR_Z_MCOUNTERH: begin
+          unique case (mhpmcounter_idx)
+            CSR_MCYCLEH[4:0]: mhpmcounterh_we[0] = 1'b1;
+            5'b00001: ;
+            CSR_MINSTRETH[4:0]: mhpmcounterh_we[2] = 1'b1;
+            default: mhpmcounterh_we[mhpmcounter_idx] = 1'b1;
+          endcase
         end
+        default: ;
```
imphil: @vogelpi to confirm: propagate X here?

rtl/ibex_cs_registers.sv
```diff
@@ -464,19 +455,25 @@ module ibex_cs_registers #(
...
-          end else if ((csr_addr & CSR_MASK_MCOUNTER) == CSR_OFF_MCOUNTERH) begin
-            mhpmcounterh_we[mhpmcounter_idx] = 1'b1;
-          end
+
+        CSR_Z_MCOUNTER: begin
+          unique case (mhpmcounter_idx)
+            CSR_MCYCLE[4:0]: mhpmcounter_we[0] = 1'b1;
+            5'b00001: ;
+            CSR_MINSTRET[4:0]: mhpmcounter_we[2] = 1'b1;
+            default: mhpmcounter_we[mhpmcounter_idx] = 1'b1;
```
imphil: nit: can you align these "case bodies" similar to how they were aligned before? I found it hard to read.

rtl/ibex_pkg.sv
```diff
@@ -285,8 +285,11 @@ parameter logic [11:0] CSR_OFF_PMP_ADDR = 12'h3B0; // pmp_addr @ 12'h3b0 - 12'h3
 
 // CSR mhpmcounter-related offsets and mask
 parameter logic [11:0] CSR_OFF_MCOUNTER_SETUP = 12'h320; // mcounter_setup @ 12'h323 - 12'h33F
+parameter logic [11:0] CSR_Z_MCOUNTER_SETUP   = 12'b0011001?????;
 parameter logic [11:0] CSR_OFF_MCOUNTER       = 12'hB00; // mcounter       @ 12'hB03 - 12'hB1F
+parameter logic [11:0] CSR_Z_MCOUNTER         = 12'b1011000?????;
 parameter logic [11:0] CSR_OFF_MCOUNTERH      = 12'hB80; // mcounterh      @ 12'hB83 - 12'hB9F
+parameter logic [11:0] CSR_Z_MCOUNTERH        = 12'b1011100?????;
 parameter logic [11:0] CSR_MASK_MCOUNTER      = 12'hFE0;
```
imphil: You can remove the MASK and OFF parameters, they are not used any more.

MarekPikula: Shall I move the comments with ranges to current parameters?

rtl/ibex_cs_registers.sv
MarekPikula: Sorry for inconvenience. Check it out now.

rtl/ibex_cs_registers.sv
```diff
@@ -234,7 +234,7 @@ module ibex_cs_registers #(
     csr_rdata_int = '0;
     illegal_csr   = 1'b0;
 
-    unique case (csr_addr_i)
+    unique casez (csr_addr)
       // mhartid: unique hardware thread id
       CSR_MHARTID: csr_rdata_int = hart_id_i;
```
vogelpi: `csr_addr_i` is of type logic where as `csr_addr`, CSR_MHARTID etc are of type enum. This will cause professional lint tools to explode.

MarekPikula: Ah, interesting. I wasn't aware of this.

MarekPikula:
I understand your point. Interesting insight.

To be honest at the moment I'm working on other things and I couldn't concentrate on it in the near future. Either way I think it would be the best if you were the one making these changes, as you know exactly what you mean, so it will be overall better in terms of time efficiency for both of us.

I finish my internship on 14th November and until that date I'll be able to easily test it on the board I have here at Trinamic. Then I'm coming back to Poland and will be probably less active for some time. I have some Altera board, so either way I'll be able to test it with Quartus, but it would be the best if I could test it before the middle of November.

vogelpi:
Thanks for getting back @MarekPikula . I am currently quite busy but I will try to get a PR ready before mid November. If you could then test it, that would be great. I will ping you. Thanks!

MarekPikula:
I guess it can be closed as #440 is working correctly.

PR #339, opened 2019-09-19
Tracing: Wrap fatal error message with `$fatal()`

vogelpi:
The syntax of this statement is not correct without the `$fatal()` SV construct. This causes errors in some tools even if the error condition is not met.


imphil:
Just FYI: This was intentionally like that. We want to have a compile error here that aborts the compilation, and no other construct did that.

PR #343, opened 2019-09-23
[Priv modes] Add support for U-Mode

### LINKED ISSUE #88: Add support for User (U) mode
Currently, Ibex is M-mode only. For some usage scenarios, support for U mode would be beneficial.

Subtasks:
- Evaluate the implementation effort for this change. As a starting point, we can look at RI5CY (https://github.com/pulp-platform/riscv) 
- Check overhead of this extension.
- Decide on configurability: should this feature be always-on, or configurable via a parameter? (Depends much on the area overhead.)
- Ensure that we have DV for this.

### LINKED ISSUE #306: Documentation: PMP documentation table rendering is broken
This table isn't being rendered correctly:

image

tomroberts-lowrisc:
- General changes to suport U-mode  (fixes #88)
- Update documentation (also fixes #306)
- Add priv mode flops to CSRs module
- Propagate correct priv mode to PMP module
- Fix various address matching bugs in PMP module
- Implement CSR priv-mode permission checking
- Implement illegal U-mode instruction checking
- Add extra mstatus bits for U-mode (MPRV and TW)

### DIFF
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -90,18 +90,24 @@ Reset Value: ``0x0000_1800``
 +-------+-----+---------------------------------------------------------------------------------+
 | Bit#  | R/W | Description                                                                     |
 +-------+-----+---------------------------------------------------------------------------------+
-| 12:11 | R   | **MPP:** Statically 2'b11 and cannot be altered (read-only).                    |
+| 21    | RW  | **TW:** Timeout Wait (WFI executed in User Mode will trap to Machine Mode).     |
++-------+-----+---------------------------------------------------------------------------------+
+| 17    | RW  | **MPRV:** Modify Privilege (Loads and stores use MPP for privilege checking).   |
++-------+-----+---------------------------------------------------------------------------------+
+| 12:11 | RW  | **MPP:** Machine Previous Privilege mode.                                       |
 +-------+-----+---------------------------------------------------------------------------------+
 | 7     | RW  | **Previous Interrupt Enable (MPIE)**, i.e., before entering exception handling. |
 +-------+-----+---------------------------------------------------------------------------------+
 | 3     | RW  | **Interrupt Enable (MIE):** If set to 1'b1, interrupts are globally enabled.    |
 +-------+-----+---------------------------------------------------------------------------------+
 
-When an exception is encountered, ``mstatus``.MPIE will be set to ``mstatus``.MIE.
-When the MRET instruction is executed, the value of MPIE will be stored back to ``mstatus``.MIE.
+When an exception is encountered, ``mstatus``.MPIE will be set to ``mstatus``.MIE, and ``mstatus``.MPP will be set to the current privilege mode.
+When the MRET instruction is executed, the value of MPIE will be stored back to ``mstatus``.MIE, and the privilege mode will be restored from ``mstatus``.MPP.
 
 If you want to enable interrupt handling in your exception handler, set ``mstatus``.MIE to 1'b1 inside your handler code.
 
+Only Machine Mode and User Mode are supported.
+Any write to ``mstatus``.MPP of an unsupported value will be interpreted as Machine Mode.
 
 Machine ISA Register (misa)
 ---------------------------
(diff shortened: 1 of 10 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -529,6 +552,7 @@ module ibex_cs_registers #(
       end // csr_save_cause_i
 
       csr_restore_mret_i: begin // MRET
```
asb: I think you want to set mpp to U mode here "When executing an x RET instruction, supposing x PP holds the value y, xIE is set to xPIE; the privilege mode is changed to y; xPIE is set to 1; and xPP is set to U (or M if user-mode is not supported)." I also note we don't set mpie to 1 either.

doc/cs_registers.rst
```diff
@@ -225,19 +231,19 @@ A particular bit in the register reads as one if the corresponding interrupt inp
...
+------------------------------------
 
 CSR Address: ``0x3A0 - 0x3A3``
 
 Reset Value: ``0x0000_0000``
 
 ``pmpcfgx`` are registers to configure PMP regions. Each register configures 4 PMP regions.
 
-+---------------------------------------+
++---------+---------+---------+---------+
```
asb: Even if bundled into this PR, this fix would be better as a separate commit.

rtl/ibex_pmp.sv
```diff
@@ -95,17 +90,11 @@ module ibex_pmp #(
       // In machine mode, any match to a locked region without sufficient permissions is a fault
       assign machine_access_fault[c][r] = region_match_both[c][r] & csr_pmp_cfg_i[r].lock &
                                           ~region_perm_check[c][r];
-      if (r == 0) begin : g_region0
```
asb: Can you split these changes (fixes?) that seem unrelated to U-mode into a separate commit or PR?

rtl/ibex_cs_registers.sv
vogelpi: @asb I think this is correct here. We need to set `mstatus`.MPP and `mstatus`.MPIE to the values in `mstack` to support recoverable NMI (instead of U mode and 1'b1 as mandated by the spec).

But for correct operation and entering U mode after booting (I guess this happens by the execution of a single MRET after booting) we may need the following changes:
1. Reset `mstatus`.MPP to U mode, and `mstatus`.MPIE to 1'b1 (currently M-mode and 1'b0). Maybe this can be omitted when using a CSR write before doing the first MRET into U mode. 
2. Set `mstack`.MPP to U mode and `mstack.MPIE` to 1'b1 when executing an MRET (currently keeping the previous value). We can also reset it straight to these values (currently reset M-mode and 1'b0).

I am not 100% sure about this. But the non-standard `mstack` CSR is not exposed to software. I thus think we need 2.

tomroberts-lowrisc:
I have hopefully addressed the comments and split the unrelated fixes into separate commits.
I have adopted the changes suggested by @vogelpi for mstack / mstatus as I think that allows us to match the spec and still support recoverable NMI.

rtl/ibex_cs_registers.sv
```diff
@@ -204,9 +215,9 @@ module ibex_cs_registers #(
   assign csr_addr           = {csr_addr_i};
   assign mhpmcounter_idx    = csr_addr[4:0];
 
-  assign illegal_csr_priv   = 1'b0; // we only support M-mode
+  assign illegal_csr_priv   = (csr_addr[9:8] > {priv_lvl_q});
   assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wreq;
-  assign illegal_csr_insn_o = illegal_csr | illegal_csr_write | illegal_csr_priv;
+  assign illegal_csr_insn_o = csr_access_i & (illegal_csr | illegal_csr_write | illegal_csr_priv);
 
   // mip CSR is purely combintational - must be able to re-enable the clock upon WFI
```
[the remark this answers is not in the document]
vogelpi: Thanks ;-)

doc/cs_registers.rst
```diff
@@ -276,6 +282,15 @@ Reset Value: ``0x0000_0000``
 | address[33:2]  |
 +----------------+
 
+Time Registers (time / timeh)
+-----------------------------
+
+CSR Address: ``0xC01 / 0xC81``
+
+The User Mode time and timeh registers are not implemented in Ibex.
```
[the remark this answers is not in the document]
vogelpi: I would also add the back ticks "``" before and after as we do for other CSRs/registers.

doc/cs_registers.rst
```diff
@@ -276,6 +282,15 @@ Reset Value: ``0x0000_0000``
...
+Any access to these registers will trap.
+It is recommended that trap handler software provides a means of accessing platform-defined mtime / mtimeh timers where available.
```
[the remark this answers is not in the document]
vogelpi: ^^

rtl/ibex_cs_registers.sv
vogelpi: Thanks, LGTM. Let's also hear @asb 's feedback.

PR #350, opened 2019-09-25
Add Docker testing environment

### LINKED ISSUE #341: Local CI tests
What do you think about making a local CI test solution based on Docker container? Currently that's how I'm testing my changes before commiting. I have simple Dockerfile doing pretty much all the initialisation part from `azure-pipelines.yml` and then I'm executing next commands by hand.

Would anyone be interested in PR with this? I could polish the Dockerfile and write simple script executing all the tests.

MarekPikula:
Fixes #341.

Follows tests in `azure-pipelines.yml`.


MarekPikula:
For now `docker-compose.yml` is based on my own Docker Hub repository, but the best would be to have an upstream Docker Hub account with this image.

MarekPikula:
Updated to the recent master since, there were some changes in `azure-pipelines.yml`.

MarekPikula:
Built and tested.

MarekPikula:
Image on Docker Hub updated.

MarekPikula:
@asb @imphil Any comments?

imphil:
Thanks for your contribution @MarekPikula. This PR is not forgotten, but I need to give this some thought and I'm unfortunately rather backlogged at the moment. I'll get back to you as soon as possible.

MarekPikula:
What do you think about adding a mention about it in _Verification_ section in documentation?

MarekPikula:
@imphil shall I update this PR to accomodate current Pipelines config, or should I abandon this PR?

imphil:
Thanks for this work @MarekPikula. I'm a bit unsure about this one, since it duplicates a lot of functionality. I see two solutions:

- Either, we find a way to reduce the code duplication.
- Or, we add this file as unsupported example, and explicitly state that.

Code duplication could be reduced by using this docker image for CI testing in Azure Pipelines. I'm not opposed to that, but don't have much time to do this work at the moment.

The other option is to add this file as "untested starting point". I'm usually not a great fan of such code, but would be OK with it in this case. If you'd like to go this route add a relevant comment please, and modify the comment in azure-pipelines.yml to be a suggestion.

MarekPikula:
Sorry for late response. Right now I'm finishing my bachelor thesis, so I'm not able to work on Ibex. I'll look at it some time in the beginning of February.

PR #355, opened 2019-09-26
[DV] Fix ovpsim compare issue

udinator:


dv/uvm/tests/core_ibex_base_test.sv
```diff
@@ -10,7 +10,7 @@ class core_ibex_base_test extends uvm_test;
   virtual core_ibex_dut_probe_if                  dut_vif;
   mem_model_pkg::mem_model                        mem;
   core_ibex_vseq                                  vseq;
-  int unsigned                                    timeout_in_cycles = 3000000;
+  int unsigned                                    timeout_in_cycles = 5000000;
```
udinator: With this timeout increase, the full regression is passing, with the exception of some hint_instr tests.

PR #360, opened 2019-10-01
New tracer

imphil:
This PR is extending #280 to provide a Verilator-compatible tracer. In the process of doing the review in #280 I found various things which could be further simplified and took the opportunity to do that. Thanks @raulbehl for you starting point in #280, it's very much appreciated.

This PR contains a series of commits, the main one is rather large unfortunately. It's probably better not to look at the diff, but at the resulting code. From the main commit message:

----
The ibex_tracer module implements an execution tracer, observing the
execution flow and writing a human-readable execution trace. The trace
information is coming from the RVFI signals, as specified at
https://github.com/SymbioticEDA/riscv-formal/blob/master/docs/rvfi.md.

The existing implementation was tailored for use in ModelSim and other
commercial simulators, and used SystemVerilog features which are not
supported in Verilator or Icarus Verilog, such as classes, queues and
non-standard format specifiers (e.g. the `-` specifier for right-aligned
output). Being unable to see an execution trace when using Verilator
significantly reduced productivity and its usefulness.

This commit refactors the tracer to only use SystemVerilog constructs
which are supported in Verilator. While doing so, multiple improvements
were made for correctness and style.

Major changes:

- Improve compatibility with Verilator. Remove many non-synthesizable
  SystemVerilog constructs, such as classes and queues.
  Use casez instead of casex for better Verilator support (Verilator
  doesn't support X).
- Make the decoded output of the tracer match objdump from binutils
  exactly. Doing so is beneficial for two reasons: we can easily
  cross-check the decoded output from the tracer against the disassembly
  produced by objdump (and we did that), and users don't need to get
  used to another slighly different disassembly format.
- A plusarg "+ibex_tracer_file_base=ibex_my_trace" can be used to set a
  different basename for the trace log file.

Smaller cleanups:

- Remove decoding of reg-reg loads, which were leftover from a PULP
  extension.
- Make better use of the data available on the RVFI. Pass all of RVFI
  to the tracer, and use the provided data instead of manually
  recreating it, e.g. to get register data or the jump target.
- Rename all "instr" abbreviations to "insn". "insn" is what RVFI uses
  (and we cannot change that), so for consistency we now always use this
  abbreviation across the file.
  
The RISC-V compliance test suite for the RV32 I, M, and C extensions has
been executed and traced. The disassembly of all traces have been
compared against traces produced by objdump to ensure identical output.


imphil:
Updated to include correct decoding for c.lui (thanks @asb for the hint), and decoding of fence.i.

tomroberts-lowrisc:
Note @udinator these changes will affect the DV framework since it relies on the tracer.

udinator:
@tomroberts-lowrisc Thanks for letting me know, I'll update the DV framework as necessary.

imphil:
@udinator I looked through the DV code, and you shouldn't need any changes. I'd still appreciate if you can run some tests with this PR to see if the tracer does work for you.

udinator:
@imphil no problem! I'll take a quick look and make sure everything is still working fine.

udinator:
@imphil Everything seems to be fine with the new trace. I had to make some minor changes to the regex we use to parse the Ibex log, but that is the only change from our side.

imphil:
Great, thanks for testing @udinator!

PR #363, opened 2019-10-02
Cleaner `csr_we_int` conditional in CSR

MarekPikula:
For write logic in ibex_cs_registers.sv:

- move `if (csr_we_int) begin` conditional outside `unique case` block to make code cleaner
- move `illegal_csr_write` assign to write logic block with assignments in appropriate places. Before it was just a general out of scope assign without too much meaning.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -204,6 +204,7 @@ module ibex_cs_registers #(
   assign csr_addr           = {csr_addr_i};
   assign mhpmcounter_idx    = csr_addr[4:0];
 
+  // See RISC-V Privileged Specification, version 1.11, Section 2.1
   assign illegal_csr_priv   = 1'b0; // we only support M-mode
   assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wreq;
   assign illegal_csr_insn_o = illegal_csr | illegal_csr_write | illegal_csr_priv;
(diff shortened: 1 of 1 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -429,68 +426,49 @@ module ibex_cs_registers #(
...
-
-      CSR_MCYCLEH: begin
-        if (csr_we_int) begin
-          mhpmcounterh_we[0] = 1'b1;
+        CSR_DPC: begin
+          // Only valid PC addresses are allowed (half-word aligned with C ext.)
+          if (csr_wdata_int[0] == 1'b0) begin
+            depc_d = csr_wdata_int;
+          end else begin
+            illegal_csr_write = 1'b1;
```
vogelpi: As far as I know, the spec does not say that writing an illegal value to the `dpc` CSR should trigger an illegal instruction exception. But the implementation must make sure only legal values are actually written. This is what WARL (write any, read legal) means, and what the current implementation is doing.

MarekPikula:
Thanks for an insight @vogelpi. I didn't know it was this strictly defined in spec. Can you point me to a specific place in spec which consider checking these address bits?

vogelpi:
Yes, it is in Section 2.1 "CSR Address Mapping Conventions" in the privileged spec.

MarekPikula:
Thanks for the reference. I just pushed version with only the `csr_we_int` conditional.

MarekPikula:
Besides, it seems that Pipelines are somewhat broken.

imphil:
CI will be fixed in #364, it's not related to your change.

rtl/ibex_cs_registers.sv
```diff
@@ -205,7 +205,7 @@ module ibex_cs_registers #(
   assign mhpmcounter_idx    = csr_addr[4:0];
 
   assign illegal_csr_priv   = 1'b0; // we only support M-mode
-  assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wreq;
+  assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wreq; // according to section 2.1 privileged spec 1.11
```
vogelpi: Can you please just add the following comment on Line 207:
```// See RISC-V Privileged Specification, version 1.11, Section 2.1```
Because this also holds for `illegal_csr_priv`.

File: rtl/ibex_cs_registers.sv [hunk 3]
```diff
@@ -429,68 +425,33 @@ module ibex_cs_registers #(
...
-      CSR_MCYCLEH: begin
-        if (csr_we_int) begin
-          mhpmcounterh_we[0] = 1'b1;
-        end
-      end
+        CSR_MCOUNTINHIBIT: mcountinhibit_we = 1'b1;
+        CSR_MCYCLE: mhpmcounter_we[0] = 1'b1;
+        CSR_MCYCLEH: mhpmcounterh_we[0] = 1'b1;
+        CSR_MINSTRET: mhpmcounter_we[2] = 1'b1;
+        CSR_MINSTRETH: mhpmcounterh_we[2] = 1'b1;
```
vogelpi: It would be nice if you could align the equal signs in these 5 statements.

rtl/ibex_cs_registers.sv
```diff
@@ -429,68 +426,33 @@ module ibex_cs_registers #(
...
-      CSR_MCYCLEH: begin
-        if (csr_we_int) begin
-          mhpmcounterh_we[0] = 1'b1;
-        end
-      end
+        CSR_MCOUNTINHIBIT: mcountinhibit_we = 1'b1;
+        CSR_MCYCLE:        mhpmcounter_we[0] = 1'b1;
+        CSR_MCYCLEH:       mhpmcounterh_we[0] = 1'b1;
+        CSR_MINSTRET:      mhpmcounter_we[2] = 1'b1;
+        CSR_MINSTRETH:     mhpmcounterh_we[2] = 1'b1;
```
vogelpi: Sorry, I was not fully clear. I meant this:

        CSR_MCOUNTINHIBIT: mcountinhibit_we   = 1'b1;
        CSR_MCYCLE:        mhpmcounter_we[0]  = 1'b1;
        CSR_MCYCLEH:       mhpmcounterh_we[0] = 1'b1;
        CSR_MINSTRET:      mhpmcounter_we[2]  = 1'b1;
        CSR_MINSTRETH:     mhpmcounterh_we[2] = 1'b1;

MarekPikula: Ah, yeah. I'm half asleep today.

vogelpi: Thanks!

PR #366, opened 2019-10-02
[DV] Update ibex log regex

udinator:
Log parsing update to match the new modifications to the tracer.
Will not merge until #360 is merged, to avoid breaking any current flows.

### DIFF
--- a/dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -30,10 +30,10 @@ def process_ibex_sim_log(ibex_log, csv):
         for line in f:
             if re.search("ecall", line):
               break
-            # Extract instruction infromation
-            m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+) " \
-                          "(?P<pc>[0-9a-f]+) (?P<bin>[0-9a-f]+) (?P<instr>.*)" \
-                          "x(?P<rd>\d+)=0x(?P<val>[0-9a-f]+)", line)
+            # Extract instruction information
+            m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+" \
+                          "(?P<pc>[0-9a-f]+)\s+(?P<bin>[0-9a-f]+)\s+(?P<instr>.*)" \
+                          ".*x(?P<rd>\d+)=0x(?P<val>[0-9a-f]+)", line)
             if m:
                 # Write the extracted instruction to a csvcol buffer file
                 rv_instr_trace = RiscvInstructiontTraceEntry()


dv/uvm/sim.py
```diff
@@ -178,8 +178,8 @@ def compare(test_list, iss, output_dir, verbose):
       logging.info("Comparing %s/DUT sim result : %s" % (iss, elf))
       run_cmd(("echo 'Test binary: %s' >> %s" % (elf, report)))
       uvm_log = ("%s/rtl_sim/%s.%d/sim.log" % (output_dir, test['test'], i))
-      rtl_log = ("%s/rtl_sim/%s.%d/trace_core_00000000.log" % (output_dir, test['test'], i))
-      rtl_csv = ("%s/rtl_sim/%s.%d/trace_core_00000000.csv" % (output_dir, test['test'], i))
+      rtl_log = ("%s/rtl_sim/%s.%d/trace_core__00000000.log" % (output_dir, test['test'], i))
+      rtl_csv = ("%s/rtl_sim/%s.%d/trace_core__00000000.csv" % (output_dir, test['test'], i))
```
imphil: Oh, that double `__` looks unintentional. I'll double-check.

imphil: Opened https://github.com/lowRISC/ibex/pull/367 to get this fixed.

dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -31,9 +31,9 @@ def process_ibex_sim_log(ibex_log, csv):
             if re.search("ecall", line):
               break
             # Extract instruction infromation
```
imphil: unrelated nit: infromation -> information

dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -31,9 +31,9 @@ def process_ibex_sim_log(ibex_log, csv):
...
-            m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+) " \
-                          "(?P<pc>[0-9a-f]+) (?P<bin>[0-9a-f]+) (?P<instr>.*)" \
-                          "x(?P<rd>\d+)=0x(?P<val>[0-9a-f]+)", line)
+            m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+" \
+                          "(?P<pc>[0-9a-f]+)\s+(?P<bin>[0-9a-f]+)\s+(?P<instr>.*)" \
+                          ".*x(?P<rd>\d+)=0x(?P<val>[0-9a-f]+)", line)
```
imphil: Now that the entries are tab-separated, you can use `line.strip().split("\t")`

That's code I've been using:


```python
def get_instructions_from_trace(trace_file):
    instructions = {}
    with open(trace_file) as f:
        line_cnt = -1
        for line in f:
            line_cnt += 1
            if line_cnt == 0:
                continue
            elements = line.strip().split("\t")
            time = elements[0]
            cycle = elements[1]
            pc = int(elements[2], base=16)
            instr = int(elements[3], base=16)
            opcode = elements[4]
            desc = ''
            if len(elements) > 5:
                desc = elements[5]
            instructions[pc] = {'instr': instr, 'opcode': opcode, 'desc': desc, 'pc': pc}
    return instructions
```

dv/uvm/sim.py
udinator: Ah sure, I've approved your related filename PR and reverted the filename in this script.

dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: Good catch, thanks for pointing it out.

dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: I think keeping our regex is better for our post-simulation comparison flows, as we only really care about the RD register and the rd_wdata, I think the regex allows us to extract that information more easily for our use-case

imphil: Sure, whatever works for your use case. (If you want more structured data, a separate tracer that makes use of the RVFI interface might be more future-proof, but that's not for now.)

PR #368, opened 2019-10-03
RISC-V Compliance test: Enable tracing

imphil:
Use the tracing-enabled core wrapper when executing the RISC-V
compliance test suite to help with debugging it.


vogelpi:
Good idea, LGTM!

PR #371, opened 2019-10-04
[DV] Top level toggle coverage

udinator:
1) Enable toggle coverage of only top level I/O pins
2) Tighten ibex log regex to exclude writes to `x0`

### DIFF
--- a/dv/uvm/cover.cfg
+++ b/dv/uvm/cover.cfg
@@ -1,1 +1,4 @@
 +tree core_ibex_tb_top.dut
+begin tgl
+  -tree core_ibex_tb_top.dut.*
+end
--- a/dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -33,7 +33,7 @@ def process_ibex_sim_log(ibex_log, csv):
             # Extract instruction information
             m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+" \
                           "(?P<pc>[0-9a-f]+)\s+(?P<bin>[0-9a-f]+)\s+(?P<instr>.*)" \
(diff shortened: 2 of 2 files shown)


dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -33,7 +33,7 @@ def process_ibex_sim_log(ibex_log, csv):
             # Extract instruction information
             m = re.search(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+" \
                           "(?P<pc>[0-9a-f]+)\s+(?P<bin>[0-9a-f]+)\s+(?P<instr>.*)" \
-                          ".*x(?P<rd>\d+)=0x(?P<val>[0-9a-f]+)", line)
+                          ".*x(?P<rd>[1-9]\d*)=0x(?P<val>[0-9a-f]+)", line)
```
taoliug: why 1-9 only? what about x20, x30?

dv/uvm/cover.cfg
```diff
@@ -1 +1,4 @@
 +tree core_ibex_tb_top.dut
+begin tgl
+  -tree core_ibex_tb_top.dut
```
taoliug: Do you need a .* at the end?

dv/uvm/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: That's covered, because the '[1-9]' will only apply to the first digit of the register number, the '\d+' will cover any digits after that.

dv/uvm/cover.cfg
udinator: Do I need that even if I'm just referring to the top level dut?

taoliug: I think you need this as you only want to disable the toggle coverage below the top level.

udinator: Yes, you're right, it's been updated.

PR #375, opened 2019-10-05
Tracer: Reference Verilator bug for miscompilation

### LINKED ISSUE #372: Tracer: Debug miscompilation in decode_load_insn()
In https://github.com/lowRISC/ibex/blob/2acb497d22a5da877a9c4350ef5693d0c767cc61/rtl/ibex_tracer.sv#L584-L613 we switched from case to if statements to get around potentially wrong behavior in Verilator. Note that the `return` in the default statement hides that the default statement is always taken; the output without the return looks even more confusing.

Upstream bug report at https://www.veripool.org/issues/1536-Verilator-Misoptimization-in-if-and-case-with-default-statement-inside-a-function

imphil:
This miscompilation is now reported and fixed, mention it in the code to
refer back to it when we can revert back to the old code (which will
take a couple months until we can require a newer Verilator version).

Fixes #372

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -583,7 +583,10 @@ module ibex_tracer (
     string      mnemonic;
 
     /*
-    TODO: Gives wrong results in Verilator for unknown reasons. Debug what's going on.
+    Gives wrong results in Verilator < 4.020.
+    See https://github.com/lowRISC/ibex/issues/372 and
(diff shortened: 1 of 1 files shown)


asb:
Though given you're closing #372 with this commit, we should probably have a new tracking bug for removing the workaround and bumping the minimum verilator version (or a meta-bug for bumping the minimum verilator version and removing any bugfix workarounds from the codebase).

PR #377, opened 2019-10-07
[RTL LSU] Change unaligned error behaviour

### LINKED ISSUE #374: DMEM fault during misaligned memory access
Hi all,
When running this set of instructions:
Screenshot from 2019-10-04 16-43-38
The following waves were generated:
Screenshot from 2019-10-04 16-47-11
The `sw` instructions at `PC=0x800058d0` has a misaligned memory address, and is thus split into two internal memory transactions. In the waves, around 395,800,000ps corresponds to the memory response to the first of these two transactions, which is when the testbench randomly asserts a memory fault here. According to Ibex documentation for this situation, the second memory transaction should continue as normal, ignoring the response data, and then the core needs to also jump to the corresponding fault handler (located at `PC=0x80014000` in our case). The LSU's FSM transitions to it's error handling states from the misaligned transaction states, and the core jumps to `0x80014000` (which encodes the instruction `j 80014080`) directly after `0x800058d0`, but while executing this jump instruction, the core jumps to `0x8001e112` instead, throwing an illegal instruction exception, and then proceeds to wait for the `rvalid` corresponding to the second part of the original misaligned transaction. This occurs around 395,835,000ps, and after this, the core jumps back to the fault handler, and fully executes it. This intermediate PC of `0x8001e112` is calculated as the incorrect jump address from load address data that's already in the execution pipeline, and is muxed through due to `lsu_addr_incr_req_i` going high while waiting for the second half of the misaligned memory transaction to complete. Maybe a potential solution for this would be to only jump to the fault handler after the second part of the misaligned transactions has completed?
Best,
Udi

tomroberts-lowrisc:
- If an error is signalled in the first part of an unaligned
  transaction, the second part is issued anyway (to avoid a
  timing loop from data_err_i -> data_req_o)
- Previously, the error was signalled and an exception was taken
  as soon as the first transaction completed. This commits makes
  the core wait until both transactions have completed.
- This prevents strange behaviour caused by the lsu and controller
  being out of sync (fixes #374).

### DIFF
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -42,7 +42,7 @@ module ibex_load_store_unit (
     output logic [31:0]  addr_last_o,          // address of last transaction      -> to controller
                                                // -> mtval
                                                // -> AGU for misaligned accesses
-    output logic         data_valid_o,         // LSU has completed transaction    -> to
+    output logic         data_valid_o,         // LSU has completed transaction    -> to ID/EX
 
     // exception signals
     output logic         load_err_o,
@@ -53,14 +53,16 @@ module ibex_load_store_unit (
 
   logic [31:0]  data_addr;
   logic [31:0]  data_addr_w_aligned;
-  logic [31:0]  addr_last_q, addr_last_d;
+  logic [31:0]  addr_last_q;
 
-  logic         data_update;
-  logic [31:0]  rdata_q, rdata_d;
-  logic [1:0]   rdata_offset_q, rdata_offset_d;
-  logic [1:0]   data_type_q, data_type_d;
-  logic         data_sign_ext_q, data_sign_ext_d;
-  logic         data_we_q, data_we_d;
+  logic         addr_update;
+  logic         ctrl_update;
+  logic         rdata_update;
+  logic [31:8]  rdata_q;
+  logic [1:0]   rdata_offset_q;
+  logic [1:0]   data_type_q;
+  logic         data_sign_ext_q;
+  logic         data_we_q;
 
   logic [1:0]   wdata_offset;   // mux control for data to be written to memory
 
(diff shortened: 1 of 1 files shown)


tomroberts-lowrisc:
Thanks @udinator for the testing and analysis. I have pushed a minor fix to add the missing ``rdata_update`` in ``WAIT_RVALID_DONE``

rtl/ibex_load_store_unit.sv
```diff
@@ -362,18 +371,14 @@ module ibex_load_store_unit (
...
-            ls_fsm_ns = data_gnt_i ? WAIT_RVALID : WAIT_GNT;
-          end
+          // Record the error status of the first part
+          lsu_err_d = data_err_i | pmp_err_q;
+          // Capture the first rdata for loads
+          rdata_update = ~data_we_q;
+          // If already granted, wait for second rvalid
+          ls_fsm_ns = data_gnt_i ? WAIT_RVALID : WAIT_GNT;
+          // Update the address for the second part, if no error
+          addr_update = data_gnt_i & ~(data_err_i | pmp_err_q);
```
vogelpi: `~(data_err_i | pmp_err_q)` could be replaced by `~lsu_err_d` here, but maybe it is not that clean as it is a `_d`-signal.

rtl/ibex_load_store_unit.sv
```diff
@@ -389,57 +394,41 @@ module ibex_load_store_unit (
...
-      WAIT_RVALID_ERR: begin
-        // Wait for the rvalid, but do nothing with it
-        if (data_rvalid_i || pmp_err_q) begin
-          ls_fsm_ns = IDLE;
-        end
-      end
-
       WAIT_RVALID_DONE: begin
+        // tell ID/EX stage to update the address (to make sure the
+        // second address can be captured correctly)
```
vogelpi: This is not only for `mtval` but also to get a possible PMP error for the second part. You may want to clarify this.

vogelpi:
And thanks @udinator for reporting the issue and for testing!

udinator:
@vogelpi - No problem!
@tomroberts-lowrisc - The update to add the `rdata_update` update has fixed the issue, and all of my tests are passing, thanks for the fix!
Best,
Udi

rtl/ibex_load_store_unit.sv
tomroberts-lowrisc: Good point - I will extend the comment

PR #380, opened 2019-10-08
Update description for running RISC-V compliance

towoe:
The RISC-V compliance test system requires an absolute path to the
executable or one which starts at `riscv-test-suite/*`.

I would be happy if someone could verify that a relative path starting in `$RISCV_COMPLIANCE_REPO_BASE` does not work.


imphil:
Yes, it needs to be an absolute path (or a relative path to a base that doesn't seem logical). Thanks for the update!

towoe:
Yes, finding the starting point of the relative path was not so straightforward as I would have hoped.
And I thought it is easier to just recommend the absolute path.

PR #383, opened 2019-10-08
Allow setting RF implementation with FuseSoC use flag

olofk:
**NOTE** This requires https://github.com/olofk/fusesoc/commit/0351dd09dfdd6a6a5e404d1375bf1fafd8b9652c which is not yet in a released FuseSoC version at this point

FuseSoC has gained support for use flags. This patch adds support
for setting the Ibex RF implementation with the ibex_rf_latch use
flag. Default is the FF implementation.

Example: Run the verilator linter with the latch-based RF
fusesoc run --target=lint --flag=ibex_rf_latch lowrisc:ibex:ibex_core

### DIFF
--- a/ibex_core.core
+++ b/ibex_core.core
@@ -22,10 +22,8 @@ filesets:
       - rtl/ibex_multdiv_slow.sv
       - rtl/ibex_prefetch_buffer.sv
       - rtl/ibex_pmp.sv
-      # XXX: Figure out the best way to switch these two implementations
-      # dynamically on the target.
-#      - rtl/ibex_register_file_latch.sv # ASIC
-      - rtl/ibex_register_file_ff.sv # FPGA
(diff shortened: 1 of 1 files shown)


imphil:
Thanks Olof! Is there a way to set/force a use flag by default for a target? E.g. we want to always use the flipflop based register file in Verilator (e.g. in https://github.com/lowRISC/ibex/blob/master/dv/riscv_compliance/ibex_riscv_compliance.core)

olofk:
The use flags support is still very rudimentary. I intend to add things like sanity checks, descriptions and default values eventually but for now I'm pushing out the basic version (which is also the parts guaranteed not to change) to gather some feedback

imphil:
Marking this WIP for the time being, but looks like a good starting point. I'd be happy to take something like this once we can set use flags in targets.

olofk:
Great. Knowing that use flags in targets are important is already good input.

PR #388, opened 2019-10-09
[RTL] Enable reading mstatus.tw

udinator:
Current documentation lists `mstatus.tw` as a `RW` field, but this readability is not reflected in RTL, this PR allows `mstatus.tw` to be read.


udinator:
No problem!

PR #395, opened 2019-10-11
[DV] Let time-consuming checker tasks detect failures

udinator:


dv/uvm/tests/core_ibex_base_test.sv
```diff
@@ -175,20 +175,43 @@ class core_ibex_base_test extends uvm_test;
 
   // Gets the next CORE_STATUS signature write and compares it against the provided core_status
   // type, throws uvm_error on mismatch
-  virtual task check_next_core_status(core_status_t core_status, string error_msg = "");
-    wait_for_mem_txn(cfg.signature_addr, CORE_STATUS);
-    signature_data = signature_data_q.pop_front();
-    `DV_CHECK_EQ_FATAL(signature_data, core_status, error_msg);
+  virtual task check_next_core_status(core_status_t core_status, string error_msg = "",
```
taoliug: if ecall is detected and some of these timeout hasn't expire, will the test fail? Would it be reasonable to set a flag in these tasks and check it when ecall is detected?

udinator: Yes, I think that would be a reasonable check to have as well, I'll add a `wait(dut_vif.ecall === 1'b1)` check as another forked off process in this task.

udinator: Tasks have been updated to monitor for error cases where ecall is detected before the per-task timeout expires.

dv/uvm/tests/core_ibex_base_test.sv
```diff
@@ -175,20 +175,63 @@ class core_ibex_base_test extends uvm_test;
...
+      end
+      begin : wait_timeout
+        timeout_flag = 1'b1;
+        clk_vif.wait_clks(timeout);
+        timeout_flag = 1'b0;
+        `uvm_fatal(`gfn, $sformatf("Did not receive core_status 0x%0x within %0d cycle timeout period",
+                                    core_status, timeout))
+      end
+      begin : ecall_check
+        wait(dut_vif.ecall === 1'b1);
```
taoliug: This is not what I mean, you can just set a global flag upon entering this task, and reset it after it's done. You can also use raise/drop objections in this task to avoid the test from ending prematurely.  When ecall is detected in the other thread, just simply check if this flag is still set.

taoliug: I feel raise/drop objection might be a more UVM way, and you can use the existing timeout mechanism to terminate the test with FATAL.

udinator: I think that if I'm using raise/drop objections in these tasks, having a flag might be unnecessary, as in this situation, the task's FATAL error will be raised anyways, as run_phase won't end until the task's internal timeout runs down.

taoliug: Yes, you will not need a flag if you use raise/drop objection

udinator: Updated these tasks to raise run_phase objections while they are active, allowing coverage of cases where ecall is detected before the task's internal timeout runs out.

PR #397, opened 2019-10-11
[DV] Fix lint warnings

udinator:


dv/uvm/tests/core_ibex_seq_lib.sv
```diff
@@ -33,13 +33,13 @@ class core_base_seq #(type REQ = uvm_sequence_item) extends uvm_sequence#(REQ);
     if(!uvm_config_db#(virtual clk_if)::get(null, "", "clk_if", clk_vif)) begin
        `uvm_fatal(get_full_name(), "Cannot get clk_if")
     end
-    void'(randomize(delay));
+    randomize(delay);
```
taoliug: What's the lint error here? I think this change will cause warning for some simulators

udinator: The lint error is basically that `randomize()` calls should not be void-casted.
Maybe doing something like `assert(randomize(delay))` would work?

taoliug: You can use this macro instead:
https://github.com/lowRISC/ibex/blob/master/dv/uvm/common/utils/dv_macros.svh#L176

udinator: Updated to use the relevant `DV_RANDOMIZE_*` macros.

PR #398, opened 2019-10-15
Debug fix

GregAC:


MarekPikula:
I confirm that after merging Ibex started working again in debug mode with PULPissimo.

GregAC:
Great thanks for the test @MarekPikula. Did you need to turn FENCE.I into a nop also or did it work without that patch?

MarekPikula:
Right now I have FENCE.I as nop. I'll check with it enabled in a sec.

rtl/ibex_controller.sv
```diff
@@ -190,7 +190,12 @@ module ibex_controller (
   // Interrupts //
   ////////////////
 
-  assign enter_debug_mode = debug_req_i & ~debug_mode_q;
+  // Enter debug mode due to an external debug_req_i or because core is in
```
imphil: because *the* core

rtl/ibex_controller.sv
```diff
@@ -190,7 +190,12 @@ module ibex_controller (
...
+  // single step mode (dcsr.step == 1). Single step must be qualified with
+  // instruction valid otherwise core will immediately enter debug mode due to
```
imphil: *the* core

rtl/ibex_controller.sv
```diff
@@ -190,7 +190,12 @@ module ibex_controller (
...
+  // a recently flushed IF (or delay in instruction returning from memory)
```
imphil: or *a* delay in *an* instruction

rtl/ibex_controller.sv
```diff
@@ -346,32 +352,21 @@ module ibex_controller (
             perf_jump_o    = jump_set_i;
           end
 
-
-          // stall IF stage to not starve debug and interrupt requests, these just
-          // need to wait until after the current (multicycle) instruction
+          // If entering debug mode or handling an IRQ need to wait until
```
imphil: If entering debug mode or handling an IRQ *the core needs* to wait until *the* current instruction ~is~ finished executing. Stall IF during that time.

MarekPikula:
I still have this error and I'm unable to continue debugging.

```
Error: Failed to execute fence.i
Error: unable to step rtos hart
```

MarekPikula:
I presume that it's because when `fence.i` is tried to be executed by OpenOCD Ibex immediately goes to exception handler without finishing what was set to be done by OpenOCD. I didn't check it yet, but if you want I can investigate further.

PR #401, opened 2019-10-16
[RTL] Fix single stepping on exceptions

GregAC:
When we single step over an instruction that causes an exception DPC
should be set to point to the exception handler (where we would have
gone were we not single stepping).

As discussed on Zulip this looks to be the correct behaviour, see response to issue on riscv-debug-spec here https://github.com/riscv/riscv-debug-spec/issues/497

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -566,10 +566,13 @@ module ibex_controller (
           end
         end // exc_req_q
 
-        // single stepping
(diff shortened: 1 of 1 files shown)


GregAC:
Oh and I've done a little bit of testing on this. Had single step set and executing an invalid instruction, seemed to the do the right thing (jump to debug handler, dpc pointed to exception handler on dret correctly execute said exception handler)

GregAC:
Thanks for the swift review @imphil and @vogelpi

PR #412, opened 2019-10-21
[RTL] Add RV32M_MultType parameter in top level

MarekPikula:
I thought that it would be nice to have this switch in top level, so that everything can be configured from here. Slow multiplication is already described in documentation, but there is no information how to enable it.


rtl/ibex_core.sv
```diff
@@ -18,6 +18,7 @@ module ibex_core #(
     parameter int unsigned MHPMCounterWidth = 40,
     parameter bit          RV32E            = 1'b0,
     parameter bit          RV32M            = 1'b1,
+    parameter bit          RV32M_MultType   = 1'b1,  // 0 -> SLOW, 1 -> FAST
```
imphil: Can you rename this parameter to `MultiplierImplementation`, and make it a string parameter ("slow", "fast")?

rtl/ibex_core.sv
```diff
@@ -478,7 +479,8 @@ module ibex_core #(
   assign unused_illegal_insn_id = illegal_insn_id;
 
   ibex_ex_block #(
-      .RV32M ( RV32M )
+      .RV32M     ( RV32M          ),
+      .MULT_TYPE ( RV32M_MultType )
```
imphil: Please name the parameter consistently through the design

rtl/ibex_core.sv
MarekPikula: What about enum? To be honest I don't really like strings in Verilog code.

MarekPikula: And if you agree to use enum how to define it in `ibex_(tracing_)core.core`?

imphil: You can use the enum value name directly in the core file, but I'd consider that even more risky than passing strings around. Support for that is pretty solid in my experience, enums passed over the command line are significantly less common and hence more fragile.

MarekPikula: Ok, sounds convincing enough for me.

MarekPikula: So it should be `datatype: string` in `.core` files?

imphil: almost :smile: `datatype: str` it is

ibex_core.core
```diff
@@ -54,6 +54,10 @@ parameters:
     datatype: bool
     paramtype: vlogparam
 
+  MultiplierImplementation:
+    datatype: str
+    paramtype: vlogparam
```
imphil: Can you add a description and a default value to this parameter?

```
description: multiplier implementation. Valid values: fast, slow
default: fast
```

ibex_core_tracing.core
```diff
@@ -40,6 +40,10 @@ parameters:
     datatype: bool
     paramtype: vlogparam
 
+  MultiplierImplementation:
+    datatype: str
+    paramtype: vlogparam
```
imphil: Please add a description here as well.

rtl/ibex_ex_block.sv
```diff
@@ -125,6 +124,8 @@ module ibex_ex_block #(
         .valid_o            ( multdiv_valid         ),
         .multdiv_result_o   ( multdiv_result        )
     );
+  end else begin : gen_multdiv_error
+    $error("Unsupported MultiplierImplementation!");
```
imphil: Unfortunately, `$error` is ignored by many synthesis tools and doesn't reliably result in a synthesis-time error. At the same time, some tools (I forgot which one) don't like `$error` at all and bail out. So as bad as it is: please remove it. (The other option is to intentionally introduce a parse error down there, which works in my experience, but is fragile again.)

MarekPikula: When I tested it with my Docker image when wrong parameter has been given Verilator indeed didn't handle `$error` as I'd imagine, but it showed some other linter errors about unconnected signals.

So yeah, I guess I'll remove it.

MarekPikula:
Also I think that there should be added test case for it for CI. I've seen some linter errors when choosing slow multiplier when run with my Docker image.

Maybe I should open separate issue about it, with all the details?

imphil:
> Also I think that there should be added test case for it for CI.

Yes, that would be great. Should be easy enough by just adding a second call to fusesoc to the azure-pipelines.yml file which sets the `--MultiplierImplementation=slow`. Please feel free to do so in a PR or file a new issue for it.

ibex_core.core
MarekPikula: BTW `description` requires quotation marks in this case.

MarekPikula:
> > Also I think that there should be added test case for it for CI.
> 
> Yes, that would be great. Should be easy enough by just adding a second call to fusesoc to the azure-pipelines.yml file which sets the `--MultiplierImplementation=slow`. Please feel free to do so in a PR or file a new issue for it.

I think it would be better to do it in separate PR as there are some linter errors present, which I didn't look into too much and I didn't look into the architecture of slow multiplier yet..

MarekPikula:
IMHO it looks better now :stuck_out_tongue_winking_eye:

PR #413, opened 2019-10-21
[DV] Update simulation terminate argument

towoe:
Add description to usage output.
Add short option '-c'.

### DIFF
--- a/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
+++ b/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
@@ -68,12 +68,13 @@ void VerilatorSimCtrl::PrintHelp() const {
             << "\n"
                "\n";
   if (tracing_possible_) {
-    std::cout << "-t|--trace             Write a trace file from the start\n";
+    std::cout << "-t|--trace                Write trace file from the start\n";
   }
-  std::cout << "-r|--rominit=VMEMFILE  Initialize the ROM with VMEMFILE\n"
-               "-m|--raminit=VMEMFILE  Initialize the RAM with VMEMFILE\n"
-               "-f|--flashinit=VMEMFILE  Initialize the FLASH with VMEMFILE\n"
-               "-h|--help              Show help\n"
+  std::cout << "-r|--rominit=VMEMFILE     Initialize the ROM with VMEMFILE\n"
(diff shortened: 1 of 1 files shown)


dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
```diff
@@ -68,12 +68,14 @@ void VerilatorSimCtrl::PrintHelp() const {
...
   }
-  std::cout << "-r|--rominit=VMEMFILE  Initialize the ROM with VMEMFILE\n"
-               "-m|--raminit=VMEMFILE  Initialize the RAM with VMEMFILE\n"
-               "-f|--flashinit=VMEMFILE  Initialize the FLASH with VMEMFILE\n"
-               "-h|--help              Show help\n"
+  std::cout << "-r|--rominit=VMEMFILE       Initialize the ROM with VMEMFILE\n"
+               "-m|--raminit=VMEMFILE       Initialize the RAM with VMEMFILE\n"
+               "-f|--flashinit=VMEMFILE     Initialize the FLASH with VMEMFILE\n"
+               "-c|--term-after-cycles=NUM  Terminate simulation after NUM "
+               "cycles\n"
```
imphil: How about
```
               "-c|--term-after-cycles=N Terminate simulation after N cycles\n"
```

towoe: :)
But now that you mentioned it, I noticed that after using only `N` the trace option was using 81 characters so I removed the `a `. Feels better to not exceed the 80 characters.

imphil: I also removed the double whitespace before "Terminate" :-)

towoe: Really? I thought the double spaces were there intentionally to add some separation.

Not to get too much off topic, but should there be a maximum line width for the generated output? We should tackle the important questions first.

imphil: :smile: 80 characters for generated output sounds about right, with the same motivation as 80 characters for source code: that's what your (old school) UNIX terminal is able to display.

Either way, let's not give it too much thought and merge as-is.

towoe: Need to measure a lot of output then.

PR #415, opened 2019-10-22
[DV] Enable timer interrupts

udinator:


dv/uvm/tests/core_ibex_seq_lib.sv
```diff
@@ -73,7 +73,7 @@ class irq_raise_single_seq extends core_base_seq#(irq_seq_item);
     start_item(irq);
     // TODO(udinator) -  constrain irq_timer to 0 for now due to timer interrupt causing spike
```
taoliug: Remove this comment? BTW, does the test pass?

udinator: Yep, tests are still fine, as mie.mtie is disabled in riscv-dv generation

PR #416, opened 2019-10-23
[RTL/Tracer] Fix compressed jump RD write

### LINKED ISSUE #414: [tracer] C.JALR does not display write to x1
Hi all,
Screenshot from 2019-10-21 11-27-57
As shown from this trace, even though `c.jalr` instruction will write `PC+2` to `x1`, this write is not reflected in the tracer log.
Best,
Udi

udinator:
Fixes #414.

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -463,7 +463,13 @@ module ibex_tracer (
 
   function void decode_cr_insn(input string mnemonic);
     if (rvfi_rs2_addr == 5'b0) begin
-      data_accessed = RS1;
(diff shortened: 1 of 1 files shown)


rtl/ibex_tracer.sv
```diff
@@ -463,7 +463,13 @@ module ibex_tracer (
 
   function void decode_cr_insn(input string mnemonic);
     if (rvfi_rs2_addr == 5'b0) begin
-      data_accessed = RS1;
+      if (rvfi_insn[12] === 1'b1) begin
```
imphil: nit: === -> ==

PR #418, opened 2019-10-23
[DV] Increase number of resets in reset_test

udinator:


dv/uvm/riscv_dv_extension/testlist.yaml
```diff
@@ -384,10 +384,10 @@
 - test: riscv_reset_test
   description: >
     Randomly reset the core once in the middle of program execution
-  iterations: 15
+  iterations: 20
   gen_test: riscv_rand_instr_test
   gen_opts: >
-    +instr_cnt=10000
+    +instr_cnt=300000
```
taoliug: How long does this run? How many resets are performed per test?

udinator: These tests run for around 1M-4M cycles, depending on the number of resets that are asserted.
I randomize the number of resets, but I'm generally seeing somewhere between 5-20 resets on average per test, at least for the tests I've run with this as of right now.

taoliug: I will recommend the following:
- Increase the number of reset per test
- Reduce or keep the instruction count per test.
- Check if you can constrain gnt/rvalid delay to hit it with just a couple of test.

udinator: I increased the number of resets by quite a bit, and reduced instr_cnt to the riscv-dv default, and added the directed load/store sequence to this test, that was enough to fully cover all LSU reset situations.

Issue #420, opened 2019-10-24
ALU: Remove dead code


vogelpi:
Inside `ibex_pkg.sv` and `ibex_alu.sv`, there are a couple of unused ALU operations defined (ALU_GT, ALU_GTU, ALU_LE, ALU_LEU). These cannot be triggered as the decoder does not support them.

These operations are remnants of the original OpenRISC design. RISC-V does not have the corresponding instructions and instead implements the ops by reversing operands to (BLT, BLTU, BGE, BGEU, see p.23 of the RISC-V spec).

I suggest to remove those fragements.

PR #424, opened 2019-10-24
[rtl/alu] Remove unused comparison operations

vogelpi:
These operations (ALU_GT, ALU_GTU, ALU_LE, ALU_LEU) are remnants of of the original OpenRISC design. RISC-V does not have these instructions and instead implements the operations by reversing operands for ALU_LT, ALU_LTU, ALU_GE, ALU_GEU.

This resolves lowRISC/ibex#420 reported by @udinator.


vogelpi:
@tomroberts-lowrisc and @GregAC, would either of you mind to launch the Ibex regression suite for this, please?

Issue #421, opened 2019-10-24
LSU: Remove/correct obsolete assertion


vogelpi:
Inside `ibex_load_store_unit.sv`, we have an assertion checking that in the RVALID state, a grant can only be received while also getting the rvalid. This assertion has become obsolete in the process of adding PMP support. Currently, the RVALID state is only entered when a grant is no longer expected.

This assertion should be removed or corrected.

PR #425, opened 2019-10-24
[rtl/lsu] Rework assertion checking response valid

vogelpi:
This commit replaces an obsolete assertion with one checking that response valid is received only if the LSU is ready to handle it.

This resolves lowRISC/ibex#421 reported by @udinator.

### DIFF
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -483,18 +483,11 @@ module ibex_load_store_unit (
   ////////////////
 
 `ifndef VERILATOR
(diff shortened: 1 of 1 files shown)


vogelpi:
@udinator , would you mind to have a look at this?

udinator:
@vogelpi - I tested this assertion with memory-operation intensive tests, and everything looked okay to me, the assertion did not fire off at anytime, and the tests all passed correctly. LGTM.

vogelpi:
Thanks @udinator !

PR #426, opened 2019-10-24
[DV] Increase number of illegal instructions generated

udinator:
Additionally:
1) Add configuration file for OVPsim, and update Makefile to pass correct config file into iss_sim target
2) Additionally, changed default ISS to OVPsim, it has been fully configured for Ibex, and can handle `mie.mtie` being enabled without any weird side-effects.

### DIFF
--- a/dv/uvm/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/riscv_dv_extension/testlist.yaml
@@ -90,10 +90,10 @@
     instruction and handle corresponding exception properly. An exception
     handling routine is designed to resume execution after illegal
     instruction exception.
-  iterations: 10
+  iterations: 15
   gen_test: riscv_rand_instr_test
   gen_opts: >
-    +illegal_instr_ratio=5
+    +illegal_instr_ratio=25
   rtl_test: core_ibex_base_test
 
 - test: riscv_hint_instr_test
--- a/dv/uvm/Makefile
+++ b/dv/uvm/Makefile
@@ -17,7 +17,7 @@ COV                 := 0
 # RTL simulator
 SIMULATOR           := "vcs"
 # ISS (spike, ovpsim)
-ISS                 := "spike"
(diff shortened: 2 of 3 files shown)


dv/uvm/riscv_dv_extension/testlist.yaml
```diff
@@ -90,10 +90,11 @@
     instruction and handle corresponding exception properly. An exception
     handling routine is designed to resume execution after illegal
     instruction exception.
-  iterations: 10
+  iterations: 20
   gen_test: riscv_rand_instr_test
   gen_opts: >
-    +illegal_instr_ratio=5
+    +instr_cnt=25000
+    +illegal_instr_ratio=100
```
taoliug: The number seems to be too much to me, have you try how long this test will run?

udinator: Around 1M cycles, give or take

taoliug: How many instruction is simulated? I don't think we need to scale the test this much to hit one corner case. If it's really needed, something needs to be fixed in constraint to make the random distribution more balanced.

udinator: Not exactly sure how many instructions are being simulated, I'm running a test right now.
I'll play around with illegal instruction constraints in the riscv-dv repo to try and fine-tune the generation model.

taoliug: The general guideline here is to fix constraint to cover the corner cases more efficiently. There are some common pitfalls which could make the randomization distribution biased, we should fix those issues.

udinator: Agreed. I removed the high instr_cnt, and took down the illegal_instr_ratio from 100 to 25, still high enough to allow for enough variety among the illegal csr operations.

Issue #100, opened 2019-06-26
[tracking] Pass RISC-V compliance test suite


imphil:
We want to execute and pass the RISC-V compliance tests on Ibex.

Setup tasks
- Create target to run the compliance tests in Verilator: #99

Known issues:
- [x] Standards-compliant trap/interrupts: #70
- [x] Implement the MSCRATCH CSR: #97

davideschiavone:
We added MSCRATCH and ported RISCY to 1.10 spec.
You will find the TB to run the compliance tests here: https://github.com/pulp-platform/riscv/tree/master/tb/core

Best
Davide

imphil:
We now pass all tests, except for four tests which require a writable mtvec. Requiring this is a bug in the compliance test suite, as a read-only mtvec is explicitly allowed by the standard.

imphil:
Follow-up discussion to fix this and other issues with riscv-compliance is happening at #214.

imphil:
We are now using the unmodified upstream test suite and have writable mtvec. But we still fail on these five tests:

```
Check      I-EBREAK-011,4d0
< 00000003
< 11111111
< 00000000
< 00000000
 ... FAIL
Check       I-ECALL-011,4d0
< 0000000b
< 11111111
< 00000000
< 00000000
 ... FAIL
Check     I-FENCE.I-011,4d0
< 00000030
< 00000012
< 00000042
< 001101b3
 ... FAIL
Check I-MISALIGN_JMP-011,36d0
< 00000002
< 00000000
< 11111111
< 22222222
< 33333333
< 44444444
< 00000002
> ffffffff
> ffffffff
> ffffffff
 ... FAIL
Check I-MISALIGN_LDST-012,10c2,10
< 91a1b1c1
< 91a1b1c1
< 91a1b1c1
< 00000002
< 00000006
< 00000003
< 00000006
---
> 00000099
> 00009900
> 00990000
> ff000000
> ffffffff
> ffffffff
> ffffffff
> ffffffff
> ffffffff
38c38
< 99999999
---
> 99000099
40,44c40,44
< 99999999
< 00000001
< 00000006
< 00000003
< 00000006
---
> 00999999
> ffffff00
> ffffffff
> ffffffff
> ffffffff
 ... FAIL
```

More investigation is needed to see what's actually going on with these tests.

imphil:
@eroom1966 did you already look into some of these issues?

towoe:
Had a quick look at the failures.
[...]

PR #427, opened 2019-10-25
Reduce mtvec alignment requirement

towoe:
RISC-V Privileged Spec v1.11 allows BASE filed of mtvec to be 4-byte
aligned.

Update the assignment and usage of mtvec to use 4-byte aligned
addresses.

This fixes `I-EBREAK` and `I-ECALL` tests of RISC-V compliance failures
reported in lowrisc/ibex#100.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -378,6 +378,7 @@ module ibex_cs_registers #(
     mepc_d       = mepc_q;
     mcause_d     = mcause_q;
     mtval_d      = mtval_q;
+    // TODO Should this assignment use boot_addr_i[31:1]? [7:2] always 6'b0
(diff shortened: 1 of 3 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -378,6 +378,7 @@ module ibex_cs_registers #(
     mepc_d       = mepc_q;
     mcause_d     = mcause_q;
     mtval_d      = mtval_q;
+    // TODO Should this assignment use boot_addr_i[31:1]? [7:2] always 6'b0
```
towoe: Suggestions?

towoe:
Thanks @vogelpi for the feedback.
I agree that the changes from this PR are not the correct fix as it will mess up the interrupts, thanks for pointing that out!
Will create an issue to track this.

PR #433, opened 2019-10-28
Example files for the Nexys4DDR board.

pcotret:
As I was playing with the Nexys4DDR board, here are the example files.
Small difference on the reset signal compared to the Arty (as it doesn't have a ChipKit connection).


examples/fpga/nexys4a7ddr/data/pins_nexys4a7ddr.xdc
```diff
@@ -0,0 +1,15 @@
+## Based on https://github.com/Digilent/digilent-xdc/blob/master/Arty-A7-100-Master.xdc
+## This file is a general .xdc for the Arty A7-100 Rev. D
```
towoe: This looks like a copy and paste error.

examples/fpga/nexys4a7ddr/README.md
```diff
@@ -0,0 +1,48 @@
+# Ibex RISC-V Core SoC Example
```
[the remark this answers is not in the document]
pcotret: That would mean a quick change in this file as well, right ?
https://github.com/lowRISC/ibex/blob/master/doc/examples.rst

pcotret:
Made a force push.

I've tried to recompile in a fresh repo, it still works ;)
lowRISC Ibex on Nexys4 DDR

examples/fpga/nexys4a7ddr/data/pins_nexys4a7ddr.xdc
```diff
@@ -0,0 +1,251 @@
+## Based on https://github.com/Digilent/digilent-xdc/blob/master/Arty-A7-100-Master.xdc
```
towoe: Could you change this to: https://github.com/Digilent/digilent-xdc/blob/master/Nexys-4-DDR-Master.xdc

towoe:
> Made a force push.
> 
Thanks @pcotret. And force pushing is the correct way to go here.

I think it is useful to not have only the Arty example in the repo but also an additional example. But I am not sure what the restrictions should be in terms of diversity. Maybe @vogelpi could state his opinion here. Thanks.

pcotret:
Well, it's up to you...

As Nexys4 and Arty have exactly the same FPGA device, it may not be useful to merge my branch (basically, it's just a matter of pin constraints in the `xdc` file).

It would make more sense to have an example on another FPGA family (Zynq, for instance)

towoe:
Thanks for the update @pcotret.
Lets just wait for some more feedback.
And before that you don't have to fix it, but you have now two commit, so a `git rebase -i HEAD~2` should help clean that up. And just a side note, the commits seem to use a mail address which is not registered with github, this is okay, but it might have not been intentionally?

imphil:
Having arty and Nexys 4 DDR /Nexys A7 is OK, but please choose either the newer or the older name.

towoe:
Thanks for the feedback Philipp!

towoe:
@pcotret in #311 a `shared` folder was created, and in #457 the Arty example makes use of this new structure. It should be now easier to re-use the functions by just specifying the dependency.

PR #437, opened 2019-10-30
New utils dir

GregAC:
As discussed in https://github.com/lowRISC/ibex/pull/311 we want a directory where things that aren't Ibex core RTL but may be used by dv stuff, example stuff and other stuff can live.

I've called it utils but happy to name it something else.


GregAC:
I had meant to move the pcount stuff into utils/pcount or similar but forgot to as part of this PR. I'll await initial feedback before doing that.

dv/riscv_compliance/ibex_riscv_compliance.core
```diff
@@ -9,13 +9,12 @@ filesets:
     depend:
       - lowrisc:dv_verilator:simutil_verilator
       - lowrisc:ibex:ibex_core_tracing
+      - lowrisc:ibex:ibex_core
```
towoe: Shouldn't "ibex_core" be already be covered by "ibex_core_tracing"?

GregAC:
Thanks for the feedback @towoe 

With regards to name, other possibilities I came up with were 'common', 'shared', 'misc'. Not really fussed about what we pick but am keen to get something like this setup soon without having to spend too long deciding over a name.

> As the moved files are all part of sim, I guess it might make sense to have a sub-folder in the new directory:

All of the RTL files aren't actually sim-only things. The Simple System PR introduces one that's sim only but the others could be used in FPGA. I hadn't actually looked at the FPGA side of things for this PR which is an oversight.

My intention was to have a .core file for all of the files you may want in sim and then a .core file for things that you may use in FPGA (and maybe one for ASIC synthesis runs).

I was avoiding a specific sim/ directory as I anticipate much of the RTL that ends up here won't be simulation only and may be useful in FPGA. Maybe flip your suggested structure on it's head to have something like:

```
utils/
   rtl/
      *.core
      [RTL that can be used for sim and fpga]
      sim/
         [RTL that is only useful for sim]
      fpga/
         [RTL that is only useful for FPGA]
         xilinx/
            [RTL that is Xilinx specific]
         altera/
            [RTL that is Altera specific]
```

We'd only actually introduce the various sub-dirs proposed there if we had need of them

tomroberts-lowrisc:
I'm leaning more towards something more like `shared`. `utils` is more reminiscent of scripting/tooling for me. Otherwise looks good, we can always switch around the sub-directory structure as we add more things.

towoe:
> My intention was to have a .core file for all of the files you may want in sim and then a .core file for things that you may use in FPGA (and maybe one for ASIC synthesis runs).
> 
> I was avoiding a specific sim/ directory as I anticipate much of the RTL that ends up here won't be simulation only and may be useful in FPGA. Maybe flip your suggested structure on it's head to have something like:
> 
> ```
> utils/
>    rtl/
>       *.core
>       [RTL that can be used for sim and fpga]
>       sim/
>          [RTL that is only useful for sim]
>       fpga/
>          [RTL that is only useful for FPGA]
>          xilinx/
>             [RTL that is Xilinx specific]
>          altera/
>             [RTL that is Altera specific]
> ```
> 
> We'd only actually introduce the various sub-dirs proposed there if we had need of them

I like this structure, but would suggest to move the "*core" files one level up. This would be more in-line with the other "core" files in the repo.

And I would prefer also "shared" over "utils" following @tomroberts-lowrisc reasoning.

GregAC:
Ok I think I like "shared" over "utils" too.

@imphil, @vogelpi any opinions?

GregAC:
As above the pcount stuff should move into here but wanted to get this done before I disappear on holiday so didn't have time for that part, it can be a new PR

imphil:
We took these changes through #311.

PR #440, opened 2019-10-31
[rtl] Rework access to performance counter CSRs

vogelpi:
This commit reworks the code section describing how the performance counter CSRs are accessed by CSR instructions. Instead of using an address mask inside the default case, and excluding CSRs in the same address range previously handled (like `mcycle(h)`, `minstret(h)`, `mcountinhibit`), all performance counter CSRs are now explicitly enumerated. This enhances readability of the code and enhances compatibility with some tools without changing behavior or causing lint problems.

This commit is based on suggestions from @MarekPikula. It replaces the proposal in lowRISC/ibex#336.


MarekPikula:
Sorry for a long delay. It works with no problem under Quartus.

vogelpi:
Thanks for your feedback @MarekPikula ! I will merge this one.

PR #451, opened 2019-11-07
Implement FENCE.I

GregAC:
We should add stimulation of fence.i to the DV and ensure this gets a decent testing. I have done some basic testing with simple system and it looks to work ok.

(I'd be happy to look into the DV work but that won't happen til I return from Holiday)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -114,7 +114,7 @@ jobs:
       export RISCV_TARGET=ibex
       export RISCV_DEVICE=rv32imc
       fail=0
-      for isa in rv32i rv32im rv32imc rv32Zicsr; do
+      for isa in rv32i rv32im rv32imc rv32Zicsr rv32Zifencei; do
         make -C build/riscv-compliance RISCV_ISA=$isa 2>&1 | tee run.log
         if [ ${PIPESTATUS[0]} != 0 ]; then
           echo -n "##vso[task.logissue type=error]"
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -490,20 +490,36 @@ module ibex_decoder #(
       /////////////
 
       OPCODE_MISC_MEM: begin
-        // For now, treat the fence (funct3 == 000) instruction as a nop.
-        // This may not be correct in a system with caches and should be
-        // revisited.
-        // fence.i (funct3 == 001) was moved to a separate Zifencei extension
-        // in the RISC-V ISA spec proposed for ratification, so we treat it as
-        // an illegal instruction.
-        if (instr[14:12] == 3'b000) begin
-          alu_operator_o     = ALU_ADD; // nop
-          alu_op_a_mux_sel_o = OP_A_REG_A;
-          alu_op_b_mux_sel_o = OP_B_IMM;
-          regfile_we         = 1'b0;
-        end else begin
-          illegal_insn       = 1'b1;
-        end
+        // For now, treat the FENCE (funct3 == 000) instruction as a NOP.  This may not be correct
+        // in a system with caches and should be revisited.
+        // FENCE.I will flush the IF stage and prefetch buffer but nothing else.
+        unique case (instr[14:12])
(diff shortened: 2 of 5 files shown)


imphil:
This implements a new ISA extension, so we need to update the docs on https://ibex-core.readthedocs.io/en/latest/introduction.html and wherever we mention an ISA string.


@tomroberts-lowrisc can you have a look at the code itself?

MarekPikula:
I tested it with Quartus and OpenOCD. It works correct.

rtl/ibex_decoder.sv
```diff
@@ -493,17 +496,27 @@ module ibex_decoder #(
...
-        if (instr[14:12] == 3'b000) begin
-          alu_operator_o     = ALU_ADD; // nop
-          alu_op_a_mux_sel_o = OP_A_REG_A;
-          alu_op_b_mux_sel_o = OP_B_IMM;
-          regfile_we         = 1'b0;
-        end else begin
-          illegal_insn       = 1'b1;
-        end
+        // FENCE.I will flush the IF stage and prefetch buffer but nothing
+        // else.
```
vogelpi: You don't need the newline here. We use 100 chars per line.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -74,8 +78,18 @@ module ibex_prefetch_buffer (
   // PMP errors are generated in the address phase, and registered into a fake data phase
   assign instr_or_pmp_err = instr_err_i | rdata_pmp_err_q[0];
 
+  // After flushing the IF and prefetch buffer the IF stage needs to begin
+  // fetching again from the instruction after the flush. This is effectively
+  // a branch to whatever PC was next to be executed.
+  // (Note that addr_o still provided a valid PC even if fifo_out_valid isn't
+  // set).
```
vogelpi: Also here, you can use up to 100 chars per line.

vogelpi: Nit: Should you use present tense here ("still provides")?

rtl/ibex_decoder.sv
```diff
@@ -493,17 +496,27 @@ module ibex_decoder #(
         // For now, treat the fence (funct3 == 000) instruction as a nop.
         // This may not be correct in a system with caches and should be
         // revisited.
-        // fence.i (funct3 == 001) was moved to a separate Zifencei extension
-        // in the RISC-V ISA spec proposed for ratification, so we treat it as
-        // an illegal instruction.
-        if (instr[14:12] == 3'b000) begin
-          alu_operator_o     = ALU_ADD; // nop
-          alu_op_a_mux_sel_o = OP_A_REG_A;
```
tomroberts-lowrisc: Could we set up the ALU operands for current PC + addr increment and make this look like a branch. That should take us to the correct address and flush the IF stage without any extra muxing on the address?

azure-pipelines.yml
```diff
@@ -123,7 +123,7 @@ jobs:
           # There's no easy way to get the test results in machine-readable
           # form to properly exclude known-failing tests. Going with an
           # approximate solution for now.
-          if [ $isa == rv32i ] && grep -q 'FAIL: 5/55' run.log; then
+          if [ $isa == rv32i ] && grep -q 'FAIL: 4/55' run.log; then
```
towoe: This already change in the master.
We can now add `rv32Zifencei` as a make target to the `isa` loop.

rtl/ibex_decoder.sv
GregAC: Good point, a far better way to implement it. I'll alter it to use that method.

GregAC:
Thanks for the feedback everyone. Latest version switches the implementation to use Tom's suggestion (just do a jump to the next PC). Practically it should do pretty much the exact same thing as before. I've done a little bit of testing with simple system and it looks ok. Proper testing should be done but I think that can be done as a seperate PR (may require riscv-dv changes haven't yet checked if it already supports fence.i generation).

PR #457, opened 2019-11-12
Arty shared

towoe:
Use the shared folder to re-use code.

@imphil does `shared/rtl/fpga/xilinx` make sense to you?


examples/sw/led/led.c
```diff
@@ -37,7 +37,7 @@ static int usleep(unsigned long usec) {
 int main(int argc, char **argv) {
   // Any data written to the stack segment will connect the lowest four bits to
   // the board leds
```
imphil: Does that comment need updating?

shared/fpga_xilinx.core
```diff
@@ -0,0 +1,18 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:ibex:fpga_xilinx_shared"
+description: "Collection of useful RTL for Xilinx based examples"
+filesets:
+  files_sv:
+    files:
+      - rtl/fpga/xilinx/prim_clock_gating.sv
+      - rtl/fpga/xilinx/clkgen_xil7series.sv
+      - rtl/ram_1p.sv
```
imphil: I'm not a huge fan of including the same file in multiple core files (that's what dependencies are for), but good enough for now.

examples/sw/led/led.c
towoe: It does, thanks for catching this Philipp.

Issue #444, opened 2019-11-04
Writing unaligned addresses to `dpc` CSR


vogelpi:
Currently, writes to the `dpc` CSR are simply ignored if the address is unaligned. The debug spec does not specify what should happen if an unaligned address is written to `dpc`. In contrast, the privileged spec says that `xepc` can only hold aligned addresses.

I suggest to align the behavior of these two registers, i.e., to make the LSB of `dpc` to be always 0 instead of ignoring writes with non-zero LSB completely.

PR #461, opened 2019-11-12
[rtl] Ignore LSB only for writes to `dpc`

vogelpi:
Without this PR, writes to `dpc` with the LSB set are completely ignored. This PR changes the implementation to ignore the LSB only. This is the same behavior as seen required for `mepc`. This resolves lowRISC/ibex#444.


vogelpi:
Thanks for your review @tomroberts-lowrisc .

PR #464, opened 2019-11-13
Update crt0 load instruction

towoe:
And moved the gitignore from sw/led to sw so it applies to simple_system as well


examples/sw/simple_system/common/crt0.S
```diff
@@ -69,7 +69,7 @@ main_entry:
   jal x1, main
 
   /* Halt simulation */
-  la x5, SIM_CTRL_BASE + SIM_CTRL_CTRL
+  li x5, SIM_CTRL_BASE + SIM_CTRL_CTRL
```
imphil: Am I missing something? Shouldn't load address be correct here?

towoe: I get the following compiler error:
```
crt0.S:72: Error: illegal operands `la x5,0x20000+0x4'
```
`SIM_CTRL_*` are defines, so not a symbol in the sense that `la` requires it.
But maybe there was a change in riscv gcc?
Would be grate if you could just test `make -C examples/sw/simple_system/hello_test`

imphil: Hm, it compiles for me with the GCC version we ship as prebuilt toolchain.

```
$ PATH=/tools/riscv/bin:$PATH riscv32-unknown-elf-gcc --version
riscv32-unknown-elf-gcc (crosstool-NG 1.24.0.37-3f461da) 9.2.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

@lenary would you mind having a look?

towoe: Thanks @imphil.
Than this is my fault for not using the recommended toolchain. I am still at tag v20180629, might have missed the requirement here.

imphil: Not at all, the code could still be wrong. I'm sure @lenary will be able to tell us how to do it correctly.

lenary: Please use `li` for immediates (including `#define`d constants), and `la` for symbol values, this should make it as clear as possible whether a symbol value is involved.

In this case, I don't think you need to add the complexity of a symbol given you already have a define for the value. Especially as the value of the define could be generated by a script (based on where something is configured to appear in memory).

This is something that I will add to the (in-progress) assembly style guide conversation.

towoe: Thanks @lenary for the clarification!

PR #465, opened 2019-11-13
[rtl] Switch to M mode on debug entry

### LINKED ISSUE #463: Debug-mode privilege level
Hi all,
I'm executing some tests after booting Ibex into User-Mode, and asserting debug stimulus, and I see the following:
Screenshot from 2019-11-12 17-28-49
In this test, our debug ROM is located at `0x80004f88`.
We can see that the core's privilege level stays at `PRIV_LVL_U` even after the core has jumped to the debug ROM and `debug_req` has dropped, while the Debug-Mode Spec says that all operation in debug mode should be taken at machine mode. As a result of the current Ibex behavior, illegal instruction exceptions are thrown when we access D-mode CSRs inside of the debug ROM.
However, `dcsr.prv` is being updated properly, and stores the value of `PRIV_LVL_M` after trapping to debug mode, it is just the current privilege mode that is being calculated incorrectly.
Best,
Udi

tomroberts-lowrisc:
- Core should operate as if in M-mode while in debug mode
- Previous priv level is restored from dcsr on DRET
- Fixes #463


tomroberts-lowrisc:
Hi @udinator - please give this a go in your U-mode / debug test. There may be some other issues with the priv mechanism that I haven't covered here.
Thanks

udinator:
Hi @tomroberts-lowrisc, I ran the same test again with the new changes, and from visual inspection of the waves, everything looked fine, and the priv_mode was being changed successfully to PRIV_LVL_M upon debug entry.

PR #466, opened 2019-11-13
Fix formatting of table in simple system

### LINKED ISSUE #454: Simple system: Fix table at the end of the documentation
See the rendering of https://github.com/lowRISC/ibex/tree/master/examples/simple_system

MarekPikula:
Fixes #454


MarekPikula:
Yeah, GitHub unfortunately doesn't like some Markdown table formatting syntax and sometimes it's not as pretty as one might imagine.

Issue #468, opened 2019-11-13
Trace simulator control


towoe:
Using the simple system example:
```
$ fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system
$ build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/Vibex_simple_system --raminit=examples/sw/simple_system/hello_test/hello_test.vmem
```
produces the following trace
```
% tail -n 4 trace_core_00000000.log 
           1260        626      00100308        8082    c.jr    x1        x1:0x001002b8
           1262        627      001002b8        000202b7        lui     x5,0x20   x5=0x00020000
           1264        628      001002bc        00428293        addi    x5,x5,4   x5:0x00020000  x5=0x00020004
           1266        629      001002c0        4305    c.li    x6,1      x6=0x00000001
```

From `crt0.S`
```
  /* Halt simulation */    
  li x5, SIM_CTRL_BASE + SIM_CTRL_CTRL    
  li x6, 1         
  sw x6, 0(x5)
```
we see the last store instruction (which terminates the simulation via `simulator_ctrl.sv +73`) is not part of the trace.
Probably not the highest priority to fix, but wanted to document what I saw.

imphil:
We might want to delay calling $finish when ending the simulation slightly delayed to allow the tracer to capture this last instruction (that's what I'm guessing is going on). Would you mind giving that a try? (The code is in simulator_ctrl.sv.)

PR #470, opened 2019-11-13
Delay simulator finish

towoe:
The tracer needs time to log the final instruction.
Introduce a counter to delay calling the simulation finish.

@imphil is that what you had in mind?

### DIFF
--- a/shared/rtl/sim/simulator_ctrl.sv
+++ b/shared/rtl/sim/simulator_ctrl.sv
@@ -36,6 +36,7 @@ module simulator_ctrl #(
   localparam SIM_CTRL_ADDR = 1;
 
   logic [7:0] ctrl_addr;
+  logic [2:0] sim_finish;
 
   integer log_fd;
 
@@ -52,6 +53,7 @@ module simulator_ctrl #(
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (~rst_ni) begin
       rvalid_o <= 0;
+      sim_finish <= 'b0;
     end else begin
       // Immeditely respond to any request
       rvalid_o <= req_i;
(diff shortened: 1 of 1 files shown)


imphil:
And I assume you tested this and it fixes the problem you reported in #468?

towoe:
> LGTM, but it looks lint-suspicious. Can you double-check Verilator lint on the type widths?

Thought it might be good to be able to easily increase the delay.

Never did the lint check :) Is `fusesoc --cores-root . run --target=lint lowrisc:ibex:ibex_core_tracing` supposed to do the trick? Fails for me, but with warning from another part.

towoe:
> And I assume you tested this and it fixes the problem you reported in #468?

Yes
```
% tail -n 4 trace_core_00000000.log
           1262        627      001002b8        000202b7        lui     x5,0x20   x5=0x00020000
           1264        628      001002bc        00428293        addi    x5,x5,4   x5:0x00020000  x5=0x00020004
           1266        629      001002c0        4305    c.li    x6,1      x6=0x00000001
           1270        631      001002c2        0062a023        sw      x6,0(x5)          x5:0x00020004  x6:0x00000001 PA:0x00020004 store:0x00000001 load:0x00000000
```

imphil:
> Is fusesoc --cores-root . run --target=lint lowrisc:ibex:ibex_core_tracing supposed to do the trick?

That will only lint things which are used in the core itself; the file you modified is not included there. You should see the lint warnings when you compile the simple system for Verilator.

towoe:
> That will only lint things which are used in the core itself; the file you modified is not included there. You should see the lint warnings when you compile the simple system for Verilator.

Thanks for explaining that. No warnings are produced for `sim_finish`. Could be the comparison which makes Verilator assume that the width is necessary.

imphil:
Thanks for checking!

PR #472, opened 2019-11-14
[rtl] Update assertions

### LINKED ISSUE #274: Assertion needs to be disabled while reset is asserted
There are a few assertions in ibex_load_store_unit which are not disabled while reset is asserted. This could cause some false firing if some signals are x when reset is asserted.

e.g. ibex_load_store_unit.sv:457

```
  assert property (
    @(posedge clk) (CS == IDLE) |-> (data_rvalid_i == 1'b0) );

```

vogelpi:
This PR updates the assertions to not fire while the core is in reset. This resolves #274. In addition, some assertions are updated and some new assertions are added.

### DIFF
--- a/rtl/ibex_fetch_fifo.sv
+++ b/rtl/ibex_fetch_fifo.sv
@@ -220,8 +220,9 @@ module ibex_fetch_fifo #(
   // Assertions //
   ////////////////
 `ifndef VERILATOR
-  assert property (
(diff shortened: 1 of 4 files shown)


vogelpi:
@udinator I would be happy if you could give this a try.

vogelpi:
Thanks a lot @udinator ! I am merging this now.

PR #474, opened 2019-11-15
CS registers tb enhancements

tomroberts-lowrisc:
All CSR addresses supported by the TB are now delared as an X
macro list. This list is then used to autogenerate the enum
type, string print helper and an array of all addresses.

The array of addresses can now be used by the randomize generator
to turn an arbitrary integer index into an address. This removes
the restriction on only generating contiguous address ranges noted
in #443.

Add missing signal csr_restore_dtret_i

### DIFF
--- a/dv/cs_registers/reg_driver/csr_listing.def
+++ b/dv/cs_registers/reg_driver/csr_listing.def
@@ -1,0 +1,32 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// List all supported registers
+
+#ifndef CSR
+#error Define CSR
+#endif
+
+CSR(PMPCfg0, 0x3A0)
+CSR(PMPCfg1, 0x3A1)
+CSR(PMPCfg2, 0x3A2)
+CSR(PMPCfg3, 0x3A3)
+CSR(PMPAddr0, 0x3B0)
+CSR(PMPAddr1, 0x3B1)
+CSR(PMPAddr2, 0x3B2)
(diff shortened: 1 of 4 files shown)


dv/cs_registers/reg_driver/csr_listing.def
```diff
@@ -0,0 +1,32 @@
+// Copyright lowRISC contributors.
```
imphil: Making this file `.h` could make it easier for build tools.

dv/cs_registers/reg_driver/register_transaction.h
```diff
@@ -9,6 +9,19 @@
 #include <random>
 #include <string>
 
+// Enumerate the supported register types
+// by macro expansion.
```
imphil: nit: could be one line.

dv/cs_registers/reg_driver/csr_listing.def
tomroberts-lowrisc: What do you think @lenary ?

dv/cs_registers/reg_driver/register_transaction.h
tomroberts-lowrisc: will fix

dv/cs_registers/reg_driver/csr_listing.def
lenary: I think it's useful to have a differentiator between 'headers' -- which contain declarations, and are only included once (or at least, only have an effect the first time they're included ) -- and this kind of data definition -- which will have a different effect every time its included. The file extension seems the best place to make this difference clear. I hope it is relatively easy to teach build tools that a `.def` file might still be a dependency of a compilation unit.

Issue #365, opened 2019-10-02
mstatus.mie reset when hitting breakpoint


MarekPikula:
I'm currently evaluating PULPissimo platform with Ibex core and I encountered some strange behaviour with interrupts. It seems that after hitting breakpoint `mie` in `mstatus` register is reset to 0.

Here is my setup:

- PULPissimo with Ibex core and [pulp-platform/riscv-dbg](https://github.com/pulp-platform/riscv-dbg) as debug core,
- ZedBoard (Xilinx Zynq, `master` version of all cores) and Terasic DE10-Nano (Altera Cyclone-V, with modifications to make the code compatible with Quartus),
- Digilent HS2 JTAG programmer,
- OpenOCD compiled basing on [instructions from PULPissimo](https://github.com/pulp-platform/pulpissimo/#gdb-and-openocd),
- [pulp-platform/pulp-riscv-gnu-toolchain](https://github.com/pulp-platform/pulp-riscv-gnu-toolchain).

OpenOCD config and log as well as gdb log (including issued commands) can be found [here](https://gist.github.com/MarekPikula/63bd5a35dcffa2361a1a7bc3de5642f9).

Code:
```c
#include <stdint.h>

int main() {
  register int mstatus_mie = (1<<3); // mstatus.mie
  asm volatile ("csrw 0x300, %0" :: "r" (mstatus_mie) );

  uint32_t mstatus[3];
  asm volatile ("csrr %0, 0x300" : "=r" (mstatus[0]));
  asm volatile ("nop");  // First breakpoint

  asm volatile ("csrr %0, 0x300" : "=r" (mstatus[1]));
  asm volatile ("nop");  // Second breakpoint

  asm volatile ("csrw 0x300, %0" :: "r" (mstatus_mie) );
  asm volatile ("csrr %0, 0x300" : "=r" (mstatus[2]));
  asm volatile ("nop");  // Third breakpoint

  return 0;
}
```

What the code does is set `mie` flag in `mstatus` register (0x300) and store it in `mstatus[0]`, then after hitting breakpoint read it to `mstatus[1]` and to confirm, `mstatus` register is written again and result stored in `mstatus[2]`.

Basically after sending `x/3x mstatus` gdb returns: `0x00001808	0x00001800	0x00001808`, so in `mstatus[0]` and `mstatus[2]` `mie` flag is set, whereas in `mstatus[1]` it is not set, although, to my understanding, it should be recovered after going out of debug mode.

[...]

PR #475, opened 2019-11-15
Fix behavior of exceptions in debug mode, document debug CSRs

### LINKED ISSUE #168: Check behavior of exceptions and special instructions related to debug mode
Currently, exceptions and special instructions are handled independent of whether the core is in debug mode or not. This is not correct for a couple of cases. For example, the Debug Spec v0.13.2 states on p.39 that while in debug mode:
- ~~WFI acts as a NOP. Currently the core just goes to sleep anyway.~~
- Exceptions do not update any registers such as `mcause`, `mepc`, `mtval`, `dpc` and `mstatus`. Currently, exception handling does not depend on debug mode.

~~Similarly, the DRET instruction must lead to an illegal instruction exception when the core is not in debug mode (p.41). Currently, DRET always restores the PC in dpc even if the core is not in debug mode, thus silently messing up the processor state.~~

### LINKED ISSUE #307: Document debug CSRs
Document dscratch0, dscratch1 and dpc CSRs in the CSR documentation, and link them from the debug documentation page.

vogelpi:
This PR consists of three small commits which implement the following changes:
- Exceptions in debug mode shall not modify any CSRs. This resolves #168 and is related to #365.
- Document debug CSRs. This resolves #307.
- Update documentation of `mhartid` CSR. Previously, the doc was still describing the cluster and core ID bit fields.

### DIFF
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -371,12 +371,5 @@ Hardware Thread ID (mhartid)
 
 CSR Address: ``0xF14``
 
-Reset Value: Defined
-
-+-------+-----+------------------------------------------------------------------+
-| Bit#  | R/W | Description                                                      |
-+-------+-----+------------------------------------------------------------------+
-| 10:5  | R   | **Cluster ID:** ID of the cluster                                |
-+-------+-----+------------------------------------------------------------------+
-| 3:0   | R   | **Core ID:** ID of the core within the cluster                   |
-+-------+-----+------------------------------------------------------------------+
+Reads directly return the value of the ``hart_id_i`` input signal.
+See also :ref:`core-integration`.
--- a/doc/cs_registers.rst
+++ b/doc/cs_registers.rst
@@ -46,7 +46,7 @@ Ibex implements all the Control and Status Registers (CSRs) listed in the follow
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x3BF  | ``pmpaddr15``      | WARL   | PMP Address Register                          |
 +---------+--------------------+--------+-----------------------------------------------+
-|  0x7B0  | ``dcsr``           | RW     | Debug Control and Status Register             |
+|  0x7B0  | ``dcsr``           | WARL   | Debug Control and Status Register             |
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x7B1  | ``dpc``            | RW     | Debug PC                                      |
 +---------+--------------------+--------+-----------------------------------------------+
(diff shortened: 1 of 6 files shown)


doc/cs_registers.rst
```diff
@@ -282,6 +282,75 @@ Reset Value: ``0x0000_0000``
...
+
++-------+-----+------------------------------------------------------------------+
+| Bit#  | R/W | Description                                                      |
++-------+-----+------------------------------------------------------------------+
+| 31:28 | R   | **xdebugver:** 4 = External spec-compliant debug support exists. |
++-------+-----+------------------------------------------------------------------+
+| 15    | RW  | **ebreakm:** EBREAK in M-Mode behaves as described in Privileged |
+|       |     | Spec (0), or enters Debug Mode (1).                              |
++-------+-----+------------------------------------------------------------------+
+| 12    | RW  | **ebreaku:** EBREAK in U-Mode behaves as described in Privileged |
```
imphil: ebreaku was RW in the 0.13.2 spec, and changed to WARL in the latest debug spec (which is actually correct). Maybe describe it as WARL here and add a note that RW was an oversight in the 0.13.2 spec?

doc/cs_registers.rst
```diff
@@ -282,6 +282,75 @@ Reset Value: ``0x0000_0000``
...
+|       |     | Spec (0), or enters Debug Mode (1).                              |
++-------+-----+------------------------------------------------------------------+
+| 8:6   | R   | **cause:** 1 = EBREAK, 3 = halt request, 4 = step                |
++-------+-----+------------------------------------------------------------------+
+| 2     | RW  | **step:** When set and not in debug mode, execute a single       |
+|       |     | instruction and enter debug mode.                                |
++-------+-----+------------------------------------------------------------------+
+| 1:0   | RW  | **prv:** Privilege level when entering debug mode, may be        |
+|       |     | modified by debugger to change privilege level in debug mode.    |
```
imphil: That description is slightly confusing, as I first read it as "the privilege level that debug code is executed as". The current spec says "Contains the privilege level the hart was operating in when Debug Mode was entered." Maybe we can just copy that sentence?

Furthermore, the spec says

> "Not all privilege levels are supported on all harts.
If the encoding written is not supported or the
debugger is not allowed to change to it, the hart
may change to any supported privilege level.

Two points there:
- The spec says prv is RW, but it's actually WARL. I've opened https://github.com/riscv/riscv-debug-spec/pull/498 to get that fixed.
- The spec allows for the core to disallow certain transitions. We currently allow all transitions to supported modes, i.e. to M and U mode. That fact is worth pointing out in the docs.

doc/cs_registers.rst
```diff
@@ -46,7 +46,7 @@ Ibex implements all the Control and Status Registers (CSRs) listed in the follow
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x3BF  | ``pmpaddr15``      | WARL   | PMP Address Register                          |
 +---------+--------------------+--------+-----------------------------------------------+
-|  0x7B0  | ``dcsr``           | RW     | Debug Control and Status Register             |
+|  0x7B0  | ``dcsr``           | WARL   | Debug Control and Status Register             |
```
imphil: Since each field has different behavior, maybe we can remove this here?

doc/debug.rst
```diff
@@ -26,3 +26,10 @@ Parameters
 +---------------------+-----------------------------------------------------------------+
 | ``DmExceptionAddr`` | Address to jump to when an exception occurs while in debug mode |
 +---------------------+-----------------------------------------------------------------+
+
+Core Debug Registers
+--------------------
+
+Ibex implements 4 core debug registers, namely :ref:`csr-dcsr`, :ref:`csr-dpc`, and two debug scratch registers.
```
imphil: 4 -> four

imphil:
LGTM but please have a look at the comments.

doc/cs_registers.rst
vogelpi: Thanks for pointing this out @imphil . I adapted the description as suggested.

doc/cs_registers.rst
vogelpi: Thanks for pointing this out and opening a PR. I think you are right. I adapted the text accordingly + added a note. Let's wait for some feedback regarding the spec before merging this one.

doc/debug.rst
vogelpi: Done.

doc/cs_registers.rst
vogelpi: I would say R is a subset of R/W, and R/W is a subset of WARL. So if, some fields are R only, it is still safe to say that the register is WARL.

Issue #213, opened 2019-08-05
prim_clock_gating.sv: Consolidate various in-tree example versions 


imphil:
We currently have multiple versions of prim_clock_gating.sv in our tree, two pure verilog ones, and one Xilinx-specific one. Figure out where to place this IP so that it can be shared.

vogelpi:
With #457 , the one generic and the Xilinx-specific implementation got moved to `shared/rtl`. I suggest we use the same one for DV and remove the copy in `dv/uvm/tb`.

PR #476, opened 2019-11-15
[dv] Remove clock gating primitive in `dv/uvm/tb`

vogelpi:
This commit removes a copy of the dummy clock gating primitive from
tree previously used for UVM-based DV. There is another, identical copy
of the same primitive in `shared/rtl` that can be used instead.

This resolves lowRISC/ibex#213.

Signed-off-by: Pirmin Vogel <vogelpi@lowrisc.org>


vogelpi:
Verilator lint is working but I would be happy if someone from the DV team could ensure nothing breaks on your side.

vogelpi:
Thanks for your feedback @udinator !

PR #479, opened 2019-11-18
[dv] Prepare CS registers TB for CI

tomroberts-lowrisc:
- Switch from Makefile calling fusesoc to fusesoc calling Makefile
- Pass parameters through DPI rather than as env variables

### DIFF
--- a/dv/cs_registers/README.md
+++ b/dv/cs_registers/README.md
@@ -8,19 +8,15 @@ It is a work in progress and only tests a handful of registers, and is missing m
 How to build and run the testbench
 ----------------------------------
 
-(from this directory)
-
 Verilator version:
 
    ```sh
-   make TOOL=verilator
-   ./build/lowrisc_ibex_tb_cs_registers_0/sim-verilator/Vtb_cs_registers
+   fusesoc --cores-root=. run --target=sim --tool=verilator lowrisc:ibex:tb_cs_registers
    ```
 VCS version:
 
    ```sh
-   make TOOL=vcs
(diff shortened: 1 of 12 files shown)


dv/cs_registers/README.md
```diff
@@ -8,19 +8,15 @@ It is a work in progress and only tests a handful of registers, and is missing m
...
 ----------------------------------
 
-(from this directory)
-
 Verilator version:
 
    ```sh
-   make TOOL=verilator
-   ./build/lowrisc_ibex_tb_cs_registers_0/sim-verilator/Vtb_cs_registers
+     fusesoc --cores-root=. run --setup --build --run --target=sim lowrisc:ibex:tb_cs_registers
```
imphil: nit: indent `fusesoc` with the start of the block (here and below)

You can remove the `--setup --build --run` part, it should be default if not supplied.

Possibly for consistency: add `--tool=verilator`?

dv/cs_registers/env/register_types.h
```diff
@@ -0,0 +1,14 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef REGISTER_TYPES_H_
+#define REGISTER_TYPES_H_
+
+struct CSRParams {
+  unsigned char PMPEnable;
```
imphil: could possibly be `bool`

PR #480, opened 2019-11-18
Proposal: Add Makefile for common tasks

towoe:
This PR is optional. Using the Makefile was helpful for me lately and maybe it could be useful for others, so please give feedback.
One reason to not add this: It might confuse people not familiar with the project structure. Instead of following the documentation, using a target of the Makfile without a set-up will fail.
But the intention of this file is not to create an automatic build flow, just a way to run some tasks for the experienced user.

Instead of copying the commands from the documentation or using the
shell history to find the commands use a Makefile as a convenience to
run some common tasks.
A shell's completion system can make use of the targets and provide tab
completion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,71 @@
+all: help
+
+.PHONY: help
+help:
+	@echo "This is a short hand for running popular tasks."
+	@echo "Please check the documentation on how to get started"
+	@echo "or how to set-up the different environments."
+
+# Use a parallel run (make -j N) for a faster build
+build-all: build-riscv-compliance build-simple-system build-arty
+
+
+# RISC-V compliance
+.PHONY: build-riscv-compliance
+build-riscv-compliance:
+	fusesoc --cores-root=. run --target=sim --setup --build \
+		lowrisc:ibex:ibex_riscv_compliance
+
+
+# Simple system
+# Use the following targets:
+# - "build-simple-system"
+# - "run-simple-system"
+.PHONY: build-simple-system
+build-simple-system:
+	fusesoc --cores-root=. run --target=sim --setup --build \
+		lowrisc:ibex:ibex_simple_system
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,32 @@
...
+build-riscv-compliance:
+	fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_riscv_compliance --RV32M=1 --RV32E=0
+
+# Simple system
+build-simple-system:
+	fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32M=1 --RV32E=0
+sw-simple-hello:
+	make -C examples/sw/simple_system/hello_test
+run-simple-system: sw-simple-hello
+	build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/Vibex_simple_system --raminit=examples/sw/simple_system/hello_test/hello_test.vmem
```
imphil: Please depend on the output, not on the step.

```makefile
sw-simple-hello: examples/sw/simple_system/hello_test/hello_test.vmem

examples/sw/simple_system/hello_test/hello_test.vmem:
	make -C examples/sw/simple_system/hello_test

run-simple-system: examples/sw/simple_system/hello_test/hello_test.vmem
	build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/Vibex_simple_system --raminit=examples/sw/simple_system/hello_test/hello_test.vmem
```

Not adding a dependency to the `Vibex_simple_system` is fine here to avoid needless rebuilds of the simulator (at least until fusesoc can do no-op builds). But you might want to add a test to the run-simple-system target to check if the `Vibex_simple_system` file exists, and error out with a help message ("run build-simple-system") otherwise.

towoe:
Thanks for the review Philipp. I did some clean up. Should there be any other targets included? I just used the ones I am familiar with. But I guess if there is a need by someone, more can be added later.

PR #484, opened 2019-11-21
[ci] Add CS registers TB to CI

tomroberts-lowrisc:
Build and run TB, checking for failures

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -96,6 +96,11 @@ jobs:
     displayName: Lint Verilog source files with Verilator
 
   - bash: |
+      # Build and run CSR testbench
+      fusesoc --cores-root=. run --target=sim --tool=verilator lowrisc:ibex:tb_cs_registers
+    displayName: Build and run CSR testbench with Verilator
+
+  - bash: |
       cd build
       git clone https://github.com/riscv/riscv-compliance.git
     displayName: Get RISC-V Compliance test suite
--- a/dv/cs_registers/env/env_dpi.cc
+++ b/dv/cs_registers/env/env_dpi.cc
@@ -35,7 +35,10 @@ void env_final() {
   delete reg_env;
 }
 
-void env_tick(svBit *stop_req) { reg_env->GetStopReq(stop_req); }
+void env_tick(svBit *stop_req, svBit *test_passed) {
+  reg_env->GetStopReq(stop_req);
+  reg_env->GetTestPass(test_passed);
+}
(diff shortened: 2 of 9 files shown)


azure-pipelines.yml
```diff
@@ -94,6 +94,24 @@ jobs:
       fi
     displayName: Lint Verilog source files with Verilator
 
+  - bash: |
+      # Build CSR testbench
+      fusesoc --cores-root=. run --setup --build --target=sim lowrisc:ibex:tb_cs_registers
+      if [ $? != 0 ]; then
+        echo -n "##vso[task.logissue type=error]"
+        echo "CS registers TB failed to build. Run 'fusesoc --cores-root=. run --setup --build --target=sim lowrisc:ibex:tb_cs_registers' to check and fix all errors."
```
asb: I know the lint task has a simple output, but I think this is probably an anti-pattern. There's risk that the printed command and what is actually executed drifts.

Using `set -x` would be an alternative. It seems there is a [negative interaction](https://docs.microsoft.com/en-us/azure/devops/pipelines/troubleshooting?view=azure-devops) with setting variables via `##vso`, but we're not doing that in these definitions currently. Or we put the command in a string and then eval it or do `bash -c`, but this is getting increasingly ugly...

imphil: We used this construct as workaround in OT mainly because people didn't have access to the actual build logs. So I'd simply remove the echo parts.

The second problem is that Verilator/fusesoc run doesn't return a non-zero exit code if the run fails. The nicest way would probably be an extension to tb_cs_registers to get this information into the return code. If that's too annoying  I'd go for a wrapper script which does essentially the tee and grep-thing.

azure-pipelines.yml
```diff
@@ -94,6 +94,24 @@ jobs:
...
+      fi
+      # Run CSR testbench
+      fusesoc --cores-root=. run --run --target=sim lowrisc:ibex:tb_cs_registers 2>&1 | tee run.log
+      # Get pass / fail status
+      if grep -q 'TEST FAILED' run.log; then
+        echo -n "##vso[task.logissue type=error]"
+        echo "CS registers TB run failed. Run 'fusesoc --cores-root=. run --setup --build --run --target=sim lowrisc:ibex:tb_cs_registers' to see output and fix failures"
+        exit 1
+      fi
+    displayName: Build CSR testbench
```
imphil: Build -> Execute (or "Build and run")

PR #490, opened 2019-11-22
Add Synopsys VCS Support for Ibex Simple System

mehrdadbiglari:
as tool to fusesoc
memory init from RTL
fix signal init values


imphil:
Thanks for this PR @mehrdadbiglari! Can you explain why you added the MEM_FILE parameter, and how the SRAM_INIT_FILE define is not enough in this case?

mehrdadbiglari:
Thanks for the feedback @imphil. I used MEM_FILE because I felt using a parameter would be a better practice for this purpose than a define. But for better compatibility now I have made another pull request using "SRAM_INIT_FILE".

PR #498, opened 2019-11-28
Reverse return code of simutil_verilator_set_mem()

imphil:
All functions in simutil_verilator return booleans to indicate
success/failure, where 1 == true, and 0 == false.
simutil_verilator_set_mem() returns an int due to DPI interface
restrictions. Before, 0 meant success, and 1 meant error.

To keep things nice and consistent, turn things around and align with
the bool meaning.

### DIFF
--- a/shared/rtl/ram_1p.sv
+++ b/shared/rtl/ram_1p.sv
@@ -51,26 +51,26 @@ module ram_1p #(
   end
 
   `ifdef VERILATOR
-    // Task for loading 'mem' with SystemVerilog system task readmemh
+    // Task for loading 'mem' with SystemVerilog system task $readmemh()
     export "DPI-C" task simutil_verilator_memload;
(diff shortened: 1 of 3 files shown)


shared/rtl/ram_1p.sv
```diff
@@ -51,26 +51,26 @@ module ram_1p #(
...
+      $readmemh(file, mem);
     endtask
 
     // TODO: Allow 'val' to have other widths than 32 bit
     function int simutil_verilator_set_mem(input int index,
-      input logic[31:0] val);
-      if (index < Depth) begin
-        mem[index] = val;
+                                           input logic[31:0] val);
+      if (index > Depth) begin
```
towoe: `>=` ?

imphil: Good catch, thanks!

PR #500, opened 2019-11-28
Make VerilatorSimCtrl class a singleton

imphil:
The VerilatorSimCtrl class was always intended to be used only once in
an application, since it sets up a global signal handler and needs to be
accessible from DPI modules. This accessibility was achieved through a
global variable `simctrl`.

With this commit the VerilatorSimCtrl is switched to a singleton class.
The instance is accessible through `VerilatorSimCtrl::GetInstance()`.
The downside of that approach is that we loose the constructor
arguments, and need to deal with a class which potentially hasn't top or
the clock and reset signals set.

### DIFF
--- a/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
+++ b/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
@@ -27,39 +27,13 @@ struct BufferDesc {
   size_t length;
 };
 
-// Static pointer to a single simctrl instance used by SignalHandler
-static VerilatorSimCtrl *simctrl = nullptr;
-
-static void SignalHandler(int sig) {
-  if (!simctrl) {
-    return;
-  }
-
-  switch (sig) {
-    case SIGINT:
-      simctrl->RequestStop(true);
-      break;
-    case SIGUSR1:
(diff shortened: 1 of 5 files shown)


dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
```diff
@@ -87,6 +62,17 @@ VerilatorSimCtrl::VerilatorSimCtrl(VerilatedToplevel &top, CData &sig_clk,
...
 
+bool VerilatorSimCtrl::SetTop(VerilatedToplevel *top, CData *sig_clk,
+                              CData *sig_rst,
+                              VerilatorSimCtrlFlags flags) {
+  top_ = top;
+  sig_clk_ = sig_clk;
+  sig_rst_ = sig_rst;
+  flags_ = flags;
+
+  return true;
```
tomroberts-lowrisc: Is there a reason for returning a bool?

imphil: I thought it might be more consistent, but looking at it again it doesn't make much sense (and I never check the return code anyways.) Changing to `void`.

PR #504, opened 2019-12-02
Add Synopsys VCS Support for Ibex Simple System

mehrdadbiglari:
Add stimuli. Fix compile error for assignment from multiple blocks.

### DIFF
--- a/examples/simple_system/README.md
+++ b/examples/simple_system/README.md
@@ -92,6 +92,26 @@ The simulator produces several output files
 * ibex_simple_system_pcount.csv - A csv of the performance counters
 * trace_core_00000000.log - An instruction trace of execution
 
+## Simulating with Synopsys VCS
+
+Similar to the Verilator flow the Simple System simulator binary can be built using:
+
+```
+fusesoc --cores-root=. run --target=sim --tool=vcs --setup --build lowrisc:ibex:ibex_simple_system --RV32M=1 --RV32E=0 --SRAM_INIT_FILE=`<sw_vmem_file>`
+```
+
+`<sw_vmem_file>` should be a path to a vmem file built as described above, use
+./examples/sw/simple_system/hello_test/hello_test.vmem to run the hello_test
+binary.
+
+To run the simulator:
+
+```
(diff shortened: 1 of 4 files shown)


examples/simple_system/README.md
```diff
@@ -92,6 +92,26 @@ The simulator produces several output files
 * ibex_simple_system_pcount.csv - A csv of the performance counters
 * trace_core_00000000.log - An instruction trace of execution
 
+## Simulating with Synopsys VCS
+
+Similar to the Verilator flow the Simple System simulator binary can be built using:
+
+```
+fusesoc --cores-root=. run --target=sim --tool=vcs --setup --build lowrisc:ibex:ibex_simple_system --RV32M=1 --RV32E=0 --vcs_options="-debug_access+r +define+SRAM_INIT_FILE=`<sw_vmem_file>`
```
imphil: This approach works, but can be slightly improved:

- Your [first attempt](https://github.com/lowRISC/ibex/pull/490/files#diff-3df86ff289fc85ea64fdbffb9a42dedfR47-R50) at this was right: Add the vcs_options to the .core file. 

- For the SRAM_INIT_FILE define, you can add it to to the `parameters` section of the core file, fusesoc will then take care to pass it to VCS in a correct way. 

  ```patch
  diff --git a/examples/simple_system/ibex_simple_system.core b/examples/simple_system/ibex_simple_system.core
  index 6cf7668..973f95a 100644
  --- a/examples/simple_system/ibex_simple_system.core
  +++ b/examples/simple_system/ibex_simple_system.core
  @@ -27,6 +27,11 @@ parameters:
      paramtype: vlogparam
      default: 0
      description: Enable the E ISA extension (reduced register set) [0/1]
  +  SRAM_INIT_FILE:
  +    datatype: str
  +    paramtype: vlogdefine
  +    description: Path to a vmem file to initialize the RAM with
  
  targets:
    sim:
  @@ -36,6 +41,7 @@ targets:
      parameters:
        - RV32M
        - RV32E
  +     - SRAM_INIT_FILE
      toplevel: ibex_simple_system
      tools:
        verilator:
  ```

mehrdadbiglari: Thanks @imphil for the hint and feedback!
I have updated the core and the README accordingly.

PR #508, opened 2019-12-02
RTL simulation script upgrade

TAD-dev:
RTL simulation script was extended to support Cadence simulator.


TAD-dev:
The script is fine. Re-request for review was just a misclick. Sorry for it.

PR #517, opened 2019-12-11
Memutils

tomroberts-lowrisc:


tomroberts-lowrisc:
@towoe and @imphil comments welcome - happy to stick with the current workaround if you think this isn't worth it

dv/verilator/memutil/cpp/verilator_memutil.cc
```diff
@@ -0,0 +1,432 @@
...
+      {"flashinit", required_argument, nullptr, 'f'},
+      {"meminit", required_argument, nullptr, 'l'},
+      {"help", no_argument, nullptr, 'h'},
+      {nullptr, no_argument, nullptr, 0}};
+
+  // Reset the command parsing index in-case other utils have already parsed
+  // some arguments
+  optind = 1;
+  while (1) {
+    int c = getopt_long(argc, argv, ":r:m:f:l:h", long_options, nullptr);
```
GregAC: Probably not something that has to be done for this PR but it'd be nice to purge the C-style argument list handling

dv/verilator/memutil/cpp/verilator_memutil.h
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Parse command line arguments
+   *
+   * This removes all recognized command-line arguments from argc/argv.
+   *
+   * The return value of this method indicates if the program should exit with
+   * retcode: if this method returns true, do *not* exit; if it returns *false*,
+   * do exit.
+   */
+  bool ParseCommandArgs(int argc, char **argv, bool &exit_app);
```
GregAC: exit_app isn't documented.

Looking at the implementation seems it used to indicate you should exit because it's been asked to just list memory regions (i.e. no error has occurred) but the two bools (one return value, one exit now but no error) feels a bit awkward.

Maybe return bool for exit/no exit and pass by reference integer retcode to return? Or have both bools pass by reference so it's more obvious what's going on (one called error, the other exit_app).

imphil: We went around in circles on this one in the previous PRs, and that's the best solution we could come up with. bool return code is the standard convention for all functions in this module. It indicates if the function was successful or not. But even a successful call can mean you want to end the program. That's what exit_app is for.

tomroberts-lowrisc: I'll fix the comment here and in simctrl

imphil:
The main goal of the previous design was to make the toplevel cc file easier by removing all copy-paste error handling from having function calls beyond the `Exec()` one. Instead of adding function calls before Exec(), we should look into "registering hooks", which are then called as part of Exec(). 

So I'd like a toplevel.cc file look like this:

```c
simctrl.RegisterExtension(memutil);
return simctrl.Exec();
```

Memutil is then an extension, looking something like this:

```
class Memutil : public Extension {
  parseCliArguments();
  finish();
  initialize();
}
```

With that in place, you can extend simutil to call the extension hook points, like `parsecliArguments()` at the right point in time. For the command line argument parsing, we'll need to find a way to hook into each other; effectively, you do partial parsing of the command line.

(I didn't think much about the function names.)

This will also solve the problems we have in other modules, e.g. the code after Exec in `ibex_simple_system.cc`.

tomroberts-lowrisc:
Thanks for the pointer @imphill, I've had a go at what you suggested.

PR #518, opened 2019-12-12
[RTL] Only restore from mstack in nmi mode

### LINKED ISSUE #492: MEPC written on `mret` instructions
Hi all,
I noticed that on an `mret` instruction, `mepc` will get overwritten with the contents of `mstack_epc` as shown below:
Screenshot from 2019-11-22 14-49-13
This is causing some test failures, as this update is not reflected in the logs, resulting in a register write mismatch since the Ibex `mepc` now holds `0`, while the `mepc` in the ISS will still contain the correct value that was explicitly written at an earlier point in the simulation (`0x80000144`).
`mepc` will still get written correctly whenever the core takes a trap, it is just this extra write that occurs on `mret` that modifies its state. According to the spec, `mepc` should never be written by the RTL apart from taking traps, so how does this align with the nonstandard `mstack` CSRs that Ibex implements?
Best,
Udi

GregAC:
Fixes #492

@udinator could you run the tests you saw failing in #492 again on this?

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -50,6 +50,7 @@ module ibex_cs_registers #(
     input  logic                 irq_external_i,
     input  logic [14:0]          irq_fast_i,
     output logic                 irq_pending_o,          // interupt request pending
+    input  logic                 nmi_mode_i,
     output logic                 csr_msip_o,             // software interrupt pending
     output logic                 csr_mtip_o,             // timer interrupt pending
     output logic                 csr_meip_o,             // external interrupt pending
@@ -580,13 +581,19 @@ module ibex_cs_registers #(
       csr_restore_mret_i: begin // MRET
         priv_lvl_d     = mstatus_q.mpp;
         mstatus_d.mie  = mstatus_q.mpie; // re-enable interrupts
-        // restore previous status for recoverable NMI
-        mstatus_d.mpie = mstack_q.mpie;
-        mstatus_d.mpp  = mstack_q.mpp;
-        mepc_d         = mstack_epc_q;
-        mcause_d       = mstack_cause_q;
-        mstack_d.mpie  = 1'b1;
-        mstack_d.mpp   = PRIV_LVL_U;
+
+        if (nmi_mode_i) begin
+          // when returning from an NMI restore state from mstack CSR
+          mstatus_d.mpie = mstack_q.mpie;
+          mstatus_d.mpp  = mstack_q.mpp;
+          mepc_d         = mstack_epc_q;
+          mcause_d       = mstack_cause_q;
+        end else begin
+          // otherwise just set mstatus.MPIE/MPP
+          // See RISC-V Privileged Specification, version 1.11, Section 3.1.6.1
+          mstatus_d.mpie = 1'b1;
+          mstatus_d.mpp  = PRIV_LVL_U;
+        end
       end // csr_restore_mret_i
 
       default:;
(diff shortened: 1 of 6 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -580,13 +581,14 @@ module ibex_cs_registers #(
       csr_restore_mret_i: begin // MRET
         priv_lvl_d     = mstatus_q.mpp;
         mstatus_d.mie  = mstatus_q.mpie; // re-enable interrupts
-        // restore previous status for recoverable NMI
-        mstatus_d.mpie = mstack_q.mpie;
-        mstatus_d.mpp  = mstack_q.mpp;
-        mepc_d         = mstack_epc_q;
-        mcause_d       = mstack_cause_q;
-        mstack_d.mpie  = 1'b1;
-        mstack_d.mpp   = PRIV_LVL_U;
```
GregAC: I didn't think there was a need to set mstack.mpie/mpp to these default values so I removed those lines. Happy to add it back in if there's a reason we need them.

vogelpi: I think it is correct to not change `mstack`.MPIE and `mstack`.MPP when `mstatus` is only restored from these conditional on `nmi_mode_i`. Just for reference, we had a discussion around this before here: https://github.com/lowRISC/ibex/pull/343#discussion_r330092789

rtl/ibex_cs_registers.sv
```diff
@@ -580,13 +581,14 @@ module ibex_cs_registers #(
...
+
+        if (nmi_mode_i) begin
+          // when returning from an NMI restore state from mstack CSR
+          mstatus_d.mpie = mstack_q.mpie;
+          mstatus_d.mpp  = mstack_q.mpp;
+          mepc_d         = mstack_epc_q;
+          mcause_d       = mstack_cause_q;
+        end
```
vogelpi: What is now missing, is driving `mstatus`.MPIE to 1 and `mstatus`.MPP to U Mode if we are taking an MRET without being in NMI mode. This is actually mandated by the spec and also the reason why we changed `mstack`MPIE/MPP when taking an MRET before (see previous comment).

vogelpi: This whole section could also be changed to:
```
if (nmi_mode_i) begin
  // when returning from an NMI restore state from mstack CSR
  mstatus_d.mpie = mstack_q.mpie;
  mstatus_d.mpp  = mstack_q.mpp;
  mepc_d         = mstack_epc_q;
  mcause_d       = mstack_cause_q;
end else begin
  // otherwise, just set mstatus.MPIE/MPP
  // See RISC-V Privileged Specification, version 1.11, Section 3.1.6.1
  mstatus_d.mpie = 1'b1;
  mstatus_d.mpp  = PRIV_LVL_U;
end
```
or alternatively
```
// restore previous state for recoverable NMI
mstatus_d.mpie = mstack_q.mpie;
mstatus_d.mpp  = mstack_q.mpp;
mstack_d.mpie  = 1'b1;
mstack_d.mpp   = PRIV_LVL_U;
if (nmi_mode_i) begin
  // restore previous state only when returning from NMI mode
  mepc_d       = mstack_epc_q;
  mcause_d     = mstack_cause_q;
end
```
Actually, the only thing that was problematic in the previous implementation was writing `mepc` when executing MRET.

GregAC: Ah hadn't realised that, thanks for pointing it out.

I went for the first option, I prefer the explicit behaviour on mret rather than relying on what's in mstack.

udinator:
@GregAC - these changes seem okay to me! As intended, `mepc` now does not update on `mret` instructions, just when taking traps.

GregAC:
Thanks for the testing @udinator

PR #528, opened 2019-12-17
[rtl] Replace X assignments

vogelpi:
This PR:
- Replaces all X assignments in the RTL with defined values. 
- Imports a new file from OpenTitan containing macros for SystemVerilog assertions (SVAs).
- Rewrites existing SVAs using these macros.
- Adds SVAs to catch invalid signal values in simulation.

### DIFF
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -7,9 +7,12 @@
  * Compressed instruction decoder
  *
  * Decodes RISC-V compressed instructions into their RV32 equivalent.
- * This module is fully combinatorial.
+ * This module is fully combinatorial, clock and reset are used for
+ * assertions only.
  */
 module ibex_compressed_decoder (
+    input  logic        clk_i,
+    input  logic        rst_ni,
     input  logic [31:0] instr_i,
     output logic [31:0] instr_o,
     output logic        is_compressed_o,
@@ -23,7 +26,8 @@ module ibex_compressed_decoder (
 
   always_comb begin
     illegal_instr_o = 1'b0;
-    instr_o         = 'X;
+    // Default instruction is NOP (ADDI x0, x0, 0)
+    instr_o         = {12'd0, 5'd0, 3'd0, 5'd0, OPCODE_OP_IMM};
 
     unique case (instr_i[1:0])
       // C0
@@ -152,13 +156,13 @@ module ibex_compressed_decoder (
                   end
 
                   default: begin
-                    illegal_instr_o = 1'bX;
+                    illegal_instr_o = 1'b1;
                   end
                 endcase
               end
 
               default: begin
-                illegal_instr_o = 1'bX;
+                illegal_instr_o = 1'b1;
               end
             endcase
           end
@@ -172,7 +176,7 @@ module ibex_compressed_decoder (
           end
 
           default: begin
-            illegal_instr_o = 1'bX;
+            illegal_instr_o = 1'b1;
           end
         endcase
       end
@@ -240,7 +244,7 @@ module ibex_compressed_decoder (
           end
 
           default: begin
-            illegal_instr_o = 1'bX;
+            illegal_instr_o = 1'b1;
           end
         endcase
       end
@@ -253,5 +257,19 @@ module ibex_compressed_decoder (
   end
 
   assign is_compressed_o = (instr_i[1:0] != 2'b11);
+
+  ////////////////
+  // Assertions //
+  ////////////////
+
+  // Selectors must be known/valid.
+  `ASSERT(IbexC1Known1, (instr_i[1:0] == 2'b01) |-> !$isunknown(instr_i[15:13]), clk_i, !rst_ni)
+  `ASSERT(IbexC1Known2,
+      ((instr_i[1:0] == 2'b01) && (instr_i[15:13] == 3'b100)) |->
+      !$isunknown(instr_i[11:10]), clk_i, !rst_ni)
+  `ASSERT(IbexC1Known3,
+      ((instr_i[1:0] == 2'b01) && (instr_i[15:13] == 3'b100) && (instr_i[11:10] == 2'b11)) |->
+      !$isunknown({instr_i[12], instr_i[6:5]}), clk_i, !rst_ni)
+  `ASSERT(IbexC2Known1, (instr_i[1:0] == 2'b10) |-> !$isunknown(instr_i[15:13]), clk_i, !rst_ni)
 
 endmodule
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -25,10 +25,11 @@ module ibex_compressed_decoder (
   ////////////////////////
 
   always_comb begin
+    // By default, forward incoming instruction, mark it as legal.
+    instr_o         = instr_i;
     illegal_instr_o = 1'b0;
-    // Default instruction is NOP (ADDI x0, x0, 0)
-    instr_o         = {12'd0, 5'd0, 3'd0, 5'd0, OPCODE_OP_IMM};
 
+    // Check if incoming instruction is compressed.
     unique case (instr_i[1:0])
       // C0
       2'b00: begin
@@ -51,6 +52,14 @@ module ibex_compressed_decoder (
             instr_o = {5'b0, instr_i[5], instr_i[12], 2'b01, instr_i[4:2],
                        2'b01, instr_i[9:7], 3'b010, instr_i[11:10], instr_i[6],
                        2'b00, {OPCODE_STORE}};
+          end
+
+          3'b001,
+          3'b011,
+          3'b100,
+          3'b101,
+          3'b111: begin
+            illegal_instr_o = 1'b1;
           end
 
           default: begin
@@ -249,9 +258,11 @@ module ibex_compressed_decoder (
         endcase
       end
 
+      // Incoming instruction is not compressed.
+      2'b11:;
+
       default: begin
-        // 32 bit (or more) instruction
-        instr_o = instr_i;
+        illegal_instr_o = 1'b1;
       end
     endcase
   end
@@ -263,6 +274,8 @@ module ibex_compressed_decoder (
   ////////////////
 
   // Selectors must be known/valid.
+  `ASSERT_KNOWN(IbexInstrLSBsKnown, instr_i[1:0], clk_i, !rst_ni)
+  `ASSERT(IbexC0Known1, (instr_i[1:0] == 2'b00) |-> !$isunknown(instr_i[15:13]), clk_i, !rst_ni)
   `ASSERT(IbexC1Known1, (instr_i[1:0] == 2'b01) |-> !$isunknown(instr_i[15:13]), clk_i, !rst_ni)
   `ASSERT(IbexC1Known2,
       ((instr_i[1:0] == 2'b01) && (instr_i[15:13] == 3'b100)) |->
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -10,11 +10,17 @@
 
 /**
  * Instruction decoder
+ *
+ * This module is fully combinatorial, clock and reset are used for
+ * assertions only.
  */
 module ibex_decoder #(
     parameter bit RV32E = 0,
     parameter bit RV32M = 1
 ) (
+    input  logic                 clk_i,
+    input  logic                 rst_ni,
+
     // to/from controller
     output logic                 illegal_insn_o,        // illegal instr encountered
     output logic                 ebrk_insn_o,           // trap instr encountered
@@ -395,7 +401,7 @@ module ibex_decoder #(
           end
 
           default: begin
-            alu_operator_o = alu_op_e'({$bits(alu_op_e){1'bX}});
+            alu_operator_o = ALU_SLTU;
           end
         endcase
       end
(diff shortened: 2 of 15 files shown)


vogelpi:
Once we have found agreement, I would be happy if
- @msfschaffner could check if this PR removes the linting errors, and
- @udinator could check that no assertion unintentionally fires.

GregAC:
Thanks for your work on this @vogelpi. Looks like there's a couple of issues with building the UVM testbench on VCS which I'm currently sorting out and will push a new version once I've got something that works.

GregAC:
Found some failures, I think all related to an assert in ibex_load_store_unit.sv

```verilog
`ASSERT_KNOWN(IbexDataOffsetKnown, data_offset, clk_i, !rst_ni)
```

I'll see if I can work out what's failing. I'll hand over to @udinator if it's not straight-forward

GregAC:
I've fixed the UVM TB build issue and an X issue (compressed instruction decoder still had an explicit X assignment in it). Wasn't sure of the best way to update this PR directly so here's the diff instead:

```diff
diff --git a/dv/uvm/ibex_dv.f b/dv/uvm/ibex_dv.f
index 410d85f..55a8090 100644
--- a/dv/uvm/ibex_dv.f
+++ b/dv/uvm/ibex_dv.f
@@ -10,6 +10,7 @@ ${PRJ_DIR}/ibex/shared/rtl/prim_clock_gating.sv

 // ibex CORE RTL files
 +incdir+${PRJ_DIR}/ibex/rtl
+${PRJ_DIR}/ibex/shared/rtl/prim_assert.sv
 ${PRJ_DIR}/ibex/rtl/ibex_pkg.sv
 ${PRJ_DIR}/ibex/rtl/ibex_tracer_pkg.sv
 ${PRJ_DIR}/ibex/rtl/ibex_tracer.sv
diff --git a/rtl/ibex_compressed_decoder.sv b/rtl/ibex_compressed_decoder.sv
index 3a285ae..ba607d6 100644
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -23,7 +23,8 @@ module ibex_compressed_decoder (

   always_comb begin
     illegal_instr_o = 1'b0;
-    instr_o         = 'X;
+    // Default instruction is NOP (ADDI x0, x0, 0)
+    instr_o         = {12'd0, 5'd0, 3'd0, 5'd0, OPCODE_OP_IMM};

     unique case (instr_i[1:0])
       // C0
diff --git a/rtl/ibex_id_stage.sv b/rtl/ibex_id_stage.sv
index 215cd1e..abd80f1 100644
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -630,7 +630,7 @@ module ibex_id_stage #(
   `ASSERT_KNOWN(IbexWbStateKnown, id_wb_fsm_cs, clk_i, !rst_ni)

   // Branch decision must be valid when jumping
-  `ASSERT(IbexBranchDecisionValid, branch_in_dec |-> !isunknown(branch_decision_i), clk_i, !rst_ni)
+  `ASSERT(IbexBranchDecisionValid, branch_in_dec |-> !$isunknown(branch_decision_i), clk_i, !rst_ni)

   // Instruction delivered to ID stage can not contain X
   `ASSERT(IbexIdInstrKnown,
```

Though still getting a few failures in debug_branch_jump_test, debug_ebreakmu_test, unaligned_load_store_test, mem_error_test and reset_test. None of them is due to an assertion failure or obvious X related issue from the logs but something more subtle may be happening.

@udinator if you've got some time today could you apply my patch above to this PR and take a look? Otherwise I'll continue investigating tomorrow.

msfschaffner:
Thanks for the clean-up - let me know once you sorted out the remaining issues. I ran lint using @GregAC's patch and that is lint-clean.

udinator:
Thanks for the cleanup @vogelpi , and for the patch @GregAC! Sanity regressions after applying the patch are clean except for a failure in debug_branch_jump_test, which I think is due to some subtle error I accidentally introduced in https://github.com/google/riscv-dv/pull/336 to the instruction generator, and shouldn't be an issue with Ibex.

rtl/ibex_compressed_decoder.sv
```diff
@@ -152,13 +159,13 @@ module ibex_compressed_decoder (
                   end
 
                   default: begin
-                    illegal_instr_o = 1'bX;
+                    illegal_instr_o = 1'b1;
```
vogelpi: I am not sure what we want to do here: If this default case is taken, the compressed decoder outputs a NOP to the ID Stage but now also asserts `illegal_instr_o`. Which is a bit weird. However the controller will write the compressed instruction to `mtval`. So, I would say this is fine. WDYT @GregAC ?

vogelpi:
Thanks everybody for the nice team work! I applied Greg's patch to this PR. If someone can approve this, we can merge it.

rtl/ibex_compressed_decoder.sv
```diff
@@ -23,7 +23,8 @@ module ibex_compressed_decoder (
 
   always_comb begin
     illegal_instr_o = 1'b0;
-    instr_o         = 'X;
+    // Default instruction is NOP (ADDI x0, x0, 0)
+    instr_o         = {12'd0, 5'd0, 3'd0, 5'd0, OPCODE_OP_IMM};
```
GregAC: I did write this but now I wonder if default instruction should be an illegal instruction? This does go to the main decoder which will produce an illegal instruction exception if the decompressor outputs an illegal instruction.

GregAC: In fact thinking along similar lines maybe illegal_instr_o = 1'b1 should be the default? Unless we've specifically decoded something it's illegal.

rtl/ibex_compressed_decoder.sv
GregAC: If we do as I suggest and make the default instruction illegal the setting illegal_instr_o = 1'b1 in all of the default cases makes sense.

rtl/ibex_compressed_decoder.sv
```diff
@@ -61,6 +62,8 @@ module ibex_compressed_decoder (
       // If this check fails, an illegal instruction exception is triggered and the controller
       // writes the actual faulting instruction to mtval.
       2'b01: begin
+        // Selector must be known
+        `ASSERT_I(IbexC1Known1, !$isunknown(instr_i[15:13]))
```
GregAC: I think we're better off using clocked assertions everywhere.  This block lacks a clock but seems reasonable to add one purely for the asserts as the ASSERT_I could cause other problems (as its sensitive to glitches)

rtl/ibex_decoder.sv
```diff
@@ -369,6 +369,8 @@ module ibex_decoder #(
         imm_b_mux_sel_o     = IMM_B_I;
         regfile_we          = 1'b1;
 
+        // Selector must be known
+        `ASSERT_I(IbexAluOpKnown, !$isunknown(instr[14:12]))
```
GregAC: Again I think we should use clocked asserts, adding clocks to modules if required

rtl/ibex_compressed_decoder.sv
msfschaffner: @GregAC is right, we should favor clocked asserts if possible. Otherwise we might get spurious triggers. I can add a note on this assertion type to the style guide.

rtl/ibex_compressed_decoder.sv
msfschaffner: That sounds like a good idea!

udinator: Agreed, I think that is a good solution for this as well!

rtl/ibex_compressed_decoder.sv
msfschaffner: I removed the corresponding recommended example from the style-guide.

rtl/ibex_compressed_decoder.sv
vogelpi: Hm, I am not convinced.

IMO, the expected functionality of the compressed decoder is (independent of whether some bits in the input can be X or not):
- By default just forward the input to the output, i.e., `instr_o = instr_i` (now at Line 259) and `illegal_instr_o = 1'b0`.
- If the incoming instruction is of compressed form, decompress it. This is what in the three cases C0, C1, C2 is happening.
- If the incoming instruction is of compressed form but illegal, signal `illegal_instr_o = 1'b1` + that the instruction is compressed. (Note: it does not matter what `instr_o` in this case is. The controller will - because of `illegal_instr_o` and `is_compressed_o` - write the 16 LSBs of the input to the compressed decoder to `mtval`.)

I thus suggest:
- Move the statement `instr_o = instr_i` from Line 259 to Line 27, i.e., by default forward the input to output instead of outputting an NOP.
- Keep the `illegal_instr_o = 1'b0` as default (at Line 25).
- If the compressed decoder cannot find out what the decompressed instruction is/ends up in a default statement in C0, C1, C2 (meaning some of the selector signals are unknown/invalid), output `illegal_instr_o = 1'b1`. In these cases, `instr_o` is still equal `instr_i`, i.e., also the normal decoder sees the real instruction (and not a perfectly fine NOP which might be misleading while debugging). But this does not really matter, the controller will trigger an illegal instruction exception and write the 16 LSBs of the input to the compressed decoder to `mtval`.

rtl/ibex_compressed_decoder.sv
vogelpi: I see. I will switch to clocked assertions.

rtl/ibex_decoder.sv
vogelpi: I will change that.

rtl/ibex_compressed_decoder.sv
sjgitty: > If we do as I suggest and make the default instruction illegal the setting illegal_instr_o = 1'b1 in all of the default cases makes sense.

If we're thinking in a security context, we might chose a different default.
But I recommend we want for our broader "how to think about illegal
conditions in a security context" guidance which will come later, and
will affect Ibex and the coding style guide.

vogelpi:
I have pushed the following changes:
1. Replace the immediate assertion with clocked assertions as requested. For this I had to feed the clock and reset into the purely combinational decoders.
2. Rework a bit the compressed decoder to by default feed the incoming instruction to the output and declare it as legal. Most of you suggested to the opposite, i.e., to by default declare it as illegal. I am not convinced that this is the way to go as outline here: https://github.com/lowRISC/ibex/pull/528#discussion_r359485040  but maybe there is just a misunderstanding. Please let me know if you don't agree with my suggestion. I have no problem to change it back if can tell me why it is better.

rtl/ibex_compressed_decoder.sv
```diff
@@ -7,24 +7,36 @@
...
 );
   import ibex_pkg::*;
 
+  // Clock and reset used for assertions only.
+  logic unused_clk;
+  logic unused_rst_n;
+
+  assign unused_clk   = clk_i;
+  assign unused_rst_n = rst_ni;
+
```
GregAC: Were the unused clk/reset causing linting issues or similar hence this was needed to be introduced? If so an extra comment explaining this would be useful. Otherwise I think the 'This module is fully combinatorial, clock and reset are used for assertions only.' line suffices for documentation about the unused clock and reset and this can be removed

msfschaffner: @GregAC, these unused assignments can be used to prevent the linter from throwing "unconnected" warnings. But I agree that a comment as you suggest would be helpful.

msfschaffner: Ok, I piped this again through our linter. Everything seems to be fine, except for these clock/reset signals. The assignments to `unused_` do get rid of the "unread signal" warnings, but since these are clock/reset signals, new errors concerning the use of clock/reset appear now (because of the assignment to the assignments to `unused_*`). So either way, with the current lint deck, we would have to add a waiver rule. 

Hence I suggest to delete the `unused_` assignments here, I will add a waiver for these clock and reset signals.

rtl/ibex_id_stage.sv
```diff
@@ -273,7 +273,7 @@ module ibex_id_stage #(
       RF_WD_EX:  regfile_wdata = regfile_wdata_ex_i;
       RF_WD_LSU: regfile_wdata = regfile_wdata_lsu_i;
       RF_WD_CSR: regfile_wdata = csr_rdata_i;
-      default:   regfile_wdata = 'X;
+      default:   regfile_wdata = regfile_wdata_ex_i;
     endcase;
```
msfschaffner: could you remove the semicolon after `endcase` here? 

I migrated to a newer linter version which now includes a rule that finds this null statement :).

rtl/ibex_compressed_decoder.sv
vogelpi: Yes, @GregAC these `unused_` signals are a way to avoid linting issues that usually works across tools. Since in this particular case of clock and reset we anyway need a waiver (thanks @msfschaffner for pointing that out), I also removed the `unused_` signals and created waivers for Verilator.

rtl/ibex_id_stage.sv
vogelpi: Thanks for catching that @msfschaffner . I have fixed it.

vogelpi:
Thanks @GregAC for approving the PR!

rtl/ibex_compressed_decoder.sv
```diff
@@ -240,18 +253,36 @@ module ibex_compressed_decoder (
...
   end
 
   assign is_compressed_o = (instr_i[1:0] != 2'b11);
 
+  ////////////////
+  // Assertions //
+  ////////////////
+
+  // Selectors must be known/valid.
+  `ASSERT_KNOWN(IbexInstrLSBsKnown, instr_i[1:0], clk_i, !rst_ni)
```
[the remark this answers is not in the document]
vogelpi: Thanks @udinator for reporting this. I did not think that the compressed decoder would ever see invalid data from the prefetch buffer. Sorry! But you are absolutely right. We only need to check the assertions if the `instr_i` is valid. I have created a PR to fix this here: #536

PR #531, opened 2019-12-18
[DV] Fix TB read responses

udinator:
Currently, when the testbench returns a read response to the core, for either a load instruction or instruction fetch response, the testbench will perform a whole word read of the requested address, after word-alignment.
However, for byte or halfword loads, this is not ideal, since it is possible that the extraneous bytes read by the testbench outside of the requested byte/halfword are not initialized in the testbench system memory.
This PR fixes this issue by having the dmem_intf_sequence only read bytes from the system memory that correspond to high bits in the `byte_enable` signal that the core outputs on LSU transactions to ensure that only valid byte addresses are accessed.

### DIFF
--- a/dv/uvm/common/ibex_mem_intf_agent/ibex_mem_intf_slave_seq_lib.sv
+++ b/dv/uvm/common/ibex_mem_intf_agent/ibex_mem_intf_slave_seq_lib.sv
@@ -16,6 +16,7 @@ class ibex_mem_intf_slave_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
   // Used to ensure that whenever inject_error() is called, the very next transaction will inject an
   // error, and that enable_error will not be flipped back to 0 immediately
   bit                     error_synch = 1'b1;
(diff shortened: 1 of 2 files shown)


dv/uvm/common/ibex_mem_intf_agent/ibex_mem_intf_slave_seq_lib.sv
```diff
@@ -54,7 +57,16 @@ class ibex_mem_intf_slave_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
         req.data = rand_data;
       end else begin
         if(req.read_write == READ) begin : READ_block
-          req.data = m_mem.read(aligned_addr);
+          if (is_dmem_seq) begin
```
taoliug: Can this be applied to imem as well?

udinator: No it cannot, because the imem interface does not use any byte_enable signal, as it assumes full word reads will always be performed.

PR #533, opened 2019-12-20
[DV] add exclusion file to upstream

udinator:


imphil:
@taoliug do you mean check in a tar file? That's something we try to avoid, as binary files don't go well with git: you can't have diffs, they are impossible to review, etc.

taoliug:
Got it, I am OK with this policy. Just that these files are generated by tool, which are not that reviewable as well.

udinator:
Closing this PR as Incisive is now supported for simulation and coverage.

PR #536, opened 2019-12-23
[rtl] Trigger C decoder assertions only if instr valid

vogelpi:
This PR ensures that the assertions in the compressed decoder do not fire if the compressed decoder sees invalid data from the prefetch buffer.

### DIFF
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -13,12 +13,18 @@
 module ibex_compressed_decoder (
     input  logic        clk_i,
     input  logic        rst_ni,
+    input  logic        valid_i,
     input  logic [31:0] instr_i,
     output logic [31:0] instr_o,
     output logic        is_compressed_o,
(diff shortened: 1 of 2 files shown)


vogelpi:
Thanks @udinator for testing and @GregAC for reviewing.

> Presumably the source of the Xs is the uninitialized flops inside the prefetch buffer?

Good point, thanks for pointing that out. I did not notice. I think the X can come from both the uninitialized flops in the prefetch buffer (`ibex_fetch_fifo.sv`) or the bus signal directly, or a combination of the two (depending on the requested address, and FIFO fill status). But I am not sure, I couln't test it with Verilator.

GregAC:
I think it's worth understanding the source of Xs. Currently we allow them from uninitialized memories and the DV environment can drive them on the external bus but if something else is causing them it may need fixing.

We can merge this PR for now and create an issue to look into the source.

vogelpi:
I created an issue and will now merge this.

PR #539, opened 2020-01-02
[DV] Continue argument parsing

towoe:
In case one of the arguments is '-h' the parsing of the arguments is
aborted and only `PrintHelp()` of *verilator_sim_ctrl.cc* is executed.
Do not abort at this point in order to forward the arguments to the
registered extensions. This allows to execute the respective
`PrintHelp()` function and print the help message. For example
`VerilatorMemUtil` needs to parse the arguments in order to print the
help message.
The execution of the simulation is still terminated as `exit_app` is set
and then evaluated after parsing of the registered extensions.


towoe:
Yes, the splitting is not ideal, I was thinking about creating a buffer to collect all messages and then do the printing in the end, but that seemed too much trouble :)

PR #541, opened 2020-01-06
[rtl] Don't elaborate unused mhpmcounter flops

### LINKED ISSUE #473: Cadence Genus doesn't remove performance counters with MHPMCounterNum = 0
It would be nice to add some ability mechanism to remove mphp registers(mhpmcounter_q) from the design via some parameter or defines. I just have synthesised the core with and without them, and can established that mhpm register occupy about 35% of the total area, but they can be optional for some tasks.

GregAC:
Refactors performance counters so only flops that are required from the
given parameters are explicitly elaborated without relying on
optimization to remove unused flops.

Fixes #473

The previous RTL produced a circuit where the value of all mhpmcounter flops would be maintained, unused bits would be reset to 0 and written 0 but before reset they could be some other value. Whilst it was possible for synthesis tools to decide this meant they could be ignored and seen as constant 0 values it requires setting optimization to 'high' or similar in some tools (see discussion on #473). My yosys flow couldn't perform the necessary optimization at all so synthesized many unused flops. This alters things so it's explicit what gets elaborated and doesn't rather than relying on apparently complex optimizations to remove them.

I still need to try this on an OpenTitan FPGA build and make sure everything is fine but testing with simple system looks good.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -187,12 +187,16 @@ module ibex_cs_registers #(
   logic [PMP_CFG_W-1:0]        pmp_cfg_rdata   [PMP_MAX_REGIONS];
 
   // Hardware performance monitor signals
-  logic [31:0] mcountinhibit_d, mcountinhibit_q, mcountinhibit;
-  logic [31:0] mcountinhibit_force;
-  logic        mcountinhibit_we;
-  logic [63:0] mhpmcounter_mask [32];
+  logic [31:0]                 mcountinhibit;
+  // Only have mcountinhibit flops for counters that actually exist
+  logic [MHPMCounterNum+3-1:0] mcountinhibit_d, mcountinhibit_q;
+  logic                        mcountinhibit_we;
+
   logic [63:0] mhpmcounter_d [32];
-  logic [63:0] mhpmcounter_q [32];
+  // mhpmcounter flops are elaborated below providing only the precise number that is required based
+  // on MHPMCounterNum/MHPMCounterWidth. This signal connects to the Q output of these flops
+  // where they exist and is otherwise 0.
+  logic [63:0] mhpmcounter [32];
   logic [31:0] mhpmcounter_we;
   logic [31:0] mhpmcounterh_we;
   logic [31:0] mhpmcounter_incr;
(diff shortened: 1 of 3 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -887,55 +889,61 @@ module ibex_cs_registers #(
...
-      for (int i=0; i<32; i++) begin
-        mhpmcounter_q[i] <= '0;
+  // Performance monitor registers
+  // Only elaborate flops that are needed from the given MHPMCounterWidth and MHPMCounterNum
+  // parameters
+  for (genvar i = 0; i < 32; i++) begin : g_mhpmcounter
+    // First 3 counters (cycle, time, instret) must always be elaborated
+    if (i < 3 + MHPMCounterNum) begin : g_mhpmcounter_exists
+      // First 3 counters must be 64-bit the rest have parameterisable width
+      localparam i_mhpmcounter_width = i < 3 ? 64 : MHPMCounterWidth;
```
vogelpi: Should the local parameter be declared as `localparam int unsigned` to avoid possible lint errors?

rtl/ibex_cs_registers.sv
```diff
@@ -887,55 +889,61 @@ module ibex_cs_registers #(
...
+
+      logic [i_mhpmcounter_width - 1:0] mhpmcounter_q;
```
vogelpi: I would remove the spaces around the `-`.

tomroberts-lowrisc: I think this should be upper camel case to match coding guidelines, with no space

rtl/ibex_cs_registers.sv
```diff
@@ -887,55 +889,61 @@ module ibex_cs_registers #(
...
+    end
+  end
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      mcountinhibit_q <= '0;
     end else begin
-      mhpmcounter_q      <= mhpmcounter_d;
-      mcountinhibit_q    <= mcountinhibit_d;
+      mcountinhibit_q <= mcountinhibit_d;
```
tomroberts-lowrisc: Not particularly important, but I think the inhibit flops could also be trimmed based on the number of implemented counters.

GregAC:
I've switched from mhpmcounter_val to mhpmcounter as a name to match what was being done with inhibit (which I think is a better naming scheme).

Added mcountinhibit changes as suggested by Tom

GregAC:
Tested on an OT FPGA build all looks fine (coremark run gives counters that look good)

PR #542, opened 2020-01-06
[examples] Add timer example to simple system

tomroberts-lowrisc:
Not particularly useful in the current system, but gives an example of
how to handle interrupts.

### DIFF
--- a/examples/simple_system/README.md
+++ b/examples/simple_system/README.md
@@ -5,8 +5,8 @@ run stand-alone binaries. It contains:
 
 * An Ibex Core
 * A single memory for instructions and data
-* A basic peripheral to write ASCII output to a file and halt simulation from
-  software
+* A basic peripheral to write ASCII output to a file and halt simulation from software
+* A basic timer peripheral capable of generating interrupts based on the RISC-V Machine Timer Registers (see RISC-V Privileged Specification, version 1.11, Section 3.1.10)
 * A software framework to build programs for it
 
 ## Prerequisites
@@ -118,5 +118,9 @@ Pass `-gui` to use the DVE GUI.
 |---------------------|--------------------------------------------------------------------------------------------------------|
 | 0x20000             | ASCII Out, write ASCII characters here that will get output to the log file                            |
 | 0x20004             | Simulator Halt, write 1 here to halt the simulation                                                    |
+| 0x30000             | RISCV timer mtime register                                                                             |
+| 0x30004             | RISCV timer mtimeh register                                                                            |
+| 0x30008             | RISCV timer mtimecmp register                                                                          |
+| 0x3000C             | RISCV timer mtimecmph register                                                                         |
 | 0x100000 – 0x1FFFFF | 1 MB memory for instruction and data. Execution starts at 0x100080, exception handler base is 0x100000 |
 
--- a/examples/sw/simple_system/common/simple_system_common.c
+++ b/examples/sw/simple_system/common/simple_system_common.c
@@ -139,3 +139,52 @@ void simple_exc_handler(void) {
 
   while(1);
 }
+
+volatile uint64_t time_elapsed;
+uint64_t time_increment;
+
+inline static void increment_timecmp(uint64_t time_base) {
+  uint64_t current_time = timer_read();
+  current_time += time_base;
+  timecmp_update(current_time);
+}
+
+void timer_enable(uint64_t time_base) {
+  time_elapsed = 0;
+  time_increment = time_base;
+  // Set timer values
+  increment_timecmp(time_base);
+  // enable timer interrupt
+  asm volatile("csrs  mie, %0\n" : : "r"(0x80));
+  // enable global interrupt
+  asm volatile("csrs  mstatus, %0\n" : : "r"(0x8));
+}
+
+void timer_disable(void) { asm volatile("csrc  mie, %0\n" : : "r"(0x80)); }
+
+uint64_t timer_read(void) {
+  uint32_t current_timeh;
+  uint32_t current_time;
+  // check if time overflowed while reading and try again
+  do {
+    current_timeh = DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0);
+    current_time = DEV_READ(TIMER_BASE + TIMER_MTIME, 0);
+  } while (current_timeh != DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0));
+  uint64_t final_time = ((uint64_t)current_timeh << 32) | current_time;
+  return final_time;
+}
+
+void timecmp_update(uint64_t new_time) {
+  DEV_WRITE(TIMER_BASE + TIMER_MTIMECMP, -1);
+  DEV_WRITE(TIMER_BASE + TIMER_MTIMECMPH, new_time >> 32);
+  DEV_WRITE(TIMER_BASE + TIMER_MTIMECMP, new_time);
+}
+
+uint64_t get_elapsed_time(void) { return time_elapsed; }
+
+void simple_timer_handler(void) __attribute__((interrupt));
+
+void simple_timer_handler(void) {
+  increment_timecmp(time_increment);
+  time_elapsed++;
+}
--- a/examples/sw/simple_system/hello_test/hello_test.c
+++ b/examples/sw/simple_system/hello_test/hello_test.c
@@ -17,5 +17,25 @@ int main(int argc, char **argv) {
 
   pcount_enable(0);
 
+  // Enable periodic timer interrupt
+  // (the actual timebase is a bit meaningless in simulation)
+  timer_enable(2000);
+
+  uint64_t last_elapsed_time = get_elapsed_time();
+
+  while (last_elapsed_time <= 4) {
+    uint64_t cur_time = get_elapsed_time();
+    if (cur_time != last_elapsed_time) {
+      last_elapsed_time = cur_time;
+
+      if (last_elapsed_time & 1) {
+        puts("Tick!\n");
+      } else {
+        puts("Tock!\n");
+      }
+    }
+    asm volatile("wfi");
+  }
+
   return 0;
 }
--- a/shared/rtl/timer.sv
+++ b/shared/rtl/timer.sv
@@ -1,0 +1,151 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Example memory mapped timer
+module timer #(
+  // Bus data width (must be 32)
+  parameter int unsigned DataWidth    = 32,
+  // Bus address width
+  parameter int unsigned AddressWidth = 32
+) (
+  input  logic                    clk_i,
+  input  logic                    rst_ni,
+  // Bus interface
+  input  logic                    timer_req_i,
+
+  input  logic [AddressWidth-1:0] timer_addr_i,
+  input  logic                    timer_we_i,
+  input  logic [ DataWidth/8-1:0] timer_be_i,
+  input  logic [   DataWidth-1:0] timer_wdata_i,
+  output logic                    timer_rvalid_o,
+  output logic [   DataWidth-1:0] timer_rdata_o,
+  output logic                    timer_err_o,
+  output logic                    timer_intr_o
+);
+
+  // The timers are always 64 bits
+  localparam int unsigned TW = 64;
+  // Upper bits of address are decoded into timer_req_i
+  localparam int unsigned ADDR_OFFSET = 10; // 1kB
+  // Register map
+  localparam bit [9:0] MTIME_LOW = 0;
+  localparam bit [9:0] MTIME_HIGH = 4;
+  localparam bit [9:0] MTIMECMP_LOW = 8;
+  localparam bit [9:0] MTIMECMP_HIGH = 12;
+
+  logic                 timer_we;
+  logic                 mtime_we, mtimeh_we;
+  logic                 mtimecmp_we, mtimcmph_we;
+  logic [DataWidth-1:0] mtime_wdata, mtimeh_wdata;
+  logic [DataWidth-1:0] mtimecmp_wdata, mtimecmph_wdata;
+  logic [TW-1:0]        mtime_q, mtime_d, mtime_inc;
+  logic [TW-1:0]        mtimecmp_q, mtimecmp_d;
+  logic                 interrupt_q, interrupt_d;
+  logic                 error_q, error_d;
+  logic [DataWidth-1:0] rdata_q, rdata_d;
+  logic                 rvalid_q;
+
+  // Global write enable for all registers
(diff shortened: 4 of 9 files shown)


shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (~rst_ni) begin
+      mtimecmp_q <= 'b0;
+    end else if (mtimecmp_we | mtimecmph_we) begin
+      mtimecmp_q <= mtimecmp_d;
+    end
+  end
+
+  // interrupt remains set until mtimecmp is written
+  assign interrupt_d  = ((mtime_q >= mtimecmp_q) | interrupt_q) & ~(mtimecmp_we | mtimecmph_we);
```
vogelpi: What is the purpose of `~(mtimecmp_we | mtimecmph_we)`? This will only avoid spurious interrupts in the very cycle `mtimecmp(h)` is being updated, but not between updating the two. I think you can remove this and instead adjust the software sequence.

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+      interrupt_q <= 'b0;
+    end else begin
+      interrupt_q <= interrupt_d;
+    end
+  end
+
+  assign timer_intr_o = interrupt_q;
+
+  // Unused error signal
+  assign timer_err_o = 1'b0;
```
vogelpi: You could signal an error if there is no address match...

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+  logic [TW-1:0]           mtimecmp_q, mtimecmp_d;
+  logic                    interrupt_q, interrupt_d;
+  logic [DataWidth-1:0]    rdata_q, rdata_d;
+  logic                    rvalid_q;
+
+  // Global write enable for all registers
+  assign timer_we = timer_req_i & timer_we_i;
+
+  // mtime increments every cycle
+  assign mtime_inc = mtime_q + {{63{1'b0}},1'b1};
```
vogelpi: `{{63{1'b0}},1'b1}` can simply be written as `64'b1`.

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+  logic                    timer_we;
+  logic                    mtime_we, mtimeh_we;
+  logic                    mtimecmp_we, mtimcmph_we;
+  logic [DataWidth-1:0]    mtime_wdata, mtimeh_wdata;
+  logic [DataWidth-1:0]    mtimecmp_wdata, mtimecmph_wdata;
+  logic [TW-1:0]           mtime_q, mtime_d, mtime_inc;
+  logic [TW-1:0]           mtimecmp_q, mtimecmp_d;
+  logic                    interrupt_q, interrupt_d;
+  logic [DataWidth-1:0]    rdata_q, rdata_d;
+  logic                    rvalid_q;
```
vogelpi: Some whitespaces can be removed here.

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+                       (mtime_we     ? mtime_wdata     : mtime_inc[31:0])};
+  assign mtimecmp_d = {(mtimecmph_we ? mtimecmph_wdata : mtimecmp_q[63:32]),
+                       (mtimecmp_we  ? mtimecmp_wdata  : mtimecmp_q[31:0])};
+
+  // Generate registers
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (~rst_ni) begin
+      mtime_q    <= 'b0;
+    end else begin
+      mtime_q    <= mtime_d;
```
vogelpi: Some whitespaces can be removed here.

examples/sw/simple_system/common/simple_system_common.c
```diff
@@ -139,3 +139,37 @@ void simple_exc_handler(void) {
...
   while(1);
 }
+
+volatile uint64_t time_elapsed;
+uint32_t time_increment;
+
+inline static void increment_timecmp(uint32_t time_base) {
+  uint64_t current_time = (uint64_t)DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0);
+  current_time <<= 32;
+  current_time |= (uint64_t)DEV_READ(TIMER_BASE + TIMER_MTIME, 0);
```
GregAC: Similar to @vogelpi's comment the spec recommends when reading you read timeh, then time, then timeh again, checking if timeh has changed. If so you had a time -> timeh overflow so should try the whole sequence again until you succeed (unlike to have more than one retry unless there's a very slow set of interrupts in the middle!)

examples/simple_system/README.md
```diff
@@ -6,6 +6,7 @@ run stand-alone binaries. It contains:
 * An Ibex Core
 * A single memory for instructions and data
 * A basic peripheral to write ASCII output to a file and halt simulation from
+* A basic timer peripheral capable of generating interrupts
   software
```
GregAC: worth adding something that says this follows the behaviour specified in the 'Machine Timer Registers' in the RISCV privileged spec

examples/sw/simple_system/hello_test/hello_test.c
```diff
@@ -17,5 +20,9 @@ int main(int argc, char **argv) {
 
   pcount_enable(0);
 
+  // Wait for 2 time periods
+  while (get_elapsed_time() != 2)
+    ;
+
```
GregAC: Just to give some more output perhaps do a loop where it outputs something periodically for a few ticks?

```c

timer_enable(2000);

uint64_t last_elapsed_time = get_elapsed_time();

while (last_elapsed != 4) {
    uint64_t cur_time = get_elapsed_time();
    if(cur_time != last_elapsed_time) {
        last_elapsed_time = cur_time;

        if(last_elapsed_time & 1) {
            puts("Tick!\n");
        } else {
            puts("Tock\n");
        }
        
    }

    asm volatile ("wfi");
}
```

Suspect you'll need a reasonably long timer to ensure you see every tick as the puts will need some time to execute (and you'll want to move the timer enable directly above my proposed loop to avoid missing a tick or two).

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+                                                       mtimecmp_q[(b*8)+:8];
+    assign mtimecmph_wdata[(b*8)+:8] = timer_be_i[b] ? timer_wdata_i[b*8+:8] :
+                                                       mtimecmp_q[DataWidth+(b*8)+:8];
+  end
+
+  // Generate write enables
+  assign mtime_we     = timer_we & (timer_addr_i == (TimerBase + MTIME_LOW));
+  assign mtimeh_we    = timer_we & (timer_addr_i == (TimerBase + MTIME_HIGH));
+  assign mtimecmp_we  = timer_we & (timer_addr_i == (TimerBase + MTIMECMP_LOW));
+  assign mtimecmph_we = timer_we & (timer_addr_i == (TimerBase + MTIMECMP_HIGH));
```
GregAC: I think it's reasonable to assume you'll only see requests for addresses within the devices range (I believe this is what the current bus does, I certainly assumed it for simulator_ctrl.sv) so you can use timer_addr_i[9:0] == MTIME_LOW etc

shared/rtl/timer.sv
```diff
@@ -0,0 +1,145 @@
...
+  assign timer_err_o = 1'b0;
+
+  // Read data
+  always_comb begin
+    rdata_d = 'b0;
+    unique case (timer_addr_i)
+      (TimerBase + MTIME_LOW):     rdata_d = mtime_q[31:0];
+      (TimerBase + MTIME_HIGH):    rdata_d = mtime_q[63:32];
+      (TimerBase + MTIMECMP_LOW):  rdata_d = mtimecmp_q[31:0];
+      (TimerBase + MTIMECMP_HIGH): rdata_d = mtimecmp_q[63:32];
```
GregAC: As my above comment on the writes I think you can drop the TimerBase here and only look at the bottom bits of the address.

examples/sw/simple_system/common/simple_system_common.c
tomroberts-lowrisc: Good point - now added as a separate function

examples/sw/simple_system/common/simple_system_common.c
```diff
@@ -139,3 +139,37 @@ void simple_exc_handler(void) {
...
+  current_time += (uint64_t)time_base;
+  DEV_WRITE(TIMER_BASE + TIMER_MTIMECMPH, current_time >> 32);
+  DEV_WRITE(TIMER_BASE + TIMER_MTIMECMP, current_time);
```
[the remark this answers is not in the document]
tomroberts-lowrisc: I was originally trying to keep the handler as small as possible and so wanted to avoid the extra instructions (the function is only used while an interrupt is already pending or is disabled so no spurious interrupt), but you're right it makes sense to do what the spec says, and it's only two extra instructions!

examples/sw/simple_system/common/simple_system_common.c
```diff
@@ -139,3 +139,37 @@ void simple_exc_handler(void) {
...
+}
+
+void timer_enable(uint32_t time_base) {
```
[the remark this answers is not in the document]
tomroberts-lowrisc: Fixed

examples/sw/simple_system/hello_test/hello_test.c
tomroberts-lowrisc: Added thanks - that's a better example (and it gives an example of wfi behavior too).

shared/rtl/timer.sv
tomroberts-lowrisc: done

shared/rtl/timer.sv
tomroberts-lowrisc: done

shared/rtl/timer.sv
tomroberts-lowrisc: Done

shared/rtl/timer.sv
tomroberts-lowrisc: Done

shared/rtl/timer.sv
tomroberts-lowrisc: The spec says it's cleared on any write to `mtimecmp` which technically includes a write to the upper half only. It would be odd for software to not write both halves though so I guess it can be removed.

shared/rtl/timer.sv
tomroberts-lowrisc: Now done

shared/rtl/timer.sv
tomroberts-lowrisc: Done

examples/simple_system/README.md
tomroberts-lowrisc: Done

shared/rtl/timer.sv
vogelpi: Thanks Tom for implementing the signaling of errors. I think the current implementation is slightly problematic as by default (unless the address matches), the module always signals an error. Maybe `error_d` should also depend on `timer_req_i`?

shared/rtl/timer.sv
vogelpi: Ah I see, and you also put a comment there. I totally missed that, sorry. In this case I would leave it at your original suggestion.

tomroberts-lowrisc: Done

shared/rtl/timer.sv
tomroberts-lowrisc: The error is only valid when rvalid is set (like the data). I've added a comment (and moved it inline with the data so it isn't flopped every cycle).

examples/simple_system/README.md
```diff
@@ -6,6 +6,7 @@ run stand-alone binaries. It contains:
 * An Ibex Core
 * A single memory for instructions and data
 * A basic peripheral to write ASCII output to a file and halt simulation from
+* A basic timer peripheral capable of generating interrupts based on the RISC-V Machine Timer Register (see RISC-V Privileged Specification, version 1.11, Section 3.1.10)
   software
```
GregAC: I think this stray software belongs with the bullet point on line 8?

examples/sw/simple_system/common/simple_system_common.c
```diff
@@ -139,3 +139,52 @@ void simple_exc_handler(void) {
...
+void timer_disable(void) { asm volatile("csrc  mie, %0\n" : : "r"(0x80)); }
+
+uint64_t timer_read(void) {
+  uint32_t current_timeh = DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0);
+  uint32_t current_time = DEV_READ(TIMER_BASE + TIMER_MTIME, 0);
+  // check if time overflowed while reading and try again
+  while (current_timeh != DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0)) {
+    current_timeh = DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0);
+    current_time = DEV_READ(TIMER_BASE + TIMER_MTIME, 0);
+  }
```
GregAC: Could avoid repetition with a do while loop

```c
  uint32_t current_timeh;
  uint32_t current_time;

  do {
    current_timeh = DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0);
    current_time = DEV_READ(TIMER_BASE + TIMER_MTIME, 0);
  } while (current_timeh != DEV_READ(TIMER_BASE + TIMER_MTIMEH, 0));

File: shared/rtl/timer.sv (hunk 1 above)
Comment (vogelpi) [reply shared/rtl/timer.sv]: Thx

File: shared/rtl/timer.sv (hunk 2 above)
Comment (vogelpi) [reply shared/rtl/timer.sv]: Thx

File: examples/sw/simple_system/common/simple_system_common.c (hunk 11 above)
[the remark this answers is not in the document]
Comment (vogelpi) [reply examples/sw/simple_system/common/simple_system_common.c]: Thanks for changing this.

File: examples/simple_system/README.md [hunk 15]
```diff
@@ -5,8 +5,8 @@ run stand-alone binaries. It contains:
 
 * An Ibex Core
 * A single memory for instructions and data
-* A basic peripheral to write ASCII output to a file and halt simulation from
-  software
+* A basic peripheral to write ASCII output to a file and halt simulation from software
+* A basic timer peripheral capable of generating interrupts based on the RISC-V Machine Timer Register (see RISC-V Privileged Specification, version 1.11, Section 3.1.10)
```
vogelpi: Nit: "Timer Register" -> "Timer Registers"

tomroberts-lowrisc: Good spot - now fixed

examples/sw/simple_system/common/simple_system_common.c
tomroberts-lowrisc: Done

examples/simple_system/README.md
tomroberts-lowrisc: Fixed

Issue #502, opened 2019-11-29
syn: Integrate with fusesoc


imphil:
Instead of having a shell script, the yosys synthesis should be called as normal fusesoc target.

rswarbrick:
This is duplicated as #1006. Since that issue has rather more description, let's close this one and track with that.

PR #543, opened 2020-01-07
[syn] Synth flow improvements

GregAC:
Adds significant functionality to synthesis flow. Timing reports are
generated using OpenSTA and an area report is generated. Flow supports
open Nangate45 library from OpenROAD.

Whilst there's plenty of room for improvement this works well enough for now. I'd like to avoid spending  much more time on this right now so if there's any major changes wanted (unless they're fundamental for the flow to work properly) I'd prefer to let this through then create a issue to make said changes.

Replacing the shell script with fusesoc certainly needs doing but we've already got an issue open for that (https://github.com/lowRISC/ibex/issues/502)


syn/ibex_core_lr_synth_conf.tcl
```diff
@@ -0,0 +1,51 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# List of inputs and outputs.  Number is timing constraint expressed as a % of
+# clock cycle, e.g.
+#
+# {instr_req_o 60.0}
+#
+# as an output means the intstr_req_o output must be stable by 60% of total
```
imphil:
```suggestion
# as an output means the instr_req_o output must be stable by 60% of total
```

syn/ibex_core_lr_synth_conf.tcl
```diff
@@ -0,0 +1,51 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# List of inputs and outputs.  Number is timing constraint expressed as a % of
```
imphil: Can you comment (and add that information to the file) on where the values in this file are coming from? Educated guess as a starting point, or is there more behind it?

syn/ibex_core_lr_synth_conf.tcl
```diff
@@ -0,0 +1,51 @@
...
+                          {irq_nm_i       10.0} \
+                          {debug_req_i    10.0} \
+                          {fetch_enable_i 0.0 }]
+
+# clock and reset IO names (at top-level)
+set lr_synth_clk_input clk_i
+set lr_synth_rst_input rst_ni
+
+# clock period in ps
+set lr_synth_clk_period 3000.0
```
imphil: Is there a motivation for this clock period?

syn/python/build_translated_names.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+import argparse
+
+from flow_utils import *
+
+def main():
+    arg_parser = argparse.ArgumentParser(
+        description=
+"""Create yosys script to translate cell names from given timing reports to human readable names
+(assumes flatted synthesis run)""")
```
imphil: Can you shorten this line, and run the file through yapf with our opentitan configuration?

syn/python/build_translated_names.py
```diff
@@ -0,0 +1,36 @@
...
+
+    cells_to_translate = set()
+
+    for csv_rpt in args.rpts:
+        (new_cells_to_translate, path_info) = extract_path_info(csv_rpt)
+        cells_to_translate.update(new_cells_to_translate)
+
+    create_translate_names_script(cells_to_translate, args.top_level, args.gen_out)
+
+    os.system(f'yosys -s {args.gen_out}/{ys_translate_script_filename}'.format())
```
imphil: `os.system` is really low-level and not recommended. Can you switch to the subprocess module https://docs.python.org/3/library/subprocess.html#module-subproces, i.e. `subprocess.run()`?

syn/syn_setup.example.sh
```diff
@@ -0,0 +1,23 @@
...
+  exit 1
+fi
+
+export LR_SYNTH_TIMING_RUN=1
+export LR_SYNTH_FLATTEN=1
+
+# SETUP CELL LIBRARY PATH
+# Uncomment the lines below and set the path to an appropriate .lib file
+# export LR_SYNTH_CELL_LIBRARY_PATH=/path/to/NangateOpenCellLibrary_typical.lib
+# export LR_SYNTH_CELL_LIBRARY_NAME=nangate
```
imphil: Instead of asking people to copy and modify this file you can provide an extension point: use this script to set the variables to some defaults, and then source another script (if it exists) with user-defined values.

syn/syn_yosys.sh
```diff
@@ -4,50 +4,56 @@
...
+#-------------------------------------------------------------------------
+source syn_setup.sh
 
 #-------------------------------------------------------------------------
 # use sv2v to convert all SystemVerilog files to Verilog
 #-------------------------------------------------------------------------
-rm -Rf syn_out
-mkdir syn_out
+
+mkdir -p $LR_SYNTH_OUT_DIR/generated
```
imphil: Add quotes to the argument.

syn/syn_yosys.sh
```diff
@@ -4,50 +4,56 @@
...
-# Usage:
-#   syn_yosys.sh 2>&1 | tee syn.std
-#
-# Above command generates two files:
-#   ibex_core_premap.v  : premap netlist (before mapping it onto std. cells)
-#   ibex_core_netlist.v : final netlist
-
-# TODO: below should be replaced by elegant fusesoc commands
+# This script drives the experimental Ibex synthesis flow. More details can be
+# found in README.md
```
imphil: Did I miss this README file?

GregAC:
>     * Are the prim_assert changes to synchronize with OpenTitan, or are they different? (I'd like to keep the two a copy of each other if possible.)

Yes they're meant to be the same at the OT changes, however I messed them up (hence the failing checks) and now realise they're not required so I'll leave that file as is.

>     * You use f-strings in Python code, which are a Python 3.6+ feature. We never specified the Python requirements for Ibex, but it makes sense to keep them in line with OpenTitan, which still supports Python 3.5. So can you please use one of the older string formatting options?

I had put Python > 3.6 in the README.md (which I had failed to add to the PR!) but I agree we should match OT requirements, formatting strings removed (would have needed work anyway as I somehow ended up with a mix of f strings and format() and one particularly stupid case where I passed an f-string to a format() call...).

syn/syn_setup.example.sh
GregAC: The issue is there's no defaults that can work. A standard cell library with timing information is required (so the cmos_cells.lib from yosys won't work). We could include the nangate library in our repository but I wasn't certain about the licensing issues around that so thought it was better to ask people to acquire it themselves. We can discuss offline.

syn/syn_yosys.sh
GregAC: Somehow got lost in my last minute fixups and force pushses (including lost off my local disk, luckily the reflog exists!)

syn/ibex_core_lr_synth_conf.tcl
GregAC: It's a bit faster than we can achieve with nangate45 and it's generally useful to push beyond what you can achieve to see where the limit is.

I've changed it to 4000.0 (250 Mhz) as that gets closer to meeting timing and does meet reg2reg timing as ultimately Ibex probably isn't a core where you're trying to clock it as fast as possible so better to give it an easier target.

GregAC:
Thanks for the swift review @imphil, think I've addressed everything bar the standard cell library issue.

syn/python/build_translated_names.py
imphil: yapf needs a bit of help here, manually split the line into two and re-running yapf should do the trick.

syn/python/translate_timing_csv.py
```diff
@@ -0,0 +1,41 @@
...
+import argparse
+
+from flow_utils import *
+
+
+def main():
+    arg_parser = argparse.ArgumentParser(
+        description=
+        """Translates CSV timing report to have human readable start and end points
+given yosys generated name translation file (generated with build_translated_names.py"""
```
imphil: as above: manually split into two lines, and re-run yapf should give better output.

PR #544, opened 2020-01-08
Verilator linter with -match

wallento:
As suggested by @imphil we are adding a wildcard match for `lint_off` rules in Verilator. This avoids adopting lines in the linter configuration file.

It is still under development in Verilator and hopefully makes one of the next two releases. Clone https://github.com/verilator/verilator/tree/issue-1649 to test it.

Upstream issue: https://github.com/verilator/verilator/issues/1649


wallento:
Feature upstream, next release (4.026) scheduled for mid February.

wallento:
It is in Verilator 4.028 now!

imphil:
This needs two things:
- A rebase to pick up recent changes to the waiver. Can you do that @wallento?
- A fix for https://github.com/lowRISC/ibex/issues/603 to ensure we inform our users that they need to update Verilator. I'll look into that.

imphil:
OK, #604 is ready with a Verilator version check. Once that's merged, this PR can be rebased on top of master, and the Verilator version can be bumped to the 4.028. This should make sure users are aware that they need to update.

lint/verilator_waiver.vlt
```diff
@@ -10,55 +10,55 @@
...
 // implementation choices for the same module.
-lint_off -msg DECLFILENAME -file "*/rtl/ibex_register_file_ff.sv"
-lint_off -msg DECLFILENAME -file "*/rtl/ibex_register_file_latch.sv"
+lint_off -rule DECLFILENAME -file "*/rtl/ibex_register_file_ff.sv"
+lint_off -rule DECLFILENAME -file "*/rtl/ibex_register_file_latch.sv"
 
 // Bits of signal are not used: boot_addr_i[7:0]
 // Boot address is 256B aligned, cleaner to pass all bits in
-lint_off -msg UNUSED -file "*/rtl/ibex_if_stage.sv" -lines 40
+lint_off -rule UNUSED -file "*/rtl/ibex_if_stage.sv" -match "'boot_addr_i'[7:0]"
```
imphil: What are those `'` for? Are there docs somewhere how the matching works? (I thought it's simple wildcard matches.)

wallento: Oddly, they are part of the output generate by Verilator. It may make sense to look into expanding it to include the dimensions, but its not on my todo list. So, nothing regex specific, but the standard output..

imphil: I'm pretty sure we copied the comment above the waiver directly from the error message, and it has no `'`. Maybe that's a recent Verilator change? Could either you @wallento or @tomroberts-lowrisc see if that's something that changed between Verilator 4.010 (roughly) and today? I would like to at least ensure that we understand what's going on, and Verilator won't change back to the original formatting in the next release.

tomroberts-lowrisc: It was commit 0f857ce423d3498ddf51f7a5955783f07a534767 (between 4.016 and 4.018)
Wilson seems to have done it deliberately so I guess it will stay that way for the foreseeable.

imphil: Thanks for checking that Tom!

imphil:
Tom or Stefan, could one of you do the rebase and also update the minimum Verilator version in our Config

imphil:
Thanks @wallento, @tomroberts-lowrisc, and @rswarbrick!

PR #547, opened 2020-01-09
[DV] update ibex log parsing to match new riscv-dv csv format

udinator:

### DIFF
--- a/dv/uvm/riscv_dv_extension/riscv_core_setting.sv
+++ b/dv/uvm/riscv_dv_extension/riscv_core_setting.sv
@@ -20,6 +20,12 @@
 // XLEN
 parameter int XLEN = 32;
 
+// Vector extension parameters - not used in Ibex
+parameter int VECTOR_EXTENSION_ENABLE = 0;
+parameter int VLEN = 512;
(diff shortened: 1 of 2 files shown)


dv/uvm/riscv_dv_extension/riscv_core_setting.sv
```diff
@@ -20,6 +20,9 @@
 // XLEN
 parameter int XLEN = 32;
 
+// VLEN - not used in Ibex
+parameter int VLEN = 512;
```
taoliug: Can you add all other vector parameters? It will avoid future break if they are used in the base class.

udinator: Done.

taoliug: Add SLEN/ELEN as well
https://github.com/google/riscv-dv/blob/master/target/rv32i/riscv_core_setting.sv#L54

udinator: Sorry, you're right I must have missed those; those have been added as well

PR #553, opened 2020-01-13
Introduce branch target ALU

GregAC:
This adds a parameter 'BranchTargetALU' to ibex_core that adds a dedicated branch target ALU for calculating branch targets, removing a stall cycle off taken branches. It also includes some timing optimisations to at least partially negate the negative timing impact of this change (based on studies using Yosys).

See the individual commit messages/comments for justification of each change.


rtl/ibex_ex_block.sv
```diff
@@ -64,7 +71,25 @@ module ibex_ex_block #(
...
   // branch handling
   assign branch_decision_o  = alu_cmp_result;
-  assign jump_target_o      = alu_adder_result_ex_o;
+
+  if (BranchTargetALU) begin : g_branch_target_alu
+    logic [32:0] bt_alu_result;
+
+    assign bt_alu_result = {{19{bt_operand_imm_i[11]}}, bt_operand_imm_i, 1'b0} + pc_id_i;
+
+    assign jump_target_o = (jt_mux_sel_i == JT_ALU) ? alu_adder_result_ex_o : bt_alu_result[31:0];
```
tomroberts-lowrisc: As just discussed, might be worth checking whether the reduced width adder + mux is smaller than just having a full width jump/branch target adder

GregAC: The full width adder is probably the way to go, even if it is a small extra bit of area it makes things simpler (jump/branch targets all from one place) and reduces muxing on the path out to instr_addr.

However I'd prefer to do it as a separate PR. I'll investigate it once I'm done with the 3rd pipeline stage work.

tomroberts-lowrisc: Sounds good to me

rtl/ibex_decoder.sv
```diff
@@ -241,6 +250,11 @@ module ibex_decoder #(
 
       OPCODE_JALR: begin  // Jump and Link Register
         jump_in_dec_o         = 1'b1;
+
+        if(BranchTargetALU) begin
```
imphil: nit: space before `(`

rtl/ibex_ex_block.sv
```diff
@@ -64,7 +71,25 @@ module ibex_ex_block #(
...
+  end else begin : g_no_branch_target_alu
+    // Unused jt_mux_sel_i/bt_operand_imm_i/pc_id_i signals causes lint errors, this avoids them
+    ibex_pkg::jt_mux_sel_e jt_mux_sel_unused;
+    logic [11:0]           bt_operand_imm_unused;
+    logic [31:0]           pc_id_unused;
+
+    assign jt_mux_sel_unused     = jt_mux_sel_i;
+    assign bt_operand_imm_unused = bt_operand_imm_i;
+    assign pc_id_unused          = pc_id_i;
```
imphil: These signals need to start with `unused_` to be picked up by our lint tools correctly.

rtl/ibex_id_stage.sv
```diff
@@ -146,7 +151,7 @@ module ibex_id_stage #(
   logic        wfi_insn_dec;
 
   logic        branch_in_dec;
-  logic        branch_set_n, branch_set_q;
+  logic        branch_set, branch_set_n;
```
imphil: The `_n` signals are leftovers from the old zero-riscy style. Our style guide uses `_n` for active low signals, can you switch to `_d` and `_q` instead?

rtl/ibex_decoder.sv
```diff
@@ -274,17 +288,28 @@ module ibex_decoder #(
...
+            // First evaluates branch condition
+            alu_op_a_mux_sel_o  = OP_A_REG_A;
+            alu_op_b_mux_sel_o  = OP_B_REG_B;
+          end else begin
+            // Then calculate jump target
+            alu_op_a_mux_sel_o  = OP_A_CURRPC;
+            alu_op_b_mux_sel_o  = OP_B_IMM;
+            imm_b_mux_sel_o     = IMM_B_B;
+            alu_operator_o      = ALU_ADD;
+            regfile_we          = 1'b0;
```
imphil: It would be good to set the three last signals in this block also in the `if` branch for symmetry?

GregAC: alu_operator_o depends upon the branch condition so you'd have to bring in the case statement to set it in the if branch and also bring it into the seperate if (BranchTargetALU) branch. So I think I'd prefer to leave things as is.

PR #558, opened 2020-01-16
FPGA example: add support for the Arty A7-35

stefanct:


towoe:
Thanks a lot for the PR Stefan.
I just had a look at the parameters and wanted to use the FuseSoC's `paramtype: cmdlinearg` and wasn't sure how this is done. While using the help option:
```
$ fusesoc --cores-root=. run --target=synth lowrisc:ibex:top_artya7_100 --help
INFO: Preparing lowrisc:ibex:fpga_xilinx_shared:0
INFO: Preparing lowrisc:prim:assert:0.1
INFO: Preparing lowrisc:ibex:ibex_core:0.1
INFO: Preparing lowrisc:ibex:top_artya7_100:0.1
usage: fusesoc run lowrisc_ibex_top_artya7_100_0.1 [-h] [--SRAM_INIT_FILE SRAM_INIT_FILE] [--part PART] [--pnr PNR]

optional arguments:
  -h, --help            show this help message and exit

Verilog defines (Compile-time global symbol):
  --SRAM_INIT_FILE SRAM_INIT_FILE
		SRAM initialization file in vmem hex format

Backend arguments:
  --part PART           FPGA part number (e.g. xc7a35tcsg324-1)
  --pnr PNR             P&R tool. Allowed values are vivado (default) and none (to just run synthesis)
```
I saw that the vivado backend already has an option to set the part.
My suggestion would now be to keep the renaming changes, but remove the second target from the `.core` file and then use an invocation like `fusesoc --cores-root=. run --target=synth --setup --build lowrisc:ibex:top_artya7_100 --part xc7a35ticsg324-1L` in the makefile.

Makefile
```diff
@@ -46,24 +46,38 @@ run-simple-system: sw-simple-hello | $(Vibex_simple_system)
...
+.PHONY: build-arty-35
+build-arty-35: sw-led
+	fusesoc --cores-root=. run --target=synth --setup --build \
+		lowrisc:ibex:top_artya7 --part xc7a35ticsg324-1L
+
+.PHONY: build-arty-100
+build-arty-100: sw-led
 	fusesoc --cores-root=. run --target=synth --setup --build \
-		lowrisc:ibex:top_artya7_100
+		lowrisc:ibex:top_artya7  --part xc7a100tcsg324-1
```
towoe: As this part is the default I would not specify it here.

Makefile
```diff
@@ -46,24 +46,38 @@ run-simple-system: sw-simple-hello | $(Vibex_simple_system)
...
+
+.PHONY: program-arty-35
+program-arty-35:
+	fusesoc --cores-root=. run --target=synth --run \
+		lowrisc:ibex:top_artya7 --part xc7a35ticsg324-1L
```
towoe: For the program (run --run) step, I think it is not required to specify `--part`. If you can confirm this, I think it is best to omit it here.

Makefile
```diff
@@ -46,24 +46,38 @@ run-simple-system: sw-simple-hello | $(Vibex_simple_system)
...
 
-.PHONY: program-arty
-program-arty:
-	fusesoc --cores-root=. pgm lowrisc:ibex:top_artya7_100
+.PHONY: program-arty-100
+program-arty-100:
+	fusesoc --cores-root=. run --target=synth --run \
+		lowrisc:ibex:top_artya7 --part xc7a100tcsg324-1L
```
towoe: "1L" might be a copy and paste error, but as with the above comment, I would just not specify it here.

examples/fpga/artya7/README.md
```diff
@@ -11,8 +11,19 @@ Please see [examples](https://ibex-core.readthedocs.io/en/latest/examples.html "
...
+  - Either a Digilent Arty A7-35 oder A7-100 board
+
 ## Build
 
+The easiest way to build and execute this example is to call the following make goals that match your hardware from the root directory.
+
+  - `build-arty-35`
+  - `program-arty-35`
+  - `build-arty-100`
+  - `program-arty-100`
```
towoe: It might be nicer to have the following:

Use the following for the Arty A7-35
````
make build-arty-35 program-arty-35
````
and for the Arty A7-100
````
make build-arty-100 program-arty-100
````

File: examples/fpga/artya7/README.md [hunk 5]
```diff
@@ -24,25 +35,34 @@ make CC=/path/to/RISC-V-compiler
...
+Run either of the following commands at the top level to build the respective hardware.
+Both variants of the Arty A7 are supported and can be selected via the `--parts` parameter.
 
 ```
-fusesoc --cores-root=. run --target=synth --setup --build lowrisc:ibex:top_artya7_100
+fusesoc --cores-root=. run --target=synth --setup --build lowrisc:ibex:top_artya7 --part xc7a35ticsg324-1L
+```
+
+```
+fusesoc --cores-root=. run --target=synth --setup --build lowrisc:ibex:top_artya7 --part xc7a100tcsg324-1L
```
towoe: "xc7a100tcsg324-1L" -> "xc7a100tcsg324-1"

examples/fpga/artya7/README.md
```diff
@@ -24,25 +35,34 @@ make CC=/path/to/RISC-V-compiler
...
-After the board is connected to the computer it can be programmed with:
+After the respective board is connected to the computer it can be programmed with the matching command:
+
+```
+fusesoc --cores-root=. run --target=synth --run lowrisc:ibex:top_artya7 --part xc7a35ticsg324-1L
+```
 
 ```
-fusesoc --cores-root=. pgm lowrisc:ibex:top_artya7_100
+fusesoc --cores-root=. run --target=synth --run lowrisc:ibex:top_artya7 --part xc7a100tcsg324-1L
```
towoe: Same as above.

Makefile
stefanct: Hm yes. Not sure why I have added that actually anymore :( I was trying lots of things till I settled on the current approach. I will remove the individual program recipes again because there is no difference between both versions. However, this is also a potential problem: if anyone got both boards and tries to build and program them both... I guess we (should) ignore that but I wanted to mention it.

towoe: Yeah, building both won't be possible, but I think we can ignore that. A solution for this might depend on FuseSoC anyway.
Building for one and then programming the other board should result in an Vivado error. So at least it will fail with an indication.

stefanct: Yes, `fusesoc` uses the information of the synthesis and supplies that to Vivado for the download (even if you specify the `--part` to presumably force it fusesoc ignores it). Vivado then checks for the ID and prints an error that they don't match (IIRC).

Makefile
stefanct: It is, but I wonder why they chose the low-power version of the smaller FPGA and the normal one for the bigger FPGA... doesn't make any sense from a technical standpoint :) Thanks for catching that!

stefanct:
Updated.
I have one further thing that we could add to this PR and that's changing the sw-led makefile to not set a path in the `CC` variable. Currently it is

    CC = /tools/riscv/bin/riscv32-unknown-elf-gcc

I'd rather just set the binary name so that it is simply picked up if `PATH` includes it as it should be in most setups and then we don't have to specify it when calling make (as mentioned in the arty example README). As of now the `make build-arty*` goals only work if the compiler is indeed at that location or one overrides it (e.g., `make build-arty-100 CC=riscv32-unknown-elf-gcc`).
However, I presume this would break someone's setup ;) Does the CI rely on that or something?

towoe:
> Updated.

Thanks!

> I have one further thing that we could add to this PR and that's changing the sw-led makefile to not set a path in the `CC` variable. Currently it is
> 
> ```
> CC = /tools/riscv/bin/riscv32-unknown-elf-gcc
> ```
> 
> I'd rather just set the binary name so that it is simply picked up if `PATH` includes it as it should be in most setups and then we don't have to specify it when calling make (as mentioned in the arty example README). As of now the `make build-arty*` goals only work if the compiler is indeed at that location or one overrides it (e.g., `make build-arty-100 CC=riscv32-unknown-elf-gcc`).
> However, I presume this would break someone's setup ;) Does the CI rely on that or something?

We can absolutely change that! Overriding like this is not very nice.
If its only this small change, just add another commit. Normally it is nicer to separate different changes also in different PRs.

stefanct:
> We can absolutely change that! Overriding like this is not very nice.
> If its only this small change, just add another commit. Normally it is nicer to separate different changes also in different PRs.

But it's so convenient to sneak in such bits ;) I have rephrased the respective paragraph in the README too to make it obvious that it is optional now.

towoe:
Thanks for the changes @stefanct!
LGTM as well. Thanks for taking a look and also reminding me @imphil.

imphil:
Thanks @towoe for the review, and thanks @stefanct for this contribution!

Issue #522, opened 2019-12-16
FPGA Optimizations


vogelpi:
This issue for tracking different work items for optimizing Ibex for FPGA deployment. @ganoam will join us in January and as part of his Master thesis look into this. Things to investigate:
- [x] Register file: Leverage RAM32M primitives or similar instead of FFs/latches.
- [x] ALU: Add a new single-cycle multiplier that makes better use of the available DSP slices. (https://github.com/lowRISC/ibex/issues/527)
- [x] CSRs: Investigate the effect of the counter width of performance monitors on resource utilization and timing (64-bit counters are known to map badly to FPGA). (https://github.com/lowRISC/ibex/issues/601), PR to infer DSP slices for performance monitor counters: #624 
- Further ideas:
  - [x] Using DSP slices to implement the bit shifting inside the ALU: #625 
    Initial explorations have shown only very little or no resource savings at all (depending on tool settings) can be gained. This is not going to be investigated further.

GregAC:
In addition to looking at Xilinx FPGAs let's investigate fitting an Ibex into the ICE40 family (http://www.latticesemi.com/iCE40). They're popular for open hardware FPGA boards and an open source tool chain exists for them.

I tried a brief experiment targeting the LP8K and it just about fit in terms of LUTs but I terminated it before it finished place & route (I suspect it will fail to fit in the device because of this). I can't remember whether I used the E register file or the full register file and whether I hacked out any CSRs. I'll see if I can find it and I'll report back more details.

Bonus points for getting it working with the IceStorm (http://www.clifford.at/icestorm/) flow.

asb:
CC @wallento who I think started looking at Ibex+ice40?

This is great as a tracking issue (thanks Pirmin!), but we should move any further discussion of the more concrete steps to linked issues (at least the reg file and MUL).

wallento:
Hi,
yes, I have been starting to port it to the Fomu during the summit, but it did not fit so far. Happy to share what I have done as a patch to the Fomu workshop repository, which is a great starting point. Will do so asap.
Cheers

ganoam:
Hi everyone,
Thank you @vogelpi for the introduction. I am looking forward to working with you.
see you soon, 
Noam

ganoam:
Hi everyone,
I have implemented a first version of the new reg file. See linked Pull request. This is my first PR, so if I did anything wrong, please let me know.


[...]

PR #559, opened 2020-01-17
[rtl] Add Single Cycle Multiplier for FPGA

ganoam:
This commit adds a single cycle multiplier implementation designed to be
synthesized into DSP block primitives.

For the artya7-100 example the multiplier is implemented by 4 DSP48E1
slices. Savings in terms of logic LUTs is minimal, however faster
multiplication capability will result in higher IPC.

### DIFF
--- a/doc/instruction_decode_execute.rst
+++ b/doc/instruction_decode_execute.rst
@@ -71,19 +71,33 @@ Multiplier/Divider Block (MULT/DIV)
 Source Files: :file:`rtl/ibex_multdiv_slow.sv` :file:`rtl/ibex_multdiv_fast.sv`
 
 The Multiplier/Divider (MULT/DIV) is a state machine driven block to perform multiplication and division.
-The fast and slow versions differ in multiplier only, both implement the same form of long division algorithm.
-The ALU block is used by the long division algorithm in both the fast and slow blocks.
+The fast and slow versions differ in multiplier only. All versions implement the same form of long division algorithm. The ALU block is used by the long division algorithm in all versions.
 
-Fast Multiplier
-  - Completes multiply in 3-4 cycles using a MAC (multiply accumulate) which is capable of a 17-bit x 17-bit multiplication with a 34-bit accumulator.
-  - A MUL instruction takes 3 cycles, MULH takes 4.
-  - This MAC is internal to the mult/div block (no external ALU use).
-  - Beware it is simply implemented with the ``*`` and ``+`` operators so results heavily depend upon the synthesis tool used.
-  - In some cases it may be desirable to replace this with a specific implementation (such as a hard macro in an FPGA or an explicit gate level implementation).
+Multiplier
+  The multiplier can be implemented in three variants controlled via the parameter ``MultiplierImplementation``.
 
-Slow Multiplier
-  - Completes multiply in clog2(``op_b``) + 1 cycles (for MUL) or 33 cycles (for MULH) using a Baugh-Wooley multiplier.
-  - The ALU block is used to compute additions.
+  Single-Cycle Multiplier
+    This implementation is chosen by setting the ``MultiplierImplementation`` parameter to "single-cycle". The single-cycle multiplier makes use of three parallel multiplier units, designed to be mapped to hardware multiplier primitives on FPGAs. It is therefore the **first choice for FPGA synthesis**.
+
+    - Using three parallel 17-bit x 17-bit multiplication units and a 34-bit accumulator, it completes a MUL instruction in 1 cycle. MULH is completed in 2 cycles.
+    - This MAC is internal to the mult/div block (no external ALU use).
+    - Beware it is simply implemented with the ``*`` and ``+`` operators so results heavily depend upon the synthesis tool used.
+    - ASIC synthesis has not yet been tested but is expected to consume 3-4x the area of the fast multiplier for ASIC.
+
+  Fast Multi-Cycle Multiplier
+    This implementation is chosen by setting the ``MultiplierImplementation`` parameter to "fast". The fast multi-cycle multiplier provides a reasonable trade-off between area and performance. It is the **first choice for ASIC synthesis**.
+
+    - Completes multiply in 3-4 cycles using a MAC (multiply accumulate) which is capable of a 17-bit x 17-bit multiplication with a 34-bit accumulator.
+    - A MUL instruction takes 3 cycles, MULH takes 4.
+    - This MAC is internal to the mult/div block (no external ALU use).
+    - Beware it is simply implemented with the ``*`` and ``+`` operators so results heavily depend upon the synthesis tool used.
+    - In some cases it may be desirable to replace this with a specific implementation such as an explicit gate level implementation.
+
+  Slow Multi-Cycle Multiplier
+    To select the slow multi-cycle multiplier, set the ``MultiplierImplementation`` parameter to "slow".
+
+    - Completes multiply in clog2(``op_b``) + 1 cycles (for MUL) or 33 cycles (for MULH) using a Baugh-Wooley multiplier.
+    - The ALU block is used to compute additions.
 
 Divider
   Both the fast and slow blocks use the same long division algorithm, it takes 37 cycles to compute (though only requires 2 cycles when there is a divide by 0) and proceeds as follows:
--- a/doc/pipeline_details.rst
+++ b/doc/pipeline_details.rst
@@ -47,9 +47,12 @@ Read the description for more information.
 |                       |                                      | takes to receive a response the longer loads and stores     |
 |                       |                                      | will stall.                                                 |
 +-----------------------+--------------------------------------+-------------------------------------------------------------+
-| Multiplication        | 2/3 (Fast Multiplier)                | Fast: 2 for MUL, 3 for MULH.                                |
-|                       |                                      | Slow: clog2(``op_b``) for MUL, 32 for MULH.                 |
-|                       | clog2(``op_b``)/32 (Slow Multiplier) | See details in :ref:`mult-div`                              |
+| Multiplication        | 0/1 (Single-Cycle Multiplier)        | 0 for MUL, 1 for MULH.                                      |
+|                       |                                      |                                                             |
+|                       | 2/3 (Fast Multi-Cycle Multiplier)    | 2 for MUL, 3 for MULH.                                      |
+|                       |                                      |                                                             |
+|                       | clog2(``op_b``)/32 (Slow Multi-Cycle | clog2(``op_b``) for MUL, 32 for MULH.                       |
+|                       | Multiplier)                          | See details in :ref:`mult-div`.                             |
 +-----------------------+--------------------------------------+-------------------------------------------------------------+
 | Division              | 1 or 37                              | 1 stall cycle if divide by 0, otherwise full long division. |
 |                       |                                      | See details in :ref:`mult-div`                              |
--- a/rtl/ibex_ex_block.sv
+++ b/rtl/ibex_ex_block.sv
@@ -9,9 +9,9 @@
  * Execution block: Hosts ALU and MUL/DIV unit
  */
 module ibex_ex_block #(
-    parameter bit    RV32M                    = 1,
-    parameter bit    BranchTargetALU          = 0,
-    parameter        MultiplierImplementation = "fast"
+    parameter bit RV32M                    = 1,
+    parameter bit BranchTargetALU          = 0,
+    parameter     MultiplierImplementation = "fast"
 ) (
     input  logic                  clk_i,
     input  logic                  rst_ni,
@@ -131,7 +131,29 @@ module ibex_ex_block #(
         .multdiv_result_o   ( multdiv_result        )
     );
   end else if (MultiplierImplementation == "fast") begin : gen_multdiv_fast
-    ibex_multdiv_fast multdiv_i (
+    ibex_multdiv_fast #(
+        .SingleCycleMultiply(0)
+    ) multdiv_i (
+        .clk_i              ( clk_i                 ),
+        .rst_ni             ( rst_ni                ),
+        .mult_en_i          ( mult_en_i             ),
+        .div_en_i           ( div_en_i              ),
+        .operator_i         ( multdiv_operator_i    ),
+        .signed_mode_i      ( multdiv_signed_mode_i ),
+        .op_a_i             ( multdiv_operand_a_i   ),
+        .op_b_i             ( multdiv_operand_b_i   ),
+        .alu_operand_a_o    ( multdiv_alu_operand_a ),
+        .alu_operand_b_o    ( multdiv_alu_operand_b ),
+        .alu_adder_ext_i    ( alu_adder_result_ext  ),
+        .alu_adder_i        ( alu_adder_result_ex_o ),
+        .equal_to_zero      ( alu_is_equal_result   ),
+        .valid_o            ( multdiv_valid         ),
+        .multdiv_result_o   ( multdiv_result        )
+    );
+  end else if (MultiplierImplementation == "single-cycle") begin: gen_multdiv_single_cycle
+    ibex_multdiv_fast #(
+        .SingleCycleMultiply(1)
+    ) multdiv_i (
         .clk_i              ( clk_i                 ),
         .rst_ni             ( rst_ni                ),
         .mult_en_i          ( mult_en_i             ),
--- a/rtl/ibex_multdiv_fast.sv
+++ b/rtl/ibex_multdiv_fast.sv
@@ -14,7 +14,9 @@
 
 `include "prim_assert.sv"
 
-module ibex_multdiv_fast (
+module ibex_multdiv_fast #(
+    parameter bit SingleCycleMultiply = 0
+  ) (
     input  logic             clk_i,
     input  logic             rst_ni,
     input  logic             mult_en_i,
@@ -36,45 +38,41 @@ module ibex_multdiv_fast (
 
   import ibex_pkg::*;
 
-  logic [ 4:0] div_counter_q, div_counter_n;
-  typedef enum logic [1:0] {
-    ALBL, ALBH, AHBL, AHBH
-  } mult_fsm_e;
-  mult_fsm_e mult_state_q, mult_state_n;
-
-  typedef enum logic [2:0] {
-    MD_IDLE, MD_ABS_A, MD_ABS_B, MD_COMP, MD_LAST, MD_CHANGE_SIGN, MD_FINISH
-  } md_fsm_e;
-  md_fsm_e md_state_q, md_state_n;
-
+  // Both multiplier variants
   logic signed [34:0] mac_res_signed;
   logic        [34:0] mac_res_ext;
-
-  logic [33:0] mac_res_q, mac_res_n, mac_res, op_remainder_n;
-  logic [15:0] mult_op_a;
-  logic [15:0] mult_op_b;
-  logic [33:0] accum;
+  logic        [33:0] accum;
   logic        sign_a, sign_b;
-  logic        div_sign_a, div_sign_b;
+  logic        mult_valid;
   logic        signed_mult;
+
+  // Shared signals (div + mult)
+  logic [33:0] mac_res_q, mac_res_d, mac_res, op_remainder_d;
+
+  // Divider signals
+  logic        div_sign_a, div_sign_b;
   logic        is_greater_equal;
   logic        div_change_sign, rem_change_sign;
   logic [31:0] one_shift;
   logic [31:0] op_denominator_q;
   logic [31:0] op_numerator_q;
   logic [31:0] op_quotient_q;
-  logic [31:0] op_denominator_n;
-  logic [31:0] op_numerator_n;
-  logic [31:0] op_quotient_n;
+  logic [31:0] op_denominator_d;
+  logic [31:0] op_numerator_d;
+  logic [31:0] op_quotient_d;
   logic [31:0] next_remainder;
   logic [32:0] next_quotient;
   logic [32:0] res_adder_h;
-  logic        mult_valid;
   logic        div_valid;
+  logic [ 4:0] div_counter_q, div_counter_d;
 
-  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_mult_state_q
+  typedef enum logic [2:0] {
+    MD_IDLE, MD_ABS_A, MD_ABS_B, MD_COMP, MD_LAST, MD_CHANGE_SIGN, MD_FINISH
+  } md_fsm_e;
+  md_fsm_e md_state_q, md_state_d;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
-      mult_state_q     <= ALBL;
       mac_res_q        <= '0;
       div_counter_q    <= '0;
       md_state_q       <= MD_IDLE;
@@ -83,23 +81,19 @@ module ibex_multdiv_fast (
       op_quotient_q    <= '0;
     end else begin
 
-      if (mult_en_i) begin
-        mult_state_q <= mult_state_n;
-      end
-
       if (div_en_i) begin
-        div_counter_q    <= div_counter_n;
-        op_denominator_q <= op_denominator_n;
-        op_numerator_q   <= op_numerator_n;
-        op_quotient_q    <= op_quotient_n;
-        md_state_q       <= md_state_n;
+        div_counter_q    <= div_counter_d;
+        op_denominator_q <= op_denominator_d;
+        op_numerator_q   <= op_numerator_d;
+        op_quotient_q    <= op_quotient_d;
+        md_state_q       <= md_state_d;
       end
 
       unique case(1'b1)
         mult_en_i:
-          mac_res_q <= mac_res_n;
+          mac_res_q <= mac_res_d;
         div_en_i:
-          mac_res_q <= op_remainder_n;
+          mac_res_q <= op_remainder_d;
         default:
           mac_res_q <= mac_res_q;
        endcase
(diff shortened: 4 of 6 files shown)


GregAC:
From a quick look I don't think there's anything particularly FPGA specific about this RTL (or anything awkward in the RTL that exists purely to ensure FPGA tools infer DSP primitives) so may be better to rename this to something more generic, ibex_multdiv_fast_sc_mul perhaps (sc_mul = single cycle multiply)? ibex_multdiv_very_fast or ibex_multdiv_faster is slightly tempting but sounds a bit silly and crucially doesn't make it obvious what's different.

GregAC:
Also how much does the RTL differ from the ibex_multdiv_fast module? If there's not a huge amount just adding a `SingleCycleMultiply` parameter to the existing ibex_multdiv_fast module and using some ifs to do the different implementations could be the best way forward.

vogelpi:
> Also how much does the RTL differ from the ibex_multdiv_fast module? If there's not a huge amount just adding a `SingleCycleMultiply` parameter to the existing ibex_multdiv_fast module and using some ifs to do the different implementations could be the best way forward.

Yes, I had the same impression. Also, it is a bit unexpected that this implementation uses 4 DSP slices.

@ganoam , I think it would be good to discuss this PR first when I am back in Zurich next week.

ganoam:
> Also how much does the RTL differ from the ibex_multdiv_fast module? If there's not a huge amount just adding a `SingleCycleMultiply` parameter to the existing ibex_multdiv_fast module and using some ifs to do the different implementations could be the best way forward.

Yes, that would definitely be possible.

> > Also how much does the RTL differ from the ibex_multdiv_fast module? If there's not a huge amount just adding a `SingleCycleMultiply` parameter to the existing ibex_multdiv_fast module and using some ifs to do the different implementations could be the best way forward.
> 
> Yes, I had the same impression. Also, it is a bit unexpected that this implementation uses 4 DSP slices.
> 
> @ganoam , I think it would be good to discuss this PR first when I am back in Zurich next week.

Yes, That might be a good idea!

davideschiavone:
Are we sure we want to do a single cycle MULH? maybe we can do a single cycle MUL, and multiple cycles MULH. That could explain the DSP slices used. Single cycle MULH may increase the critical path significantly.
Also, I would set the parameter to 0 by default rather that 1 as the main target is ASIC as far as I know. 
Best
Davide

vogelpi:
> Are we sure we want to do a single cycle MULH? maybe we can do a single cycle MUL, and multiple cycles MULH. That could explain the DSP slices used. Single cycle MULH may increase the critical path significantly.

Thanks for your valuable input @davideschiavone . You are right, it makes sense to do single-cycle MUL and multi-cycle MULH. @ganoam will further look into this.

ganoam:
Hi Guys, 
Thank you all for your feedback. 

> Also, it is a bit unexpected that this implementation uses 4 DSP slices.

The reason for this is that our DSP slices (DSP48E1 in our Arta7 example) are 18 x 25 multipliers.
So we have to partition our multiplication for example like this: 
`C = A * B = (A[15:0] * B[15:0]) + (A[31:16] * B[15:0]) + (A[15:0] * B[31:16]) + (A[31:16] * B[31:16])`
This is done automatically by the synthesis tools, so I don't know exactly how it is partitioned. but in any case, in order to cover a 32 bit multiplication completely, 4 multiplication units are necessary.

The existing 'fast' multiplier does this iteratively, using just one single DSP slice for FPGAs. That's also the reason why it takes 3 cycles for MUL and 4 for MULH: for MUL, the last operation doesn't need to be computed.

> Are we sure we want to do a single cycle MULH? maybe we can do a single cycle MUL, and multiple cycles MULH. That could explain the DSP slices used. Single cycle MULH may increase the critical path significantly.

I am not sure how to do this in a smart way. In order to cover the 32 LSBs in a single cycle, we already need to partition the multiplication into 3 parts, as mentioned above, so savings with respect to doing both in a single cycle would be quite low. The last operation in the above example could be done only for the MULH instruction and would use only one DSP slice. 

For FPGA implementations I think it would be possible to do both in one cycle. A quick synthesis experiment does result in a slightly longer critical path, but it still easlily meets the timing constraints. For ASIC Implementations it would probably be a problem though.

vogelpi:
> The last operation in the above example could be done only for the MULH instruction and would use only one DSP slice.

The idea would be to leverage one of the other three DSP slices for this fourth multiplication.

As for potential gains: When we are doing a MULH in two cycles, we also need a 32-bit register to store the output of the normal MUL in the first cycle. When doing everything in one cycle, this can be avoided. However, in the current implementation, this register is shared with the divider I think. So it's already there. I would hope to get some insight into this from your investigation.

ganoam:
Hi Guys, 
I have investigated our options and commented in issue #522 .
cheers, Noam

rtl/ibex_multdiv_fast.sv
```diff
@@ -69,12 +63,16 @@ module ibex_multdiv_fast (
...
   logic        div_valid;
+  logic [ 4:0] div_counter_q, div_counter_n;
+
+  typedef enum logic [2:0] {
+    MD_IDLE, MD_ABS_A, MD_ABS_B, MD_COMP, MD_LAST, MD_CHANGE_SIGN, MD_FINISH
+  } md_fsm_e;
+  md_fsm_e md_state_q, md_state_n;
 
-  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_mult_state_q
+  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_register_update
```
GregAC: Just drop the names for always blocks, the style guide only requires them for generate blocks

doc/pipeline_details.rst
```diff
@@ -47,8 +47,9 @@ Read the description for more information.
 |                       |                       | takes to receive a response the longer loads and stores     |
 |                       |                       | will stall.                                                 |
 +-----------------------+-----------------------+-------------------------------------------------------------+
-| Multiplication        | 2/3 (Fast Multiplier) | 2 for MUL, 3 for MULH.                                      |
+| Multiplication        | 1-3 (Fast Multiplier) | 2 for MUL, 3 for MULH (ASIC), 1 for MUL, 2 for MULH (FPGA). |
```
GregAC: Cycles for MUL/MULH depend purely on what parameters you've chosen. Up to the person using Ibex what they deem suitable for ASIC. So better to say it simply depends upon the MultiplierImplementation setting.

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+      if (!rst_ni) begin
+        mult_state_q <= MULL;
+      end else begin
+        if (mult_en_i) begin
+          mult_state_q <= mult_state_n;
+        end
+      end
+    end
+
+    `ASSERT_KNOWN(IbexMultStateKnown, mult_state_q, clk_i, !rst_ni)
```
vogelpi: I think this assertion can stay at the very bottom. It is the same for both versions of the implementation.

rtl/ibex_multdiv_fast.sv
```diff
@@ -36,27 +38,19 @@ module ibex_multdiv_fast (
...
-    ALBL, ALBH, AHBL, AHBH
-  } mult_fsm_e;
-  mult_fsm_e mult_state_q, mult_state_n;
-
-  typedef enum logic [2:0] {
-    MD_IDLE, MD_ABS_A, MD_ABS_B, MD_COMP, MD_LAST, MD_CHANGE_SIGN, MD_FINISH
-  } md_fsm_e;
-  md_fsm_e md_state_q, md_state_n;
-
+  // Both multiplier variations
```
vogelpi: variations -> variants

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+
+    assign mac_res_signed = $signed(summand1) + $signed(summand2) + $signed(summand3);
+
+    assign mac_res_ext    = $unsigned(mac_res_signed);
+    assign mac_res        = mac_res_ext[33:0];
+
+    assign sign_a = signed_mode_i[0] & op_a_i[31];
+    assign sign_b = signed_mode_i[1] & op_b_i[31];
+
+    //the first two multipliers are only used in stage 1 (MULL). We can assign them statically
```
vogelpi: "// The..." Finish the second sentence with a full stop.

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+    assign mult2_sign_b = sign_b;
+    assign mult2_op_a = op_a_i[`OP_L];
+    assign mult2_op_b = op_b_i[`OP_H];
+
+    // used in MULH
+    assign accum[17:0] = mac_res_q[33:16];
+    assign accum[33:18] = {16{signed_mult & mac_res_q[33]}};
+
+    always_comb begin : mult_fsm
+      //defautl values == MULL stage
```
vogelpi: // Default values == MULL state

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+        MULL: begin
+          //AHBL
+          mult3_sign_a = sign_a;
+          mult3_sign_b = 1'b0;
+          mult3_op_a = op_a_i[`OP_H];
+          mult3_op_b = op_b_i[`OP_L];
+
+          summand1 = {18'h0, mult1_res[`OP_H]};
+          summand2 = mult2_res;
+          summand3 = mult3_res;
```
vogelpi: There is no need for these assignments, they are in the "default" section above.

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+
+          if (operator_i == MD_OP_MULL) begin
+            mac_res_n = {2'b0, mac_res[`OP_L],mult1_res[`OP_L]};
+            mult_valid = mult_en_i;
+            mult_state_n = MULL;
```
vogelpi: I would set these above in the "default" section as well. Then you just need an `if (operator_i != MD_OP_MULL)` here.

rtl/ibex_multdiv_fast.sv
```diff
@@ -36,27 +38,19 @@ module ibex_multdiv_fast (
...
+  // Both multiplier variations
   logic signed [34:0] mac_res_signed;
   logic        [34:0] mac_res_ext;
+  logic        [33:0] accum;
+  logic        sign_a, sign_b;
+  logic        mult_valid;
+  logic        signed_mult;
 
+  // Shared signals (div + mult)
   logic [33:0] mac_res_q, mac_res_n, mac_res, op_remainder_n;
```
vogelpi: Since you quite heavily rework the entire multiplier, I would recommend to also adapt the register signals from `_n/_q` to `_d/_q` as recommended in our style guide.

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+          summand1 = '0;
+          summand2 = accum;
+          summand3 = mult3_res;
+
+          mult_state_n = MULL;
+          mult_valid = 1'b1;
+        end
+
+        default: begin
+         mult_state_n = MULL;
```
vogelpi: Nit: indentation

rtl/ibex_multdiv_fast.sv
```diff
@@ -107,18 +101,241 @@ module ibex_multdiv_fast (
...
+    assign sign_a = signed_mode_i[0] & op_a_i[31];
+    assign sign_b = signed_mode_i[1] & op_b_i[31];
+
+    //the first two multipliers are only used in stage 1 (MULL). We can assign them statically
+    //ALBL
+    assign mult1_sign_a = 1'b0;
+    assign mult1_sign_b = 1'b0;
+    assign mult1_op_a = op_a_i[`OP_L];
+    assign mult1_op_b = op_b_i[`OP_L];
+    //ALBH
```
vogelpi: Add a line break

doc/pipeline_details.rst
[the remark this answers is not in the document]
vogelpi: Actually it is 0/1 for the Single-Cycle Multiplier. We are counting the stall cycles here, i.e., for how many cycles the pipeline is blocked in addition to the first cycle.

rtl/ibex_ex_block.sv
```diff
@@ -131,7 +131,29 @@ module ibex_ex_block #(
...
+        .op_b_i             ( multdiv_operand_b_i   ),
+        .alu_operand_a_o    ( multdiv_alu_operand_a ),
+        .alu_operand_b_o    ( multdiv_alu_operand_b ),
+        .alu_adder_ext_i    ( alu_adder_result_ext  ),
+        .alu_adder_i        ( alu_adder_result_ex_o ),
+        .equal_to_zero      ( alu_is_equal_result   ),
+        .valid_o            ( multdiv_valid         ),
+        .multdiv_result_o   ( multdiv_result        )
+    );
+  end else if (MultiplierImplementation == "single_cc") begin: gen_multdiv_single_cc
```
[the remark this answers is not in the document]
ganoam: After discussing with @vogelpi, I have left this as it is for now.

rtl/ibex_multdiv_fast.sv
ganoam: done. Thank you

doc/pipeline_details.rst
ganoam: I have updated the doc.

doc/instruction_decode_execute.rst
```diff
@@ -75,11 +75,23 @@ The fast and slow versions differ in multiplier only, both implement the same fo
...
+    - This MAC is internal to the mult/div block (no external ALU use).
+    - Beware it is simply implemented with the ``*`` and ``+`` operators so results heavily depend upon the synthesis tool used.
+
+  Fast Multiplier for FPGA
+    Setting the parameter ``SingleCycleMultiply`` to 1 instantiates the fast multiplier intended for use with FPGA targets with unused hardware multiplier primitives to spare.
+
+    - ASIC synthesis has not yet been tested but is expected to consume 3-4x the area of the fast multiplier for ASIC.
+    - Using three parallel 17-bit x 17-bit multiplication units and a 34-bit accumulator, it completes a MUL instruction in 1 cycle. MULH is completed in 2 cycles.
+    - This MAC is internal to the mult/div block (no external ALU use).
+    - Beware it is simply implemented with the ``*`` and ``+`` operators so results heavily depend upon the synthesis tool used.
```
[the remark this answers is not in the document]
ganoam: I have updated the documentation. Please check it...

doc/instruction_decode_execute.rst
```diff
@@ -75,11 +75,23 @@ The fast and slow versions differ in multiplier only, both implement the same fo
 The ALU block is used by the long division algorithm in both the fast and slow blocks.
```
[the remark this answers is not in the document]
ganoam: updated.

doc/pipeline_details.rst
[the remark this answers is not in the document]
ganoam: updated.

rtl/ibex_multdiv_fast.sv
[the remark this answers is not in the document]
ganoam: done. thanks

rtl/ibex_multdiv_fast.sv
ganoam: Done. thanks

rtl/ibex_multdiv_fast.sv
ganoam: Done Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: Done. Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: Done. Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: Done. Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: Done. Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: Done. Thanks.

rtl/ibex_multdiv_fast.sv
ganoam: done

doc/instruction_decode_execute.rst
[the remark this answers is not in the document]
vogelpi: Ok, thanks.

doc/instruction_decode_execute.rst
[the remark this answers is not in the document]
vogelpi: Thanks for updating the doc. It's better now but inconsistent regarding the parameter values (integers here vs. strings in `doc/integration.rst`). Please align the documentation to the sourcecode.

doc/pipeline_details.rst
```diff
@@ -30,6 +30,7 @@ In the table below when an instruction stalls for X cycles X + 1 cycles pass bef
 Some instructions stall for a variable time, this is indicated as a range e.g. 1 - N means the instruction stalls a minimum of 1 cycle with an indeterminate maximum cycles.
 Read the description for more information.
 
+<<<<<<< HEAD
```
vogelpi: This should be removed.

doc/pipeline_details.rst
```diff
@@ -47,9 +48,12 @@ Read the description for more information.
 |                       |                                      | takes to receive a response the longer loads and stores     |
 |                       |                                      | will stall.                                                 |
 +-----------------------+--------------------------------------+-------------------------------------------------------------+
-| Multiplication        | 2/3 (Fast Multiplier)                | Fast: 2 for MUL, 3 for MULH.                                |
-|                       |                                      | Slow: clog2(``op_b``) for MUL, 32 for MULH.                 |
-|                       | clog2(``op_b``)/32 (Slow Multiplier) | See details in :ref:`mult-div`                              |
+| Multiplication        | 0/1 (Single-Cycle Multiplier         | 0 for MUL, 1 for MULH.                                      |
+|                       +--------------------------------------+-------------------------------------------------------------+
```
vogelpi: Please remove the additional horizontal lines. I just does not render nicely.

Please just leave one "empty line" between to multiplier implementations/rows.

doc/instruction_decode_execute.rst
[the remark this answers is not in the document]
ganoam: Oh man, sorry about that. Done.

rtl/ibex_multdiv_fast.sv
vogelpi: You are right. They must be inside the generated block. Thanks for adding a comment.

rtl/ibex_multdiv_fast.sv
vogelpi: nit: it's "state" not "stage"

rtl/ibex_multdiv_fast.sv
vogelpi: Thx

rtl/ibex_multdiv_fast.sv
```diff
@@ -83,42 +81,250 @@ module ibex_multdiv_fast (
...
+    // used in MULH
+    assign accum[17:0] = mac_res_q[33:16];
+    assign accum[33:18] = {16{signed_mult & mac_res_q[33]}};
+
+    always_comb begin
+      // Default values == MULL stage
+      mac_res_d = {2'b0, mac_res[`OP_L],mult1_res[`OP_L]};
+      mult_valid = mult_en_i;
+      mult_state_d = MULL;
+      //AHBL
```
vogelpi: Add newline.

rtl/ibex_multdiv_fast.sv
```diff
@@ -83,42 +81,250 @@ module ibex_multdiv_fast (
...
+    assign mult2_op_a = op_a_i[`OP_L];
+    assign mult2_op_b = op_b_i[`OP_H];
+
+    // used in MULH
+    assign accum[17:0] = mac_res_q[33:16];
+    assign accum[33:18] = {16{signed_mult & mac_res_q[33]}};
+
+    always_comb begin
+      // Default values == MULL stage
+      mac_res_d = {2'b0, mac_res[`OP_L],mult1_res[`OP_L]};
```
vogelpi: nit: space after second comma is missing

rtl/ibex_multdiv_fast.sv
vogelpi: Thanks for doing this!

rtl/ibex_multdiv_fast.sv
```diff
@@ -83,42 +81,250 @@ module ibex_multdiv_fast (
...
+        ALBH: begin
+          // al*bh<<16
+          mult_op_a = op_a_i[`OP_L];
+          mult_op_b = op_b_i[`OP_H];
+          sign_a    = 1'b0;
+          sign_b    = signed_mode_i[1] & op_b_i[31];
+          // result of AL*BL (in mac_res_q) always unsigned with no carry, so carries_q always 00
+          accum     = {18'b0,mac_res_q[31:16]};
+          if (operator_i == MD_OP_MULL) begin
+            mac_res_d = {2'b0,mac_res[`OP_L],mac_res_q[`OP_L]};
```
vogelpi: add spaces after commas

rtl/ibex_multdiv_fast.sv
```diff
@@ -83,42 +81,250 @@ module ibex_multdiv_fast (
...
+
+        AHBL: begin
+          // ah*bl<<16
+          mult_op_a = op_a_i[`OP_H];
+          mult_op_b = op_b_i[`OP_L];
+          sign_a    = signed_mode_i[0] & op_a_i[31];
+          sign_b    = 1'b0;
+          if (operator_i == MD_OP_MULL) begin
+            accum        = {18'b0,mac_res_q[31:16]};
+            mac_res_d    = {2'b0,mac_res[15:0],mac_res_q[15:0]};
```
vogelpi: spaces after commas

rtl/ibex_multdiv_fast.sv
```diff
@@ -160,175 +366,96 @@ module ibex_multdiv_fast (
...
 
       MD_CHANGE_SIGN: begin
-        md_state_n  = MD_FINISH;
+        md_state_d  = MD_FINISH;
         if (operator_i == MD_OP_DIV) begin
-          op_remainder_n = (div_change_sign) ? {2'h0,alu_adder_i} : mac_res_q;
+          op_remainder_d = (div_change_sign) ? {2'h0,alu_adder_i} : mac_res_q;
         end else begin
-          op_remainder_n = (rem_change_sign) ? {2'h0,alu_adder_i} : mac_res_q;
+          op_remainder_d = (rem_change_sign) ? {2'h0,alu_adder_i} : mac_res_q;
```
vogelpi: spaces after commas

doc/pipeline_details.rst
ganoam: Ok, done.

rtl/ibex_multdiv_fast.sv
ganoam: Done, thanks.

rtl/ibex_multdiv_fast.sv
ganoam: done, thanks.

rtl/ibex_ex_block.sv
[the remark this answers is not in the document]
vogelpi: We have had this discussion [before](https://github.com/lowRISC/ibex/pull/412#discussion_r338181250) and after some discussion decided to go for strings. I suggest to stick with this decision and not go back to using non-descriptive integers or even enums. Unless of course you have some very good arguments @GregAC .

I would go for `fast`, `slow` and `single-cycle` as these are simple, descriptive names. But I am open. I just don't want that @ganoam has to update the RTL and doc 5 times before we agree on something.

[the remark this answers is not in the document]
GregAC: It's mostly a general disliking of using strings unless strictly required (e.g. for a filename)

There is a potential issue around our use of string parameters, as we don't use an explicit string type tools that don't support them may silently take the strings then internally implicitly convert to some integer (such 0 for all strings or the ASCII code of the first character) leading to unexpected behaviour, there is some discussion of this here: https://github.com/google/verible/issues/167

In particular this could lead to a simulation/synthesis mismatch (verilator sim correctly uses string parameter, FPGA/ASIC implementation run doesn't and you end up with a different kind of multiplier without realising).

doc/pipeline_details.rst
GregAC: Looks good, thanks

rtl/ibex_ex_block.sv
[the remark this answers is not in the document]
GregAC: Anyway happy to leave things as is for this PR

rtl/ibex_multdiv_fast.sv
ganoam: done

rtl/ibex_multdiv_fast.sv
ganoam: done

rtl/ibex_multdiv_fast.sv
ganoam: done

doc/pipeline_details.rst
vogelpi: thx

doc/instruction_decode_execute.rst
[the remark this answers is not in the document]
vogelpi: Thanks for fixing this.

rtl/ibex_ex_block.sv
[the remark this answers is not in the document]
vogelpi: Thanks @GregAC for getting back and pointing to the other thread. I see that Philipp is involved there and that our multiplier selection is the example being discussed :-) 
I think it is reasonable to stick to strings for this PR. If needed we can re-discuss this in a new issue.

PR #566, opened 2020-01-24
Support for Riviera 2019 10

danielmlynek:
Initial support for Riviera 2019.10 and some issue fixed in ibex it self.
There is also riscv-dv changed to work with Riviera added.

### DIFF
--- a/dv/uvm/riscv_dv_extension/riscv_core_setting.sv
+++ b/dv/uvm/riscv_dv_extension/riscv_core_setting.sv
@@ -80,7 +80,7 @@ int kernel_program_instr_cnt = 400;
 // ----------------------------------------------------------------------------
 
 // Implemented previlieged CSR list
-parameter privileged_reg_t implemented_csr[] = {
+const privileged_reg_t implemented_csr[] = {
     // Machine mode mode CSR
     MVENDORID,        // Vendor ID
     MARCHID,          // Architecture ID
@@ -151,14 +151,12 @@ parameter privileged_reg_t implemented_csr[] = {
 // --------------------------------------------------------------------------
 // Supported interrupt/exception setting, used for functional coverage
 // --------------------------------------------------------------------------
-
-parameter interrupt_cause_t implemented_interrupt[] = {
+const interrupt_cause_t implemented_interrupt[] = {
   M_SOFTWARE_INTR,
   M_TIMER_INTR,
   M_EXTERNAL_INTR
 };
-
-parameter exception_cause_t implemented_exception[] = {
+const exception_cause_t implemented_exception[] = {
   INSTRUCTION_ACCESS_FAULT,
   ILLEGAL_INSTRUCTION,
   BREAKPOINT,
--- a/dv/uvm/tests/core_ibex_seq_lib.sv
+++ b/dv/uvm/tests/core_ibex_seq_lib.sv
@@ -121,8 +121,7 @@ endclass
 
 // Simple debug sequence
 // debug_req is just a single bit sideband signal, use the interface to drive it directly
-class debug_seq extends core_base_seq;
-
+class debug_seq extends core_base_seq#(irq_seq_item);
   virtual core_ibex_dut_probe_if dut_vif;
 
   `uvm_object_utils(debug_seq)
--- a/dv/uvm/yaml/rtl_simulation.yaml
+++ b/dv/uvm/yaml/rtl_simulation.yaml
@@ -88,3 +88,19 @@
       -covbaserun test
     wave_opts: >
       -input <cwd>/ius.tcl
+
+- tool: riviera
+  env_var: ALDEC_PATH
+  compile:
+    cmd:
+      - "vlib <out>/work"
+      - "vlog -work <out>/work
+        +incdir+<ALDEC_PATH>/vlib/uvm-1.2/src
+        -l uvm_1_2
+        -f ibex_dv.f"
+
+  sim:
+    cmd: >
+      vsim -c <sim_opts> <cov_opts> -lib <out>/work +access +r+w -l <out>/logfile.log -do "run -all; endsim; quit -force" +ibex_tracer_file_base="trace_core"
+    cov_opts: >
+      -acdb_file <out>/cov.acdb


imphil:
Thanks @danielmlynek for this PR! Adding Riviera support is a great addition. Unfortunately we need to untangle this one a bit.

* All changes in the `vendor/google_riscv-dv/` directory need to go to the https://github.com/google/riscv-dv repository. Can you open a PR there with only the changes you have in the vendor directory?
* Once that's done and merged we can copy the changes into the Ibex repository. We do that regularly with a script (like in https://github.com/lowRISC/ibex/pull/564). Please leave a comment here and we'll do that for you.
* I'm not convinced we need all the Riviera special-case defines. We'll need to go through their motivation one-by-one to see which things are tool-specific bugs/deficiencies, and where we actually are relying on non-standard behavior and should fix that for all tools instead of introducing a special-case for Riviera.

As first step, please open a PR at https://github.com/google/riscv-dv and then we'll continue here once that's done.

I'm sorry that this is slightly more involved, let me know if it gets too painful and we can help out where needed.

sjgitty:
@taoliug to ensure visibility

taoliug:
Thanks for adding this support. One general comment about riscv-dv changes, I'd prefer to limit the usage of ifdef as much as possible. We can try to find some syntax that works for all simulators.  As  Philipp mentioned, please make changes in riscv-dv repo first.

danielmlynek:
Thank you for the responses.

We are planing to have PR with riscv-dv soon, but a the moment we are not ready for that. We have put this version of riscv-dv here to have verify IBEX it self.
Lets focus on changes in IBEX and forget for now on riscv-dv. You do not develop riscv-dv on your own in github fork or in your ibex repo. I understand that, you do not want to have issues with merging, rebasing it etc....

You can run ibex with no riscv-dv, you have scripts in place to run verilator in this mode (fusesoc, edalize). Lets focus on have Riviera running in this mode now, and on preparing changes needed in IBEX for future.

I agree that the best is not to have simulator dependent ifdefs in the code and we are going in that direction. At the moment all changes was done under `ifdef _VCP so they are easy to find and manage. `_VCP is automatically defined by Riviera simulator. 
I can update the PR so ifdef _VCP are only in place where they are necessary.

Changes are in IBEX are:

- [ ] \ibex\dv\uvm\yaml\rtl_simulation.yaml - support for Riviera simulator was added. 
 \ibex\dv\uvm\riviera_sim.tcl - run simulation  command used by above yaml "run -all "

I think that this 2 can be merged, proper macros for new tool need to be added

    

- [ ] \ibex\dv\uvm\tests\core_ibex_seq_lib.sv - Riviera issue workaround added DZI370
This is our issue and will be fixed, but maybe you can think of using version from ifdef _VCP for all simulators.

- [ ]     \ibex\dv\uvm\riscv_dv_extension\riscv_core_setting.sv  issue DKO4049.
Riviera do not support parameters of dynamic/associate/queue array types. SV standard does not say explicitly that such code is legal. IMHO this code is at least controversial as dynamic nature of such array types are against static nature of parameters.  It would be better to change this "implemented_csr" from parameter to const.

- [ ]     Due to not allowed by LRM assignments to enum found in ibex code we need to suppress VCP2694 (in compilation command vlog). There should be some explicit casting done to make this SV standard compatible. Maybe you should fix to be independent on LRM extensions.

- [ ]     examples/simple_system/rtl/ibex_simple_system.sv

    There is a bug in this file in IBEX sources - variable cannot have mix of continuous and procedural assignment. Under _VCP there is a proposed fix for that issue.

- [ ]     In ibex\ibex_dzi\rtl\ibex_compressed_decoder.sv and many similar files we had to to add "`include "prim_assert.sv""  to proper compile by fusesoc scripts. 
    in ibex\ibex_dzi\shared\prim_assert.core we need to mark  that prim_assert.sv is include file: "rtl/prim_assert.sv : {is_include_file : true}"

ibex_compressed_decoder.svfile needs the specified prim_assert.sv header. You should consider to add this file here for all tools. I think that including needed files with include guards instead adding it to compilation command is basic SV coding rule. Issues here is because FUSESOC compiles all files in separate vlogs and defines from files compiled in previous vlog invocation will not be visible in those compiled later.


- [ ]     dv/uvm/ibex_dv.f

    We had to put "32'h8000_0000" into quotation as Riviera scripter do not allow ' to be placed here in define value. As often special characters are hard o understand by scripter. I think that there could be found some solution here which will work in all tools. Alternatives working in Riviera are:
    "32'h8000_0000"
    32\\'h8000_0000
    2147483648

dv/uvm/ibex_dv.f
```diff
@@ -1,8 +1,7 @@
 // Copyright lowRISC contributors.
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
-
-+define+BOOT_ADDR=32'h8000_0000
```
danielmlynek: ' character makes issues for TCL scripter, hex format changed to  to decimal

dv/uvm/tests/core_ibex_seq_lib.sv
```diff
@@ -119,8 +119,7 @@ endclass
 
 // Simple debug sequence
 // debug_req is just a single bit sideband signal, use the interface to drive it directly
-class debug_seq extends core_base_seq;
-
+class debug_seq extends core_base_seq#(irq_seq_item);
```
danielmlynek: DZI370 Riviera workaround

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -213,7 +219,7 @@ module ibex_simple_system (
       .timer_wdata_i  (device_wdata[Timer]),
       .timer_rvalid_o (device_rvalid[Timer]),
       .timer_rdata_o  (device_rdata[Timer]),
-      .timer_err_o    (device_err[Timer]),
```
danielmlynek: changes in this file added to avoid mix of continuous and procedural assignment to device_err

dv/uvm/riscv_dv_extension/riscv_core_setting.sv
```diff
@@ -151,14 +151,12 @@ parameter privileged_reg_t implemented_csr[] = {
...
-
-parameter interrupt_cause_t implemented_interrupt[] = {
+const interrupt_cause_t implemented_interrupt[] = {
   M_SOFTWARE_INTR,
   M_TIMER_INTR,
   M_EXTERNAL_INTR
 };
-
-parameter exception_cause_t implemented_exception[] = {
+const exception_cause_t implemented_exception[] = {
```
danielmlynek: const of dynamic array is more natural and widely supported that parameter of such type

shared/rtl/prim_assert.sv
```diff
@@ -172,3 +175,5 @@
 `ifdef FPV_ON                                    \
    `COVER(__name, __prop, __clk, __rst)          \
 `endif
+
+`endif //PRIM_ASSERT_SV
```
danielmlynek: include guards added

dv/uvm/yaml/rtl_simulation.yaml
```diff
@@ -88,3 +88,22 @@
...
+      - "vlog -work <out>/work
+        +incdir+<ALDEC_PATH>/vlib/uvm-1.2/src
+        +incdir+${PRJ_DIR}/ibex/shared/rtl
+        -l uvm_1_2
+        -err VCP2694 W1
+        -f ibex_dv.f"
+
+  sim:
+    cmd: >
+      vsim -c <sim_opts> <cov_opts> -lib <out>/work +access +r+w -l <out>/logfile.log -do "run -all; endsim; quit -force" +ibex_tracer_file_base="trace_core"
```
imphil: Is the plan to use `riviera_sim.tcl` here, or can the TCL file be removed?

examples/simple_system/rtl/ibex_simple_system.sv
imphil: Thanks for catching this. I would suggest a slightly different implementation, however: Instead of using a new signal `device_err_timer`, can you
* Remove the assignment from the `initial` statement.
* Explicitly `assign` all undriven signals:

   ```verilog
   // Tie-off unused error signals
   assign device_err[Ram] = 1'b0;
   assign device_err[SimCtrl] = 1'b0;
   ```

rtl/ibex_compressed_decoder.sv
```diff
@@ -10,6 +10,9 @@
  * This module is fully combinatorial, clock and reset are used for
  * assertions only.
  */
+
+ `include "prim_assert.sv"
```
imphil: nit: remove the space in front of the include here, and in the other files.

rtl/ibex_multdiv_slow.sv
```diff
@@ -8,6 +8,9 @@
  *
  * Baugh-Wooley multiplier and Long Division
  */
+
+ `include "prim_assert.sv" //we need include this file bcs FuseSoC generate separate compilation, so assert macros is in different vlog, and we can't see them here
```
imphil: you can remove the comment here.

shared/rtl/prim_assert.sv
imphil: Thanks for the include guards. Can you add a newline to the end of the file?

dv/uvm/yaml/rtl_simulation.yaml
```diff
@@ -88,3 +88,22 @@
...
+- tool: riviera
+  env_var: ALDEC_PATH
+  compile:
+    cmd:
+      - "vlib <out>/work"
+      - "vlog -work <out>/work
+        +incdir+<ALDEC_PATH>/vlib/uvm-1.2/src
+        +incdir+${PRJ_DIR}/ibex/shared/rtl
+        -l uvm_1_2
+        -err VCP2694 W1
```
imphil: Can you add a comment why this flag is needed/what it does?

danielmlynek: This argument is needed to lower severity for compiler error on illegal assignment to enum to warning.
All line where illegal assignment are done is now marked with warning ie:
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 781  1

This argument will be no longer needed when explicit cast will be used on such illegal assignment ie in file :
 /home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv, line 201
` operating_mode = core_init_mstatus[12:11];`
need to be changed to:
` operating_mode = priv_lvl_e'(core_init_mstatus[12:11]);`

dv/uvm/yaml/rtl_simulation.yaml
danielmlynek: I've removed riviera_sim.tcl

dv/uvm/yaml/rtl_simulation.yaml
danielmlynek: i've added the requested note. Maybe you can consider to fix the sources so this argument will not be neded

danielmlynek:
I've done requested by @imphil changes, but they are not structured as separate commits.
I'm not sure how it should be done  - please let me know so I know how to do it in future.
For now you can fix it

dv/uvm/yaml/rtl_simulation.yaml
imphil: It seems like our lint flow excludes DV code, I've filed https://github.com/lowRISC/ibex/issues/569 to get that sorted. Thanks for letting us know!

imphil:
Thanks a lot @danielmlynek. I pulled out the RTL changes and opened https://github.com/lowRISC/ibex/pull/570 to get them merged as quickly as possible. Once that's merged I'll rebase this PR on top of the RTL changes to keep only the DV changes, which need to be reviewed by someone else.

imphil:
I've now updated this PR on top of the RTL changes that were merged into master, leaving only the DV changes in this PR. 

@danielmlynek I've taken the liberty to shuffle the changes around into three commits with your name on them and a commit message that explains what's going on. You can see those at https://github.com/lowRISC/ibex/pull/566/commits. 

To continue working on the `pullreq_2019_10` branch on your machine, you need to `git fetch --all` first, and then do a `git reset --hard origin/pullreq_2019_10` (assuming the fork remote is called `origin` on your machine, check with `git remote -v`; if not, replace `origin` as necessary). If you only do a `git pull` you'll get into a huge mess of merge conflicts. Please let me know if you get lost anywhere and I can help you out of it.


@udinator @taoliug Can you please have a look at the changes? The discussion is reflected in the commit descriptions, if you read though them individually you should get all the necessary information. The changes in 7ff0c7726d10b309286ee189093ce1107060e814 should probably be replicated elsewhere in riscv-dv.

dv/uvm/riscv_dv_extension/riscv_core_setting.sv
udinator: This change seems fine to me. @taoliug, what about you?

dv/uvm/tests/core_ibex_seq_lib.sv
udinator: Seems fine to me as well, this shouldn't affect simulation at all.

imphil:
I've removed the error-to-warning conversion in  dv/uvm/yaml/rtl_simulation.yaml, together with the incdir setting (which is now part of the ibex_core.f file). 

@danielmlynek can you test if the current version still works on Riviera?

danielmlynek:
will work after #569 will be fixed

dv/uvm/riscv_dv_extension/riscv_core_setting.sv
taoliug: Looks good as long as it doesn't break vcs/ius simulation. @udinator, please confirm.

udinator: This modification doesn't break any simulations, I can still run tests successfully with this change applied.

imphil:
Thanks for checking @udinator, and thanks a lot for your work getting this PR ready  @danielmlynek! It's always interesting how many bugs get uncovered by running the same code base against different tools.

Is this PR all that's needed to get the Ibex verification running on Riviera, or are there more changes to https://github.com/google/riscv-dv needed?

danielmlynek:
> Is this PR all that's needed to get the Ibex verification running on Riviera
No-UVM tests based on fusesoc should be now working in Riviera.
Tests using google riscv-dv will fail. Original riscv-dv  does not compile in Riviera-PRO 2019.10.
We have internal version of riscv-dv wich runs in Riviera-PRO 2019.10 - it was a part of this PR but you have refuse it. Some riscv-dv based tests had strange results and we do not known if this is ok or not.
We are continuously working to support riscv-dv in Riviera. We can consider to create aldec fork of riscv-dv so you can check Riviera in this verification flow.

imphil:
I'm sorry, there seems to be a misunderstanding. The riscv-dv changes are very much appreciated, but this repository is the wrong one to integrate them into. All changes to riscv-dv need to be proposed as pull request in https://github.com/google/riscv-dv, where the maintainers are very happy to work with you to get your changes integrated. Please let me know if you run into any issues and I'll help you get things sorted.

Once the changes are integrated into the riscv-dv repository, we can pull them back into the ibex repository here.

Issue #569, opened 2020-01-28
Add missing casts to enums (most likely in DV code only)


imphil:
@danielmlynek discovered this in https://github.com/lowRISC/ibex/pull/566:

All line where illegal assignment are done is now marked with warning ie:
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 781  1

This argument will be no longer needed when explicit cast will be used on such illegal assignment ie in file :
 /home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv, line 201
` operating_mode = core_init_mstatus[12:11];`
need to be changed to:
` operating_mode = priv_lvl_e'(core_init_mstatus[12:11]);`

@danielmlynek do you know of more cases where such a cast is missing?

@msfschaffner I suppose currently the DV code isn't running through Lint, which should have caught that. Can you think of an easy way to lint that code as well? (Either here, or in OpenTitan.)

danielmlynek:
Riviera compiler returns this warning for below lines - so all of them need to be reviewed to see if needs explicit casting
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 201  1
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 781  1
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 88  32
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 91  32
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 94  32
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 97  32
WARNING VCP2694 "Assignment to enum variable from expression of different type." "/home/danielm/gitlab/ibex/dv/uvm/tests/core_ibex_test_lib.sv" 100  32
[...]

PR #572, opened 2020-01-28
Riviera compile warnings

udinator:
The first commit fixes a VCS compile-time issue introduced by #570 where the `prim_assert.sv` file wasn't recognized by the compiler.

The second commit addressed #569 and fixes the missing enum casts.
@danielmlynek - can you pull these changes into your Riviera environment and check the compilation again?

### DIFF
--- a/dv/uvm/ibex_dv.f
+++ b/dv/uvm/ibex_dv.f
@@ -7,6 +7,7 @@
 +define+BOOT_ADDR=2147483648 // 32'h8000_0000
 +define+TRACE_EXECUTION
 +define+RVFI
++incdir+${PRJ_DIR}/ibex/shared/rtl
 
 ${PRJ_DIR}/ibex/shared/rtl/prim_clock_gating.sv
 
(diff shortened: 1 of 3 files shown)


dv/uvm/yaml/rtl_simulation.yaml
```diff
@@ -19,6 +19,7 @@
          -l <out>/compile.log
          -sverilog -ntb_opts uvm-1.2
          +define+UVM_REGEX_NO_DPI -timescale=1ns/10ps -licqueue
+         +incdir+../../shared/rtl
```
imphil: Can we put that option onto the ibex_core.f file to make all simulators use it?

udinator: Sure, I can do that

dv/uvm/ibex_dv.f
```diff
@@ -7,6 +7,7 @@
 +define+BOOT_ADDR=2147483648 // 32'h8000_0000
 +define+TRACE_EXECUTION
 +define+RVFI
++incdir+../../shared/rtl
```
imphil: How about using the same PROJ_DIR variable here as in the next line?

udinator: ah yes, that's a good point, not sure how I missed using this the first time around. It's been modified to better match the rest of the .f file.

Issue #567, opened 2020-01-24
Reads of `mip` CSR should not depend on corresponding fields in `mie`


pfmooney:
Currently, the bits in `mip` are ANDed with those in `mie`:
https://github.com/lowRISC/ibex/blob/43752a6c193893d4750daceff5f284f22da52f77/rtl/ibex_cs_registers.sv#L241-L245
This is at odds with the specification, which states the mechanisms by which those bits would be cleared:
-  The MTIP bit is read-only and is cleared by writing to the memory-mapped machine-mode timer compare register
-  MEIP is set and cleared by a platform-specific interrupt controller

And while it does state that "An interrupt _i_ will be taken if bit _i_ is set in both `mip` and `mie`, and if interrupts are globally enabled.", it makes no mention of the asserted bit(s) in `mip` depending on their `mie` counterpart(s).

Looking for other interpretations of this issue, I found that qemu-riscv treats `mip` similarly, where `mie` used to determine if CPU execution should be interrupted, but not whether bits are asserted in `mip`.  A survey of some other open-source RISC-V cores ([e200](https://github.com/SI-RISCV/e200_opensource), [scr1](https://github.com/syntacore/scr1)) yielded the same result: reads of `mip` yielding results not ANDed through `mie`.

vogelpi:
Thanks @pfmooney for opening this issue. I think you are right, `mip` should not depend on `mie` but the two are combined to decide whether an interrupt is actually taken. This should be fixed. Hopefully I can do it this week.

I am also including @davideschiavone , I think RI5CY has the same issue.

vogelpi:
I am also adding @udinator and @taoliug as this needs to be accounted for in verification as well.

davideschiavone:
Thanks @vogelpi  - CLINT porting on RI5CY is actually happening right now so thanks a lot for the tip!

davideschiavone:
Dear all, Arjan from SiLabs actually noticed that this PR may be wrong.
@vogelpi , can you partecipated at the discussion here https://github.com/openhwgroup/cv32e40p/pull/231 ?

Thanks a lot

GregAC:
I think we are compliant with the bit of the spec Arjan highlighted.

Both the global core clock enable the wake from WFI in the controller depend upon irq_pending:

https://github.com/lowRISC/ibex/blob/f90faa6ca365e02047d456ac3ad084ecc457ada0/rtl/ibex_core.sv#L285

[...]

PR #575, opened 2020-01-29
[rtl] Decouple `mip` and `mie` CSRs

vogelpi:
This PR modifies the `mip` CSR to not depend on the `mie` CSR. While the values of both these CSRs are combined to decide whether an interrupt shall be handled, the RISC-V spec does not state that the content of of `mip` should depend on `mie`. This commit better aligns Ibex with other open-source RISC-V cores.

This resolves lowRISC/ibex#567 reported by @pfmooney.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -52,12 +52,9 @@ module ibex_cs_registers #(
     input  logic                 irq_timer_i,
     input  logic                 irq_external_i,
     input  logic [14:0]          irq_fast_i,
-    output logic                 irq_pending_o,          // interupt request pending
     input  logic                 nmi_mode_i,
-    output logic                 csr_msip_o,             // software interrupt pending
-    output logic                 csr_mtip_o,             // timer interrupt pending
-    output logic                 csr_meip_o,             // external interrupt pending
-    output logic [14:0]          csr_mfip_o,             // fast interrupt pending
+    output logic                 irq_pending_o,          // interrupt request pending
+    output ibex_pkg::irqs_t      irqs_o,                 // interrupt requests qualified with mie
     output logic                 csr_mstatus_mie_o,
     output logic [31:0]          csr_mepc_o,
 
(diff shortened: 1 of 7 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -173,7 +173,6 @@ module ibex_cs_registers #(
   logic  [5:0] mcause_q, mcause_d;
   logic [31:0] mtval_q, mtval_d;
   logic [31:0] mtvec_q, mtvec_d;
-  Interrupts_t mip;
```
GregAC: I wonder if there would be any value in retaining this mip signal and pushing Interrupts_t into ibex_pkg. Then when passing around these interrupts we can have a single signal rather than breaking it out into the different interrupts?

rtl/ibex_cs_registers.sv
```diff
@@ -647,7 +635,14 @@ module ibex_cs_registers #(
...
 
-  assign irq_pending_o = csr_msip_o | csr_mtip_o | csr_meip_o | (|csr_mfip_o);
+  // Qualify incoming interrupt requests with mie CSR for controller and to re-enable clock upon
+  // WFI (must be purely combinational). These values may differ from the content of the mip CSR.
+  assign msip_o = irq_software_i & mie_q.irq_software;
+  assign mtip_o = irq_timer_i    & mie_q.irq_timer;
+  assign meip_o = irq_external_i & mie_q.irq_external;
+  assign mfip_o = irq_fast_i     & mie_q.irq_fast;
+
+  assign irq_pending_o = msip_o | mtip_o | meip_o | (|mfip_o);
```
GregAC: Related to my other comment if we retain the mip signal (or something similar to it) we could turn these lines into

assign mip_o = mip;
assign irq_pending = |mip;

Or we have two signals, one is the raw interrupts the other is the interrupts with the enables

```verilog
assign raw_interrupts.irq_software = irq_software_i;
....

assign mip = raw_interrupts & mie_q;
```

GregAC: Though not entirely sure such code works well with packed structs, I feel like it should be ok but not sure what the standard says.

rtl/ibex_cs_registers.sv
vogelpi: Thanks for your review @GregAC . I think this is a good idea and we are also passing packed structs between the CSRs and PMPs already. I will now rework this PR to do the same for the interrupts.

vogelpi:
@GregAC I reworked the PR. Do you want to have another quick look before I merge it maybe?

pfmooney:
Thanks all

Issue #557, opened 2020-01-16
Nested NMI interrupts


udinator:
Hi,
When debugging a test case in which I asserted two nested NMI interrupts to force the core to respond to the second NMI interrupt while executing the handler for the first NMI, resulting in this waveform:
![Screenshot from 2020-01-16 13-01-12](https://user-images.githubusercontent.com/16736281/72562455-560e1200-3860-11ea-82bc-b7b538baff8d.png)
At the location of the cursor, Ibex is in the middle of executing the NMI interrupt handler and sees the input interrupt lines being changed to another set of incoming interrupts; these new interrupts are processed when `csr_save_cause_i` goes high at `3,843,000ns`.  Note that `nmi_mode_q` in the controller is high before the second set of interrupts due to the NMI detection.
However, this second set of interrupts also includes a NMI, but Ibex sets `mcause` to `0b11_1110 = 0x3e` indicating a fast interrupt, causing a testbench failure since we should expect `mcause` to be `0b11_1111 = 0x3f` for the NMI. This mismatch is due to the `IRQ_TAKEN` state in the controller, specifically this section:
![Screenshot from 2020-01-16 13-21-00](https://user-images.githubusercontent.com/16736281/72563686-0d0b8d00-3863-11ea-8229-7ef814b02245.png).
Since `nmi_mode_q` is already high from the first NMI that Ibex was initially handling, the core does not recognize the NMI asserted inside of the interrupt handler, and as such delegates it to be a fast interrupt, causing a mismatch in `mcause`.

Another potential issue that could arise is from this section in the CSR register file:
![Screenshot from 2020-01-16 13-25-38](https://user-images.githubusercontent.com/16736281/72564000-b2befc00-3863-11ea-87e4-a4b7110aded3.png)
Since Ibex restores CSR state from the `mstack` CSRs only in the case of NMI interrupts, I don't think the core would be able to correctly return from a series of nested NMIs, as the saved state from the first NMI in the `mstack` CSRs would be overwritten by the state of the second NMI.

Given these potential issues, my question is if this situation of nested NMI interrupts is something that Ibex can handle at all?

vogelpi:
Hi @udinator,
thanks for filing this issue and the detailed report.

[...]

PR #577, opened 2020-01-29
Ignore all interrupts in NMI mode, clarify documentation

vogelpi:
This is related to lowRISC/ibex#557.

### DIFF
--- a/doc/exception_interrupts.rst
+++ b/doc/exception_interrupts.rst
@@ -5,8 +5,10 @@ Exceptions and Interrupts
 
 Ibex implements trap handling for interrupts and exceptions according to the RISC-V Privileged Specification, version 1.11.
 
+When entering an interrupt/exception handler, the core sets the ``mepc`` CSR to the current program counter and saves ``mstatus``.MIE to ``mstatus``.MPIE.
 All exceptions cause the core to jump to the base address of the vector table in the ``mtvec`` CSR.
 Interrupts are handled in vectored mode, i.e., the core jumps to the base address plus four times the interrupt ID.
+Upon executing an MRET instruction, the core jumps to the program counter previously saved in the ``mepc`` CSR and restores ``mstatus``.MPIE to ``mstatus``.MIE.
 
 The base address of the vector table is initialized to the boot address (must be aligned to 256 bytes, i.e., its least significant byte must be 0x00) when the core is booting.
 The base address can be changed after bootup by writing to the ``mtvec`` CSR.
@@ -47,10 +49,13 @@ After reset, all interrupts are disabled.
 To enable interrupts, both the global interrupt enable (MIE) bit in the ``mstatus`` CSR and the corresponding individual interrupt enable bit in the ``mie`` CSR need to be set.
 For more information, see the :ref:`cs-registers` documentation.
 
-If multiple interrupts are pending, the highest priority is given to the interrupt with the highest ID.
+If multiple interrupts are pending, they are handled in the priority order defined by the RISC-V Privileged Specification, version 1.11 (see Machine Interrupt Registers, Section 3.1.9).
+The highest priority is given to the interrupt with the highest ID, except for timer interrupts, which have the lowest priority.
 
 The NMI is enabled independent of the values in the ``mstatus`` and ``mie`` CSRs, and it is not visible through the ``mip`` CSR.
 It has interrupt ID 31, i.e., it has the highest priority of all interrupts and the core jumps to the trap-handler base address (in ``mtvec``) plus 0x7C to handle the NMI.
+When handling the NMI, all interrupts including the NMI are ignored.
+Nested NMIs are not supported.
 
 All interrupt lines are level-sensitive.
 It is assumed that the interrupt handler signals completion of the handling routine to the interrupt source, e.g., through a memory-mapped register, which then deasserts the corresponding interrupt line.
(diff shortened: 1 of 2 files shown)


doc/exception_interrupts.rst
```diff
@@ -94,13 +98,14 @@ Ibex can trigger an exception due to the following exception causes:
 The illegal instruction exception, instruction access fault, LSU error exceptions and ECALL instruction exceptions cannot be disabled and are always active.
 
 
-Handling
---------
+Nested Interrupt/Exception Handling
+-----------------------------------
 
 Ibex does support nested interrupt/exception handling.
```
udinator: Could you specify what the max number of nested interrupts would be that Ibex can support?

vogelpi: Well, since the saving of these registers is has to happen in software, the number of nested interrupts supported is infinite.

udinator: Sorry, let me clarify what I mean, I phrased that question poorly. At just the hardware level, Ibex has the standard set of CSRs to handle CSRs as well as the `mstack` CSRs to support one level of nesting. Say that the core is currently handling a nested interrupt so both of these sets of CSRs contain relevant state, and then the core receives another nested interrupt in the middle. Assuming software doesn't store this state separately, it seems like the stored state from the initial interrupt would get overridden, correct?
Would this mean that the core can support at most one nested interrupt at the hardware level before having to overwrite state (assuming software doesn't do anything to help)?
Please feel free to correct me if any of this is inaccurate, as the spec is very vague on nested interrupts and related topics.

vogelpi: The `mstack` CSRs are only there for the NMI. The other CSRs are only ever restored from them when executing an MRET in NMI mode. They are not accessible by software. Thus the hardware "nesting" is only supported for one NMI. I tried to clarify this in the documentation.

doc/exception_interrupts.rst
```diff
@@ -47,10 +49,13 @@ After reset, all interrupts are disabled.
 To enable interrupts, both the global interrupt enable (MIE) bit in the ``mstatus`` CSR and the corresponding individual interrupt enable bit in the ``mie`` CSR need to be set.
 For more information, see the :ref:`cs-registers` documentation.
 
-If multiple interrupts are pending, the highest priority is given to the interrupt with the highest ID.
+If multiple interrupts are pending, they are handled in the priority order defined by the RISC-V Privileged Specification, version 1.11 (see Machine Interrupt Registers, Section 3.1.9).
+The highest priority is given to the interrupt with the highest ID, except for timer interrupts which have the lowest priority.
```
imphil:
```suggestion
The highest priority is given to the interrupt with the highest ID, except for timer interrupts, which have the lowest priority.
```

doc/exception_interrupts.rst
```diff
@@ -47,10 +49,13 @@ After reset, all interrupts are disabled.
...
 
 The NMI is enabled independent of the values in the ``mstatus`` and ``mie`` CSRs, and it is not visible through the ``mip`` CSR.
 It has interrupt ID 31, i.e., it has the highest priority of all interrupts and the core jumps to the trap-handler base address (in ``mtvec``) plus 0x7C to handle the NMI.
+When handling an NMI, all interrupts including the NMI are ignored.
```
imphil: suggestion: an -> the (there's only one)

doc/exception_interrupts.rst
```diff
@@ -94,13 +99,20 @@ Ibex can trigger an exception due to the following exception causes:
...
---------
+Nested Interrupt/Exception Handling
+-----------------------------------
 
-Ibex does support nested interrupt/exception handling.
-Exceptions inside interrupt/exception handlers cause another exception.
-However, exceptions during the critical part of your exception handlers, i.e. before having saved the ``mepc`` and ``mstatus``, will cause those CSRs to be overwritten.
-Interrupts during interrupt/exception handlers are thus disabled by default, but can be explicitly enabled if desired.
+Ibex does support nested interrupt/exception handling in software.
+However, interrupts/exceptions during the critical part of the handler, i.e. before software has saved the ``mepc`` and ``mstatus``, will cause those CSRs to be overwritten.
```
imphil: will -> would (can't happen as discussed in the next sentence)

doc/exception_interrupts.rst
```diff
@@ -94,13 +99,20 @@ Ibex can trigger an exception due to the following exception causes:
...
+The hardware thus automatically disables interrupts upon entering an interrupt/exception handler.
```
imphil: Can you swap this sentence with the previous one? First explain what's happening, and then explain why it is implemented this way.

doc/exception_interrupts.rst
```diff
@@ -94,13 +99,20 @@ Ibex can trigger an exception due to the following exception causes:
...
+If desired, software can explicitly enable interrupts by setting ``mstatus``.MIE to 1'b1 from within the handler.
+However, software should only do this after saving ``mepc`` and ``mstatus``.
+There is no limit on the maximum number of nested interrupts.
+Note that, after enabling interrupts by setting ``mstatus``.MIE to 1'b1, the current handler will be interrupted also by lower priority interrupts.
```
imphil: Would it be possible to include a couple lines of pseudo-code how to handle nested interrupts in software?

```c
isr_handle_nested_interrupts() {
  // Save mpec and mstatus to stack
  mepc_bak = mepc;
  mstatus_bak = mstatus;

  // Re-enable interrupts (optional)
  mstatus.MIE = 1;

  // Handle interrupt
  // This code block can be interrupted by other interrupts, including lower-priority ones.

  // XXX: Include example code how to get only higher priority interrupts.

  // XXX: Do you need to disable interrupts before restoring?

  // Restore mepc and mstatus
  mpec = mepc_bak;
  mstatus = mstatus_bak;
}

doc/exception_interrupts.rst
vogelpi: Done.

doc/exception_interrupts.rst
vogelpi: Done.

doc/exception_interrupts.rst
vogelpi: Done.

doc/exception_interrupts.rst
vogelpi: Yep, good idea.

doc/exception_interrupts.rst
vogelpi: Excellent idea! I adapted your code snipped a bit and added that to the documentation.

vogelpi:
Thanks for your comments @udinator and @imphil , I implemented your suggestions. Please let me know if you are happy with the changes.

vogelpi:
Thanks @imphil !

PR #582, opened 2020-02-03
Fix incdirs of src_files.yml

dalance:
This PR fixes `incdirs` of `src_files.yml`.
#570 added `` `include "prim_assert.sv"``, but no incdir added.


imphil:
Good catch, thanks!

PR #583, opened 2020-02-04
Extend riscv-compliance description

towoe:
Include all tests in the description which can be executed for Ibex and
riscv-compliance.


imphil:
Thanks!

PR #584, opened 2020-02-04
[doc] Riviera-PRO instructions for Simple System

GregAC:
@dawidzim does this look ok to you?

### DIFF
--- a/examples/simple_system/README.md
+++ b/examples/simple_system/README.md
@@ -113,6 +113,18 @@ To run the simulator:
 
 Pass `-gui` to use the DVE GUI.
 
+## Simulating with Riviera-PRO
+
+To build and run Simple System run:
+
+```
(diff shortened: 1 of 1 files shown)


examples/simple_system/README.md
```diff
@@ -113,6 +113,21 @@ To run the simulator:
 
 Pass `-gui` to use the DVE GUI.
 
+## Simulating with Riviera-PRO
+
+To build and run Simple System run:
+
+```
+fusesoc --cores-root=. run --target=sim --tool=rivierapro lowrisc:ibex:ibex_simple_system --RV32M=1 --RV32E=0 \
+  --SRAM_INIT_FILE=<sw_vmem_file> --vsim_options=+ibex_tracer_file_base=test_results
```
imphil: ibex_tracer_file_base should actually be a plusarg known to fusesoc. That's currently not the case and something we should fix. In the meantime (and for consistency with other examples in this file), can you remove the plusarg?

imphil:
And can you file an issue to make ibex_tracer_file_base a proper plusarg in the .core files of the tracer and the simple system?

dawidzim:
LGTM

PR #588, opened 2020-02-06
[dv] enable PMP

udinator:
-enable the ibex PMP module to be included in simulations
-add basic PMP test
-set some initial default value to the rtl PMP configuration parameters
-simulations with this setup pass successfully with cosimulating with OVPsim

### DIFF
--- a/dv/uvm/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/riscv_dv_extension/testlist.yaml
@@ -574,3 +574,13 @@
   rtl_test: core_ibex_invalid_csr_test
   sim_opts: >
     +require_signature_addr=1
+
+- test: riscv_pmp_test
+  desc: >
+    Random PMP settings
+  iterations: 1
+  gen_test: riscv_rand_instr_test
+  gen_opts: >
+    +instr_cnt=6000
+    +num_of_sub_program=2
+  rtl_test: core_ibex_base_test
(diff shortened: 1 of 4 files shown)


dv/uvm/riscv_dv_extension/testlist.yaml
```diff
@@ -574,3 +574,13 @@
   rtl_test: core_ibex_invalid_csr_test
   sim_opts: >
     +require_signature_addr=1
+
+- test: riscv_pmp_test
+  desc: >
+    Random PMP settings
+  iterations: 1
+  gen_test: riscv_rand_instr_test
+  gen_opts: >
```
taoliug: is there any option needed to enable pmp cfg?

udinator: This is done through the support_pmp field in the core_setting file

taoliug: I am wondering if there can be a test specific option to tun off pmp while support_pmp = 1. This test entry doesn't look different from other random test.

udinator: The issue is that I can't turn off PMP because it is a compile time parameter to the design, which would require two separate compiled vcs binaries. My thinking is to have all other tests run the default values of the PMP setup, since the default configuration will allow any memory access without any restrictions.

PR #590, opened 2020-02-07
:zap: Reduce latency of slow multiplier

stmach:
- The slow multiplier is modified to terminate iterations early instead
  of always going the full 32 iterations.
- Multiplications now terminate early after clog2(`op_b`) iterations.
- The slow multiplier can be further optimized by swapping the smaller
  operand into `op_b` when in the `MD_IDLE` state.

### DIFF
--- a/doc/pipeline_details.rst
+++ b/doc/pipeline_details.rst
@@ -30,57 +30,57 @@ In the table below when an instruction stalls for X cycles X + 1 cycles pass bef
 Some instructions stall for a variable time, this is indicated as a range e.g. 1 - N means the instruction stalls a minimum of 1 cycle with an indeterminate maximum cycles.
 Read the description for more information.
 
-+-----------------------+-----------------------+-------------------------------------------------------------+
-| Instruction Type      | Stall Cycles          | Description                                                 |
-+=======================+=======================+=============================================================+
-| Integer Computational | 0                     | Integer Computational Instructions are defined in the       |
-|                       |                       | RISCV-V RV32I Base Integer Instruction Set.                 |
-+-----------------------+-----------------------+-------------------------------------------------------------+
-| CSR Access            | 0                     | CSR Access Instruction are defined in 'Zicsr' of the        |
-|                       |                       | RISC-V specification.                                       |
-+-----------------------+-----------------------+-------------------------------------------------------------+
-| Load/Store            | 1 - N                 | Both loads and stores stall for at least one cycle to await |
-|                       |                       | a response.  For loads this response is the load data       |
-|                       |                       | (which is written directly to the register file the same    |
-|                       |                       | cycle it is received).  For stores this is whether an error |
-|                       |                       | was seen or not.  The longer the data side memory interface |
-|                       |                       | takes to receive a response the longer loads and stores     |
-|                       |                       | will stall.                                                 |
-+-----------------------+-----------------------+-------------------------------------------------------------+
-| Multiplication        | 2/3 (Fast Multiplier) | 2 for MUL, 3 for MULH.                                      |
-|                       |                       | See details in :ref:`mult-div`                              |
-|                       | 32 (Slow Multiplier)  |                                                             |
-+-----------------------+-----------------------+-------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


rtl/ibex_multdiv_slow.sv
```diff
@@ -221,19 +221,20 @@ module ibex_multdiv_slow (
               accum_window_d = res_adder_l;
               op_a_shift_d   = op_a_shift_q << 1;
               op_b_shift_d   = op_b_shift_q >> 1;
+              md_state_d     = !(op_b_shift_q >> 1) ? MD_LAST : MD_COMP;
             end
             MD_OP_MULH: begin
               accum_window_d = res_adder_h;
               op_a_shift_d   = op_a_shift_q;
               op_b_shift_d   = op_b_shift_q >> 1;
+              md_state_d     = !(op_b_shift_q >> 1) ? MD_LAST : MD_COMP;
```
vogelpi: If we do an early termination here, the result is not yet properly shifted. You still have to cycle through all interations for `MULH`:
```
md_state_d     = (multdiv_state_q == 5'd1) ? MD_LAST : MD_COMP;
```

stmach:
Yikes, @vogelpi you're right! I reverted the changes for MULH and adapted the doc accordingly.

towoe:
I was curious what `:zap:` meant in the title. Realized it is an emoji. Not sure if the guidelines advice against the usage of eomjis for the commit message, but I would suggest to remove it. @vogelpi @imphil what do you think?

imphil:
It's common in the PULP project (and some others), I find it useless (and as you found out, confusing), but wouldn't block merging this PR based on it.

GregAC:
Personally I'd prefer to keep emoticons out of commit messages (as they're not going to render in many places, leaving potentially confusing text), a tag such as [rtl] can be useful so at a glance you have a rough idea of what a particular commit may have touched.

stmach:
Feel free to adapt the commit message as necessary when merging, or let me know so I can change it to e.g. use the [rtl] tag.

towoe:
Thanks for the feedback @imphil and @GregAC.
@stmach I think the easiest thing to do would be if you change the commit message and force push, from my perspective `[rtl]` is not a requirement for this specific commit.
But before you edit the message, lets wait and see if @vogelpi has some other review requests.

doc/pipeline_details.rst
```diff
@@ -30,57 +30,57 @@ In the table below when an instruction stalls for X cycles X + 1 cycles pass bef
...
+|                       |                                      | (which is written directly to the register file the same    |
+|                       |                                      | cycle it is received).  For stores this is whether an error |
+|                       |                                      | was seen or not.  The longer the data side memory interface |
+|                       |                                      | takes to receive a response the longer loads and stores     |
+|                       |                                      | will stall.                                                 |
++-----------------------+--------------------------------------+-------------------------------------------------------------+
+| Multiplication        | 2/3 (Fast Multiplier)                | Fast: 2 for MUL, 3 for MULH.                                |
+|                       |                                      | Slow: clog2(``op_b``) for MUL, 32 for MULH                  |
+|                       | clog2(``op_b``)/32 (Slow Multiplier) | See details in :ref:`mult-div`                              |
```
vogelpi: Can you please
- move the content in the middle column from Line 52 to Line 51, and
- add a full stop after MULH on Line 51?
The rendered documentation currently looks weird.

stmach: OK will do. FYI the original also had the middle column text on Lines 50 and 52 with 51 empty, not sure how that affects rendering of the doc exactly.

vogelpi: It looks like this:

Screenshot from 2020-02-10 13-00-29

stmach: Hmm yes, the whole clog2 thing is a bit long, you could trade off accuracy for legibility and use "up to 32" for the slow multiplier part. Not sure whether that's desired though.

imphil: I've filed https://github.com/lowRISC/ibex/issues/595 to make the table easier to edit in the future. For now, do whatever is easiest to get the right rendering, and don't worry to much about it looking "ugly" in source code.

PR #591, opened 2020-02-07
[rtl] Fix assertion issues

### LINKED ISSUE #548: Illegal instruction assertion fire
Hi all,
I'm seeing a potential issue where, upon decoding an illegal instruction that contains `x` within it, Ibex will display that the instruction is illegal, but assertions in the LSU will fire, causing the simulation to error out. The simulation log is shown below,
Screenshot from 2020-01-09 14-08-58
We can see that at `PC=0x80006232` Ibex recognizes that there is an illegal instruction, and the core does in fact respond correctly by trapping to the proper exception handler, but then the `IbexDataOffsetKnown` assertion in the LSU constantly fires due to an `x` from this instruction propagating through the design, failing the simulation.
What do you think about additionally qualifying these assertions that use `$isunknown` with a check ensuring the instruction is not illegal?
Udi

GregAC:
Fixes #548

Fixes the long standing assertion problem from the linked issue as well as a new problem that was introduced with the branch target ALU.

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -567,7 +567,10 @@ module ibex_core #(
       .load_err_o            ( lsu_load_err        ),
       .store_err_o           ( lsu_store_err       ),
 
-      .busy_o                ( lsu_busy            )
+      .busy_o                ( lsu_busy            ),
+
+      .illegal_insn_id_i     ( illegal_insn_id     ),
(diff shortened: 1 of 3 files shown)


GregAC:
I want to rebase this on #592 (which adds default clk/reset to prim_assert) once that's gone in. I'll fix the lint issue causing the CI failure when I do that.

PR #604, opened 2020-02-11
Document and check tool versions

### LINKED ISSUE #603: Document and enforce minimum Verilator version
We continue to use features in Verilator which require a sufficiently up-to-date version of Verilator. We need to

- Document our minimum-supported Verilator version.
- Somehow signal to users that they are running a version of Verilator which is too old. Ideally, this check happens in the fusesoc run to display this information when it is actually needed.

I'll give it a try with a fusesoc hook.

imphil:
Provide machine-readable minimum tool versions in the `tool_requirements.py` file. (Currently for Verilator only.) Use this information in two ways:
- In the written documentation.
- As a check in fusesoc to error out if outdated tools are used.

Please see the individual commit messages for more details.

Fixes #603


doc/system_requirements.rst
```diff
@@ -0,0 +1,33 @@
+System and Tool Requirements
+============================
+
+The Ibex CPU core is written in SystemVerilog.
+We try to achieve a balance between the used language features (as described in our `style guide <https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md>`_) and reasonably wide tool support.
+
+We regularly use and test the following tools for working with the RTL code of Ibex.
```
asb: Is this statement perhaps a bit strong? We've heard positive success stories from the tools listed, but the core Ibex developers don't use all of them in their day-to-day flow, and we don't have automated testing set up for them?

imphil: Slightly reworded that one, thanks!

doc/system_requirements.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Tools with known issues
+-----------------------
+
+Not all EDA tools have enough SystemVerilog support to be able to work with the Ibex code base.
+Users of such tools are encouraged to file issues with the vendor.
+As a workaround, tools like `sv2v <https://github.com/zachjs/sv2v>`_ can pre-process the source code to an older version of Verilog.
+
+- Intel (Altera) Quartus Prime Lite and Standard are *not* supported due to insufficient SystemVerilog support
+  (`issue #117 <https://github.com/lowRISC/ibex/issues/117>`_).
```
vogelpi: Maybe you could also add a reference to PR lowRISC/ibex#336 which discusses some more issues than Issue #117 and also contains a link to Marek's documentation of Quartus issues at the top.

imphil: I extended the issue to summarize the findings. Let's point people there to have a single point of reference.

PR #606, opened 2020-02-14
[rtl] Alter multdiv to better match style guide

GregAC:
Use of case inside in always_ff block does not meet style guide
recomendations. Refactored to remove this.

### DIFF
--- a/rtl/ibex_multdiv_fast.sv
+++ b/rtl/ibex_multdiv_fast.sv
@@ -46,8 +46,12 @@ module ibex_multdiv_fast #(
   logic        mult_valid;
   logic        signed_mult;
 
-  // Shared signals (div + mult)
-  logic [33:0] mac_res_q, mac_res_d, mac_res, op_remainder_d;
+  // Flop used for intermediate value holding during div & mul calculation
+  logic [33:0] intermediate_val_q, intermediate_val_d;
+  // Results that become intermediate value depending on whether mul or div is being calculated
+  logic [33:0] mac_res_d, op_remainder_d;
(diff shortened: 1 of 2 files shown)


rtl/ibex_multdiv_fast.sv
```diff
@@ -89,19 +92,16 @@ module ibex_multdiv_fast #(
...
 
-      unique case(1'b1)
-        mult_en_i:
-          mac_res_q <= mac_res_d;
-        div_en_i:
-          mac_res_q <= op_remainder_d;
-        default:
-          mac_res_q <= mac_res_q;
-       endcase
+      if (mult_en_i | div_en_i) begin
```
tomroberts-lowrisc: I don't think its currently in the style guide, but I would propose splitting this out further, i.e. only one if-condition per always_ff. There's a risk that clock-gate inference fails with anything more than that. I can do a style-guide PR if people agree.

GregAC: Sounds sensible to me, I'll update this PR to do that in this instance.

You can also go a bit further and state that either there's no reset in the always_ff block or everything that's written in the always_ff block must be reset, i.e.

```systemverilog
  if (~rst_ni) begin
    signal_1_q <= '0
  end else if (signal_en) begin
    signal_1_q <= signal_1_d;
    signal_2_q <= signal_2_d;
  end
end
```

Would not be allowed

tomroberts-lowrisc: Yep, agreed

PR #609, opened 2020-02-19
Ibex coremark

GregAC:


examples/sw/benchmarks/coremark/ibex/core_portme.c
```diff
@@ -0,0 +1,230 @@
+/*
+        File : core_portme.c
+*/
+/*
+        Author : Shay Gal-On, EEMBC
+        Legal : TODO!
+*/
```
GregAC: Uncertain if we should include the lowRISC license header on this too, it's a copy of a file from the coremark repository with some changes.

I note it doesn't have a lowRISC license header in the OT repository, this also applies to the other portme files

imphil: Just put a normal lowRISC header on it.

examples/sw/benchmarks/coremark/ibex/core_portme.c
```diff
@@ -0,0 +1,230 @@
...
+
+void coremark_exc_handler(void) {
+  ee_printf("EXCEPTION!!!\n");
+  ee_printf("============\n");
+  ee_printf("MEPC:   0x%0X\n", get_mepc());
+  ee_printf("MCAUSE: 0x%0X\n", get_mcause());
+  ee_printf("MTVAL:  0x%0X\n", get_mtval());
+  sim_halt();
+
+  while(1);
```
imphil:
```suggestion
  while (1);
```

vendor/patches/eembc_coremark/0001-no-minimum-run-time.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/core_main.c b/core_main.c
+index 6161974..cca596c 100644
+--- a/core_main.c
++++ b/core_main.c
+@@ -295,10 +295,16 @@ MAIN_RETURN_TYPE main(int argc, char *argv[]) {
+ 	if (time_in_secs(total_time) > 0)
+ 		ee_printf("Iterations/Sec   : %d\n",default_num_contexts*results[0].iterations/time_in_secs(total_time));
+ #endif
+-	if (time_in_secs(total_time) < 10) {
```
imphil: How about adding a `define` for this functionality, giving people to override it in the Makefile? If you do so, please also add this piece of information to the README.

examples/sw/benchmarks/coremark/ibex/core_portme.mak
```diff
@@ -0,0 +1,85 @@
+#File : core_portme.mak
+
+OUTFILES = $(OPATH)coremark.dis $(OPATH)coremark.map
```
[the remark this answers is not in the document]
GregAC: The main coremark makefile uses it similarly (no slash), so I'd prefer to leave it as is. Otherwise when you fail to have a trailing slash you may get slightly confusing behaviour as half of the build system will work and the other half won't.

vendor/patches/eembc_coremark/0001-no-minimum-run-time.patch
GregAC: I think we're better off without. The reason to put this code back in is to meet the 'letter of the law' wrt to publishing coremark results (absolutely no alterations to this file amongst others). The define is still a file alteration so doesn't help you if you want 'letter of the law' coremark results.

GregAC:
Thanks for the review @imphil

examples/sw/benchmarks/coremark/ibex/core_portme.mak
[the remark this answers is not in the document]
rswarbrick: Also, the `$(OPATH)coremark.dis` style is technically better, even if it is a bit confusing: what happens if OPATH is the empty string?

azure-pipelines.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
   - bash: |
       fork_origin=$(git merge-base --fork-point origin/master)
-      changed_files=$(git diff --name-only $fork_origin | grep -v /vendor/ | grep -E '\.(cpp|cc|c|h)$')
+      changed_files=$(git diff --name-only $fork_origin | grep -v '^vendor' | grep -E '\.(cpp|cc|c|h)$')
```
GregAC: Decided to alter this so it'll only remove things where vendor is at the beginning of the path.

With the plain 'vendor' any file that has 'vendor' in the path or filename will get left out of lint. This could cause a file that isn't vendored into the repository to be left out of lint CI.

A rule that we don't put 'vendor' on any path or filename that isn't vendored is reasonable but the issue was if we fail to apply that rule then we'll end up with a file in the repository not getting the lint check and we won't get anything that flags that up to us.

This way round if we want something that's vendored that isn't under the top-level vendor/ we'll see a CI failure and can consider the best way forward.

May be worth altering this to '^vendor/' to make it strictly only filter stuff in the vendor/ directory.

GregAC:
@imphil happy for this to go in?

azure-pipelines.yml
imphil: This rule was copied from OpenTitan, where vendor directories are all over the place. But for Ibex LGTM.

GregAC: Interesting, surprised the /vendor/ style works in OT but not in Ibex which is why I switched to 'vendor' (this PR was failing CI because it was linting the vendored coremark source as the grep had failed to filter things with vendor in).

examples/sw/benchmarks/coremark/ibex/core_portme.c
imphil: Please remove everything starting from the `// This file is ...`, and replace it with two lines  under the `Copyright lowRISC contributors.`

```
Copyright 2018 Embedded Microprocessor Benchmark Consortium (EEMBC). 
Original Author: Shay Gal-on
```

That's all information we need to maintain in a shorter form.

GregAC:
Ah can do, github was helpfully hiding all of those comments so I missed them first time round

imphil:
Yeah, I thought so. It helps looking only at the single commit, otherwise the GitHub UI gets super slow in my experience.

PR #611, opened 2020-02-20
Tracer: Mark all functions "automatic"

dawidzim:
hey!
this function should be automatic because `string desc = "";` is cleared only once at start

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -88,7 +88,7 @@ module ibex_tracer (
   localparam MEM = (1 << 3);
   logic [3:0] data_accessed;
 
-  function void printbuffer_dumpline();
+  function automatic void printbuffer_dumpline();
     string rvfi_insn_str;
 
     if (file_handle == 32'h0) begin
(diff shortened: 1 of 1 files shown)


rtl/ibex_tracer.sv
```diff
@@ -655,7 +655,7 @@ module ibex_tracer (
     end
   endfunction
 
-  function string get_fence_description(logic [3:0] bits);
+  function automatic string get_fence_description(logic [3:0] bits);
```
imphil: For consistency: can you either make all functions automatic, or split the initializer from the variable declaration in this function (I'd do that):

```verilog
string desc;
desc = "";
```

GregAC: This feels like something that needs addressing in the style guide (had a quick look don't think we discuss this).

I think I'd be in favour of all functions being marked 'automatic'. Explicit initializers does give you the same thing but a tool needs to do some analysis to determine there's no static state that needs to be maintained between calls. Seems like the explicit 'automatic' makes life easier in that case.

In particular for any synthesisable RTL I don't think we'd ever want state to be maintained between functions (or discourage/forbid such a coding style) and want to make that as obvious as possible (though from a quick search I don't think we're using functions in any synthesisable RTL currently, within Ibex at least).

I'll open up a style guide issue to discuss.

imphil: So for now: would you mind marking all functions in this file automatic @dawidzim?

dawidzim: Yea sure, I can do this

Issue #612, opened 2020-02-20
CSR misa "I" bit value for "E" extension


towoe:
The "I" bit of the CSR misa is always set:
https://github.com/lowRISC/ibex/blob/f90faa6ca365e02047d456ac3ad084ecc457ada0/rtl/ibex_cs_registers.sv#L113

But I think it should no be set if RV32E is used.
The priv spec says it 3.1.1 p. 16

> The “I” bit will be set for RV32I, RV64I, RV128I base ISAs, and the “E” bit will be set for RV32E.

and

>The “E” bit is read-only.
>Unless misa is hardwired to zero, the “E” bit always reads as the
>complement of the “I” bit. An implementation that supports both RV32E and RV32I can select
>RV32E by clearing the “I” bit.

Any objections to the following change?
```verilog
| 32'(!RV32E) << 8)
```

imphil:
Sounds about right! Please feel free to do a PR!

PR #613, opened 2020-02-20
[rtl] Change misa for RV32E

towoe:
Following RISC-V privileged architecture version 1.11,
the "E" bit of misa should return the complement of the "I" bit.

Set the "I" bit only if RV32E is not used.

Closes lowRISC/ibex#612.


imphil:
Thanks!

PR #620, opened 2020-02-21
Separate 0x from hex number

dawidzim:
Hey!
I add `if` statement for separate `0x` from hex number, because when `0x` is argument of `atohex`, this function returns wrong number

### DIFF
--- a/dv/uvm/core_ibex/env/core_ibex_env_cfg.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_env_cfg.sv
@@ -32,7 +32,8 @@ class core_ibex_env_cfg extends uvm_object;
     void'($value$plusargs("max_interval=%0d", max_interval));
     void'($value$plusargs("require_signature_addr=%0d", require_signature_addr));
     void'($value$plusargs("signature_addr=%s", signature_addr_str));
-    if(signature_addr_str.substr(0,1)=="0x") signature_addr_str=signature_addr_str.substr(2,signature_addr_str.len()-1);
+    if(signature_addr_str.substr(0,1) == "0x") 
+      signature_addr_str = signature_addr_str.substr(2,signature_addr_str.len()-1);
     signature_addr = signature_addr_str.atohex();
   endfunction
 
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/env/core_ibex_env_cfg.sv
```diff
@@ -32,6 +32,7 @@ class core_ibex_env_cfg extends uvm_object;
     void'($value$plusargs("max_interval=%0d", max_interval));
     void'($value$plusargs("require_signature_addr=%0d", require_signature_addr));
     void'($value$plusargs("signature_addr=%s", signature_addr_str));
+    if(signature_addr_str.substr(0,1)=="0x") signature_addr_str=signature_addr_str.substr(2,signature_addr_str.len()-1);
```
luismarques: Many languages also accept the `0X` prefix (uppercase), consider also accepting that.
You probably also want to format things so that the `if` body isn't on the same line.

danielmlynek: It is not the matter of accepting "0x" prefix in atohex by us. 
This is the matter of what SV language standarts says  - and it does not mention about special support for "0x".
So if you want to have standart compliant code you need either change the code to ignore '0x" or change the macros for running simulation by not using "0x" prefix.

rswarbrick: This plusarg string is only ever set in dv/uvm/core_ibex/Makefile and it seems that it isn't set consistently (for instruction gen, we pass "8ffffffc" and for simulation we pass "0x8ffffffc"). Can't we just fix the makefile and drop "0x" support entirely?

udinator: You're right that it is not consistently set in the various Makefile steps. I'll standardize this and remove the '0x'.

udinator:
@dawidzim - as per the discussion above, I have submitted a PR that actually eliminates passing the `0x` prefix of the signature_addr into the rtl testbench, which I think directly addresses your concern. I think this change should also mean that this PR can be closed, as the extra step to strip 0x will not be necessary anymore. Thank you for bringing this issue up!

dawidzim:
LGTM

PR #623, opened 2020-02-24
Riscv-formal Verilog source

towoe:
Provide a conversion of SystemVerilog to Verilog for riscv-formal.

Update RVFI to include IXL (following upstream update). For this I made the local parameter MXL a global parameter CSR_MISA_MXL so it can be used in `ibex_core.sv`.

A new top level folder `formal` is used. This reflects the usage of `formal` in OpenTitan `hw`, instead of putting it into `dv/formal`.
The conversion is done with a Makefile and does not use FuseSoC, which is somewhat not good as the list for the source files is replicated, but this makes it easier to be used by riscv-formal.
It also stores output artifacts in `formal/build`. The intention here is to decouple it from the FuseSoC system, but I can see that it can also at some confusion.

### DIFF
--- a/formal/.gitignore
+++ b/formal/.gitignore
@@ -1,0 +1,2 @@
+build/
+ibex.v
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -80,6 +80,7 @@ module ibex_core #(
     output logic        rvfi_halt,
     output logic        rvfi_intr,
     output logic [ 1:0] rvfi_mode,
+    output logic [ 1:0] rvfi_ixl,
     output logic [ 4:0] rvfi_rs1_addr,
     output logic [ 4:0] rvfi_rs2_addr,
     output logic [ 4:0] rvfi_rs3_addr,
@@ -959,6 +960,7 @@ module ibex_core #(
   logic        rvfi_stage_halt      [RVFI_STAGES-1:0];
   logic        rvfi_stage_intr      [RVFI_STAGES-1:0];
   logic [ 1:0] rvfi_stage_mode      [RVFI_STAGES-1:0];
+  logic [ 1:0] rvfi_stage_ixl       [RVFI_STAGES-1:0];
   logic [ 4:0] rvfi_stage_rs1_addr  [RVFI_STAGES-1:0];
   logic [ 4:0] rvfi_stage_rs2_addr  [RVFI_STAGES-1:0];
   logic [ 4:0] rvfi_stage_rs3_addr  [RVFI_STAGES-1:0];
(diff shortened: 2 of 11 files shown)


rtl/ibex_core.sv
```diff
@@ -744,10 +746,11 @@ module ibex_core #(
       rvfi_order             <= rvfi_order + 64'(rvfi_valid);
       rvfi_insn              <= rvfi_insn_id;
       rvfi_mode              <= {priv_mode_id};
+      rvfi_ixl               <= CSR_MISA_MXL;
       rvfi_rs1_addr          <= rvfi_rs1_addr_id;
       rvfi_rs2_addr          <= rvfi_rs2_addr_id;
       rvfi_pc_rdata          <= pc_id;
-      rvfi_pc_wdata          <= pc_if;
+      rvfi_pc_wdata          <= pc_set ? jump_target_ex : pc_if;
```
imphil: What's the exact definition of rvfi_pc_wdata we're going for? Should it only be the (unconditional) jump destination, or also the (conditional) branch destination?

formal/.gitignore
```diff
@@ -0,0 +1,2 @@
+build/*
```
imphil:
```suggestion
build/
```

rtl/ibex_core.sv
towoe: From riscv-formal:
>This is the program counter (pc) before (rvfi_pc_rdata) and after (rvfi_pc_wdata) execution of this instruction. I.e. this is the address of the retired instruction and the address of the next instruction.

So the reason for the destination should not matter.
I will have a look at it again, but I think there is some difference in stage/valid signaling for branching and this will result in the branch destination being correctly handled by `pc_if`.

towoe:
Thanks for the review Philipp!
> It's a shame we're now duplicating the sv2v things in multiple places, but it's a good enough start for now, so let's leave that as-is.

Yeah, I am not a fan of this as well.
I am currently again in the need of sv2v, so maybe I will work on this. I am fine with leaving this PR open and update it once the support is available or we merge and I will update it later.
But using maybe FuseSoC to handle the sv2v invocation will increase the burden of dependencies for riscv-formal, but I guess this is fine as there are already required setup steps.

rtl/ibex_core.sv
imphil: I was asking because in the tracer I sometimes have to manually reconstruct the target address, and in other cases I was able to use rvfi_pc_wdata. Given that snippet you posted, it should be possible to always use rvfi_pc_wdata probably. You can have a look at the tracer and see where it doesn't use rvfi_pc_wdata to find cases to double-check.

imphil:
I wouldn't let this PR linger around, let's get it merged with the current "workaround" sv2v solution, and update once we have something better. Otherwise it will bitrot rather quickly.

rswarbrick:
Incidentally, if you want to include assertions in the actual design, the latest sv2v has an `-E assert` option (on master as of a few days ago), which disables assert stripping.

towoe:
> I wouldn't let this PR linger around, let's get it merged with the current "workaround" sv2v solution, and update once we have something better. Otherwise it will bitrot rather quickly.

True, rebasing did already need some work. And Yosys fails with an error.
I will work on this now. Running all the tests will take some time as well, will push once they work again.

towoe:
> Incidentally, if you want to include assertions in the actual design, the latest sv2v has an `-E assert` option (on master as of a few days ago), which disables assert stripping.

Thanks for the hint Rupert. I already saw your commit and tested it for something else, great work!

rtl/ibex_core.sv
towoe: Now I get what you mean. `rvfi_pc_wdata` is the actual PC of the next instruction which will follow the current valid instruction.
What you do in the tracer is to calculate the target address if the jump would be taken, regardless of the comparison of the register data.
So if the branch is not happening `rvfi_pc_wdata` will just be the increment of the current PC and not the target address as encoded in the instruction and the current PC.
If the branch is taken those two address are then equal.

imphil:
Thanks Tobi!

towoe:
Thanks for the review!

PR #624, opened 2020-02-25
Modifiy Performance Counter for DSP Inference

ganoam:
This commit modifies the cs_registers module to allow inference of DSP
slices to absorb counter logic. A preprocessor variable is introduced
to include the required synthesis pragma. DSP inferrence is supported
for Xilinx FPGA devices featuring DSP48E1 slices or similar and for
counter widths up to 48 bit.

The resulting relative logic utilizations with respect to the non-DSP
32 bit counter implementation are 84% and 90% respectively for 32 and
48 bit counters. For 32 bit counters only, the corresponding
flip-flops can be incorporated into the DSP's output pipeline
register, resulting in a reduction of the number of flip-flops to 50%.
In order to infer DSP slices for performance counters, define the
preprocessor variable XILINX_DSP_COUNTERS.

### DIFF
--- a/doc/performance_counters.rst
+++ b/doc/performance_counters.rst
@@ -122,3 +122,17 @@ The remaining event selector CSRs are tied to 0, i.e., no events are counted by 
 +----------------------+-------------+-------------+--------------+
 | ``mhpmevent10(h)``   | 0x32A       | 0x0000_0400 |           10 |
 +----------------------+-------------+-------------+--------------+
+
+FPGA Targets
+------------
+
+For FPGA targets the performance counters constitute a particularily large structure.
+Implementing the maximum 29 event counters 32, 48 and 64 bit wide results in relative logic utilizations of the core of 100%, 111% and 129% respectively.
+The relative numbers of flip-flops are 100%, 125% and 150%.
+It is recommended to implement event counters of 32 bit width where possible.
+
+For Xilinx FPGA devices featuring the `DSP48E1` DSP slice or similar, counter logic can be absorbed into the DSP slice for widths up to 48 bits.
+The resulting relative logic utilizations with respect to the non-DSP 32 bit counter implementation are 83% and 89% respectively for 32 and 48 bit DSP counters.
+This comes at the expense of 1 DSP slice per counter.
+For 32 bit counters only, the corresponding flip-flops can be incorporated into the DSP's output pipeline register, resulting in a reduction of the number of flip-flops to 50%.
+In order to infer DSP slices for performance counters, define the preprocessor variable ``FPGA_XILINX``.
--- a/examples/fpga/artya7/top_artya7.core
+++ b/examples/fpga/artya7/top_artya7.core
@@ -31,6 +31,12 @@ parameters:
     default: "../../../../../examples/sw/led/led.vmem"
     paramtype: vlogdefine
 
+  FPGA_XILINX:
+    datatype: str
+    description: Identifies Xilinx FPGA targets to set DSP pragmas for performance counters.
+    default: 1
+    paramtype: vlogdefine
+
 targets:
   synth:
     default_tool: vivado
@@ -40,6 +46,7 @@ targets:
     toplevel: top_artya7
     parameters:
       - SRAM_INIT_FILE
+      - FPGA_XILINX
     tools:
       vivado:
         part: "xc7a100tcsg324-1"  # Default to Arty A7-100
--- a/lint/verilator_waiver.vlt
+++ b/lint/verilator_waiver.vlt
@@ -58,6 +58,52 @@ lint_off -rule UNUSED -file "*/rtl/sim/simulator_ctrl.sv" -match "*'wdata_i'[31:
 // entire 32-bit address around to make the code a bit cleaner.
 lint_off -rule UNUSED -file "*/rtl/timer.sv" -match "*'timer_addr_i'[31:10]*"
 
+// Bits of signal are not used for MHPMCounterNum < 29: mhpmcounter_we[31:MHPMCounterNum+3]
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_cs_registers.sv" -match "*'mhpmcounter_we'[31:MHPMCounterNum+3]*"
+
+// Bits of signal are not used: mhpmcounter_we[1]
+// Bits of signal are not used: mhpmcounterh_we[1]
+// Bits of signal are not used: mhpmcounter_incr[1]
+//
+// cleaner to write all bits even if not all are used
+//
+lint_off -rule UNUSED -file "*/rtl/ibex_cs_registers.sv" -match "*'mhpmcounter_we'[1]*"
+lint_off -rule UNUSED -file "*/rtl/ibex_cs_registers.sv" -match "*'mhpmcounterh_we'[1]*"
+lint_off -rule UNUSED -file "*/rtl/ibex_cs_registers.sv" -match "*'mhpmcounter_incr'[1]*"
+
+// Signals are unused if MHPMCounterNum == 0: clk_i, rst_ni
+// Signal is unused if MHPMCounterNum == 0: counter_val_i[31:0]
+//
+// If no counters are implemented, no flops are elaborated. No clock, reset or
+// next-state logic is used.
+//
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'clk_i'"
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'rst_ni'"
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counter_val_i'"
+
+// Bits of signal are not used for MHPMCounterNum < 29: counter_inc_i[28:MHPMCounterNum]
+// Bits of signal are not used for MHPMCounterNum < 29: counterh_we_i[28:MHPMCounterNum]
+// Bits of signal are not used for MHPMCounterNum < 29: counter_we_i[28:MHPMCounterNum]
+//
+// cleaner to write all bits even if not all are used
+//
+// lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counter_inc_i'[28:*]*"
+// lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counterh_we_i'[28:*]*"
+// lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counter_we_i'[28:*]*"
+
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*counter_inc_i*"
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*counterh_we_i*"
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*counter_we_i*"
+
+// Bits of signal are not used for MHPMCounterWidth < 64: counter_upd[63:MHPMCounterWidth]
+// Bits of signal are not used for MHPMCounterWidth < 64: counter_load[63:MHPMCounterWidth]
+//
+// cleaner to write all bits even if not all are used
+//
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counter_upd'[63:*]*"
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -match "*'counter_load'[63:*]*"
+
 // Signal is not used: test_en_i
 // testability signal
 lint_off -rule UNUSED -file "*/rtl/ibex_register_file_ff.sv" -match "*test_en_i*"
--- a/rtl/ibex_counters.sv
+++ b/rtl/ibex_counters.sv
@@ -1,0 +1,86 @@
+module ibex_counters #(
+  parameter int MaxNumCounters = 29,
+  parameter int NumCounters = 0,
+  parameter int CounterWidth = 32
+) (
+  input                             clk_i,
+  input                             rst_ni,
+
+  input  logic [MaxNumCounters-1:0] counter_inc_i,
+  input  logic [MaxNumCounters-1:0] counterh_we_i,
+  input  logic [MaxNumCounters-1:0] counter_we_i,
+  input  logic [31:0]               counter_val_i,
+  output logic [63:0]               counter_val_o [MaxNumCounters]
+);
+  logic [63:0] counter [MaxNumCounters];
+
+  assign counter_val_o = counter;
+
+  for (genvar i = 0; i < MaxNumCounters; i++) begin : g_counter
+    // Only elaborate flops that are needed from the given CounterWidth and NumCounters.
+    if (i < NumCounters) begin : g_counter_exists
+
+      logic [63:0] counter_upd;
+      logic [63:0] counter_load;
+      logic we;
+      logic [CounterWidth-1:0] counter_d;
+
+      // Update
+      always_comb begin
+
+        // Write
+        we = counter_we_i[i] | counterh_we_i[i];
+        counter_load[63:32] = counter[i][63:32];
+        counter_load[31:0]  = counter_val_i;
+        if (counterh_we_i[i]) begin
+          counter_load[63:32] = counter_val_i;
+          counter_load[31:0]  = counter[i][31:0];
+        end
+
+        // Increment
+        counter_upd = counter[i] + 64'h1;
+
+        // Next value logic
+        if (we) begin
+          counter_d = counter_load[CounterWidth-1:0];
+        end else if (counter_inc_i[i])begin
+          counter_d = counter_upd[CounterWidth-1:0];
+        end else begin
+          counter_d = counter[i][CounterWidth-1:0];
+        end
+      end
+
+`ifdef FPGA_XILINX
+      // Set DSP pragma for supported xilinx FPGAs
+      localparam dsp_pragma = CounterWidth < 49  ? "yes" : "no";
+      (* use_dsp = dsp_pragma *) logic [CounterWidth-1:0] counter_q;
+`else
+      logic [CounterWidth-1:0] counter_q;
+`endif
+
+      // Counter flop
+`ifdef FPGA_XILINX
+      // DSP output register requires synchronous reset.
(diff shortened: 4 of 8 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -906,13 +888,54 @@ module ibex_cs_registers #(
...
+          mhpmcounter_d = mhpmcounter_upd[i][IMHPMCounterWidth-1:0];
+        end else begin
+          mhpmcounter_d = mhpmcounter[i][IMHPMCounterWidth-1:0];
+        end
+
+      end
 
+`ifdef XILINX_DSP_COUNTERS
+      // DSP pipeline registers require synchronous reset.
+      always @(posedge clk_i) begin
```
imphil: Hm. If this is safe to do, we could do it unconditionally. If not, we shouldn't do it at all. 

@GregAC @tomroberts-lowrisc What's your take on that?

rtl/ibex_cs_registers.sv
```diff
@@ -906,13 +888,54 @@ module ibex_cs_registers #(
...
+`else
       always @(posedge clk_i or negedge rst_ni) begin
-        if(~rst_ni) begin
+`endif
+        if(!rst_ni) begin
```
imphil: while-your-at-it-nit: 
```suggestion
        if (!rst_ni) begin
```

rtl/ibex_cs_registers.sv
```diff
@@ -906,13 +888,54 @@ module ibex_cs_registers #(
       // First 3 counters must be 64-bit the rest have parameterisable width
       localparam int unsigned IMHPMCounterWidth = i < 3 ? 64 : MHPMCounterWidth;
 
+`ifdef XILINX_DSP_COUNTERS
+      localparam dsp_pragma = IMHPMCounterWidth < 49  ? "yes" : "no";
```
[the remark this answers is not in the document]
ganoam: Unfortunately it is not ignored. The counter is implemented using two DSP slices, and almost all logic savings are lost. I have not investigated what exactly happens here, since the documentation states that 48 bits is the maximum width DSP counters.

I agree it's ugly. :( But it seems to be necessary.

examples/fpga/artya7/top_artya7.core
```diff
@@ -40,6 +46,7 @@ targets:
     toplevel: top_artya7
     parameters:
       - SRAM_INIT_FILE
+      - FPGA_XILINX
```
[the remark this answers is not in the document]
ganoam: Yes, I will change that with the next push.

rtl/ibex_cs_registers.sv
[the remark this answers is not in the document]
tomroberts-lowrisc: There's a small LUT cost for including the pragma when the counters are > 48, but maybe it's better to just bear that cost for the sake of cleaning the code?

rtl/ibex_cs_registers.sv
tomroberts-lowrisc: I'm not sure it's guaranteed that you'll get clock edges during reset in all systems - hence sync resets tend to be applied using a "first_cycle_out_of_reset" type signal. Probably doesn't matter for FPGAs though.

GregAC: If you synthesise for something that doesn't have a synchronous reset primitive you'll end up with an extra mux to implement it. Uncertain what how timing/area compares between sync reset and async reset flops too.

I'd prefer to avoid synchronous reset if at all possible, but seems in this case it's required to infer DSP slices? I think if we're happy with the Xilinx specific `ifdef stuff in this file, we leave it in.

GregAC: If you synthesise for something that doesn't have a synchronous reset primitive you'll end up with an extra mux to implement it. Uncertain what how timing/area compares between sync reset and async reset flops too.

I'd prefer to avoid synchronous reset if at all possible, but seems in this case it's required to infer DSP slices? I think if we're happy with the Xilinx specific `ifdef stuff in this file, we leave it in.

rtl/ibex_cs_registers.sv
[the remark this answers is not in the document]
ganoam: Actually there is a very small LUT gain. But also a cost of two DSPs per counter.

rtl/ibex_cs_registers.sv
ganoam: It's not required. If we use asynchronous reset, the DSPs are still inferred, but with some additional combinational logic, to still output '0 asynchronously altough the actual register might not yet have been reset.

Cost is n+1 LUTs for every n bit counter.

GregAC: The issue I'm worried about in reverse then!

If we've got 48 bit counters and say 8 of them that's 392 LUTs so reasonably significant area wastage.

On the other hand I dislike the Xilinx specific logic under an `ifdef.

Maybe a more radical restructuring is called for with the counters living in a sub-module? Then we can have the precise style needed for Xilinx tools to infer DSP slices and another version with a more cleanly written style for everything else (like how we have an _fpga specific version of the register file now) or potentially just one counter module but at least confine the XILINX `ifdef to there (it's annoying to have it in the cs_registers as it just adds confusion to anyone trying to understand the micro-architecture, at least inside a counters module it's more isolated).

I'd envision the counter module top-level looking something like this:

```systemverilog
module #(
  parameter int NumCounters = 0,
  parameter int CounterWidth = 32) ibex_counters (
  input clk_i,
  input rst_ni,

  input [NumCounters-1:0] counter_inc_i,
  input [NumCounters-1:0] counter_rst_i,
  output [CounterWidth-1:0] counter_val_o [NumCounters-1:0]
);

...
```

You'd do two separate instantiations to handle the fixed 64 bits for the mcycle and minstret and then the varying width for the rest of the performance counters.

GregAC: Also has the handy property that if you have a hierarchical area report you can easily see what's being used by the counters specifically

ganoam: I think that would be a good solution. I will try it.

ganoam:
some comments are comming soon...

ganoam:
Hey guys!
Here we go again.
As suggested by @imphil , I have written a new module containg just the counter logic, and the messy bit with the preprocessor variable. In order for the original implementation to still be visible I have for now pushed a separate commit instead of amending the existing one.

Some notes:
* @Imphil has suggested to implement `mcycle` and `minstret` in one module and the up to 29 variable counters in another module. I have observed, that doing that comes with an utilization overhead with respect to implementing them in separate modules.
 Another reason to implement them in separate modules is that their width and elaboration could easily be controlled separately, if so desired.
 The code for implementing the two counters in a single module is still in the file, commented out.

 * There was also the suggestion to write two separate modules for fpga and asic implementations. I have written an implementation that more closely resembles the original. You can find the implementation here: (ibex_coutners_asic.sv)[https://github.com/ganoam/ibex/blob/fpga-opt-perf-mon/rtl/ibex_counters_asic.sv] For 32-bit counters, there seems to be an issue with optimization of this module. I suspect, the logic connected to the `counterh_we_i` signal is not removed, since they update and flop elaboration logic reside in different generate blocks.

The following table reports utilization results for the different designs, with the maximum of 29 counters. I will omit FF and DSP figures, as they did not change with respect to previously reported results. `baseline` denotes the design as it currently is on the master branch. `original` denotes the version presented in the previous commit. `combined` and `separate` denote the most recent commit, sharing one module for minstret and mcycle and implementing them separately. `asic` denotes results generated using the file mentioned above.

 ```
           | LUTs (width=32) | LUTs (width=48) | LUTS (width=64)
----------------------------------------------------------------
No DSP Pragma
----------------------------------------------------------------
baseline   | 3281            | 3638            | 4235
original   | 3261            | 3706            | 4220
combined   | 3384            | 3708            | 4233
separate   | 3181            | 3601            | 4033
asic       | 3674            | 3778            | 4166
----------------------------------------------------------------
With DSP Pragma
----------------------------------------------------------------
original   | 2726            | 2920            |
combined   | 2784            | 2939            |
separate   | 2734            | 2937            |
asic *     | 3627            | 3675            |
----------------------------------------------------------------

```
* `asic` design with DSP pragmas inserted at the appropriate locations.

The following table reports utilization results for the different designs, implementing just mcycle and minstret counters. (MHPMCounterNum=0)
```
           | LUTS
------------------
baseline   | 2339
original   | 2341
combined   | 2351
separate   | 2401
asic       | 2352
```

File: doc/performance_counters.rst [hunk 5]
```diff
@@ -122,3 +123,10 @@ The remaining event selector CSRs are tied to 0, i.e., no events are counted by
 +----------------------+-------------+-------------+--------------+
 | ``mhpmevent10(h)``   | 0x32A       | 0x0000_0400 |           10 |
 +----------------------+-------------+-------------+--------------+
+
+FPGA Targets
+------------
+
+For FPGA targets the performance counters constitute a particularily large structure. Implementing the maximum 29 event counters 32, 48 and 64 bit wide results in relative logic utilizations of the core of 100%, 114% and 129% respectively. The relative numbers of flip-flops are 100%, 125% and 150%. It is recommended to implement event counters of 32 bit width where possible.
```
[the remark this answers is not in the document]
vogelpi: This still has not been fixed. Please @ganoam adapt your documentation changes accordingly.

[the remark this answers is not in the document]
ganoam: Done. (To do: check numbers, once the PR will be merged.)

rtl/ibex_cs_registers.sv
ganoam: done.

vogelpi:
> In order for the original implementation to still be visible I have for now pushed a separate commit instead of amending the existing one.

That is fine, but once the PR got approved, the PR should be cleaned up before merging.
> 
>     * @imphil has suggested to implement `mcycle` and `minstret` in one instantiation and the up to 29 variable counters in another instantiation. I have observed that some utilization improvement can be achieved by instantiating the two counters separately. Also, it mighy be convenient to implement them in separate modules such that their width and elaboration could easily be controlled separately.

Sounds good.
>       The code for implementing the two counters in a single module is still in the file, commented out.

This is not needed. Please remove it from the PR if you do not intend to get this merged. 
> 
>     * There was also the suggestion to write two separate modules for fpga and asic implementations. I have written an implementation that more closely resembles the original. You can find the implementation here: (ibex_coutners_asic.sv)[https://github.com/ganoam/ibex/blob/fpga-opt-perf-mon/rtl/ibex_counters_asic.sv] For 32-bit counters, there seems to be an issue with optimization of this module. I suspect, the logic connected to the `counterh_we_i` signal is not removed, since they update and flop elaboration logic reside in different generate blocks. I have not investigated further, since my Xilinx-specific implementation seemed to yield better results, also for non-DSP implementations. I can still pursue the situation, if desired.
>

Sorry, but this part is a bit confusing: You tried out another implementation that follows more closely the current implementation on master. You found out that this leads to results that are worse than what you propose. Right?

If yes: that's good to know. But there is no need to share that implementation. Let's keep it simple.
If no: Please clarify.

Regarding the ASIC implementation: Once we have found agreement on the FPGA implementation of the counters (i.e. this PR), we should check the ASIC using Greg's yosys flow. If the result is not good, we can split out `ibex_counter_asic.sv` and `ibex_counter_fpga.sv`.

Please, let's get this PR ready quickly and move on.

ganoam:
Ok, I have implemented the requested changes, squashed the commits and adapted the documentation. 

> You tried out another implementation that follows more closely the current implementation on master. You found out that this leads to results that are worse than what you propose. Right?
> 
> If yes: that's good to know. But there is no need to share that implementation. Let's keep it simple.
> If no: Please clarify.

Yes, that is what I meant. Sorry about the confusion. 

I have just realized, that I may have squashed the commits too early. If you still want to have a look at the previous version, it is still on my exploration branch: [ibex_cs_registers_commit.sv](https://github.com/ganoam/ibex/blob/fpga-opt-perf-mon/rtl/ibex_cs_registers_commit.sv).

lint/verilator_waiver.vlt
```diff
@@ -58,6 +58,40 @@ lint_off -rule UNUSED -file "*/rtl/sim/simulator_ctrl.sv" -match "*'wdata_i'[31:
...
+
+// Bits of signal are not used for MHPMCounterNum < 29: counterh_we_i[29:MHPMCounterNum]
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -lines 10
+
+// Bits of signal are not used for MHPMCounterNum < 29: counter_we_i[29:MHPMCounterNum]
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -lines 11
+
+
```
vogelpi: Remove 1 newline.

lint/verilator_waiver.vlt
```diff
@@ -58,6 +58,40 @@ lint_off -rule UNUSED -file "*/rtl/sim/simulator_ctrl.sv" -match "*'wdata_i'[31:
...
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -lines 9
+
+// Bits of signal are not used for MHPMCounterNum < 29: counterh_we_i[29:MHPMCounterNum]
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -lines 10
+
+// Bits of signal are not used for MHPMCounterNum < 29: counter_we_i[29:MHPMCounterNum]
+// cleaner to write all bits even if not all are used
+lint_off -rule UNUSED -file "*/rtl/ibex_counters.sv" -lines 11
```
vogelpi: Can you please change to the `-match` syntax here? As opposed to specifying the line number, this requires less work to keep it up to date with the RTL and is supported as of Verilator v.4.028.

ganoam: done .
Is there a way to parameterize the waiver (ignore bits `[63:MHPMCounterWidth]` for example)?
I did not find a way, except for wildcards, which in the end is actually the same thing as ignoring the whole signal. - maybe I should just do that?

rtl/ibex_counters.sv
```diff
@@ -0,0 +1,88 @@
...
+  parameter int CounterWidth = 32
+) (
+  input clk_i,
+  input rst_ni,
+
+  input logic [MaxNumCounters-1 :0] counter_inc_i,
+  input logic [MaxNumCounters-1 :0] counterh_we_i,
+  input logic [MaxNumCounters-1 :0] counter_we_i,
+  input logic [31:0] counter_val_i,
+  output logic [63:0] counter_val_o [0:MaxNumCounters-1]
```
GregAC: Style nit: Could you line these up like the port definitions you see in other modules (`logic` will be aligned, `[` will be aligned, names will be aligned)

rtl/ibex_counters.sv
```diff
@@ -0,0 +1,88 @@
...
+);
+  logic [63:0] counter [0:MaxNumCounters-1];
+
+  assign counter_val_o = counter;
+
+  for (genvar i = 0; i < MaxNumCounters; i++) begin : g_counter
+    // Only elaborate flops that are needed from the given CounterWidth and NumCounters.
+    if (i<NumCounters) begin : g_counter_exists
```
GregAC: Style nit: space around the `<`

rtl/ibex_counters.sv
GregAC: Style nit: Don't need space before `:`

GregAC: Style nit: Unpacked array for `counter_valid_o` should be defined as `[63:0] counter_val_o [MaxNumCounters]`

rtl/ibex_counters.sv
GregAC: Style Nit: As with port definition just use `[MaxNumCounters]` for unpacked array.

rtl/ibex_counters.sv
```diff
@@ -0,0 +1,88 @@
...
+
+        // Next value logic
+        if (we) begin
+          counter_d = counter_load[CounterWidth-1:0];
+        end else if (counter_inc_i[i])begin
+          counter_d = counter_upd[CounterWidth-1:0];
+        end else begin
+          counter_d = counter[i][CounterWidth-1:0];
+        end
+
```
vogelpi: Remove newline.

lint/verilator_waiver.vlt
vogelpi: Thanks for changing the waiver rules.
> Is there a way to parameterize the waiver (ignore bits [63:MHPMCounterWidth] for example)? I did not find a way, except for wildcards, which in the end is actually the same thing as ignoring the whole signal. - maybe I should just do that?

I don't know if there is a way evaluate the parameter to generate the matching string. I think what you came up with is alright.

rtl/ibex_counters.sv
ganoam: Done. Thank you!

rtl/ibex_counters.sv
ganoam: Done, thanks.

rtl/ibex_counters.sv
ganoam: Done, thanks.

ganoam:
Sorry about the delay. I have not seen the reviews two days ago....

lint/verilator_waiver.vlt
ganoam: OK, thank you.

I had to change it for `counter_inc_i`, `counter_we_i` and `counterh_we_i`, since for `MHPMCounterNum=0` the whole signal must be ignored and apparently the corresponding warning is not captured by ignoring all bits.

vogelpi: SGTM, I am resolving this one.

examples/fpga/artya7/top_artya7.core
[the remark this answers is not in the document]
vogelpi: This has been changed to FPGA_XILINX in all other places. I think this comment can be resolved.

rtl/ibex_counters.sv
vogelpi: It's still here.

ganoam: oh, there was another one. sorry.

vogelpi:
Sorry, I hit "Rebase and Merge" without refreshing the browser window. We now got a merge commit :-(

PR #626, opened 2020-02-26
Improve docs for getting started in verification.rst

rswarbrick:
The important change is that we now tell the user to configure spike
with --enable-misaligned. This is needed because Ibex handles
misaligned loads and stores rather than jumping to a trap handler (the
default behaviour in spike).

This patch also points a user at the manual from the riscv-dv
repository, rather than duplicating the list of steps here.

### DIFF
--- a/doc/verification.rst
+++ b/doc/verification.rst
@@ -80,10 +80,43 @@ Getting Started
 Prerequisites & Environment Setup
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- VCS RTL simulator (needed to support UVM 1.2)
-- RISCV-DV Prerequisites - https://github.com/google/riscv-dv#prerequisites
-- GCC setup - https://github.com/google/riscv-dv#compile-generated-programs-with-gcc
-- ISS setup - https://github.com/google/riscv-dv#run-iss-instruction-set-simulator-simulation - note that commit log must be enabled in spike by passing ``--enable-commitlog`` to the configure script.
+In order to run the co-simulation flow, you'll need:
+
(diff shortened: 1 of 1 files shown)


doc/verification.rst
```diff
@@ -80,10 +80,17 @@ Getting Started
 Prerequisites & Environment Setup
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- VCS RTL simulator (needed to support UVM 1.2)
-- RISCV-DV Prerequisites - https://github.com/google/riscv-dv#prerequisites
-- GCC setup - https://github.com/google/riscv-dv#compile-generated-programs-with-gcc
-- ISS setup - https://github.com/google/riscv-dv#run-iss-instruction-set-simulator-simulation - note that commit log must be enabled in spike by passing ``--enable-commitlog`` to the configure script.
+In order to run the co-simulation flow, you'll need an RTL simulator,
+an instruction set simulator and a working RISCV compiler.
```
GregAC: Could add a link to the lowrisc toolchains that are available here: https://github.com/lowRISC/lowrisc-toolchains/releases?

doc/verification.rst
```diff
@@ -80,10 +80,17 @@ Getting Started
...
+
+The flow is currently tested with VCS as the simulator and Spike_ as
+the instruction set simulator. See the `Getting Started
+<riscv-dv-getting-started_>`_ section of the RISCV DV manual for
+instructions for how to install dependencies, but note that Spike must
```
GregAC: instructions *on* how to install dependencies

PR #628, opened 2020-02-27
Avoid unneccessary rebuilding in dv/uvm/core_ibex/Makefile

rswarbrick:
Before this patch, running the Makefile's default target deleted
everything and then ran the whole flow. This sometimes does
unnecessary work (if I've just changed the design, there's no need to
rebuild and re-run the instruction generator). It also definitely
won't work with Make's -j flag, since it depends on the targets being
built in order.

This patch keeps the same stages in the Makefile, but makes each stage
generate a stamp file, adding dependencies between the stages. This
way, you can make a small change to the design and re-run the
simulation without having to generate the random inputs again.

This doesn't make much difference if you're running lots of tests with
no LSF (since VCS is very slow, its runtime for simulation completely
dominates), but it can make a significant difference if you're
debugging a single test, have made a change to the design and want to
re-run.

One significant change is that running 'make' doesn't automatically
delete existing files any more. To make this possible (and useful!),
we generate random data and test results in a directory keyed by the
seed. For example

  make SEED=123

will generate results in out/seed-123/regr.log (rather than
out/regr.log as before).

To make sure we rebuild things properly if you change something like
the number of iterations or the tests to run, we dump some variables
describing the mode in which we were running. If these don't match the
nnext time around, we'll rebuild stuff if necessary.

Advanced (or hurried) users of the existing Makefile might have done
things like change the design and then run

  make SEED=123 compile rtl_sim

Now, the rtl_sim target depends on its logical dependencies. On the
plus side, this means that you won't accidentally simulate out-of-date
code. On the minus side, cunning tricks to avoid having to re-run
stuff after touching a design file won't work. (If you're feeling
really determined to do something like that, it's still possible with
make -t).


dv/uvm/core_ibex/Makefile
```diff
@@ -114,45 +211,123 @@ RISCV_DV_OPTS=--custom_target=${DV_DIR}/riscv_dv_extension \
...
-	mkdir -p ${OUT}/rtl_sim
-	python3 ./sim.py \
-     --o=${OUT} \
+.PHONY: compile
+compile: $(OUT)/rtl_sim/.compile.stamp
+
+###############################################################################
+# Run ibex RTL simulation with generated programs
+#
+# Because we compile a TB once rather than for each seed, we have to copy in
```
udinator: One question based off of some messing around I did with the new Makefile format.
After running `make ITERATIONS=1 TEST=riscv_rand_instr_test SEED=123 instr`, I ran `make ITERAITONS=1 TEST=riscv_rand_instr_test compile`, which then placed the output vcs binaries into `out/rtl_sim`.
Then, I ran `make ITERATIONS=1 TEST=riscv_rand_instr_test SEED=123 sim`, which then recompiled the entire testbench into `out/seed-123/rtl_sim`, instead of just copying the `out/rtl_sim` directory over, which is what this sections seems to be suggesting.
Is the full recompilation in this situation intended?

rswarbrick: The full recompilation definitely isn't intended! Oops! I'll have a look at that this morning.

About the `.metadata` directory suggestion, where exactly do you think that should go? At the moment, we're creating:
  - out/seed-123/instr_gen/.gen.stamp
  - out/seed-123/instr_gen/.compile.stamp
  - out/seed-123/instr_gen/.iss.stamp
  - out/rtl_sim/.compile.stamp
  - out/seed-123/rtl_sim/.compile.stamp
  - out/seed-123/rtl_sim/.run.stamp

This is 3 different directories (dictated by test flags, the design, and "test flags & design", respectively). Are you proposing a parallel directory tree at out/.metadata? That seems like it would be more confusing than helpful. Or have I misunderstood?

rswarbrick: Right, I think the recompilation is fixed. I'll push up a patch with the fix in a second. I've actually made some more changes which make what's going on a bit clearer for the user: the checks for whether variables have changed now print out a message if a variable differs.

The thing that caused the recompilation is that compiling the TB depended on `COMMON_OPTS` which contains the test list (which you, quite reasonably, weren't setting in your command lines above). The new patch now distinguishes between `COMMON_OPTS` and `TEST_OPTS`. The latter is used for targets that depend on which tests we're running.

rswarbrick:
With the most recent version of this patch, compiling the testbench lists all tests, which is rather verbose and possibly a bit confusing. That's because `sim.py` is calling `process_regression_list` unnecessarily and we're no longer specifying the test list when compiling the testbench.

I've fixed that separately in PR #635.

dv/uvm/core_ibex/Makefile
udinator: Sorry, I don't think my suggestion was very clear, let me try to elaborate a bit.
I think the out/rtl_sim/.compile.stamp file is fine as-is, since only that one metadata file is generated in a 'non-seed' directory.
As for all of the .stamp files generated in every seed-### directory, my suggestion would be create a directory `out/seed-123/.metadata`, and tag the files as follows:
* out/seed-123/.metadata/.instr_gen.gen.stamp
* out/seed-123/.metadata/.instr_gen.compile.stamp
* out/seed-123/.metadata/.instr_gen.iss.stamp
* out/seed-123/.metadata/.rtl_sim.compile.stamp
* out/seed-123/.metadata/.rtl_sim.run.stamp

rswarbrick:
Hi @udinator, I think I've made the path changes you requested. As well as the stamp files, I put the Makefile fragments in the metadata directory, since this is also just Make cruft.

PR #637, opened 2020-02-28
[dv] Increase timeouts to fix failures

GregAC:
I was seeing various failures due to these timeouts. Upon looking at the waves seems things were working fine it just needed more time to get to the store instructions that output the CSR info wait_for_csr_write is waiting for. So I added a large timeout increase here (it'll still die pretty quickly if we actually need the timeout).

Surprised this has only just become an issue, perhaps we're seeing a greater range of delays on the iside interface so the instruction sequence leading up to the SW that outputs what we're waiting for takes longer?

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -186,11 +186,11 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
   endfunction
 
   virtual task wait_for_core_setup();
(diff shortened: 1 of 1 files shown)


udinator:
Thanks for noticing this issue Greg! For this particular situation, we really only need to increase the wait time for the write to MSTATUS, and I've actually already opened a two part PR for this, tacked on to a PR to update the vendored in riscv-dv repository; I've just added you and Rupert as reviewers for that PR also. 
I'm fine using either your PR or the one I've opened as both as virtually identical, what do you think?

PR #622, opened 2020-02-21
Add support for Qrun Questa

danghai:
Hi, I make PR to support for qrun simulator. I try to run; however, I got a lot of warning for compilation, and optimization. I attach the output file. Could you help me take the look? I hope qrun can support for running `ibex`. Thank you
[qrun_compile_optimize.log](https://github.com/lowRISC/ibex/files/4238391/qrun_compile_optimize.log)

PR #640, opened 2020-03-01
Fix qrun compile warning

danghai:
Relate to #622

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_slave_seq_lib.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_slave_seq_lib.sv
@@ -84,7 +84,7 @@ class ibex_mem_intf_slave_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
     end
   endtask : body
 
-  virtual function inject_error();
+  virtual function void inject_error();
     this.enable_error = 1'b1;
   endfunction
 


vendor/google_riscv-dv/src/riscv_pmp_cfg.sv
```diff
@@ -160,7 +160,7 @@ class riscv_pmp_cfg extends uvm_object;
       // RV64 - pmpaddr is bits [55:2] of the whole 56 bit address, prepended by 10'b0
       // Return {10'b0, shifted_addr[53:0]}
       64: begin
-        return {10'b0, shifted_addr[53:0]};
+        return {10'b0, shifted_addr[XLEN - 11:0]};
```
rswarbrick: I assume this patch means that Questa is doing some sort of lint check that goes down this branch even when XLEN is 32. If so, can you add a note in a comment explaining why we say XLEN - 11, rather than "53"? Otherwise, this is liable to get accidentally reverted again next time someone changes the code.

danghai: I got this warning when I try qrun questa:
```
** Warning: /home/danghai/hhoangda/ibex/vendor/google_riscv-dv/src/riscv_pmp_cfg.sv(163): (vopt-2697) MSB 53 of part-select into 'shifted_addr' is out of bounds.
```

I think it is out of bounds, it should XLEN-11 = 52

danghai:
Hi @rswarbrick ,
```
Is there anything else that you want to track that needs the timestamps too?
```
I usually run makefile with default output directory. I just attach the time for default output directory to avoid confuse. It makes sure I can rebuild things properly for different default output directory.

```
Edit: Aargh, apparently I don't know how to use GitHub PRs yet. This is supposed to be attached to the 2nd commit ("Add time for output report")
```
I think you can use cherry pick to get my commit?

vendor/google_riscv-dv/src/riscv_pmp_cfg.sv
rswarbrick: Oh, I see. Firstly, isn't 64 - 11 equal to 53? Secondly, this code is vendored in from the riscv-dv repository (see https://github.com/google/riscv-dv). I think patches to this code probably need to go through there. (Sorry for not spotting the path in the first place).

rswarbrick:
Right, I've just gone through this PR again. Thanks for spotting the missing `void` keyword in `ibex_mem_intf_slave_seq_lib.sv`: we should definitely fix that.

Also, I completely agree with the change to `sim.py`. Actually, I was going to re-jig that script this morning (making it PEP8 compliant and doing general tidy-ups). So I'll grab that commit and start a "tidy up sim.py" PR based on that. Once it's done, I'll link to the PR here.

The other changes, I'm a little less convinced about: I don't quite understand the XLEN change you're suggesting and I think I disagree with the Makefile change.

If you think the Makefile change is good, could you rebase it onto master and split it into a separate PR? It's not related to Questa compilation warnings, so I think it makes sense to consider separately.

vendor/google_riscv-dv/src/riscv_pmp_cfg.sv
danghai: I think It is range from [0:52]. I recently check the https://github.com/google/riscv-dv. It already is fixed but It does not integrate to ibex yet. So, I will remove this commit.

danghai:
@rswarbrick , I remove 2 commits. Thanks

rswarbrick:
Oh, sorry, I forgot to link to the sim.py tidyup PR #648. The first patch there is cherry-picked from this PR's original set.

danghai:
@rswarbrick . It looks good to me.

PR #642, opened 2020-03-02
Verilator warnings

rswarbrick:
Fix a few last verilator warnings. The first two patches are just waivers. The third seems to be an actual bug.

### DIFF
--- a/dv/riscv_compliance/lint/verilator_waiver.vlt
+++ b/dv/riscv_compliance/lint/verilator_waiver.vlt
@@ -1,0 +1,32 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Lint waivers for processing riscv_compliance RTL with Verilator
+//
+// This should be used for rules applying to things like testbench
+// top-levels. For rules that apply to the actual design (files in the
+// 'rtl' directory), see verilator_waiver_rtl.vlt in the same
+// directory.
+//
+// See https://www.veripool.org/projects/verilator/wiki/Manual-verilator#CONFIGURATION-FILES
+// for documentation.
+//
+// Important: This file must included *before* any other Verilog file is read.
+// Otherwise, only global waivers are applied, but not file-specific waivers.
+
+`verilator_config
+
+// We have some boolean top-level parameters in e.g. simple_system.sv.
+// When building with fusesoc, these get set with defines like
+// -GRV32M=1 (rather than -GRV32M=1'b1), leading to warnings like:
+//
+//   Operator VAR '<varname>' expects 1 bits on the Initial value, but
+//   Initial value's CONST '32'h1' generates 32 bits.
+//
+// This signoff rule ignores errors like this. Note that it only
+// matches when you set a 1-bit value to a literal 1, so it won't hide
+// silly mistakes like setting it to 2.
+//
+lint_off -rule WIDTH -file "*/rtl/ibex_riscv_compliance.sv"
+         -match "*expects 1 bits*Initial value's CONST '32'h1'*"
--- a/dv/riscv_compliance/rtl/riscv_testutil.sv
+++ b/dv/riscv_compliance/rtl/riscv_testutil.sv
@@ -90,8 +90,17 @@ module riscv_testutil (
     end
   end
 
-  // only word writes are supported
-  assign dev_err_o = (~dev_we_i | dev_be_i != 4'hf) & dev_req_i;
+  // The interface is write-only, and only supports 32-bit writes. If
+  // either of these checks fails, raise dev_err_o on the next cycle.
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      dev_err_o <= 1'b0;
+    end else begin
+      dev_err_o <= (~dev_we_i | dev_be_i != 4'hf) & dev_req_i;
+    end
+  end
+
+  // Since the interface is write-only, tie rdata to 0.
   assign dev_rdata_o = 32'h0;
 
 
--- a/dv/riscv_compliance/rtl/riscv_testutil.sv
+++ b/dv/riscv_compliance/rtl/riscv_testutil.sv
@@ -106,8 +106,6 @@ module riscv_testutil (
   logic [31:0] read_addr_d, read_addr_q;
   always_comb begin
     state_d = state_q;
-    host_req_o = 1'b0;
-
     unique case (state_q)
       WAIT: begin
         if (read_signature_and_terminate) begin
@@ -119,8 +117,6 @@ module riscv_testutil (
       end
 
       READ: begin
-        host_req_o = 1'b1;
-        host_addr_o = read_addr_q;
         if (host_gnt_i) begin
           read_addr_d = read_addr_q + 4;
           if (read_addr_d == end_signature_addr_q) begin
@@ -141,6 +137,11 @@ module riscv_testutil (
       end
     endcase
   end
+
+  // These are the address and read request bits, respectively of the
+  // TestUtilHost master port.
+  assign host_addr_o = read_addr_q;
+  assign host_req_o = (state_q == READ);
 
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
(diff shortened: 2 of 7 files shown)


dv/riscv_compliance/rtl/riscv_testutil.sv
```diff
@@ -90,8 +90,17 @@ module riscv_testutil (
     end
   end
 
-  // only word writes are supported
-  assign dev_err_o = (~dev_we_i | dev_be_i != 4'hf) & dev_req_i;
+  // The interface is write-only, and only supports 32-bit writes. If
+  // either of these checks fails, raise dev_err_o on the next cycle.
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+	if (! rst_ni) begin
```
imphil: Tabs! :smile:

rswarbrick: Oops. Fixed! (as is my editor...)

dv/riscv_compliance/rtl/riscv_testutil.sv
```diff
@@ -142,6 +147,11 @@ module riscv_testutil (
     endcase
   end
 
+  // These are the address and read request bits, respectively of the
+  // TestUtilHost master port.
+  assign host_addr_o = read_addr_q;
+  assign host_req_o = (state_q == READ);
```
GregAC: Any particular reason these got moved out of the alway_comb block above?

host_addr_o did need a default assignment inside the block but otherwise I don't think this has altered the functioning of these signals?

lint/verilator_waiver.vlt
```diff
@@ -85,3 +85,19 @@ lint_off -rule UNUSED -file "*/rtl/ibex_decoder.sv" -match "*instr_alu*"
...
+// -GRV32M=1 (rather than -GRV32M=1'b1), leading to warnings like:
+//
+//   Operator VAR '<varname>' expects 1 bits on the Initial value, but
+//   Initial value's CONST '32'h1' generates 32 bits.
+//
+// This signoff rule ignores errors like this. Note that it only
+// matches when you set a 1-bit value to a literal 1, so it won't hide
+// silly mistakes like setting it to 2.
+
+lint_off -rule WIDTH -file "*/rtl/ibex_simple_system.sv"
```
GregAC: I think we want to keep verilator_waiver.vlt for waivers relating specifically to the Ibex RTL, waivers for things like simple_system should live elsewhere (e.g. under examples/simple_system), though I'm uncertain if multiple waiver files are supported.

Was the normal lint check flagging up issues in simple_system?

imphil: quick answer here: yes, multiple waivers should work just fine.

dv/riscv_compliance/rtl/riscv_testutil.sv
rswarbrick: That's the point of the patch. It's explained in the commit message, but I'll reproduce it here:

    Avoid two combinatorial loop warnings in riscv_compliance suite
    
    We do this by pulling the definition of host_addr_o and host_req_o out
    of an always_comb process in riscv_testutil.sv.
    
    When set inside the process, Verilator warns about a combinatorial
    loop. This happens because a read request could go out on the bus and
    appear again (combinatorially) on the slave interface, setting
    read_signature_and_terminate. This doesn't actually happen (because
    read_signature_and_terminate only takes effect when we are in state
    WAIT), but Verilator's sensitivity tracking isn't fine-grained enough
    to notice.

lint/verilator_waiver.vlt
rswarbrick: Ok, I'll change this.

rswarbrick: Hmm, does this mean we need a separate signoff file for ibex_riscv_compliance too? It's a little sad, because the same signoff (with a different filename) will appear in both.

imphil: There are no strict rules. But it would be nice to keep the core ibex one separated from all the other ones. I wouldn't mind having all the other waivers in a shared file.

dv/riscv_compliance/rtl/riscv_testutil.sv
GregAC: Makes sense, I don't think Verilator could do a better job here if it views each process as its own blob of logic with inputs and outputs and does the loop detection at that level, read_signature_and_terminate is an input, host_req_o/host_addr_o is an output so there's a loop involving them (it's only when you turn it into gates you find the loop doesn't exist).

lint/verilator_waiver.vlt
GregAC: Well I don't think we strictly need the lint waivers, the actual lint check (as used by the CI) only looks at Ibex RTL so presumably these waivers are only about suppressing warnings when you do a normal build?

rswarbrick: You're right that these won't cause a lint check to fail, but I believe that one should never see warnings when building stuff unless something has gone wrong. (Yes, I know EDA tools fail at this terribly, but that's no reason for us to be rubbish too! :-) )

GregAC: Fair enough, happy to do that.

A single shared file for everything that isn't core ibex RTL should be fine for now. We can revisit that decision if it starts getting messy.

How shall we name things perhaps:

* lint/verilator_waiver_rtl.vlt
* lint/verilator_waiver_other.vlt

With some comments at the top of each explaining what should be in the file?

rswarbrick: Cool, sounds good to me.

rswarbrick:
Right, I've split the waiver file in two. I've done that in a separate patch from the "waive last warning" patch, since I think the result is a bit cleaner. Of course, that means that the initial version of the "verilator_waiver_other.vlt" file doesn't actually contain any rules. I don't think that really matters though.

ibex_core.core
```diff
@@ -77,7 +82,7 @@ parameters:
 targets:
   default:
     filesets:
-      - tool_verilator ? (files_lint_verilator)
+      - tool_verilator ? (files_lint_verilator_all)
```
imphil: Only include the `verilator_waiver.vlt` here (as it was before).

The other waiver needs to be included in the other systems. If you want to keep it in the current directory (`lint/verilator_waiver_other.vlt`), you need to create a separate `lint/xxx.core` file for it with a core named e.g. `lowrisc:ibex:waivers_other`, and depend on it in the simple system and compliance test cores). (You cannot have a file path starting with `..` in a core file.)

rswarbrick: Ah, right, I understand. I've just pushed up another version of the patch set. The first commit adds waiver files in `examples/simple_system/lint` and `dv/riscv_compliance/lint` and wires everything up. I hope I've got the fusesoc incantations right?

dv/riscv_compliance/rtl/riscv_testutil.sv
```diff
@@ -90,8 +90,17 @@ module riscv_testutil (
     end
   end
 
-  // only word writes are supported
-  assign dev_err_o = (~dev_we_i | dev_be_i != 4'hf) & dev_req_i;
+  // The interface is write-only, and only supports 32-bit writes. If
+  // either of these checks fails, raise dev_err_o on the next cycle.
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (! rst_ni) begin
```
imphil:
```suggestion
    if (!rst_ni) begin
```

dv/riscv_compliance/lint/verilator_waiver.vlt
```diff
@@ -0,0 +1,32 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Lint waivers for processing non-core RTL with Verilator
```
imphil: Maybe

```suggestion
// Lint waivers for processing riscv_compliance RTL with Verilator
```

examples/simple_system/lint/verilator_waiver.vlt
```diff
@@ -0,0 +1,32 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Lint waivers for processing non-core RTL with Verilator
```
imphil: Suggestion as above.

dv/riscv_compliance/rtl/riscv_testutil.sv
rswarbrick: Hi Greg, I don't think I can land this patch at the moment because it's marked as you requesting changes. Do you think there's anything else that needs doing?

PR #643, opened 2020-03-02
Make exiting from simple_system tests work with Spike

rswarbrick:


shared/rtl/sim/simulator_ctrl.sv
```diff
@@ -33,7 +39,7 @@ module simulator_ctrl #(
 );
 
   localparam CHAR_OUT_ADDR = 0;
-  localparam SIM_CTRL_ADDR = 1;
+  localparam SIM_CTRL_ADDR = 2;
```
GregAC: Please update the examples/simple_system/README.md file as well with the new address

examples/sw/simple_system/common/link.ld
```diff
@@ -19,6 +19,16 @@ _stack_start   = ORIGIN(stack) + LENGTH(stack);
...
 
+/* The tohost address is used by Spike for a magic "stop me now" message. This
+   is set to equal SIM_CTRL_CTRL (see simple_system_regs.h), which has that
+   effect in simple_system simulations. Note that it must be 8-byte aligned.
+
+   We don't read data back from Spike, so fromhost is set to some dummy value:
+   we place it just above the top of the stack.
+ */
+tohost   = 0x20008;
+fromhost = _stack_start + 0x10;
```
GregAC: So Spike may write something to `fromhost`?

rswarbrick: I think that Spike supports some form of host <-> target protocol for debugging using OpenOCD. I guess that this is how they send information in the other direction.

rswarbrick: I think that Spike supports some form of host <-> target protocol for debugging using OpenOCD. I guess that this is how they send information in the other direction.

shared/rtl/sim/simulator_ctrl.sv
rswarbrick: Good catch! Done.

PR #645, opened 2020-03-02
Add a wrapper script to run a simple_system binary under Spike

rswarbrick:
The main point of this is that it's somewhere we can store what it
calls "ss_opts": the options that we must pass to Spike to tell it the
memory layout of the simple_system environment.

While we're at it, I've tried to make this reasonably pleasant to use.
So you can do

   spike-simple-system.sh <elf-file>

just to run the thing. To pass stuff to Spike without the
bash script getting in the way, use '--':

   spike-simple-system.sh -- my option -h --here <elf-file>

### DIFF
--- a/examples/simple_system/spike-simple-system.sh
+++ b/examples/simple_system/spike-simple-system.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+
+set -u
+set -e
+
+# This is a simple bash script to allow you to run a binary compiled
+# for the simple_system environment using Spike.
+#
+
+error() {
+    echo >&2 "$@"
+    exit 1
+}
+
+usage () {
+    echo >&2 "Usage:  spike-simple-system [-h] [-v] [-n] [--] [options] <elf>"
+    echo >&2
+    echo >&2 "  Where <elf> is the path to an ELF binary that was"
+    echo >&2 "  compiled for the simple_system environment. Any arguments "
+    echo >&2 "  in <options> will be passed to Spike after those to set up "
+    echo >&2 "  the environment for simple_system tests."
+    echo >&2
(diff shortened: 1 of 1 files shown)


examples/simple_system/spike-simple-system.sh
```diff
@@ -0,0 +1,108 @@
...
+    echo >&2 "  Where <elf> is the path to an ELF binary that was"
+    echo >&2 "  compiled for the simple_system environment. Any arguments "
+    echo >&2 "  in <options> will be passed to Spike after those to set up "
+    echo >&2 "  the environment for simple_system tests."
+    echo >&2
+    echo >&2 "  If the -v argument is passed, the Spike command will be "
+    echo >&2 "  echoed to stderr before it runs. If the -n argument is passed "
+    echo >&2 "  (which implies -v), we don't actually run the command."
+    echo >&2
+    echo >&2 "  This will write the Spike instruction log to stderr."
```
GregAC: Is it spike that writes instruction log to stderr or can we choose where it goes? 

For a standalone run at least it'd be useful to redirect it to a file. You can of course do this with an stderr redirect when you run the script but passing a filename as a script option would be a bit easier. It would then be nice if any actual errors go to stderr whilst instruction log doesn't (i.e. if you run this script from a terminal and tell it to log instruction trace somewhere if spike encounters an error that it writes to stderr you want to see that appear on the terminal rather than inside the instruction log).

rswarbrick: Yes, this is Spike. I'm going to merge this commit now and have a poke around with Spike tomorrow to see if I can convince it to generate its output somewhere less obnoxious.

PR #646, opened 2020-03-02
Improvements in ibex_log_to_trace_csv.py

rswarbrick:
This is a sequence of patches spurred on by the work I'm doing trying to make it easy to run simple_system binaries with Spike and then compare trace logs with Ibex.

The first patch fixes a simple bug (it wasn't possible to run the Python script except from directories exactly two components below the ibex root).

The second patch makes the command line interface a bit nicer (relevant to me, because that's how I'm running it!). Now, running without any arguments converts data from stdin to stdout (rather than exploding with a mysterious error).

The final two patches tidy things up in general, getting the code PEP8 compliant and simplifying log checking. Honestly, this is mostly just indentation changes, but there were also some dead variables to delete and I made imports explicit (which allows static analysis tools to spot when you mis-speled a variable name).

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -145,30 +145,36 @@ def check_ibex_uvm_log(uvm_log, core_name, test_name, report, write=True):
       signature
 
     """
-    pass_cnt = 0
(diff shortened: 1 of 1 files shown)


udinator:
These changes all look good to me!
One request: since you modified the indentation of this python script to 4-spaces to match PEP8 standard, whenever you have a bit of time could you open a PR with the same indentation updates to the dv/uvm/core_ibex/sim.py?

PR #648, opened 2020-03-03
Sim.py cleanups

rswarbrick:
Here's a big pile of patches to sim.py, making it PEP8 compliant and then (more interestingly) doing some simple refactors and cleanups. It's now a bit longer, but this is mostly a pile of comments explaining what the code does etc.

The patches also slightly change the behaviour of a couple of arguments. I've fixed up the Makefile to match, and think that this makes it a bit cleaner in general. (For example, boolean settings now get set with --foo rather than --foo=1).

The other big change is to how we actually run simulations. I suspect that this didn't use to work with LSF and my patches might well fix that. I don't have access to an LSF cluster, so can't check, but it looks a bit more sensible, at least.


dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,512 @@
...
+    # Compile the TB
+    logging.info("Compiling TB")
+    for cmd in compile_cmds:
+        cmd = subst_vars(cmd,
+                         {
+                             'out': output_dir,
+                             'cmp_opts': opts
+                         })
+        logging.debug("Compile command: %s" % cmd)
+        run_cmd(cmd)
```
udinator: Should `lsf_cmd`be prepended to the compile command?

dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,512 @@
...
+        if entry.get('tool') == simulator:
+            return entry
+
+    raise RuntimeError("Cannot find RTL simulator {}".format(simulator))
+
+
+def get_simulator_cmd(simulator, yaml_path, enables):
+    '''Get compile and run commands for the testbench
+
+    simulator is the name of the simulator to use. yaml_path is the path to a
```
udinator: Just a more nitty comment: could you retain a similar formatting of the function argument/return value docstrings to how it was before, with a colon mapping each argument to its description?

dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,512 @@
...
+    simulator (currently supported: spike and ovpsim). output_dir is the base
+    output directory (which should contain logs from both the ISS and the test
+    run itself). report is the path to the regression report file we're
+    writing.
+
+    Returns True if the test run passed and False otherwise.
+
+    '''
+    test_name = test['test']
+    elf = os.path.join(output_dir, '{}.{}.o'.format(test_name, idx))
```
udinator: I think this should be `os.path.join(output_dir, 'instr_gen/asm_tests/{}.{}.o'.format(test_name, idx))`

dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,512 @@
...
+
+    logging.info("Comparing %s/DUT sim result : %s" % (iss, elf))
+
+    with open(report, 'a') as report_fd:
+        report_fd.write('Test binary: {}\n'.format(elf))
+
+    out_dir = os.path.join(output_dir, 'rtl_sim',
```
udinator: try naming this `rtl_dir`, otherwise this corrupts the path of the ISS files below and will error out of the post_compare step

dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,512 @@
...
+                           '{}.{}'.format(test_name, idx))
+
+    rtl_log = os.path.join(out_dir, 'trace_core_00000000.log')
+    rtl_csv = os.path.join(out_dir, 'trace_core_00000000.csv')
+    uvm_log = os.path.join(out_dir, 'sim.log')
```
udinator: as above, you would then use `rtl_dir` as the first argument to os.path.join(...)

dv/uvm/core_ibex/sim.py
rswarbrick: That happens at the bottom of get_test_sim_cmd

dv/uvm/core_ibex/sim.py
rswarbrick: Yes, you're completely right. Will fix.

dv/uvm/core_ibex/sim.py
rswarbrick: Yep, you're right. Actually, I think the ISS files below should have been based on `output_dir` anyway, but the similar sounding names are icky, so I'll fix it.

dv/uvm/core_ibex/sim.py
rswarbrick: Done.

dv/uvm/core_ibex/sim.py
rswarbrick: Re the comment formatting, is this an internal Google style? I don't think I've seen it before. We can do that if you think it's important, but it will make the comments longer and I don't think they become any more informative. Since tidying up the patch set properly would take half an hour's work or so, I'm going to hold off on this for now: if you think this needs doing, let me know and I'll update the patches.

udinator: Yep, the comment formatting is as per the open-source Google Python coding style guide, in particular [this section](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods).
This function docstring formatting isn't the most important thing, as you already are providing some explanations of the arguments and so on, so I'm ok if you just leave it as is for this PR, and we can just leave this for some time in the future.

udinator:
> Hi Udi! Thank you very much for the careful review. I'm feeling a little stupid about the messed-up directory variables: I thought I'd tested it, but the test in question was failing with UVM errors, so never got as far as the ISS comparison. Oops!

No worries! Out of curiosity, which test was failing?

> 
> It's a bit difficult for me to run VCS today, so I'll give this patch set a proper test and shake out any remaining bugs I find tomorrow morning: please don't feel that you have to look at it again until after that.

No problem at all, thanks for addressing the comments!

dv/uvm/core_ibex/sim.py
rswarbrick: Ah, good point. Although, in my defence, this is true about the version on master too! :-) I'll fix that up and add a last patch to the set.

rswarbrick:
> > [...] I thought I'd tested it, but the test in question was failing with UVM errors, so never got as far as the ISS comparison. Oops!
> 
> No worries! Out of curiosity, which test was failing?

I see failures in riscv_debug_branch_jump_test and riscv_irq_in_debug_mode_test. I think I'll probably have a look into them this afternoon.

dv/uvm/core_ibex/sim.py
```diff
@@ -1,301 +1,529 @@
...
+                raise RuntimeError('Cannot substitute {} in command because '
+                                   'the environment variable ${} is not set.'
+                                   .format(needle, env_var))
+            string = string.replace(needle, value)
+
+    return string
+
+
+def subst_cmd(cmd, enable_dict, opts_dict, env_vars):
+    '''Substitute options and environment variables in cmd
```
danghai: I think you should put the description for argument list

rswarbrick: I'm sorry, I don't understand your message. I believe that every argument is (quite carefully) documented in this function's docstring.

danghai: I mean it has description for arguments (Ex cmd, enable_dict, opts_dict, env_vars). For example:
https://github.com/lowRISC/ibex/blob/master/vendor/google_riscv-dv/run.py

rswarbrick: Do you believe that the three following paragraphs of the docstring fail to document one of the arguments to the function? In which case, please could you say which one?

udinator: @danghai, are you referring to the format of the comment docstrings?

udinator:
@rswarbrick, regarding the two failing tests you are seeing, the debug_branch_test is a known issue on my end, and has to do solely with the riscv-dv generator, and it is on my TODO list to fix the generator to get this test working.
The irq_in_debug_mode_test I am not sure about however, I'll take a look at it as well.

dv/uvm/core_ibex/sim.py
danghai: @udinator yes

rswarbrick:
@udinator: Is there anything else that you think should be changed before this PR can land?

dv/uvm/core_ibex/sim.py
udinator: @danghai - I will open a PR in the near future to update the docstring formatting.

udinator:
> @udinator: Is there anything else that you think should be changed before this PR can land?

@rswarbrick , nope, this PR LGTM. Just some quick nitty things for future PRs though, could you make sure to:
1) keep PRs to a single commit? Just to reduce reviewer time as well as to clearly differentiate changes from each other
2) could you make sure to keep any function docstrings (for Python/scripting code) aligned with the format defined in the Google open-source Python style guide? For this PR it's not a big deal, I'll open a PR in the near future to update the formatting, but just something to keep in mind.
Thanks for all the effort!

rswarbrick:
@udinator: Thanks for the review. Re point (1), I don't believe that squashing commits together is a good idea (it makes bisection much less useful!), but I'll try to make smaller related PRs in future instead of a single PR for a big patch set.

udinator:
> @udinator: Thanks for the review. Re point (1), I don't believe that squashing commits together is a good idea (it makes bisection much less useful!), but I'll try to make smaller related PRs in future instead of a single PR for a big patch set.

Cool, that would be great! Just because of the time difference, relatively smaller PRs are more likely to get approved and merged much more quickly.

Issue #652, opened 2020-03-04
gdb support


jrrk2:
By default the ibex repository does not include riscv-dbg (gdb) support. Such support is available within the opentitan repository as part of a much larger project. I have modified the basic ibex example to include gdb support for a total of 4001 LUTs, by crudely copying relevant modules from opentitan and modifying to suit the original example style. I believe this contribution would be attractive to potential commercial customers who will naturally assume debug support as a prerequisite. The question is, how to format these changes to ease merging with the no debug ibex FPGA example (tested on Arty-A7-A100)

[my fork](https://github.com/lowRISC/ibex/pull/654)

imphil:
Thanks for your work on this @jrrk2. 

I'm not convinced the Ibex repository is the right place to put such an example. Adding debug support needs much more SoC infrastructure than we have currently, including an extended "bus" implementation. I'd therefore leave the simple system as simple as it is for now, and redirect people to more fully-featured SoC implementations if they are looking for examples on how to integrate a RISC-V Debug Implementation with Ibex, with OpenTitan and (probably) PULPissimo being two examples for that.

Maybe it would be a viable path forward to link these sample integrations in the docs somewhere? And maybe we can also include a more minimal Ibex example that you put together in #654 that you keep in a personal repository on GitHub?

jrrk2:
Closing this issue because the solution is discussed in https://github.com/lowRISC/ibex/pull/654

PR #654, opened 2020-03-04
Support gdb debugging in base ibex example

jrrk2:
The base ibex example does not include riscv-dbg (gdb) support. This patch (which probably isn't minimal) provides a solution for gdb on the arty-a7 board. This board requires an external Olimex adapter to avoid hacking the riscv tap length and register numbers. Also implemented: changes to prevent out of range warnings in Vivado simulation of PLLE2_ADV primitive, and a simple testbench (top_tb.sv).


imphil:
Thanks for this PR @jrrk2. Let's first figure out if we want to go down that route in #652, and then come back to this PR.

jrrk2:
Somehow I submitted two pull requests, it should be just one (the simplified one).

imphil:
@jrrk2 Given the amount of code needed I'm not convinced that the Ibex repository is the right place for an exemplary debug system integration. I've outlined more of that thinking in https://github.com/lowRISC/ibex/issues/652#issuecomment-596545822.

I agree that we need better ways to point users to showcases of how Ibex can be integrated with a debug system. The OpenTitan repository can serve as that example, as does the pulpissimo repository. I've created https://github.com/lowRISC/ibex/pull/759 to document this aspect better, please let me know there if there are things we should add to the documentation.

Thanks for your contribution to Ibex. Even though we couldn't merge the PR ultimately, I'm sure others will find it a useful point of reference for their work.

jrrk2:
Hi Philipp,
  I understand what you are saying, but it’s quite difficult to use opentitan as an exemplar of how to do things unless you buy into the entire infrastructure. This is because register generation, for example, happens at chip level, it precludes reusing individual components. The pulpissimo is non functional at the moment due to some bitrot in sub-modules.

Regards,
Jonathan

PR #656, opened 2020-03-05
Loop fix

GregAC:
Two commits on this, one is a lint waiver fix that is unrelated to the change but without it I'm getting lint error locally, maybe a Verilator version issue (I'm using 4.028) and otherwise I don't know how it got past CI in the first place?

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -116,7 +116,8 @@ module ibex_controller (
   logic illegal_dret;
   logic illegal_umode;
   logic exc_req_lsu;
-  logic special_req;
+  logic special_req_all;
+  logic special_req_branch;
   logic enter_debug_mode;
   logic ebreak_into_debug;
   logic handle_irq;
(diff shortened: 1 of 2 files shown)


rtl/ibex_controller.sv
```diff
@@ -187,10 +188,29 @@ module ibex_controller (
...
+
+  // generic special request signal, applies to all instructions
+  assign special_req_all = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
       exc_req_d | exc_req_lsu;
 
+  // special request that can specifically occur during branch instructions
+  assign special_req_branch = (illegal_insn_d | instr_fetch_err) & (ctrl_fsm_cs != FLUSH);
+
+  `ASSERT(SpecialReqBranchGivesSpecialReqAll,
+    special_req_branch |-> special_req_all);
```
tomroberts-lowrisc: Not sure but I think the semicolons are included in the macro - we had an issue about duplicated semicolons recently

GregAC: Ah yes well spotted.

GregAC:
Further thought will be required for the third pipe stage. Potentially even for the non error case as the branch will wait until the load/store is resolved until executing, so there's probably a combinational path from data_rvalid_i to pc_set because of it.

I had previously been looking at this area for timing reasons, the illegal instruction signal takes a while to compute so factoring it into pc_set was annoying. I thought we could do something like allow the pc set to happen then just flush things the following cycle. We may end up doing something like that for the third pipe stage. Assume load/store will complete the cycle the branch is being computed and do the PC set. If we're waiting then we've handily prefetched the branch, if we get an error response then flush everything.

tomroberts-lowrisc:
Makes sense - there's no problem following the branch then taking the exception in the next cycle, since it only happens on errors.

PR #659, opened 2020-03-06
[ml] add condensed test targets for ML (and constraints)

udinator:
Currently only the riscv_rand_test target fully works with RTL simulation, the other two test targets both fail RTL simulation due to some testbench error, which I think was introduced by a previous commit, will debug this issue.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ml_testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ml_testlist.yaml
@@ -30,6 +30,29 @@
 # gcc_opts        : gcc compile options
 # --------------------------------------------------------------------------------
 
+
+# --------------------------------------------------------------------------------
+# ML Parameter Constraints - riscv_rand_test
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/riscv_dv_extension/ml_testlist.yaml
```diff
@@ -65,3 +114,196 @@
...
+    +stream_name_10=riscv_load_store_stress_instr_stream
+    +stream_freq_10=4
+    +dist_control_mode=1
+    +dist_shift=10
+    +dist_arithmetic=10
+    +dist_logical=10
+    +dist_compare=10
+    +dist_branch=10
+    +dist_synch=10
+    +dist_csr=10
```
taoliug: All distribution config can be removed now, they are obsolete

udinator: Done.

PR #666, opened 2020-03-09
[rtl] IF stage timing fix

tomroberts-lowrisc:
- ready_i input to the prefetch buffer factored both it's own valid_o
  output and the pc_set branch signal, neither of which are required.
- Refactoring the ready_i signal to just id_in_ready_i improves timing
  significantly for the icache.
- Also removed offset_in_init signal which appeared to serve no purpose.

### DIFF
--- a/rtl/ibex_if_stage.sv
+++ b/rtl/ibex_if_stage.sv
@@ -75,7 +75,8 @@ module ibex_if_stage #(
 
   import ibex_pkg::*;
 
-  logic              have_instr;
+  logic              instr_valid_id_d, instr_valid_id_q;
+  logic              instr_new_id_d, instr_new_id_q;
 
   // prefetch buffer related signals
   logic              prefetch_busy;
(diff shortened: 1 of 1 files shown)


rtl/ibex_if_stage.sv
```diff
@@ -186,35 +186,40 @@ module ibex_if_stage #(
...
   );
 
-  // IF-ID pipeline registers, frozen when the ID stage is stalled
-  assign if_id_pipe_reg_we = have_instr & id_in_ready_i;
+  // The ID stage becomes valid as soon as any instruction is registered in the ID stage flops.
+  // Note that the current instruction is squashed by the incoming pc_set_i signal.
+  // Valid is held until it is explicitly cleared (due to an instruction completing or an exception)
+  assign instr_valid_id_d = (fetch_valid & id_in_ready_i & ~pc_set_i) |
+                            (instr_valid_id_q & ~instr_valid_clear_i);
+  // Record when a new instruction enters the ID stage (only used for RVFI signalling).
```
vogelpi: The part "(only used for RVFI signaling)" of this comment is not true, this signals becomes the write enable for the actual ID instr flops below.

tomroberts-lowrisc: Good point - moved the comment to the output where it is actually relevant.
Thanks!

Issue #670, opened 2020-03-09
Questa: Defaulting port to var rather than wire


imphil:
(Spin-off from #622 by @danghai)

Questa reports:

> ** Warning: /home/danghai/hhoangda/ibex/rtl/ibex_controller.sv(60): (vlog-13314) Defaulting port 'irqs_i' kind to 'var' rather than 'wire' due to default compile option setting of -svinputport=relaxed.

(We get a similar message for other ports where we have a `struct` as input port.)

The code in question is:

```verilog
// in ibex_controller.sv
input  ibex_pkg::irqs_t       irqs_i,


// in ibex_pkg.sv
// Interrupt requests
typedef struct packed {
  logic        irq_software;
  logic        irq_timer;
  logic        irq_external;
  logic [14:0] irq_fast; // 15 fast interrupts,
                         // one interrupt is reserved for NMI (not visible through mip/mie)
} irqs_t;
```

I don't see anything wrong with this code, it looks like the warning is not helping here. 

@danghai can you run `verror 13314` in Questa to learn more about the error? It can probably be silenced unless the output shows something that we should fix on our side.

PR #677, opened 2020-03-10
Resolve warning Questa: Defaulting port to var rather than wire

danghai:
Relate #670 .
verror 13314:

```
vlog Message # 13314:
An input port that is declared with a type but without the 'var' keyword
should default to 'wire' for strict LRM conformance. The default compile
option of -svinputport=relaxed is not LRM compliant, and only defaults
4-state scalar or 4-state single dimension vector types to 'wire'.
Consider using -svinputport=net for strict conformance. See the vlog -help
text for other behaviors available with the -svinputport option.
```

The warning messages are removed when I add option `-svinputport=relaxed`


dv/uvm/core_ibex/yaml/rtl_simulation.yaml
```diff
@@ -110,6 +110,7 @@
   compile:
     cmd:
       - "qrun -f ibex_dv.f -uvmhome uvm-1.2
+              -svinputport=relaxed
```
imphil: Can you try `-svinputport=net` here to see if the strict (LRM-compliant) mode works as well? I think it should, but it's easy to get lost in the SV spec.

.gitignore
```diff
@@ -23,3 +23,6 @@ __pycache__
 # This is the default output directory in dv/uvm/core_ibex and
 # contains auto-generated files from building and running tests.
 /dv/uvm/core_ibex/out
+
+# This is generated by Questa tool when running DV simulations
+/dv/uvm/core_ibex/modelsim.ini
```
imphil: This probably can be made slightly wider to catch a `modelsim.ini` anywhere in the tree.

```suggestion
modelsim.ini
```

rswarbrick: In case you're wondering about the `ucli.key` rule above, that's probably too specific too (I wrote it and haven't got around to changing it).

danghai:
I try to run the full flow in `Makefile`. I got the result so far:
[regr.log](https://github.com/lowRISC/ibex/files/4313779/regr.log)
Compile and Optimize log:
[qrun_compile_optimize.log](https://github.com/lowRISC/ibex/files/4313790/qrun_compile_optimize.log)
1 sample Sim log (all are the same, 2 warning messages)
[sim.log](https://github.com/lowRISC/ibex/files/4313794/sim.log)

dv/uvm/core_ibex/yaml/rtl_simulation.yaml
danghai: It has the same as result. So I changed it by `svinputport=net`

.gitignore
danghai: @rswarbrick what do you mean `ucli.key`?

PR #680, opened 2020-03-10
[rtl] Fixes for single-cycle mutiply

GregAC:
* single cycle multiplier configuration needlessly stalled for a cycle
* mult_hold wasn't being set in mult state machine in single cycle
  multiplier

### DIFF
--- a/rtl/ibex_multdiv_fast.sv
+++ b/rtl/ibex_multdiv_fast.sv
@@ -111,6 +111,10 @@ module ibex_multdiv_fast #(
     end
   end
 
+  `ASSERT_KNOWN(DivEnKnown, div_en_internal);
+  `ASSERT_KNOWN(MultEnKnown, mult_en_internal);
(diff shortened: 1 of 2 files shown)


GregAC:
Looks like CI hasn't properly triggered here? Only seeing the docs side of things

rtl/ibex_multdiv_fast.sv
```diff
@@ -111,6 +111,10 @@ module ibex_multdiv_fast #(
     end
   end
 
+  `ASSERT_KNOWN(DivEnKnown, div_en_internal);
+  `ASSERT_KNOWN(MultEnKnown, mult_en_internal);
+  `ASSERT_KNOWN(MultDivEnKnown, multdiv_en);
```
vogelpi: These assertions could actually be moved to the bottom (they are not inside a generate block).

GregAC: Personally I prefer to keep assertions close to where the signals they concern are used

PR #684, opened 2020-03-11
[rtl/sw] Add multiply and divide wait counters

GregAC:


GregAC:
Yup I always forget about that one!

PR #694, opened 2020-03-12
Add missing space after code-block directive

rswarbrick:
Without this, Sphinx spits out a warning message and swallows the
block.


imphil:
Ugh, I should have seen that. Thanks for catching that!

PR #695, opened 2020-03-12
Notes on the ICache specification

rswarbrick:
This also adds a couple of comments splitting up the ports in
ibex_icache.sv that I found helpful when working out what everything
did.

### DIFF
--- a/doc/icache.rst
+++ b/doc/icache.rst
@@ -9,7 +9,25 @@ NOTE - This module is currently DRAFT
 The optional Instruction Cache (I$) is designed to improve CPU performance in systems with high instruction memory latency.
 The I$ integrates into the CPU by replacing the prefetch buffer, interfacing directly between the bus and IF stage.
 
-For details of the memory and IF stage interfaces see :ref:`instruction-fetch`.
+High-level operation
+--------------------
+
+The I$ passes instructions to the core using a ready / valid interface.
+Inside the cache is an address counter, which increments for every instruction fetched (by 2 or 4 bytes, depending on whether the instruction contents show it to be compressed).
+When the core takes a branch, it resets the counter to a new address by raising the ``branch_i`` signal and supplying the new address on ``addr_i``.
+The next instruction returned by the cache will be the instruction at this new address.
+
+The I$ communicates with instruction memory using an interface that matches the IF stage (allowing the cache to be enabled or disabled without needing to change the Ibex toplevel's interface). For more details of this interface, see :ref:`instruction-fetch`.
+
+To avoid the cache fetching needlessly when the core is asleep (after a ``wfi`` instruction), it has a ``req_i`` input. Shortly after this goes low, the cache will stop making memory transactions.
+
+If the ``icache_enable_i`` input is low, the cache operates in pass-through mode, where every requested instruction is fetched from memory and no results are cached.
+
+In order to invalidate the cache, the core can raise the ``icache_inval_i`` line for one or more cycles, which will start an internal cache invalidation.
+No fetches are cached while the invalidation is taking place (behaving as if ``icache_enable_i`` is low).
+
+While the I$ is busy, either (pre)fetching data or invalidating its memory, it raises the ``busy_o`` signal.
+This can be used to avoid the cache's clock being gated when it is doing something.
 
 
 Configuration options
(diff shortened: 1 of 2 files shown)


doc/icache.rst
```diff
@@ -149,3 +167,53 @@ Cache invalidation
...
+This isn't an attempt to describe the cache's performance characteristics.
+
+The I$ has a single clock (``clk_i``) and asynchronous reset (``rst_ni``).
+
+Data is requested from the instruction memory with the ports prefixed by ``instr_``. These work as described in :ref:`instruction-fetch`.
+Note that there's one extra port on the I$, which doesn't appear at the ``ibex_core`` top-level.
+This is ``instr_pmp_err_i``.
+If the PMP block disallows a fetch for a certain address, it will squash the outgoing memory request entirely and set ``instr_pmp_err_i``.
+If that happens, the cache behaves as if the request was granted but returned some time later with an error.
+Note that it is possible for ``instr_gnt_i`` and ``instr_pmp_err_i`` to be high on the same cycle (if the memory system ties ``instr_gnt_i`` to ``instr_req_o``, for example).
```
tomroberts-lowrisc: This is kind-of the other way around.
It's possible `instr_gnt_i` will never go high on a PMP error if the thing connected downstream ties `instr_gnt_i` to `instr_req_o`. On the other hand, if `instr_gnt_i` is wired up more like a conventional ready signal, it's possible that `instr_gnt_i` and `instr_pmp_err_i` would be high at the same time.

doc/icache.rst
```diff
@@ -115,7 +133,7 @@ The fill buffers perform several functions in the I$ and constitute most of it's
 - The fill buffers are used as data storage for hitting requests as well as for miss tracking so all lookup requests require a fill buffer.
 - A fill buffer makes multiple external requests to memory to fetch the required data to fill a cache line (tracked via ``fill_ext_cnt_q``).
 - Returning data is tracked via ``fill_rvd_cnt_q``.
-Not all requests will fetch all their data, since requests can be cancelled due to a cache hit or an intervening branch.
+  Not all requests will fetch all their data, since requests can be cancelled due to a cache hit or an intervening branch.
```
tomroberts-lowrisc: It looks like I was trying to make bullet points here, but failed. I guess either leave it as it is and I'll fix it later, or switch the `-`'s to `*`'s.

doc/icache.rst
```diff
@@ -149,3 +167,53 @@ Cache invalidation
...
+Data is requested from the instruction memory with the ports prefixed by ``instr_``. These work as described in :ref:`instruction-fetch`.
+Note that there's one extra port on the I$, which doesn't appear at the ``ibex_core`` top-level.
+This is ``instr_pmp_err_i``.
+If the PMP block disallows a fetch for a certain address, it will squash the outgoing memory request entirely and set ``instr_pmp_err_i``.
+If that happens, the cache behaves as if the request was granted but returned some time later with an error.
+Note that it is possible for ``instr_gnt_i`` and ``instr_pmp_err_i`` to be high on the same cycle (if the memory system ties ``instr_gnt_i`` to ``instr_req_o``, for example).
+In that case, the error signal takes precedence.
+
+Fetched instructions are returned to the core using ports ``ready_i``, ``valid_o``, ``rdata_o``, ``addr_o`` and ``err_o``.
+This interface uses ready/valid handshaking in the usual way (a transaction is signalled by ready and valid being high; if ready or valid goes high, it will remain high until the transaction goes through; if valid goes high, the other output signals will remain stable until the transaction goes through).
```
tomroberts-lowrisc: This is sort-of true, but the stability won't hold if there is an intervening branch.

doc/icache.rst
```diff
@@ -149,3 +167,53 @@ Cache invalidation
...
+
+When the cache comes out of reset, it may start fetching memory as soon as ``req_i`` goes high.
+However, the address counter is not initialised on reset.
+This means the behaviour of the I$ is unspecified unless ``branch_i`` is asserted on or before the first cycle that ``req_i`` is asserted after reset.
+If that is not true, there's nothing to stop the cache fetching from random addresses.
+
+The ``req_i`` port signals to the cache that the core is awake and will start requesting instructions soon.
+If ``req_i`` becomes false, the cache may do a few more instruction fetches to fill a cache line, but will stop fetching when that is done.
+The cache will not do any instruction fetches after this until ``req_i`` goes high again.
+The cache assumes that the core will not assert ``ready_i`` when ``req_i`` is low.
```
tomroberts-lowrisc: It doesn't really matter what `ready_i` does when `req_i` is low from the cache's perspective. If there are still some valid instructions buffered then if `ready_i` is high they will drain - if `ready_i` is low, they won't.

doc/icache.rst
```diff
@@ -149,3 +167,53 @@ Cache invalidation
...
+
+The ``icache_enable_i`` signal controls whether the cache stores fetched data.
+If the signal is false, fetched data will be discarded on a branch (or after enough addresses have been fetched in a straight line).
```
tomroberts-lowrisc: This might be a bit ambiguous. Buffered data is discarded on a branch only. The cache will cease prefetching when it runs out of resources, but it's not discarding any data at that point.

rtl/ibex_icache.sv
```diff
@@ -21,22 +21,26 @@ module ibex_icache #(
...
-    // IF stage interface
+    // IF stage interface: Pass fetched instructions to the core
     input  logic                ready_i,
     output logic                valid_o,
     output logic [31:0]         rdata_o,
     output logic [31:0]         addr_o,
     output logic                err_o,
 
-    // Instruction memory / interconnect interface
+    // Instruction memory / interconnect interface: Fetch instruction
```
tomroberts-lowrisc: Just make it 1 line - we have 100 character limit for sv files.

doc/icache.rst
rswarbrick: Ok, that makes sense. I've redrafted things a bit to make `ready_i` look less like a request. I've still kept the "don't do that" requirement: I think if this happens, it's a core bug (which we should have an assertion to spot!)

doc/icache.rst
rswarbrick: Ok, I'm going to try rejigging this by explicitly talking about a prefetch buffer.

rtl/ibex_icache.sv
rswarbrick: Done.

doc/icache.rst
rswarbrick: Ok, that makes sense. (I also made a mistake about the ready signal that I've fixed).

Issue #696, opened 2020-03-12
Questa: [SVCHK] - Extra checking for conflicts with always_comb and always_latch variables is done at vopt time.


imphil:
In Questa we get the following warning:

```
 18 ** Warning: /home/danghai/ibex/rtl/ibex_cs_registers.sv(730): (vlog-2583) [SVCHK] - Extra checking for conflicts with always_comb     and always_latch variables is done at vopt time.
 19 ** Warning: /home/danghai/ibex/rtl/ibex_cs_registers.sv(732): (vlog-2583) [SVCHK] - Extra checking for conflicts with always_comb     and always_latch variables is done at vopt time.
 20 ** Warning: /home/danghai/ibex/rtl/ibex_cs_registers.sv(739): (vlog-2583) [SVCHK] - Extra checking for conflicts with always_comb     and always_latch variables is done at vopt time.
 21 ** Warning: /home/danghai/ibex/rtl/ibex_cs_registers.sv(742): (vlog-2583) [SVCHK] - Extra checking for conflicts with always_comb     and always_latch variables is done at vopt time.
 22 ** Warning: /home/danghai/ibex/rtl/ibex_cs_registers.sv(745): (vlog-2583) [SVCHK] - Extra checking for conflicts with always_comb     and always_latch variables is done at vopt time.

```

_Originally posted by @danghai in https://github.com/lowRISC/ibex/issues/669#issuecomment-597823073_

My gut feeling is that this warning can be safely ignored. @danghai can you give us the output of `verror 2583` to see if Questa suggests anything that can be done here on the RTL side? Otherwise I'd propose we find a way to silence this warning.

danghai:
The verror message:

```
vlog Message # 2583:
vlog does not check for conflicts with always_comb and always_latch variables 
and it is deferred to vopt time. Run vopt for additional design-level checks.
```

PR #697, opened 2020-03-12
Questa qrun: Extra checking for conflicts with always_comb and always_latch variables is done at vopt time

danghai:
I make PR to suppress warning vlog-2583 for silence warning
Fixed: #696


imphil:
Looks like a safe "solution" to me, thanks!

PR #699, opened 2020-03-12
Remove property from assert message

### LINKED ISSUE #669: Questa complains about newlines in ASSERT macros
(Spin-off from #622 by @danghai)

Compiling Ibex with Questa gives the following warning:

```
** Warning: ** while parsing macro expansion: 'ASSERT' starting at /home/danghai/hhoangda/ibex/rtl/ibex_compressed_decoder.sv(286)
** at /home/danghai/hhoangda/ibex/rtl/ibex_compressed_decoder.sv(286): (vlog-2643) Unterminated string literal continues onto next line.
```

The code in question is

```verilog
// in ibex_compressed_decoder.sv
  `ASSERT(IbexInstrLSBsKnown, valid_i |->
      !$isunknown(instr_i[1:0]))

// in prim_assert.sv
`define ASSERT(__name, __prop, __clk = `ASSERT_DEFAULT_CLK, __rst = `ASSERT_DEFAULT_RST) \
`ifdef INC_ASSERT                                                                        \
  __name: assert property (@(posedge __clk) disable iff (__rst !== '0) (__prop))         \
    else `ASSERT_RPT(`PRIM_STRINGIFY(__name), `PRIM_STRINGIFY(__prop))                   \
`endif

// Converts an arbitrary block of code into a Verilog string
`define PRIM_STRINGIFY(__x) `"__x`"

`define ASSERT_RPT(__name, __msg)                                         \
`ifdef UVM                                                                \
  assert_rpt_pkg::assert_rpt($sformatf("[%m] %s: %s (%s:%0d)",            \
                             __name, __msg, `__FILE__, `__LINE__));       \
`else                                                                     \
  $error("[ASSERT FAILED] [%m] %s: %s", __name, __msg);                   \
`endif

```

Given this code, I'm assuming the error message actually comes from the fact that `prop` (the second argument to `ASSERT`) is converted to a string before being used in `ASSERT_RPT`, and this use of `prop` with newlines isn't liked by Questa.

According to SV2017, the definition of a string literal is 

image

Which indicates our code is actually wrong. I see a couple of options:
- Remove the reporting of the property in `ASSERT_RPT`, and only report the message.
- Add backslashes to the end-of-line of each multi-line property. I'm not sure if that's actually valid for the non-string case?
- Avoid newlines in properties when writing `ASSERT` macros.

@danghai:
- First, let's check if the hypothesis holds: Does the warning go away if you remove the `ASSERT_RPT()` call from the definition of the `ASSERT` macro?
- Can you try to add a backslash at the end of the property in `ibex_compressed_decoder.sv`, e.g. 

   ```systemverilog
     `ASSERT(IbexInstrLSBsKnown, valid_i |-> \
         !$isunknown(instr_i[1:0]))
   ```

Questa version info: QuestaSim-64 qrun 2019.4 Utility 2019.10 Oct 15 2019

CC @msfschaffner @sriyerg

imphil:
Assert macros in prim_assert.sv report the asserted property as part of
the error message if the assertion fails. Doing that requires us to
stringify the property. If the property is multi-line in the source
code, this produces a multi-line string. According to the
SystemVerilog standard, a multi-line string requires a backslash (`\`)
at the end of a line. Adding this backslash through a macro is not
possible (at least to my knowledge), giving us no way to produce a
standards-compliant implementation of the `ASSERT` macros that report
the property as string.

This commit therefore removes the property from the error message if an
assertion fails. It makes these messages less useful, but the
information can still be recovered in a less convenient way through the
name of the assertion, and the file and line numbers.

Fixes #669

### DIFF
--- a/shared/rtl/prim_assert.sv
+++ b/shared/rtl/prim_assert.sv
@@ -81,7 +81,7 @@
 // It can be called as a module (or interface) body item.
 `define ASSERT(__name, __prop, __clk = `ASSERT_DEFAULT_CLK, __rst = `ASSERT_DEFAULT_RST) \
 `ifdef INC_ASSERT                                                                        \
-  __name: assert property (@(posedge __clk) disable iff (__rst !== '0) (__prop))         \
+  __name: assert property (@(posedge __clk) disable iff ((__rst) !== '0) (__prop))       \
     else `ASSERT_RPT(`PRIM_STRINGIFY(__name), `PRIM_STRINGIFY(__prop))                   \
 `endif
 // Note: Above we use (__rst !== '0) in the disable iff statements instead of
@@ -92,7 +92,7 @@
 // Assert a concurrent property NEVER happens
 `define ASSERT_NEVER(__name, __prop, __clk = `ASSERT_DEFAULT_CLK, __rst = `ASSERT_DEFAULT_RST) \
 `ifdef INC_ASSERT                                                                              \
-  __name: assert property (@(posedge __clk) disable iff (__rst !== '0) not (__prop))           \
+  __name: assert property (@(posedge __clk) disable iff ((__rst) !== '0) not (__prop))         \
     else `ASSERT_RPT(`PRIM_STRINGIFY(__name), `PRIM_STRINGIFY(__prop))                         \
 `endif
 
@@ -106,7 +106,7 @@
 //  Cover a concurrent property
 `define COVER(__name, __prop, __clk = `ASSERT_DEFAULT_CLK, __rst = `ASSERT_DEFAULT_RST) \
 `ifdef INC_ASSERT                                                                       \
-  __name: cover property (@(posedge __clk) disable iff (__rst !== '0) (__prop));        \
+  __name: cover property (@(posedge __clk) disable iff ((__rst) !== '0) (__prop));      \
 `endif
 
 //////////////////////////////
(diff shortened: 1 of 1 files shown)


imphil:
@sriyerg I'll do the initial commit and review through this repository, but will propose the same change to the OT repository once we converge on a solution here.

imphil:
@danghai can you test the changes in this PR against Questa?

danghai:
There is no warning for this issue:
[qrun_compile_optimize.log](https://github.com/lowRISC/ibex/files/4326965/qrun_compile_optimize.log)

The another warning relates to `-svinputport`

shared/rtl/prim_assert.sv
```diff
@@ -34,13 +34,13 @@
...
-                             __name, __msg, `__FILE__, `__LINE__));       \
-`else                                                                     \
-  $error("[ASSERT FAILED] [%m] %s: %s", __name, __msg);                   \
+// ASSERT_RPT is available to change the reporting mechanism when an assert fails
+`define ASSERT_RPT(__name)                                   \
+`ifdef UVM                                                   \
+  assert_rpt_pkg::assert_rpt($sformatf("[%m] %s (%s:%0d)",   \
+                             __name, `__FILE__, `__LINE__)); \
+`else                                                        \
+  $error("[ASSERT FAILED] [%m] %s", __name);                 \
```
sriyerg: Could we add the `__FILE__` and `__LINE__` info to $error as well?

sriyerg:
I've added more people for review so that it serves as a heads-up for those that might be the most impacted by this change (reduced ease of debugging assertion failures).

shared/rtl/prim_assert.sv
weicaiyang: these 2 macros are from UVM?

imphil: These macros are defined in the SV spec, so we can use them here as well. I've updated the patch to include them.

imphil:
@sriyerg and @weicaiyang thanks for your reviews. 

@danghai thanks for your bug report and your testing, very much appreciated! It's always an interesting journey to make code work well with different tools, but we're getting there one step at a time.

shared/rtl/prim_assert.sv
weicaiyang: I see. Thanks @imphil

PR #704, opened 2020-03-13
[dv] coverage generation

udinator:
Add a step to the simulation flow for merged coverage generation.
Additionally changed the naming of coverage databases created by VCS to `test_<test_name>_<iterations>`  to make unique each database name and avoid overwriting, since seeds are generated in the top level Makefile.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -89,7 +89,7 @@ all: sim
 
 instr: iss_sim
 
-sim: post_compare
+sim: post_compare cov
 
 .PHONY: clean
 clean:
@@ -101,7 +101,7 @@ COMMON_OPTS := $(if $(call equal,$(VERBOSE),1),--verbose,)
 # Options for all targets that depend on the tests we're running.
 TEST_OPTS := $(COMMON_OPTS) \
              --seed=${SEED} \
-             --test"=${TEST}" \
+             --test="${TEST}" \
              --testlist=${TESTLIST} \
              --iterations=${ITERATIONS}
 
@@ -386,7 +386,7 @@ $(OUT-SEED)/regr.log: \
 post_compare: $(OUT-SEED)/regr.log
 
 ###############################################################################
-# Generate functional coverage
+# Generate RISCV-DV functional coverage
 fcov:
 	python3 ${GEN_DIR}/cov.py \
           --core ibex \
@@ -395,12 +395,15 @@ fcov:
           --isa rv32imc \
           --custom_target riscv_dv_extension
 
-# Load verdi to review coverage
-cov_vcs:
-	cd ${OUT}/rtl_sim; verdi -cov -covdir test.vdb &
-
-cov_ius:
-	if [ ! -d "${OUT}/rtl_sim/cov_work/scope/merged_cov" ]; \
-	  then imc -execcmd "merge -out ${OUT}/rtl_sim/cov_work/scope/merged_cov ${OUT}/rtl_sim/cov_work/scope/test_*"; \
-    fi
-	imc -load ${OUT}/rtl_sim/cov_work/scope/merged_cov &
+# Merge coverage in the <out> directory
+cov:
+	@rm -rf ${OUT}/rtl_sim/test.vdb
+	@./sim.py \
+		--steps=cov \
+		${TEST_OPTS} \
+		--simulator="${SIMULATOR}" \
+		--o="${OUT}" \
+		--lsf_cmd="${LSF_CMD}";
+	@if [ -d "test.vdb" ]; then \
+		mv -f test.vdb ./${OUT}/rtl_sim/; \
+	fi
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/Makefile
```diff
@@ -134,8 +134,7 @@ $(gen-dirs): %:
 # dump-vars,P,X,VS) will expand to a list of 'file' commands that write each
 # variable to P in Makefile syntax, but with "last-X-" prepended. At the start
 # of the file, we also define last-X-vars-loaded to 1. You can use this to
-# check whether there was a dump file at all.
-#
+# check whether there was a dump file at all.  #
```
rswarbrick: There's a stray '#' here.

dv/uvm/core_ibex/Makefile
```diff
@@ -395,12 +394,15 @@ fcov:
...
-cov_vcs:
-	cd ${OUT}/rtl_sim; verdi -cov -covdir test.vdb &
-
-cov_ius:
-	if [ ! -d "${OUT}/rtl_sim/cov_work/scope/merged_cov" ]; \
-	  then imc -execcmd "merge -out ${OUT}/rtl_sim/cov_work/scope/merged_cov ${OUT}/rtl_sim/cov_work/scope/test_*"; \
-    fi
-	imc -load ${OUT}/rtl_sim/cov_work/scope/merged_cov &
+# Merge coverage in the <out> directory and load into GUI
+cov:
```
rswarbrick: I think we should split up "merge coverage" and "open a GUI" into two separate steps. Is that possible with VCS's "urg" tool?

If so, this would probably be nicer if the target was a file that gets generated by merging the coverage (`test.vdb`?). Then you add a phony target that depends on that file. See the `$(OUT-SEED)/regr.log` and `post_compare` targets for an example.

dv/uvm/core_ibex/sim.py
```diff
@@ -413,6 +413,43 @@ def compare(test_list, iss, output_dir):
     return fails == 0
 
 
+#TODO(udinator) - support IUS, DSim, and Riviera
+def gen_cov(base_dir, simulator, lsf_cmd):
+    """Generate a merged coverage directory, and open up the appropriate GUI
```
rswarbrick: As with the comment in the Makefile, can we make this two separate steps? What if I want to run it in a CI flow with no X11 connection?

dv/uvm/core_ibex/sim.py
```diff
@@ -413,6 +413,43 @@ def compare(test_list, iss, output_dir):
...
+                dir_list.append(vdb_path)
+    if dir_list == []:
+        logging.info("No coverage data available, exiting...")
+        sys.exit(RET_SUCCESS)
+
+    if simulator == 'vcs':
+        cov_cmd = "urg -full64 -format both -dbname test.vdb " \
+                  "-report %s/rtl_sim/urgReport -dir " % base_dir
+        for cov_dir in dir_list:
+            cov_cmd += "%s " % cov_dir
```
rswarbrick: Minor nit: I think you've got a trailing space here. Probably drop the trailing space after `-dir` and then change this to `" %s"`.

dv/uvm/core_ibex/Makefile
udinator: Oops, nice catch

dv/uvm/core_ibex/Makefile
udinator: Ah sorry, I actually meant to remove the `load into GUI` part of this comment, this makefile step is only meant to generate a merged coverage directory. Opening the GUI can be left up to the user, since there are a wide variety of tools out there.

dv/uvm/core_ibex/sim.py
udinator: Same as above, I forgot to remove this section of the comment.

dv/uvm/core_ibex/sim.py
udinator: Good catch, I've updated the code accordingly

dv/uvm/core_ibex/Makefile
```diff
@@ -135,7 +135,6 @@ $(gen-dirs): %:
 # variable to P in Makefile syntax, but with "last-X-" prepended. At the start
 # of the file, we also define last-X-vars-loaded to 1. You can use this to
 # check whether there was a dump file at all.
-#
```
rswarbrick: This is a nit, but deleting this line messes up the paragraph structure in the comments. Either please don't do that, or reflow the comments (so that "Note" would be on the same line as "all").

dv/uvm/core_ibex/Makefile
rswarbrick: Cool, that sounds good to me.

dv/uvm/core_ibex/Makefile
udinator: Ah, not quite sure why I deleted this line! It has been inserted back into the Makefile.

PR #712, opened 2020-03-17
[rtl/icache] Fix PMP error logic

tomroberts-lowrisc:
Instruction requests triggering PMP errors have their external request
suppressed. The beat counting logic therefore needs to know that these
requests will never receive any rvalid data responses.

This fix stops the external request counter from incrementing, and marks
all external requests complete as soon as any error is received.

The data in the cache line beyond the error is not required since the
core cannot access it without consuming the error first.


rswarbrick:
Looks sensible to me. Good point about why we can just cancel the cache line entirely.

vogelpi:
Does this change have an impact on the "Detailed behaviour" section in the documentation?

tomroberts-lowrisc:
> Does this change have an impact on the "Detailed behaviour" section in the documentation?
Good point - I have tweaked the wording a bit to accurately reflect the new behaviour.

PR #713, opened 2020-03-18
[rtl] Fix mtval for unaligned instr errors

### LINKED ISSUE #709: Wrong mtval recorded for instruction access fault
The RISCV Privileged ISA specification specifies:
"For instruction-fetch access or page-fault exceptions on systems with variable-length instructions, `mtval` will contain the virtual address of the portion of the instruction that caused the fault while `mepc` will point to the beginning of the instruction."

In Ibex currently, both `mepc` and `mtval` will point to the beginning of the instruction, even if the fault originated in the second half of a misaligned uncompressed instruction.

I see two solutions:
1) Wire `mtval` to zero on instruction access faults to indicate that this reporting feature is not supported.
2) Capture the second address somehow in this particular case. (It will always be the ID stage PC + 2)

tomroberts-lowrisc:
mtval should record which half of the instruction caused the error
rather than just recording the PC.
An extra signal is added in the IF stage to indicate when an error is
caused by the second half of an unaligned instruction. This signal is
then used to increment the PC by 2 for mtval capture on an error.

Fixes #709

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -107,6 +107,7 @@ module ibex_core #(
   logic [15:0] instr_rdata_c_id;       // Compressed instruction sampled inside IF stage
   logic        instr_is_compressed_id;
   logic        instr_fetch_err;        // Bus error on instr fetch
+  logic        instr_fetch_err_plus2;  // Instruction error is misaligned
   logic        illegal_c_insn_id;      // Illegal compressed instruction sent to ID stage
   logic [31:0] pc_if;                  // Program counter in IF stage
   logic [31:0] pc_id;                  // Program counter in ID stage
@@ -364,6 +365,7 @@ module ibex_core #(
       .instr_rdata_c_id_o       ( instr_rdata_c_id       ),
       .instr_is_compressed_id_o ( instr_is_compressed_id ),
       .instr_fetch_err_o        ( instr_fetch_err        ),
+      .instr_fetch_err_plus2_o  ( instr_fetch_err_plus2  ),
       .illegal_c_insn_id_o      ( illegal_c_insn_id      ),
       .pc_if_o                  ( pc_if                  ),
       .pc_id_o                  ( pc_id                  ),
@@ -436,6 +438,7 @@ module ibex_core #(
       .exc_cause_o                  ( exc_cause                ),
 
       .instr_fetch_err_i            ( instr_fetch_err          ),
+      .instr_fetch_err_plus2_i      ( instr_fetch_err_plus2    ),
       .illegal_c_insn_i             ( illegal_c_insn_id        ),
 
       .pc_id_i                      ( pc_id                    ),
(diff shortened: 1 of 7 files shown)


tomroberts-lowrisc:
I did come up with an alternative fix for this:
Instead of adding 2 to the ID stage PC, we instead mux in the IF stage PC and mask the bottom bits.
In the end I stuck with the adder since it's only a bit more logic than a mux, and avoids putting even more fanout on the IF stage PC (and therefore the branch target) which is already pretty bad.

rtl/ibex_core.sv
```diff
@@ -107,6 +107,7 @@ module ibex_core #(
   logic [15:0] instr_rdata_c_id;       // Compressed instruction sampled inside IF stage
   logic        instr_is_compressed_id;
   logic        instr_fetch_err;        // Bus error on instr fetch
+  logic        instr_err_plus2;        // Instruction error is misaligned
```
vogelpi: This is minor: I would change the name of the new signal to `instr_fetch_err_plus2` where you already have an `instr_fetch_err` just to make clear that these are not to completely different things.

rtl/ibex_fetch_fifo.sv
```diff
@@ -33,7 +33,8 @@ module ibex_fetch_fifo #(
     input  logic        out_ready_i,
     output logic [31:0] out_addr_o,
     output logic [31:0] out_rdata_o,
-    output logic        out_err_o
+    output logic        out_err_o,
+    output logic        out_err_plus2_o
```
vogelpi: Here it is obviously fine to keep it as `out_err_plus2_o`.

rtl/ibex_id_stage.sv
```diff
@@ -51,6 +51,7 @@ module ibex_id_stage #(
 
     input  logic                      illegal_c_insn_i,
     input  logic                      instr_fetch_err_i,
+    input  logic                      instr_err_plus2_i,
```
vogelpi: Also here, I would change to `instr_fetch_err_plus2_i`.

rtl/ibex_if_stage.sv
```diff
@@ -88,6 +89,7 @@ module ibex_if_stage #(
   logic       [31:0] fetch_rdata;
   logic       [31:0] fetch_addr;
   logic              fetch_err;
+  logic              err_plus2;
```
vogelpi: Also here, I would change to `fetch_err_plus2`.

rtl/ibex_core.sv
tomroberts-lowrisc: Agreed I was being a bit lazy and avoiding a load of code realignment. Now done!

GregAC:
Does the IF stage stage PC have that much fan-out? I would have thought it'd be the ID stage PC that has the most fanout.

tomroberts-lowrisc:
> Does the IF stage stage PC have that much fan-out? I would have thought it'd be the ID stage PC that has the most fanout.

Maybe not fanout, but the IF stage PC is effectively the branch target + some muxing. Has a fair bit of fanout in the prefetch buffer / icache, pmp and cs_registers as well as driving the instruction fetch address output.

GregAC:
Ah yes of course, makes sense.

LGTM!

PR #714, opened 2020-03-18
Multi config ci

GregAC:


GregAC:
Whilst the immediate goal (run CI on 3rd pipe stage/BT ALU/Single cycle mult) was simple I wanted a more robust solution to supporting multiple Ibex configurations, this is what I have come up with. The key parts are the ibex_configs.hjson file and the ./util/ibex_config.py python script. The python script can read and validate the configs file then output it in multiple formats for a named config (currently just fusesoc command line options but its designed to be easily extended).

It lacks documentation but then we lack documentation on how to run things like lint and riscv-compliance etc anyway. So for this PR at least the makefile updates can probably count as sufficient for now but something to be addresses (I'll open an issue).

We also need to define what exactly a supported configuration is and what configurations we decide to support, but again this decision isn't required for this PR.

ibex_configs.hjson
```diff
@@ -0,0 +1,18 @@
+{
```
GregAC: I need a copyright header...

GregAC: I went through a few options for the configuration file, initially I was going to do something super simple you could use awk to pull configuration information from simply, but you quickly end up with complex awk commands or an awkward config file, so I thought just writing some python to do it was a better idea.

I then considered doing a nice readable ASCII art table with the configs, I decided against this for two reasons:

1. Whilst parsing is easy if you keep to strict formatting rules it's brittle
2. As you add more parameters the table ends up getting rather wide and the 'easily-readable' bit is lost so there's not point in doing it in the first place

So I've gone for hjson, the python script checks every config specifies the same configuration options. I considered having a default section then specific overrides for different configs but thought it was better to able to look at a single config and immediately know what parameters are getting set to what values.

GregAC:
@rswarbrick we'll want to use this config system with the DV setup, so I'd be particularly interested on any feedback on how easy/hard that will be in it's current form.

GregAC:
@rswarbrick we'll want to use this config system with the DV setup, so I'd be particularly interested on any feedback on how easy/hard that will be in it's current form.

rswarbrick:
Should the lint fixes and .core files commits be squashed? From one viewpoint, the .core file patch introduces a bug (lint fails), which the lint patch fixes... (I don't have strong feelings about this, but I'm keeping bisectability in mind)

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+                                      ' has differing parameters ' +
+                                      ','.join(parameter_difference))
+
+
+def config_dict_to_fusesoc_opts(config_dict):
+    fusesoc_cmd = []
+    for parameter, value in config_dict.items():
+        fusesoc_cmd.append('--' + parameter + '=' + str(value))
+
+    return ' '.join(fusesoc_cmd)
```
rswarbrick: For bonus points, you probably want to use `shlex.quote` here, assuming that the output is supposed to be something that the shell can run.

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def get_config_file_location():
+    """Returns the location of the config file, IBEX_CONFIG_FILE environment
+    variable overrides the default"""
+
+    if 'IBEX_CONFIG_FILE' in os.environ:
+        return os.environ['IBEX_CONFIG_FILE']
+
+    return DEFAULT_CONFIG_FILE
```
rswarbrick: This could be `return os.environ.get('IBEX_CONFIG_FILE', DEFAULT_CONFIG_FILE)`, which is rather shorter.

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+    argparser = argparse.ArgumentParser(
+        description="""Outputs Ibex configuration
+    parameters for a named config in a number of formats.  If not specified on
+    the command line the config will be read from {0}.  This default can be
+    overridden by setting the IBEX_CONFIG_FILE environment variable
+    """.format(get_config_file_location()))
+
+    argparser.add_argument('config_name',
+                           help="""The name of the Ibex
+    configuration to output""")
```
rswarbrick: This is rather ugly. Better:

    help=('The name of the Ibex '
          'configuration to output')

(and you only need to wrap at column 80).

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+                  'not found in',
+                  args.config_filename,
+                  file=sys.stderr)
+
+            sys.exit(1)
+
+        print(config_outputs[args.output_type](config_dicts[args.config_name]))
+    except ConfigException as ce:
+        print('ERROR: failure to read configuration from',
+              args.config_name,
```
rswarbrick: config_filename, no?

util/ibex_config.py
GregAC: This was post running it through the pylint script in the OT repo, it does sometimes produce odd outputs, I'll go with your suggestion (no python lint CI to fail so doesn't matter if the linter doesn't agree), after taking a look at the style guide.

util/ibex_config.py
GregAC: Indeed, good point

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+import hjson
+
+DEFAULT_CONFIG_FILE = 'ibex_configs.hjson'
+
+
+class ConfigException(Exception):
+    pass
+
+
+def verify_config(name, config_dict):
```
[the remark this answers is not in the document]
GregAC: I'll read it and have a think.

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+    try:
+        config_hjson = hjson.load(config_file)
+
+        for k, v in config_hjson.items():
+            verify_config(k, v)
+
+        verify_config_parameters(config_hjson)
+
+        return config_hjson
+    except hjson.HjsonDecodeError as e:
```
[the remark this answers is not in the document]
GregAC: will do

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+            if first:
+                parameters.add(parameter)
+
+            parameters_this_config.add(parameter)
+
+        if first:
+            first = False
+        else:
+            parameter_difference = parameters ^ parameters_this_config
+            if len(parameter_difference) != 0:
```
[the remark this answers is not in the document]
GregAC: agreed

util/ibex_config.py
GregAC: Well it's something that should be appended to a fusesoc command line rather than run directly but shlex.quote should do something useful all the same, I'll add it in.

util/ibex_config.py
GregAC: yes indeed, possibly the only error handling path I haven't actually tested.

Makefile
```diff
@@ -15,7 +17,8 @@ build-all: build-riscv-compliance build-simple-system build-arty-100 \
 .PHONY: build-riscv-compliance
 build-riscv-compliance:
 	fusesoc --cores-root=. run --target=sim --setup --build \
-		lowrisc:ibex:ibex_riscv_compliance
+		lowrisc:ibex:ibex_riscv_compliance \
+		$(shell ./util/ibex_config.py $(IBEX_CONFIG) fusesoc_opts)
```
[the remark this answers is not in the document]
GregAC: I view this makefile as a quick and easy way to run commands as well as useful examples of how to do things. As such I'd prefer to leave this as is due to its simplicitly.

A target that outputs the fusesoc flags could be useful for debugging.

tomroberts-lowrisc:
Looks like a good approach to me.

Makefile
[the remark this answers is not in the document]
rswarbrick: Yes, thinking about it, these targets are all `.PHONY` anyway, so tracking whether we rebuilt things properly doesn't really matter.

util/ibex_config.py
rswarbrick: :-) Yay for Python's "whaddya mean, static analysis?" semantics.

azure-pipelines.yml
```diff
@@ -172,37 +166,7 @@ jobs:
...
-            fail=1
-          fi
-        fi
-      done
-      exit $fail
-    displayName: "Run RISC-V Compliance test for Ibex RV32IMC"
+  # Run Ibex RTL CI per supported configuration
+  - template : ci/ibex-rtl-ci-steps.yml
+    parameters:
+      ibex_configs : ['small-fastmult', 'maxperf-scmult']
```
imphil: It's easier to extend these lists (and produce smaller diffs) if you go for the "longer list style":

```suggestion
      ibex_configs: 
      - small-fastmult
      - maxperf-scmult
```

ibex_configs.hjson
```diff
@@ -0,0 +1,18 @@
...
+  small-fastmult : {
```
imphil: Please add a description field to tell users a bit more about this configuration in text form.

ibex_configs.hjson
imphil: Can we please use normal JSON or YAML here? hjson is a very OpenTitan thing [and actually not super active development now that JSON5 has arrived] and parsers are not as widely available as for JSON and for YAML. (I'd actually go for YAML, since that's the default config file format also for RISC-V DV and fusesoc.)

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python3
```
imphil: Please add a license header.

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class ConfigException(Exception):
+    pass
+
+
+def verify_config(name, config_dict):
+    """Check config_dict matches expectations:
+        - It's a mapping object e.g. OrderedDict
+        - It's values are either strings or integers"""
```
imphil: Please have a look at http://google.github.io/styleguide/pyguide.html#383-functions-and-methods for the recommended format doc comments.

util/ibex_config.py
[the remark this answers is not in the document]
imphil: The easier and more scalable option is using an existing schema validator. For JSON and YAML that would be jsonschema. But in general I wouldn't invest much into validating here. It's a trivial script and already quite "blown up" by infrastructure around its primary goal.

util/ibex_config.py
```diff
@@ -0,0 +1,145 @@
...
+        return os.environ['IBEX_CONFIG_FILE']
+
+    return DEFAULT_CONFIG_FILE
+
+
+def main():
+    config_outputs = {'fusesoc_opts': config_dict_to_fusesoc_opts}
+
+    argparser = argparse.ArgumentParser(
+        description="""Outputs Ibex configuration
```
imphil: Outputs -> Output

ci/ibex-rtl-ci-steps.yml
```diff
@@ -0,0 +1,55 @@
+parameters:
+  ibex_configs : []
```
imphil:
```suggestion
  ibex_configs: []
```

ibex_configs.hjson
```diff
@@ -0,0 +1,18 @@
...
+    RV32E                    : 0
+    RV32M                    : 1
+    BranchTargetALU          : 0
+    WritebackStage           : 0
+    MultiplierImplementation : "fast"
+  }
+
+  maxperf-scmult : {
```
imphil: The naming of the multipliers is confusing here. Is `fast` slower than `sc`? I know this matches our code currently, but maybe we don't need to deepen the use of these confusing terms. Can we maybe use the # of cycles as naming? E.g. `4cmul`, or `1cmul`?

imphil:
Thanks Greg! The Azure bits look nice, most small comments are on the config script.

util/ibex_config.py
GregAC: I think 'outputs' reads better to my ear.

ibex_configs.hjson
GregAC: I have switched to yaml

util/ibex_config.py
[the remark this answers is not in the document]
GregAC: So I've basically left the validation as is, I feel the structure is actually what @rswarbrick link is getting at in principle, the verify functions are intended to be internal, so any external user must use the 'parse' function (as the link would put it) which either gives you a valid configuration or results in an exception.

Briefly considering schema validation but haven't implemented for two reasons:
1. Will spend a while working out how to do it and I'm trying to avoid too much extra effort on this
2. From my brief reading it wasn't obvious how to enforce the all configurations must have the same parameter set rule via schemas. May be possible but see point 1.

GregAC:
Realised I haven't put the latest python through pylint, will do that after coffee :)

GregAC:
> Should the lint fixes and .core files commits be squashed? From one viewpoint, the .core file patch introduces a bug (lint fails), which the lint patch fixes... (I don't have strong feelings about this, but I'm keeping bisectability in mind)

Maybe, I quite like avoid having commits mix too many unrelated changes together (makes reading git blame more useful). Yes breaks bisectability but given we don't really aim to keep that on the Ibex codebase it probably doesn't matter (I think for that to be relevant you're either strict on it with CI enforcement to check or it isn't something you aim for, a half-way house doesn't help).

ibex_configs.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Ibex configurations files, holds the parameter sets that are tested under CI.
+# Each configuration must specify the same set of parameters
+
+# Two-stage pipeline without additional branch target ALU and 2 cycle multiplier
+# (3 cycles for mulh)
```
GregAC: I've messed up these cycle counts, it's 3 for mul and 4 for mulh.

Should also explain that stall cycles are actually one less (so single cycle mul means no stalling, 3 cycle mul means 2 cycles of stall on mul)

python-requirements.txt
```diff
@@ -2,6 +2,8 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+hjson
```
imphil: not needed any more?

GregAC: hjson forever!

Well spotted, I shall remove it.

ibex_configs.yaml
```diff
@@ -0,0 +1,25 @@
...
+small-3cmult:
+  RV32E                    : 0
+  RV32M                    : 1
+  BranchTargetALU          : 0
+  WritebackStage           : 0
+  MultiplierImplementation : "fast"
+
+# Three-stage pipeline with additional branch traget ALU and 1 cycle multiplier
+# (2 cycles for mulh) so mul does not stall (mulh stall 1 cycles). This is the
+# maximum performance configuration.
```
imphil: Possible follow-up: put this information into the YAML itself, e.g. into a `description` field (or `_description` to avoid name clashes) to make it available from the ibex_config.py script. (No need to do that as part of this PR though, but please file a follow-up. We can then also use this information to auto-generate parts of the docs for supported configurations.)

ibex_core_tracing.core
```diff
@@ -33,26 +33,30 @@ parameters:
     paramtype: vlogdefine
 
   RV32E:
-    datatype: bool
+    datatype: int
     paramtype: vlogparam
 
   RV32M:
-    datatype: bool
+    datatype: int
```
imphil: Any reason for that change?

GregAC: It's used inconsistently across the cores and I found with 'bool' I couldn't get it to work, end up with errors of the form

```
fusesoc run lowrisc_ibex_ibex_core_tracing_0.1: error: argument --RV32E: ignored explicit argument 'True'
```

`--RV32E=1` `--RV32E=True` and `--RV32E=true` all give the same result.

Potentially a fusesoc issue that I should file!

ibex_configs.yaml
GregAC: Good idea, could do it as part of a documentation PR

ibex_core_tracing.core
tomroberts-lowrisc: I noticed that also. It wants `--RV32E`` only to set it to true (no argument).

GregAC: Aha, that may well be preferable, though I should also alter all of the other .core files to be consistent. Think I'll do that.

imphil: Please file a fusesoc bug/feature request to allow the forms you tried (I thought it worked like that in the past, but I'm not sure). It certainly makes sense.

GregAC:
Ok I've switched all of the int parameters to use bool as that seemed cleaner

PR #715, opened 2020-03-18
[verilator] Fix --term-after-cycles

luismarques:
The option to limit the verilator simulation's maximum number of cycles was
implemented in terms of the `time_` variable, which counts half cycles. This
patch makes the limit respect the requested number of full cycles.

The patch divides `time_` by 2, instead of multiplying `term_after_cycles_`
by 2, as that better conveys the intended semantics, and slightly increases
the range of the limit. Related computations were modified for consistency.

### DIFF
--- a/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
+++ b/dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
@@ -283,10 +283,10 @@ void VerilatorSimCtrl::Run() {
   UnsetReset();
   Trace();
   while (1) {
-    if (time_ >= initial_reset_delay_cycles_ * 2) {
+    if (time_ / 2 >= initial_reset_delay_cycles_) {
       SetReset();
     }
-    if (time_ >= reset_duration_cycles_ * 2 + initial_reset_delay_cycles_ * 2) {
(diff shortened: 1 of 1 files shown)


dv/verilator/simutil_verilator/cpp/verilator_sim_ctrl.cc
```diff
@@ -315,7 +315,7 @@ void VerilatorSimCtrl::Run() {
                 << std::endl;
       break;
     }
-    if (term_after_cycles_ && time_ > term_after_cycles_) {
+    if (term_after_cycles_ && time_ / 2 >= term_after_cycles_) {
```
GregAC: nit: Personally I like to some parenthesis in nested expressions to make intent more obvious rather than relying on operator precedence alone, e.g.

```c
if (term_after_cycles_ && (time_ / 2 >= term_after_cycles_)) {
````

However I can't immediately find any guidance on this in the C++ style guide, so up to you.

luismarques: Thanks. I added the parenthesis.

imphil:
> Had noticed this but never got around to fixing it.

Please try to file issues for such things if possible; even if you don't get around fixing it, it's at least helpful to avoid doing the debugging twice.

PR #721, opened 2020-03-23
[rtl] Remove stall cycle with BT ALU

tomroberts-lowrisc:
- Branches don't need to stall with the branch target ALU

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -628,8 +628,8 @@ module ibex_id_stage #(
             end
             branch_in_dec: begin
               // cond branch operation
-              id_fsm_d      =  branch_decision_i ? MULTI_CYCLE : FIRST_CYCLE;
-              stall_branch  =  branch_decision_i;
+              id_fsm_d      =  (!BranchTargetALU && branch_decision_i) ? MULTI_CYCLE : FIRST_CYCLE;
+              stall_branch  =  ~BranchTargetALU & branch_decision_i;
               branch_set_d  =  branch_decision_i;
               perf_branch_o =  1'b1;
             end


tomroberts-lowrisc:
@GregAC apologies if you were already looking at this, but it overlapped with my fixed execution time work.

GregAC:
I had left this in intentionally. After any jump or branch there is a stall due to a pipeline bubble. So for a jump in particular it still counts as a multi-cycle instruction, one cycle executing the jump, one or more cycles waiting for the pipeline bubbles. With the branch target ALU a taken branch should do exactly the same as an unconditional jump from a stalling perspective

This is a bit weird but aimed to keep the behaviour of the existing RTL as much as possible (which treated the pipeline bubble after the jump/branch as a stall).

Looking at it again I don't think we've actually got the same behaviour between jump/branch anyway, without BT ALU a branch will set stall_branch the first cycle, then in `MULTI_CYCLE` perform the pc change then go back to `FIRST_CYCLE`, whilst a jump will perform the pc change then go to `MULTI_CYCLE` possibly only staying there a single cycle regardless of how long the iside takes to perform the jump.

So in conclusion happy with this change but I think we need to alter the jump behaviour as well, probably just remove the `jump_in_dec` case from the state machine entirely and remove the `stall_jump` signal as I'm not sure it's doing anything useful.

Out of interest did you see any actual issues because of this or did you just see it and thought it looked wrong?

tomroberts-lowrisc:
Ah I see that makes sense. No issue found, just thought it looked weird.

To my mind the stall due to the pipeline bubble is separate from the instruction. If we add branch prediction, that stall might not be there at all so probably makes sense to remove it from the instruction.

The jumps stall to allow the writeback to happen for JAL (first cycle sets the jump, second calculates and stores PC+4). It makes sense for the jump to do it that way around since it's more performant - the writeback second cycle happens in the pipeline bubble so has no cost.

I have a rough version of the RTL where I extend the BT ALU to work for jumps too, after which the second cycle isn't required since the writeback can happen in parallel.

GregAC:
All makes sense, had forgotten about the need for writeback for JAL.

Had actually intended to extend the BT ALU for jumps, looks like I forgot, thanks for looking at sorting it out.

GregAC:
In case it was uncertain - This all LGTM following the discussion (i.e. this RTL change in isolation without the jump changes I initially proposed).

PR #727, opened 2020-03-25
[rtl] Branch signal timing fix

tomroberts-lowrisc:
- Before this fix, the branch signal was qualified by the illegal
  instruction signal and the illegal csr signal.
- This patch removes both of these since the decoder already masks
  branches with illegal isntruction, and a branch cannot be a CSR op.
- This improves the worst path in the design significantly without the
  branch target ALU.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -209,7 +209,7 @@ module ibex_controller #(
       exc_req_d | exc_req_lsu;
 
   // special request that can specifically occur during branch instructions
-  assign special_req_branch = (illegal_insn_d | instr_fetch_err) & (ctrl_fsm_cs != FLUSH);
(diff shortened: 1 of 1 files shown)


GregAC:
What happens when you have a branch with an illegal condition? This is the relevant bit from the decoder:

https://github.com/lowRISC/ibex/blob/bdfe8de0f7f0833d9d78877f328d46fb35e104e8/rtl/ibex_decoder.sv#L242-L257

From a quick look at the RTL I think `branch_in_dec` gets set as does `illegal_insn`. The `illegal_insn` means the branch isn't executed but after this change the branch will execute rather than taking an illegal instruction exception.

If this is the case I think we can still do something like this fix but we may want a specific illegal_branch_insn signal or similar just to catch the illegal branch cases?

tomroberts-lowrisc:
See further down in the same always_comb block - `branch_in_dec_o` is cleared by `illegal_insn`

PR #732, opened 2020-03-25
[dv] remove return value from the compare() step

udinator:
Currently, if post_compare step fails, the entire makefile will error out, and coverage will not be generated.
Since coverage generation technically does not rely on the result of the post_compare stage, I now have each stage of the sim flow return RET_SUCCESS, to indicate that the step itself was able to run to completion; incorrect comparison results etc... will be logged to the terminal upon any failures and will now not prevent coverage from being generated.

### DIFF
--- a/dv/uvm/core_ibex/sim.py
+++ b/dv/uvm/core_ibex/sim.py
@@ -556,10 +556,9 @@ def main():
         rtl_sim(sim_cmd, matched_list, seed, args.sim_opts,
                 output_dir, bin_dir, args.lsf_cmd, check_return_code)
 
-    # Compare RTL & ISS simulation result.;
+    # Compare RTL & ISS simulation result.
     if steps['compare']:
-        if not compare(matched_list, args.iss, args.o):
-            return RET_FAIL
+        compare(matched_list, args.iss, args.o)
 
     # Generate merged coverage directory and load it into appropriate GUI
     if steps['cov']:


rswarbrick:
I'm not a great fan of this patch: in the software world I hail from(!), tests generally exit with status code 1 if they fail. We could quite easily have got the behaviour you need with some Makefile changes without discarding the error code.

Anyway, it's landed now, but it would have been good if you'd waited for a review. This came in at 8:20pm UK time and was merged at 10pm. If you don't want a review from me, please don't add me to the reviewer list.

udinator:
Understood.
I'd meant for this PR to be a sort of temporary solution to just quickly unblock the full coverage flow for some external users of the Ibex simulation flow.
You're right that generally the tests should exit with some status code, and masking them all off to `RET_SUCCESS` with this PR is somewhat messy. I'm looking into a more permanant  (and higher quality!) solution with the Makefile that will still exit with the correct status code but also not block the Makefile from moving into the coverage stage at the end.

rswarbrick:
Ok, great. I'd probably suggest a Makefile target which checks that the sim ran to completion (passing the test or not). Coverage runs will just depend on that. Runs that expect the test to pass can depend on a stamp file (`test.passed` or whatever) which depends on the test completing and then greps for a PASS code or similar.

The default Make target would then depend on both the coverage and the "test passed" targets. If you want to ensure you get coverage results even if the tests fail (which I don't think really makes sense, but hey), use `make -k`.

Personally, I'd change `sim.py` to fail on failing tests again because that makes more sense to me. To do that, you'll also need to prefix the command that runs it in the Makefile with a `-` sign. But I guess that's optional.

PR #742, opened 2020-03-27
[rtl/pmp] Fix PMP error prioritization

tomroberts-lowrisc:
- Region matches should be prioritized from 0 - N as stated in the RISCV
  Privileged Spec v1.11

### DIFF
--- a/rtl/ibex_pmp.sv
+++ b/rtl/ibex_pmp.sv
@@ -36,8 +36,6 @@ module ibex_pmp #(
   logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_low;
   logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_both;
   logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_perm_check;
-  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   machine_access_fault;
-  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   user_access_allowed;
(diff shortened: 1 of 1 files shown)


rtl/ibex_pmp.sv
```diff
@@ -87,14 +85,26 @@ module ibex_pmp #(
...
     end
-    assign access_fault[c] = (priv_mode_i[c] == PRIV_LVL_M) ? |machine_access_fault[c] :
-                                                              ~|user_access_allowed[c];
+
+    // Access fault determination / prioritization
+    always_comb begin
+      // Default is allow for M-mode, deny for other modes
+      access_fault[c] = (priv_mode_i[c] != PRIV_LVL_M);
+
+      // PMP regions are statically prioritized, from 0 - N
```
imphil: - regions -> entries (to match language in spec)
- N -> N-1?

rtl/ibex_pmp.sv
```diff
@@ -87,14 +85,26 @@ module ibex_pmp #(
...
+      // The lowest PMP region number which matches an address determines accessability
```
imphil: "lowest-numbered PMP entry"?

rtl/ibex_pmp.sv
tomroberts-lowrisc: Will do

rtl/ibex_pmp.sv
tomroberts-lowrisc: Will do

PR #743, opened 2020-03-27
[rtl] Unify top-level parameter declaration

tomroberts-lowrisc:
- Make parameter declaration order and default values in
  ibex_core_tracing.sv match the documentation


rswarbrick:
I agree. Is it possible to pull this data out of the RTL when making the documentation, or is that just too much of a faff?

GregAC:
Given it shouldn't change much and messing it up or more of an annoyance than a problem probably just best to manually keep them in sync.

tomroberts-lowrisc:
I kind-of think it's self-evident that they should match, and easy enough to maintain manually.

PR #744, opened 2020-03-27
[doc/icache] Document the err_plus2_o signal

tomroberts-lowrisc:
Add a few sentences to describe the behaviour/meaning of the err_plus2_o
signal, and how it is used by the core.

### DIFF
--- a/doc/icache.rst
+++ b/doc/icache.rst
@@ -241,6 +241,10 @@ This increment depends on the instruction data (visible at ``rdata_o``): if the 
 If the bottom bits of the instruction are ``2'b11`` then the instruction is considered to be uncompressed and the address will be incremented by 4.
 Since the contents of ``rdata_o`` are not specified if an instruction fetch has caused an error, the core must signal a branch before accepting another instruction after it sees ``err_o``.
 
+Because a single instruction can span two 32bit memory addresses, an extra signal (``err_plus2_o``) indicates when an error is caused by the second half of an unaligned uncompressed instruction.
+This signal is only valid when ``valid_o`` and ``err_o`` are set, and will only be set for uncompressed instructions.
+The core uses this signal to record the correct address in the ``mtval`` CSR upon an error.
+
 Since the address counter is not initialised on reset, the behaviour of the I$ is unspecified unless ``branch_i`` is asserted on or before the first cycle that ``req_i`` is asserted after reset.
 If that is not true, there's nothing to stop the cache fetching from random addresses.
 


doc/icache.rst
```diff
@@ -240,6 +240,9 @@ On other cycles, the address counter will be incremented every time an instructi
 This increment depends on the instruction data (visible at ``rdata_o``): if the bottom bits of the instruction are not ``2'b11``, the instruction is considered to be compressed and the address will be incremented by 2.
 If the bottom bits of the instruction are ``2'b11`` then the instruction is considered to be uncompressed and the address will be incremented by 4.
 Since the contents of ``rdata_o`` are not specified if an instruction fetch has caused an error, the core must signal a branch before accepting another instruction after it sees ``err_o``.
+The address in error is that signalled on ``addr_o``.
+Because a single instruction can span two 32bit memory addresses, an extra signal (``err_plus2_o``) indicates when the error is caused by the second half of an unaligned uncompressed instruction.
```
rswarbrick: Can compressed instructions ever signal `err_plus2_o`? Or is that squashed? Also, can `err_plus2_o` be true without `err_o`? (Is "instruction bad" equal to `err_o | err_plus2_o` or is it `err_o`, and I can figure out the offending address by looking at `err_plus2_o`?)

doc/icache.rst
```diff
@@ -240,6 +240,9 @@ On other cycles, the address counter will be incremented every time an instructi
 This increment depends on the instruction data (visible at ``rdata_o``): if the bottom bits of the instruction are not ``2'b11``, the instruction is considered to be compressed and the address will be incremented by 2.
 If the bottom bits of the instruction are ``2'b11`` then the instruction is considered to be uncompressed and the address will be incremented by 4.
 Since the contents of ``rdata_o`` are not specified if an instruction fetch has caused an error, the core must signal a branch before accepting another instruction after it sees ``err_o``.
+The address in error is that signalled on ``addr_o``.
```
rswarbrick: I found this quite confusing to read. Would it make more sense just to say that a transaction on the `icache <-> core` bus pertains to data starting at `addr_o`? Then I feel like explaining the difference between `err_o` and `err_plus2_o` is kind of separate. (Maybe even a separate paragraph: this is my fault, but I think that the text is pretty dense!)

tomroberts-lowrisc: I've had another go, creating a separate paragraph and removing the first sentence, since the meaning of ``addr_o`` is already specified elsewhere.

doc/icache.rst
tomroberts-lowrisc: Will clarify in the doc

PR #746, opened 2020-03-27
[bitmanip] Add ZBT Instruction Group

ganoam:
This commits implements the Bit Manipulateion Extension ZBT instruction
group: cmix, cmov, fsr[i] and fsl. Those are instructions depend on
three ALU operands. Completeion of these instructions takes 2 clock
cycles.

Architectural additions:

        * Multicycle Stage Register in ID stage.
                multicycle_op_stage_reg

        * Decoder generates alu_multicycle signal, to stall pipeline

        * For all ternary instructions:
                1. cycle: connect alu operands a and b to rs1 and rs2
                          respectively
                2. cycle: connect operands a and be to rs3 and rs2
                          respectively

Instruction Details:
        * cmov:
                1. store operand a (rs1) in stage reg.
                2. return stage reg output (rs2)  or rs3.

                if rs2 != 0 the output (rs1) is already known in the
                  first cycle. -> variable latency implementation is
                  possible.

        * cmix:
                1. store rs1 & rs2 in stage_reg
                2. return stage_Reg_q | (rs2 & ~rs3)

                reusing bwlogic from zbb

        * fsl/fsr:
                1. store rs1
                2. - if(shift_amt[5] == 0: shift {rs3, rs1}
                   - else:                 shift {rs1, rs3}

TODO:

        * Update Doc.

        * Include version of bitmnanip draft in Doc.

        * cmov -> variable latency:
                pro: performance,
                con: potential security issue

        * Use multiplier mac register instead of separate
                register?!?!?!

### DIFF
--- a/doc/instruction_decode_execute.rst
+++ b/doc/instruction_decode_execute.rst
@@ -64,9 +64,9 @@ Other blocks use the ALU for the following tasks:
 * It computes memory addresses for loads and stores with a Reg + Imm calculation
 * The LSU uses it to increment addresses when performing two accesses to handle an unaligned access
 
-Support for the RISC-V Bitmanipulation Extension is enabled via the parameter ``RV32B``.
+Support for the RISC-V Bitmanipulation Extension (Document Version 0.92, November 8, 2019) is enabled via the parameter ``RV32B``.
 This feature is *EXPERIMENTAL* and the details of its impact are not yet documented here.
-Currently only the Zbb base extension is implemented.
+Currently the Zbb and Zbt sub-extensions are implemented.
 All instructions are carried out in a single clock cycle.
 
 .. _mult-div:
--- a/rtl/ibex_alu.sv
+++ b/rtl/ibex_alu.sv
@@ -13,10 +13,16 @@ module ibex_alu #(
     input  logic [31:0]       operand_a_i,
     input  logic [31:0]       operand_b_i,
 
+    input  logic              instr_first_cycle_i,
+
     input  logic [32:0]       multdiv_operand_a_i,
     input  logic [32:0]       multdiv_operand_b_i,
 
     input  logic              multdiv_sel_i,
+
+    input  logic [31:0]       imd_val_q_i,
+    output logic [31:0]       imd_val_d_o,
+    output logic              imd_val_we_o,
 
     output logic [31:0]       adder_result_o,
     output logic [33:0]       adder_result_ext_o,
@@ -57,12 +63,7 @@ module ibex_alu #(
 
       // MinMax OPs (RV32B Ops)
       ALU_MIN,  ALU_MINU,
-      ALU_MAX,  ALU_MAXU,
-
-      // Logic-with-negate OPs (RV32B Ops)
-      ALU_XNOR,
-      ALU_ORN,
-      ALU_ANDN: adder_op_b_negate = 1'b1;
+      ALU_MAX,  ALU_MAXU: adder_op_b_negate = 1'b1;
 
       default:;
     endcase
@@ -72,8 +73,14 @@ module ibex_alu #(
   assign adder_in_a    = multdiv_sel_i ? multdiv_operand_a_i : {operand_a_i,1'b1};
 
   // prepare operand b
-  assign operand_b_neg = {operand_b_i,1'b0} ^ {33{adder_op_b_negate}};
-  assign adder_in_b    = multdiv_sel_i ? multdiv_operand_b_i : operand_b_neg;
+  assign operand_b_neg = {operand_b_i,1'b0} ^ {33{1'b1}};
+  always_comb begin
+    unique case(1'b1)
+      multdiv_sel_i:     adder_in_b = multdiv_operand_b_i;
+      adder_op_b_negate: adder_in_b = operand_b_neg;
+      default :          adder_in_b = {operand_b_i, 1'b0};
+    endcase
+  end
 
   // actual adder
   assign adder_result_ext_o = $unsigned(adder_in_a) + $unsigned(adder_in_b);
@@ -86,26 +93,85 @@ module ibex_alu #(
   // Shift //
   ///////////
 
+  // The shifter structure consists of a 33-bit shifter: 32-bit operand + 1 bit extension for
+  // arithmetic shifts and one-shift support.
+  // Rotations and funnel shifts are implemented as multi-cycle instructions.
+  // For funnel shifs, operand_a_i is tied to rs1 in the first cycle and rs3 in the
+  // second cycle. operand_b_i is always tied to rs2.
+  //
+  // Rotation pseudocode:
+  //   shift_amt = rs2 & 31;
+  //   multicycle_result = (rs1 >> shift_amt) | (rs1 << (32 - shift_amt));
+  //                       ^-- cycle 0 -----^ ^-- cycle 1 --------------^
+  //
+  // For funnel shifts, the order of applying the shift amount or its complement is determined by
+  // bit [5] of shift_amt.
+  // Funnel shift Pseudocode: (fsl)
+  //  shift_amt = rs2 & 63;
+  //  shift_amt_compl = 32 - shift_amt[4:0]
+  //  if (shift_amt >=33):
+  //     multicycle_result = (rs1 >> shift_amt_cmpl[4:0]) | (rs3 << shift_amt[4:0]);
+  //                         ^-- cycle 0 ---------------^ ^-- cycle 1 ------------^
+  //  else if (shift_amt <= 31 && shift_amt > 0):
+  //     multicycle_result = (rs1 << shift_amt[4:0]) | (rs3 >> shift_amt_compl[4:0]);
+  //                         ^-- cycle 0 ----------^ ^-- cycle 1 -------------------^
+  //  For shift_amt == 0, 32, both shift_amt[4:0] and shift_amt_compl[4:0] == '0.
+  //  these cases need to be handled separately outside the shifting structure:
+  //  else if (shift_amt == 32):
+  //     multicycle_result = rs3
+  //  else if (shift_amt == 0):
+  //     multicycle_result = rs1.
+
   logic       shift_left;
   logic       shift_ones;
   logic       shift_arith;
   logic       shift_rot;
+  logic       shift_funnel;
   logic       shift_none;
   logic       shift_op_rev;
   logic       shift_op_rev8;
   logic       shift_op_orc_b;
-  logic [4:0] shift_amt;
+  logic [5:0] shift_amt;
+  logic [5:0] shift_amt_compl; // complementary shift amount (32 - shift_amt)
+  logic       shift_multicycle;
 
-  assign shift_amt      = operand_b_i[4:0];
+  // bit shift_amt[5]: word swap bit: only considered for FSL/FSR.
+  // if set, reverse operations in first and second cycle.
+  assign shift_amt[5] = operand_b_i[5] && shift_funnel;
+  assign shift_amt_compl = 32 - operand_b_i[4:0];
 
-  assign shift_left     = RV32B ? (operator_i == ALU_ROL) || (operator_i == ALU_SLO) ||
-                                    (operator_i == ALU_SLL) :
-                                  (operator_i == ALU_SLL);
+  assign shift_amt[4:0] = instr_first_cycle_i ?
+      (operand_b_i[5] && shift_funnel ? shift_amt_compl[4:0] : operand_b_i[4:0]) :
+      (operand_b_i[5] && shift_funnel ? operand_b_i[4:0] : shift_amt_compl[4:0]);
+
+  // left shift if this is:
+  // * a standard left shift (slo, sll)
+  // * a rol in the first cycle
+  // * a ror in the second cycle
+  // * fsl: without word-swap bit: first cycle, else: second cycle
+  // * fsr: without word-swap bit: second cycle, else: first cycle
+  always_comb begin
+    unique case (operator_i)
+      ALU_SLL: shift_left = 1'b1;
+      ALU_SLO: shift_left = RV32B ? 1'b1 : 1'b0;
+      ALU_ROL: shift_left = RV32B ? instr_first_cycle_i : 0;
+      ALU_ROR: shift_left = RV32B ? !instr_first_cycle_i : 0;
+      ALU_FSL: shift_left =
+          RV32B ? (shift_amt[5] ? !instr_first_cycle_i : instr_first_cycle_i) : 1'b0;
+      ALU_FSR: shift_left =
+          RV32B ? (shift_amt[5] ? instr_first_cycle_i : !instr_first_cycle_i) : 1'b0;
+      default: shift_left = 1'b0;
+    endcase
+  end
+
   assign shift_ones     = RV32B ? (operator_i == ALU_SLO) || (operator_i == ALU_SRO) : 1'b0;
   assign shift_arith    = (operator_i == ALU_SRA);
   assign shift_rot      = RV32B ? (operator_i == ALU_ROL) || (operator_i == ALU_ROR) : 1'b0;
+  assign shift_funnel   = RV32B ? (operator_i == ALU_FSL) || (operator_i == ALU_FSR) : 1'b0;
+  assign shift_multicycle = shift_funnel || shift_rot;
+
   assign shift_none     = RV32B ? (operator_i == ALU_REV)  || (operator_i == ALU_REV8) ||
-                                    (operator_i == ALU_ORCB) :
+                                      (operator_i == ALU_ORCB) :
                                   1'b0;
 
   assign shift_op_rev   = RV32B ? (operator_i == ALU_REV) : 1'b0;
@@ -113,11 +179,7 @@ module ibex_alu #(
   assign shift_op_orc_b = RV32B ? (operator_i == ALU_ORCB) : 1'b0;
 
   logic [31:0] shift_result;
-  logic [31:0] shift_extension_rvb;
-  logic        shift_extension;
   logic [32:0] shift_result_ext;
-  logic [63:0] shift_result_ext_rvb;
-
 
   always_comb begin
     shift_result = operand_a_i;
@@ -127,27 +189,15 @@ module ibex_alu #(
       shift_result = operand_a_rev;
     end
 
-    if (RV32B) begin
-      // rotation: extend with a copy of the operand
-      // shift-ones: extend with ones
-      // arithmetic shift: sign-extend.
-      // else: zero-extend.
-      shift_extension_rvb = shift_rot ?
-                            shift_result :
-                            {32{shift_ones || (shift_arith && operand_a_i[31])}};
-
-      shift_result_ext_rvb = {shift_extension_rvb, shift_result} >> shift_amt;
-    end else begin
-      shift_extension  = (shift_arith && shift_result[31]);
-      shift_result_ext = $signed({shift_extension, shift_result}) >>> shift_amt;
-    end
+    shift_result_ext = $signed({shift_ones || (shift_arith && shift_result[31]), shift_result})
+        >>> shift_amt[4:0];
 
     // shift, if this is a shift operation
     if (!shift_none) begin
-        shift_result = RV32B ? shift_result_ext_rvb[31:0] : shift_result_ext[31:0];
+      shift_result = shift_result_ext[31:0];
     end
 
-    // shift left: bytewise reverse. (orcomnine with '0)
+    // shift left: bytewise reverse. (orcombine with '0)
     // orc_b: bytewise reverse and orcombine.
     if (shift_op_orc_b || shift_left) begin
       shift_result = (shift_op_orc_b ? shift_result : 32'h 0) |
@@ -182,7 +232,6 @@ module ibex_alu #(
   logic cmp_signed;
 
   always_comb begin
-    cmp_signed = 1'b0;
     unique case (operator_i)
       ALU_GE,
       ALU_LT,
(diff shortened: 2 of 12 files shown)


ganoam:
Hi everyone, 
A first version of the ALU featuring both zbb and zbt instructions.

Impact on timing and area is as of yet quite significant [Syntheis Report Sheets](https://docs.google.com/spreadsheets/d/1dDDtzMgS5quMXtv-8z2uD--jBIHMSnaWrqj1EIDp1Nw/edit#gid=1507987886).

A point that came to mind only when I was writing the Commit message is, Instead of creating a new register in the id stage we would probably better share the mac register from the multiplier. Or maybe make the multiplier use the one in the id-stage. 

I will try that next week.

rtl/ibex_decoder.sv
```diff
@@ -781,7 +794,17 @@ module ibex_decoder #(
                     alu_operator_o = ALU_ORCB;        // Byte-wise Reverse and Or-Combine
                   end
                 end
-                default: ;
+                default: begin
```
vogelpi: I would prefer not to map this logic into the default but rather map an `if/else` around the original `unique case`. We know that some tools like Altera/Intel Quartus just "optimize" everything inside the default. I am not saying we need to be compatible but if we are aware of problems it's is better to avoid them in the first place.

But I am open, would be interested to hear other people's voices, maybe @GregAC, @tomroberts-lowrisc or @imphil ?

rtl/ibex_decoder.sv
```diff
@@ -362,7 +366,11 @@ module ibex_decoder #(
                   default: illegal_insn = 1'b1;
                 endcase
               end
-              default: illegal_insn = 1'b1;
+
+              default: begin
+                illegal_insn = RV32B ? !instr[26] : 1'b1;         // fsri
```
vogelpi: Also here, I would put the logic in a `if/else` around the original `unique case` (see more details below).

rtl/ibex_decoder.sv
```diff
@@ -846,7 +869,49 @@ module ibex_decoder #(
             alu_operator_o        = ALU_ADD;
           end
 
-          default: ;
+          default: begin
```
vogelpi: Again I would not put this logic in the default. But before changing everything, let's hear other peoples voices.

rtl/ibex_alu.sv
```diff
@@ -128,26 +137,46 @@ module ibex_alu #(
...
+          end else begin
+            shift_op_lsb = multicycle_op_stage_reg_i;
+            shift_op_msb = shift_result;
+          end
+        end
+
+        default: begin
+          shift_op_lsb = shift_result;
+          shift_op_msb = {32{shift_ones || (shift_arith && operand_a_i[31])}};
+        end
```
vogelpi: If you move stuff into the default, a comment would be helpful.

rtl/ibex_alu.sv
```diff
@@ -128,26 +137,46 @@ module ibex_alu #(
...
+      endcase
+
+      shift_result_ext = {shift_op_msb, shift_op_lsb} >> shift_amt[4:0];
```
tomroberts-lowrisc: Just a random thought, but would it be possible to do the shift over two cycles?
From the spec for `fsl`:
```
shamt ? (A << shamt) |  (B >> (XLEN - shamt)) : A
```
Could shift the A operand in the first cycle, then do the B shift and OR in the second.

I am wondering how much of the area increase is the 64bit shifter (could do a quick experiment just to see what the area looks like with everything else the same, but without the 64bit shifter).

ganoam: yes, we could do this of course. 

The 64 bit shifter is actually already used in the zbb extension. There we use it for rotate shifts. The impact of the 64-bit shifter in contrast to the 32 bit shifter can be seen by comparing rows 60/61 with 64/65:

60/61: zbb not implemented, (so also no rol/ror), but shifter implemented something like:
```
shift_out = {32{msb}, operand[31:0]} >> shift_amt;
```
64/65:
```
shift_out = {msb, operand[31:0]} >>> shift_amt;
```

In this example there is a saving of 300-400 luts. For fsl/fsr, we would however have to compute `32-shift_amt` as well (we could of course share the adder). 

Also, rol/rot would become 2-cycle instructions.

I will try that

rtl/ibex_decoder.sv
ganoam: Hi, yes I remember we already talked about this. I will change it. However, there are already cases of this structure on the master branch for example in the decoder:
image
maybe these should be corrected as well then.

vogelpi: These are fine. You can leave them as is.

rtl/ibex_pkg.sv
```diff
@@ -86,7 +86,15 @@ typedef enum logic [5:0] {
...
-  ALU_SLTU
+  ALU_SLTU,
+
+  // Ternary Bitmanip Operations
+  // RV32B
+  ALU_CMOV,
+  ALU_CMIX,
+  ALU_FSL,
+  ALU_FSR
+
```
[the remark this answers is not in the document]
ganoam: done. thanks

rtl/ibex_id_stage.sv
```diff
@@ -410,6 +435,8 @@ module ibex_id_stage #(
       .alu_operator_o                  ( alu_operator         ),
       .alu_op_a_mux_sel_o              ( alu_op_a_mux_sel_dec ),
       .alu_op_b_mux_sel_o              ( alu_op_b_mux_sel_dec ),
+      .alu_multicycle_o                ( alu_multicycle_dec   ),
+
```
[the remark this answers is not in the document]
ganoam: Done, thanks!

rtl/ibex_core.sv
```diff
@@ -18,7 +18,7 @@ module ibex_core #(
     parameter int unsigned MHPMCounterWidth         = 40,
     parameter bit          RV32E                    = 1'b0,
     parameter bit          RV32M                    = 1'b1,
-    parameter bit          RV32B                    = 1'b0,
+    parameter bit          RV32B                    = 1'b1,
```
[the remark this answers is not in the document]
ganoam: Yup, thanks.

rtl/ibex_core.sv
```diff
@@ -751,6 +768,10 @@ module ibex_core #(
   assign rvfi_rs1_data_id = id_stage_i.rf_rdata_a_fwd;
   assign rvfi_rs2_addr_id = rf_raddr_b;
   assign rvfi_rs2_data_id = id_stage_i.rf_rdata_b_fwd;
+  assign rvfi_rs3_addr_id = rf_raddr_a;
+  assign rvfi_rs3_data_id = id_stage_i.rf_rdata_a_fwd;
+
+
```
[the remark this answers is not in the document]
ganoam: Done, thanks.

rtl/ibex_core.sv
```diff
@@ -285,12 +289,16 @@ module ibex_core #(
   logic [31:0] rvfi_insn_id;
   logic [4:0]  rvfi_rs1_addr_id;
   logic [4:0]  rvfi_rs2_addr_id;
+  logic [4:0]  rvfi_rs3_addr_id;
```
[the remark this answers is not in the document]
ganoam: :)

rtl/ibex_decoder.sv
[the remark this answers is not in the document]
ganoam: done. thanks

rtl/ibex_decoder.sv
ganoam: done. thanks

rtl/ibex_alu.sv
ganoam: this has changed.

rtl/ibex_decoder.sv
ganoam: done, thanks

rtl/ibex_decoder.sv
ganoam: done.

ganoam:
Update: After lots of experiments, I recommend this version for merge. 

As already discussed, all ternary instructions (funnel shifts, cmix and cmov) are implemented with 1 cycle of latency due to only two register file ports. 

A 33 bit shifter is implemented (32 bit operand + 1 for arithmetic and one-shifts). Accordingly, Rotations as well as funnel shifts are implemented as multi-cycle instructions. A

This commit includes changes to the multdiv module, moving the intermediate value register into the id_stage module, allowing it to be shared with the ALU.

The synthesis results of this run are listed [here](https://docs.google.com/spreadsheets/d/1dDDtzMgS5quMXtv-8z2uD--jBIHMSnaWrqj1EIDp1Nw/edit#gid=1890157149) (-> sheet: zbt-pr-candidates, -> design: 33bitshifter_adderbinmux_optimized). 

During analysis of my various designs, I have repeatedly found a path among the most critical paths, which should not be present. Among others it is present [here](https://github.com/ganoam/ibex/blob/bitmanip-devel-zbt/reports/bitmanip-zbt-pr/33bitshifter_adderbinmux_optimized/RV32B0_3stage_btalu_2_ns/post_synth_timing.rpt) as the critical path. It leads from [...] register_file -> ***multdiv*** -> alu -> ex_block/branch_decision_o -> controller -> [...]
As far as I can tell, it can be removed by feeding the early `branch_in_dec` signal from the id_stage into ibex_controller and masking branch_set_i in line 414:
```
if (((branch_in_dec_i && branch_set_i) || jump_set_i) && ~special_req_branch) begin
```
￼
thanks @vogelpi  for the hint.
As this is not related to this PR, does not really impact timing or area of my designs and might need further investigation, I have not included this change in this PR.

rtl/ibex_ex_block.sv
```diff
@@ -37,6 +38,11 @@ module ibex_ex_block #(
     input  logic [31:0]           multdiv_operand_b_i,
     input  logic                  multdiv_ready_id_i,
 
+    // intermediate val reg
+    output logic                  intermediate_val_reg_en_o,
+    output logic [33:0]           intermediate_val_reg_in_o,
+    input  logic [33:0]           intermediate_val_reg_out_i,
```
vogelpi, line 44: Having `_in_o` and `_out_i` is not ideal. What about switching to:
```
imd_val_we_o
imd_val_d_o,
imd_val_q_i
```
?

rtl/ibex_alu.sv
```diff
@@ -319,6 +443,10 @@ module ibex_alu #(
     assign minmax_result   = '0;
     assign bitcnt_result   = '0;
     assign pack_result     = '0;
+    assign multicycle_result  = '0;
+    // Rvb support signals
```
vogelpi, line 447: RV32B

rtl/ibex_ex_block.sv
```diff
@@ -55,6 +61,19 @@ module ibex_ex_block #(
...
   logic        multdiv_en;
+  logic [31:0] alu_intermediate_val_reg_in;
+  logic        alu_intermediate_val_reg_en;
+  logic [33:0] multdiv_intermediate_val_reg_in;
+  logic        multdiv_intermediate_val_reg_en;
+
+  // Intermediate Value Register Mux
+
+  assign intermediate_val_reg_in_o =
+      multdiv_en ? multdiv_intermediate_val_reg_in : {2'b0, alu_intermediate_val_reg_in};
```
vogelpi, line 72: If possible, I would use the signal `multdiv_sel_i` instead of `multdiv_en`. The former comes directly out of the decoder, the later is a combination many signals in ID stage and controller (it factors in if the instruction is still executed or not). I think this is irrelevant here but might lead to unnecessarily long paths.

rtl/ibex_ex_block.sv
```diff
@@ -134,46 +157,52 @@ module ibex_ex_block #(
...
+        .alu_adder_i           ( alu_adder_result_ex_o           ),
+        .equal_to_zero         ( alu_is_equal_result             ),
+        .intermediate_val_i    ( intermediate_val_reg_out_i      ),
+        .intermediate_val_o    ( multdiv_intermediate_val_reg_in ),
+        .intermediate_val_en_o ( multdiv_intermediate_val_reg_en ),
+        .multdiv_ready_id_i    ( multdiv_ready_id_i              ),
+        .valid_o               ( multdiv_valid                   ),
+        .multdiv_result_o      ( multdiv_result                  )
     );
   end
```
vogelpi, line 208: There a is an assignment of `ex_valid_o` at the bottom that should also be adapted, as now also ALU instructions can be multicycle.

rtl/ibex_alu.sv
```diff
@@ -86,38 +95,90 @@ module ibex_alu #(
...
+  // Rotations and funnel shifts are implemented as multi-cycle instructions.
+  // For funnel shifs, operand_a_i is tied to rs1 in the first cycle and rs3 in the
+  // second cycle. operand_b_i is always tied to rs2.
+  //
+  // Rotation pseudocode:
+  //   shift_amt = rs2 & 31;
+  //   multicycle_result = (rs1 >> shift_amt) | (rs1 << (32 - shift_amt));
+  //                       ^-- cycle 0 -----^ ^-- cycle 1 --------------^
+  //
+  // For funnel shifts, the order of applying the shift amount or its compliment is determined by
```
rswarbrick, line 109: Nit: "complement"

rtl/ibex_id_stage.sv
```diff
@@ -341,6 +350,20 @@ module ibex_id_stage #(
...
+  always_ff @(posedge clk_i or negedge rst_ni) begin : intermediate_val_reg
+    if (!rst_ni) begin
+      intermediate_val_reg_q <= '0;
+    end else if (intermediate_val_reg_en_ex_i) begin
+      intermediate_val_reg_q <= intermediate_val_reg_in_ex_i;
+    end
+  end
+
+  assign intermediate_val_reg_out_ex_o = intermediate_val_reg_q;
```
gautschimi, line 365: assuming a bit manipulation instruction is writing this shared intermediate_val_reg. can it be that this will lead to unnecessary switching activity in the mulitplier? (even though no multiplication is currently carried out)

functionally it does not matter, but in terms of power consumption it could be worth to silence the inputs to the multiplier when it is not used. e.g. something like:
```
assign intermediate_val_reg_mult_o = mult_en ? intermediate_val_reg_q : '0;
assign intermediate_val_reg_alu_o = alu_en ?  intermediate_val_reg_q : '0;
```

doc/instruction_decode_execute.rst
```diff
@@ -64,7 +64,7 @@ Other blocks use the ALU for the following tasks:
 * It computes memory addresses for loads and stores with a Reg + Imm calculation
 * The LSU uses it to increment addresses when performing two accesses to handle an unaligned access
 
-Support for the RISC-V Bitmanipulation Extension is enabled via the parameter ``RV32B``.
+Support for the RISC-V Bitmanipulation Extension (Document Verision draft, November 30, 2019) is enabled via the parameter ``RV32B``.
```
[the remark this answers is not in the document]
ganoam: Yes. Done, thanks.

rtl/ibex_alu.sv
```diff
@@ -86,38 +95,90 @@ module ibex_alu #(
...
+
+  // left shift if this is:
+  // * a standard left shift (slo, sll)
+  // * a rol in the first cycle
+  // * a ror in the second cycle
+  // * fsl: without word-swap bit: first cycle, else: second cycle
+  // * fsr: without word-swap bit: second cycle, else: first cycle
+  always_comb begin
+    shift_left = 1'b0;
+    case (operator_i)
```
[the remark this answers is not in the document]
ganoam: done, thanks.

rtl/ibex_alu.sv
```diff
@@ -72,8 +75,14 @@ module ibex_alu #(
...
 
   // prepare operand b
-  assign operand_b_neg = {operand_b_i,1'b0} ^ {33{adder_op_b_negate}};
-  assign adder_in_b    = multdiv_sel_i ? multdiv_operand_b_i : operand_b_neg;
+  assign operand_b_neg = {operand_b_i,1'b0} ^ {33{1'b1}};
+  always_comb begin
+    unique case(1'b1)
+      multdiv_sel_i: adder_in_b = multdiv_operand_b_i;
+      adder_op_b_negate: adder_in_b = operand_b_neg;
+      default : adder_in_b = {operand_b_i, 1'b0};
```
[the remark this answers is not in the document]
ganoam: Done, thanks.

rtl/ibex_decoder.sv
```diff
@@ -593,6 +605,9 @@ module ibex_decoder #(
 
     opcode_alu         = opcode_e'(instr_alu[6:0]);
 
+    use_rs3  = 1'b0;
+    alu_multicycle_o        = 1'b0;
```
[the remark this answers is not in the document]
ganoam: Done, thanks.

rtl/ibex_alu.sv
```diff
@@ -86,38 +95,90 @@ module ibex_alu #(
...
+  //  shift_amt = rs2 & 63;
+  //  if (shift_amt >=33):
+  //     multicycle_result = (rs1 >> (32 - shift_amt[4:0]) | (rs3 << shift_amt[4:0]);
+  //                         ^-- cycle 0 ----------------^ ^-- cycle 1 ------------^
+  //  else if (shift_amt == 32):
+  //     // this case is handled separately outside the shifting structure. (shift_amt[4:0]
+  //     multicycle_result = rs3
+  //  else:
+  //     multicycle_result = (rs1 << shift_amt[4:0]) | (rs3 >> (32 - shift_amt[4:0]));
+  //                         ^-- cycle 0 ----------^ ^-- cycle 1 --------------------^
```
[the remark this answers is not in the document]
ganoam: Thanks :) I still had something to clarify.

rtl/ibex_ex_block.sv
ganoam: Yup. I like that a lot.

rtl/ibex_core.sv
```diff
@@ -106,17 +108,20 @@ module ibex_core #(
...
-  logic [31:0] pc_wb;                  // Program counter in WB stage
+  logic        instr_fetch_err;                // Bus error on instr fetch
+  logic        instr_fetch_err_plus2;          // Instruction error is misaligned
+  logic        illegal_c_insn_id;              // Illegal compressed instruction sent to ID stage
+  logic [31:0] pc_if;                          // Program counter in IF stage
+  logic [31:0] pc_id;                          // Program counter in ID stage
+  logic [31:0] pc_wb;                          // Program counter in WB stage
+  logic [33:0] intermediate_val_reg_in_ex;     // Intermediate register for multicycle Ops
+  logic [33:0] intermediate_val_reg_out_ex;    // Intermediate register for multicycle Ops
+  logic        intermediate_val_reg_en_ex;
```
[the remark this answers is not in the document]
ganoam: I changed it according to your next comment.

rtl/ibex_alu.sv
ganoam: Done, thanks.

rtl/ibex_ex_block.sv
ganoam: Yes, good Idea, thanks.

rtl/ibex_ex_block.sv
ganoam: Hm is it alright to use `alu_imd_val_we` for that, i.e. the reg enable for the intermediate register? That's a direct indicator for the ALU being in the first cycle of a multicycle instruction.
Otherwise I would run out a dedicated signal.

rtl/ibex_alu.sv
ganoam: Done, thanks

rtl/ibex_id_stage.sv
ganoam: Yes, I think that will happen. It would affect lines 361-376. 

But also, the same could happen due to `adder_result_ext_o` from the ALU, which is probably much more active. Should I do that as well while I am at it?

rtl/ibex_decoder.sv
```diff
@@ -135,7 +139,7 @@ module ibex_decoder #(
   assign zimm_rs1_type_o = { 27'b0, instr[`REG_S1] }; // rs1
 
   // source registers
-  assign rf_raddr_a_o = instr[`REG_S1]; // rs1
+  assign rf_raddr_a_o = use_rs3 ? instr[`REG_S3] : instr[`REG_S1]; // rs3 / rs1
```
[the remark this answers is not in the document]
ganoam: done.

rtl/ibex_ex_block.sv
vogelpi: That should work. You could do something like:
```
  assign ex_valid_o = multdiv_sel    ? multdiv_valid : 
                      alu_imd_val_we ? 1'b0          : 1'b1;
```

ganoam: In the latest version, I have done it so:
```
assign ex_valid = muldiv_sel ? multdiv_valid  : !alu_imd_val_we;
```
Which does the same thing.

rtl/ibex_id_stage.sv
gautschimi: I think it would make sense. it would be good to make sure that the multiplier is not switching while the ALU or LSU is active, and vice versa. but it might better to address these issues in a separate merge request as it is not really related to bit manipulation support

vogelpi: That is a good point!

But I would not address this in this current PR because:
- Silencing these inputs might lead to additional muxes or mux levels. Some of them might be on the critical path. We should carefully check where to best do the silencing. In the given example, the silencing is best done inside the multdiv block: the `imd_val_q` signal feeds into two other signals. For both of them, there already exist muxes to at least partially set them to 0. So it is more about making sure we use the correct input of exisiting muxes when idle. 
- We should check also other places in the design (LSU, other multdiv inputs).

I will create an issue to track this.

vogelpi: The issue is here: #783

ganoam:
Last minute correction: Mention zbt in documentation.

PR #749, opened 2020-03-30
[dv] fix timing issue in single_step test

udinator:
In the `riscv_debug_single_step_test` currently, we wait for 2000 clocks after each round of single stepping has been completed, which was previously fine since this meant that debug requests would start being sent out immediately at the very end of the `<init>` code section (which is the very beginning of the `<main>` code section).
This no longer applies however, since due to a restructuring of the generated program to enable PMP, there is now a jump instruction at the very end of the `<init>` section.  As the single stepping test cannot handle single stepping over branch/jump instructions as of right now, we must ensure to wait until at least this instruction has retired to start sending debug requests.
Resultantly, the `clk_vif.wait_clks(2000)` function call has now been moved to the start of each round of single stepping, which keeps the waiting time between each round the same, but provides the added effect of guaranteeing that this initial jump instruction is skipped over.


udinator:
SGTM, I will edit the commit message to give an explanation of this modification.

PR #750, opened 2020-03-31
[dv] Define `UVM for VCS sim

GregAC:
Without `UVM defined assertion failures don't result in regression
failure as they do not produce UVM reports.

### DIFF
--- a/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
+++ b/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
@@ -18,6 +18,7 @@
       - "vcs -f ibex_dv.f  -full64
          -l <out>/compile.log
          -sverilog -ntb_opts uvm-1.2
+         +define+UVM
          +define+UVM_REGEX_NO_DPI -timescale=1ns/10ps -licqueue
          -LDFLAGS '-Wl,--no-as-needed'
          -Mdir=<out>/vcs_simv.csrc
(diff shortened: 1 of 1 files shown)


GregAC:
I think they have, that is a good point actually, I should kick off a regression and see what's failing (have only done a couple of limited tests with this PR)!

imphil:
Can you either do a blind modification for the other simulators as well, or open an issue to get xcelium and riviera pro modified as well?

GregAC:
I note others simulators all seem to have specific UVM options already, this may give you the `define UVM implicitly, IUS has `-uvm`, riviera pro has `-l uvm_1_2`, qrun has `-uvmhome uvm-1.2`. dsim doesn't seem to have a clear uvm specific option.

So I'd be slightly wary of adding it blindly, though I can't see it would cause an issue.

@udinator any thoughts?

GregAC:
I've run a regression and don't see any issues (haven't looked at every failure, but similar failure rate before this change and those I have looked at aren't assertion failures).

udinator:
Apologies for seeing this PR so late, the change for VCS LGTM.

With respect to the other simulators, do those options essentially just define the `UVM` macro from the command line? I'm not too familiar with all of their options...

udinator:
Update: seems like IUS does not support the use of `$clog2` to define `localparam` values, so I'm unable to even compile the RTL files (IUS stopped being supported by Cadence back in 2016 after Xcelium was released!) to test my previous theory.
My hunch is still that we would need to manually define the `UVM` macro from the command line options, but feel free to correct me if this is not the case.

GregAC:
I'm happy to go with your hunch. I've updated the PR to define UVM across all of the simulator compiles.

imphil:
IUS isn't supported for our code base anyways due to missing language feature, so please file a follow-up to get it properly removed, and don't worry about it.

PR #752, opened 2020-03-31
Rv32b ci

GregAC:


dv/riscv_compliance/ibex_riscv_compliance.core
```diff
@@ -34,6 +34,11 @@ parameters:
     paramtype: vlogparam
     default: false
     description: Enable the E ISA extension (reduced register set)
+  RV32B:
+    datatype: bool
+    paramtype: vlogparam
+    default: false
+    description: Enable the B ISA extension (bit manipulation EXPERIMENTAL) [0/1]
```
imphil: The 0/1 is wrong/confusing, please remove it.

I know this goes slightly beyond this PR, but can you align the descriptions of these parameters between the files? `ibex_simple_system.core` has the `[0/1]` e.g. for the RV32E parameter, which should be removed as well.

ibex_configs.yaml
```diff
@@ -20,6 +21,16 @@ small-3cmult:
 maxperf-1cmult:
   RV32E                    : False
   RV32M                    : True
+  RV32B                    : False
+  BranchTargetALU          : True
+  WritebackStage           : True
+  MultiplierImplementation : "single-cycle"
+
+# maxpref-1cmult config above with bitmanip extension
```
imphil: Since we're planning to advertise these configurations as "supported configurations": Maybe add an explicit "experimental-" or something like that to the config names? (I know we don't do that for the third pipeline stage currently, but at least there the implementation is complete, while for bitmanip we're very early and very incomplete at this point.)

dv/riscv_compliance/ibex_riscv_compliance.core
GregAC: Ah whoops, copy/paste issue.

For this PR I was just aiming to match the existing style across .core files. We can clean up as part of https://github.com/lowRISC/ibex/issues/753

ibex_configs.yaml
GregAC: As we hadn't really documented much of this yet or official put down what it means to be 'supported' I was intentionally being quite vague.

I'll put both the maxperf configurations under a comment calling them 'EXPERIMENTAL CONFIGURATIONS', and add 'experimental-' as you suggest. That along with the other EXPERIMENTAL labels and lack of documentation should give the correct impression for now.

PR #711, opened 2020-03-17
[dv] initial icache testbench

udinator:
The initial commit of the testbench for the ibex_icache module generated by the `uvmdvgen` tool.
The testbench files and .hjson supporting files for `dvsim` tool are all contained within `dv/uvm/icache/[data/doc/dv]`.

There are several things to note with this PR:
1) A directory `vendor/lowrisc_ip` has been created, which contains the `dv_base` testbench library, the `dv_utils` package, and various common tools, including the `uvmdvgen` tool for auto-generating testbenches and most importantly, the `dvsim` tool by which to actually simulate this icache testbench. For the time being, these tools have been manually copied over from the lowRISC/opentitan repository until they are spun off into a separate repo themselves. These tools and utilities from lowRISC/opentitan are the bulk of this PR and should not warrant a review.
2) Several fusesoc .core files have been created in the existing Ibex testbench environment and agent directories, these are required since the icache testbench will use the currently existing ibex memory interface agent. These core files are dual purpose, as they will also be used when the Ibex testbenches switches over to using the `dvsim` .hjson based flow.
3) To run builds and simulations, the `dvsim` tool must be invoked from the `ibex/` top level directory.

This PR provides all of the testbench components and run flow utilities needed to compile the icache testbench and run tests once they have been implemented.

Signed-off-by: Udi <udij@google.com>

PR #754, opened 2020-03-31
Fill out an initial DV plan for the icache

rswarbrick:
This currently has quite a lot of TBD sections, but should hopefully
describe my plan for the UVM testing reasonably well.


rswarbrick:
This took a few days since PR #711 landed because it seemed much easier to make sure I had the UVM hierarchy right by writing some code. So far, I have the core and memory agents implemented (but no error checking, so it's not exactly a testbench yet).

rswarbrick:
Handy link for Github's rendering of the markdown: https://github.com/rswarbrick/ibex/blob/icache-dv-plan/dv/uvm/icache/doc/ibex_icache_dv_plan.md

dv/uvm/icache/doc/tb.svg
```diff
@@ -1 +0,0 @@
....109375 -0.59375q1.796875 0 2.828125 1.515625l0 -1.296875l1.46875 0l0 13.375l-1.640625 0zm-5.046875 -8.5625q0 1.859375 0.78125 2.796875q0.78125 0.9375 1.875 0.9375q1.046875 0 1.796875 -0.890625q0.765625 -0.890625 0.765625 -2.703125q0 -1.9375 -0.796875 -2.90625q-0.796875 -0.96875 -1.875 -0.96875q-1.0625 0 -1.8125 0.90625q-0.734375 0.90625 -0.734375 2.828125zm17.574646 8.78125q-1.359375 -1.703125 -2.296875 -4.0q-0.9375 -2.296875 -0.9375 -4.765625q0 -2.15625 0.703125 -4.140625q0.828125 -2.3125 2.53125 -4.59375l1.171875 0q-1.09375 1.890625 -1.453125 2.703125q-0.546875 1.25 -0.875 2.625q-0.390625 1.703125 -0.390625 3.421875q0 4.375 2.71875 8.75l-1.171875 0zm9.712677 -7.03125l1.6875 0.203125q-0.40625 1.484375 -1.484375 2.3125q-1.078125 0.8125 -2.765625 0.8125q-2.125 0 -3.375 -1.296875q-1.234375 -1.3125 -1.234375 -3.671875q0 -2.453125 1.25 -3.796875q1.265625 -1.34375 3.265625 -1.34375q1.9375 0 3.15625 1.328125q1.234375 1.3125 1.234375 3.703125q0 0.15625 0 0.4375l-7.21875 0q0.09375 1.59375 0.90625 2.453125q0.8125 0.84375 2.015625 0.84375q0.90625 0 1.546875 -0.46875q0.640625 -0.484375 1.015625 -1.515625zm-5.390625 -2.65625l5.40625 0q-0.109375 -1.21875 -0.625 -1.828125q-0.78125 -0.953125 -2.
```
udinator: This testbench diagram will need to remain, and should be modified to reflect the full icache testbench; for a reference diagram, you can look at [the UART tb.svg diagram](https://github.com/lowRISC/opentitan/blob/master/hw/ip/uart/doc/dv_plan/tb.svg)

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
 * **DV**
-  * Verify all IBEX_ICACHE IP features by running dynamic simulations with a SV/UVM based testbench
-  * Develop and run all tests based on the [testplan](#testplan) below towards closing code and functional coverage on the IP and all of its sub-modules
+  * Verify all Ibex ICache IP features by running dynamic simulations with a SV/UVM based testbench
+  * Develop and run all tests based on the [testplan](#testplan) below towards closing code and functional coverage.
 * **FPV**
-
-## Current status
-<!-- TODO for now -->
-* [Design & verification stage]({{< relref "doc/project/hw_dashboard" >}})
```
udinator: this should link to [opentitan/doc/project/hw_dashboard](https://github.com/lowRISC/opentitan/blob/master/doc/project/hw_dashboard.md), to conform with all other DV plan documents

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
-  * [HW development stages]({{< relref "doc/project/hw_stages" >}})
```
udinator: similarly, this should link to [opentitan/doc/project/hw_stages.md](https://github.com/lowRISC/opentitan/blob/master/doc/project/hw_stages.md) to conform with all other DV plan documents

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
-* [Simulation results](https://reports.opentitan.org/hw/ip/ibex_icache/dv/latest/results.html)
```
udinator: we're going to want to keep this link in this dv plan, as once you start running regressions with dvsim, the sim results (and coverage too!) will be made publicly visible by being uploaded to this link at the opentitan web server.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: This brings up another question, which I think would be more directed towards @tomroberts-lowrisc and @sjgitty - are we going to have a hw_dashboard entry for the Icache design side?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: As a fully rendered example that you can look at, see the [AES dv plan](https://docs.opentitan.org/hw/ip/aes/doc/dv_plan/).

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+  * [TBD]
 
 ## Design features
-For detailed information on IBEX_ICACHE design features, please see the [IBEX_ICACHE technical specification]({{< relref "doc/icache.rst" >}}).
+
+The ICache design is documented in the [Instruction Cache](https://ibex-core.readthedocs.io/en/latest/icache.html) section of the Ibex User Manual.
 
 ## Testbench architecture
-IBEX_ICACHE testbench has been constructed based on the [DV_LIB testbench architecture]({{< relref "vendor/lowrisc_ip/dv_lib/" >}}).
```
udinator: I think a short blurb referencing the dv_lib tb architecture should be kept in the dv plan, @sriyerg what do you think?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
 
-### Block diagram
-![Block diagram](tb.svg)
```
udinator: This `tb.svg` diagram should be kept in this document, and be modified to reflect Icache testbench, the AES dv plan linked above is an example of what this should look like once fully rendered.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+Since this acts as a passive slave (rather than the core agent, which is an active master), we split this out into a very simple separate agent as well.
+
+![Agent connections](agent-connections.svg)
+
+All three agents report events to the scoreboard.
+The core agent reports every branch sent to the cache, invalidation requests, enabling/disabling the cache and every instruction fetched (address and contents).
+The memory agent reports every change of seed (see the [Memory Agent](#memory-agent) section below).
+The busy agent reports every edge of the busy signal.
+
+### Agents
```
udinator: Since it appears that a new agent will be constructed for the memory interface, and the naming has been changed on the existing icache core interface, I think it might be necessary to re-generate the testbench with the `uvmdvgen` tool (or you can simply replace all references to the `ibex_mem_intf_...` components with your new memory agent, and update the naming everwhere).

asb:
Thanks for all your feedback on this PR Udi. Just as a meta-point, I think Rupert took the approach of (for now) avoiding the custom syntax that relies on tooling that only exists in the OpenTitan repo. There's a discussion to be had about that (is that the right path? should we instead import OT tooling? Or instead commit this DV plan to OT?). But in order to keep things focused, I propose we park that question and circle back to it after agreeing the substance of the plan.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+The testbench intentionally avoids knowing detailed information about the cache's performance characteristics (for example, cache size, line size or number of ways).
+This means that the testbench cannot compare the DUT with a reference model, nor can it model the exact requests that the DUT will make of instruction memory: the whole point of a cache is that it might avoid an instruction fetch.
+
+The cache has two main interfaces, which face the core and instruction bus respectively.
+We model this with two agents: the *core agent* and the *memory agent*.
+The core agent will emulate a core making instruction fetch requests from the cache and the memory agent will emulate the instruction bus.
+
+In fact, there's one more logical interface: the `busy_o` signal from the cache.
+This signal is passed up to top-level in the design and warns the chip not to clock-gate or reset the core (because there are bus transactions in flight, or a memory invalidation in progress).
+Since this acts as a passive slave (rather than the core agent, which is an active master), we split this out into a very simple separate agent as well.
```
udinator: If I'm understanding this, there will be a simple agent for just this one signal?
Would another option be to either include the `busy_o` signal in the iCache core interface or pass a generic interface for this signal into the icache core agent, since this signal is simply used to signal in-flight transactions or invalidations?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+
+The memory agent emulates the instruction bus, supplying data to the cache.
+This must be deterministic (in order for the scoreboard to tell whether the cache fetched the right data), but must also be able to change with time (in order to check invalidation works correctly).
+To support this, the memory contents are modelled with a 32-bit seed value.
+
+The architectural state of the instruction bus and memory (contents at each address; PMP ranges; ranges that will cause a memory error) are all derived from this seed.
+The precise functions can be found in [`dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv`](https://github.com/lowRISC/ibex/blob/master/dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv).
+
+The memory agent is an active slave, responding to instruction fetches from the cache with either a PMP error (on the same cycle as the request) or instruction data (with an in-order request pipeline).
+As well as the instruction fetch requests, its sequence contains occasional transactions that change the underlying seed.
```
udinator: It sounds like when the seed is first set, the memory model (plus pmp regions, etc...) will be randomly generated based on this value, correct? Does this then mean that when the cache is invalidated due to `fence.i` reset the entire memory model will be re-generated due to the seed of the memory model changing?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+These transactions are reported to the scoreboard (so it can know the set of possible seeds for any given fetch).
+To avoid too many valid seeds in the scoreboard, this should happen at a similar rate to cache invalidations in the core agent.
```
udinator: Will the memory sequence change the underlying seed outside of invalidation/reset scenarios?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
 
 ### Self-checking strategy
 #### Scoreboard
-The `ibex_icache_scoreboard` is primarily used for end to end checking.
-It creates the following analysis ports to retrieve the data monitored by corresponding interface agents:
-* analysis port1:
-* analysis port2:
-<!-- explain inputs monitored, flow of data and outputs checked -->
+The `ibex_icache_scoreboard` is used to check that instructions fetched from the cache lie at the correct addresses and contain the correct data.
+It has two analysis ports: one for the memory agent and another for the core agent.
```
udinator: If the `busy_o` agent is not going to communicate with the scoreboard, is there a need for a full agent to be developed around this one signal?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
+* Interface assertions: TBD
+* Unknown checks on DUT outputs: TBD
 
 ## Building and running tests
-We are using our in-house developed [regression tool]({{< relref "vendor/lowrisc_ip/dvsim" >}}) for building and running our tests and regressions.
-Please take a look at the link for detailed information on the usage, capabilities, features and known issues.
-Here's how to run a basic sanity test:
+
+Tests are built and run with the [`dvsim`](https://github.com/lowRISC/ibex/tree/master/vendor/lowrisc_ip/dvsim) tool (vendored in from the OpenTitan project).
+This is wrapped with a Makefile.
```
udinator: Is it possible to make this Makefile target agnostic? I'm integrating Ibex with dvsim as well, so it would be good for this Makefile to work with both Ibex/Icache instead of just with one of the two possible targets.
Maybe something like:
```console
$ cd dv/uvm
$ make IP=icache TEST=ibex_icache_sanity
```

Just an idea though, so feel free to propose anything you think might work!

udinator:
@asb - Sure, that makes sense to me. I tried to limit the discussion to just involve the tooling that is already used (or going to be used) for Icache DV work (dv_lib libraries, dvsim, uvmdvgen, so on).
We can circle back to this later down the road.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
-* analysis port1:
-* analysis port2:
-<!-- explain inputs monitored, flow of data and outputs checked -->
+The `ibex_icache_scoreboard` is used to check that instructions fetched from the cache lie at the correct addresses and contain the correct data.
+It has two analysis ports: one for the memory agent and another for the core agent.
+
+The scoreboard can keep track of what addresses it expects from responses to the core by modelling the cache's internal address counter and following branches.
+
+To check correctness of the fetched data, the scoreboard will keep track of the set of seeds that have been set by the memory agent since the last cache invalidation began.
+When it sees an instruction being passed back to the core, it can check that the response matches one of the possible seeds.
```
udinator: Is it possible to also compare the data as well as the seed?
From what it sounds like, the seed acts as a timestamp of sorts to check that the cache is pulling data from the correct "set" of valid data, but I think it might also be good to add an extra layer of checking by comparing the data itself as well.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: I think you will also have to also modify the fusesoc core files as well to reflect the new testbench hierarchy and dependencies.

dv/uvm/icache/doc/tb.svg
rswarbrick: @udinator What information does this diagram convey? It's a lovely list of classes and, yes, I can update that one with icache names, but I don't understand why anyone would want to. Is this in a style guide somewhere, or is there a good reason to do it?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Well, that will only work for code that has been vendored in to OpenTitan. I'm not really convinced that OpenTitan-specific integration stuff belongs here. If we eventually decide it does, let's add it back in as a separate PR.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: As above. I'm not particularly keen on adding static links to OpenTitan documentation from Ibex code - the dependency arrows are supposed to go the other way around.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: As above. Yes, I agree that we need to figure out how to publish these results from a different repository from OpenTitan, but that's surely a different problem from "write documentation explaining how we're testing the icache", which this PR seeks to address.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Who is the target audience for this document? I decided it was people who want to know whether the icache works or not. I shouldn't imagine those people care about our class structure. People who need to actually understand the test code will presumably look at the source, at which point it will be obvious what it's derived from.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: See reply to your comment on the hunk that deletes tb.svg.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: That's a very good point. I was thinking that, since the `busy_o` signal is kind of a different interface from the rest of the cache <-> core signals, it should have a different agent, but I agree that it's slightly overkill. I'll push up a version of the document that gets rid of it.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: I've done this already, thanks (but obviously not in this patch).

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: An invalidation doesn't change the memory seed (after all: in the real system, an invalidation is different from a memory write).

The idea is that the memory seed changes occasionally, asynchronously from what the core is doing. The memory sequence will inject occasional "change the seed" transactions and the scoreboard will check that results from fetches match at least one of the seeds that has been seen since the last invalidation. Incidentally, the "entire memory model" is just a 32-bit integer: everything else is calculated with hash functions from that seed.

Is there a bit of this text that is confusing? It makes sense to me, but then again I wrote it, so I'm probably not a very good judge!

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Yes.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: I wrote that text before adding the `busy_o` agent above but then forgot to change it: good catch! However, I agree that we don't need the extra agent, so maybe I should just claim I was being prescient :-D.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Yes, that's the idea.

In the simple case, let's suppose that we haven't yet seen any seed changes, and I know that the memory seed should have been `h'1234`. Then the scoreboard looks at the retrieved address and data. It knows what the address should have been (and checks it's right). Assuming that it is, it can then compute the memory hash function (using some shared code) with seed `h'1234` and the given address and check it equals the data we saw fetched.

I'll update this text to make it a bit clearer what's going on.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: I agree that this makes sense in the future, but for now I'd prefer to get each working separately. Let's try to generalise the Makefiles once each works. The easiest way would be adding `dv/uvm/Makefile` and just calling out to the Makefiles below, but maybe we can share more code than that.

rswarbrick:
Hi @udinator, Thanks for the comments. I'll push up a new version this morning to drop the extra agent and clarify how the scoreboard checks returned data.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: SGTM.

dv/uvm/icache/doc/tb.svg
udinator: Yes, this diagram is intended to provide a high level visual representation of how the testbench is constructed (which is an important part of a DV plan document); the template in question is used throughout the rest of the OpenTitan DV environments to provide a testbench diagram.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: Yes, that is a good point, as per the comment made by @asb, we should have a later discussion to decide what should be done for status reporting and integration with other OT tooling, for the time being I think it is okay to omit it (this applies to all three of the links).

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: Ah I understand now, I just had some confusion whether the seed would be changed asynchronously from the core sending an invalidation request or not.

To clarify something in your comment, if data in the "memory model" is calculated based off of a hash of the seed, will it be possible for the result of an instruction fetch to match more than 1 seed?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: resolving this comment, as you answer it above

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: No: it will match exactly one seed. But if the cache is enabled (and we're not modelling eviction strategies etc.), there are a set of seeds that it might have come from (I'll get the old version of the data if cached; the new version if not).

udinator: Sure, that makes sense to me, thanks for the clarification.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: SGTM

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: Sure, this can be worked out in the future.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
udinator: Sure, that checking approach seems like a solid plan to me.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
```diff
@@ -1,99 +1,142 @@
...
-IBEX_ICACHE testbench has been constructed based on the [DV_LIB testbench architecture]({{< relref "vendor/lowrisc_ip/dv_lib/" >}}).
 
-### Block diagram
-![Block diagram](tb.svg)
+The testbench intentionally avoids knowing detailed information about the cache's performance characteristics (for example, cache size, line size or number of ways).
+This means that the testbench cannot compare the DUT with a reference model, nor can it model the exact requests that the DUT will make of instruction memory: the whole point of a cache is that it might avoid an instruction fetch.
+
+The cache has two main interfaces, which face the core and instruction bus respectively.
+We model this with two agents: the *core agent* and the *memory agent*.
+The core agent will emulate a core making instruction fetch requests from the cache and the memory agent will emulate the instruction bus.
```
udinator, line 23: One thing that came up in a discussion earlier today - is there a way that we can create a base/generic memory agent that can then be extended for both ibex and the icache? The current ibex agent drives/monitors most of the signals relevant for the icache, so this could be extended in the icache to add the additional signals you plan to include in the cache memory interface.
This would definitely increase the reusability of the existing code, I'm just a little hesitant of having a new memory interface that would duplicate a lot of the already existing logic when we could stick this in a base driver/monitor to both benefit from it.
@sriyerg what do you think?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Yes, understood. Since I'm folding this single pin into another agent, I'll just add it to the core <-> icache interface. I think that's even a principled decision (as well as being convenient!) because it is indeed part of the interface between the cache and the core.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: For now, it's just wrapping up the dvsim command line arguments. Let's come back to this when we've actually got a working testbench and are deciding how to present it to potential future users.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Ok. I'm not really convinced, but I don't mind putting the auto-generated text back if you believe it's important. I'll push up a patch with this in a few minutes.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: For what it's worth, I've actually implemented most of this (because I wanted to make sure this would work). The interface is properly different from the memory interface that the core DV uses, because of the PMP line (this doesn't make it out to top-level). This has to be driven at a different time from the usual request/grant signals. I'm not sure that we'd gain anything by trying to combine these two different agents: while I haven't implemented analysis transactions in the monitor yet, I've done the slave stuff and the only code that I could directly copy from the slave memory interface was the 25 lines that drive the grant line.

rswarbrick:
@sriyerg I understand your note about the testplan, but don't think it really makes sense to write that without a rough block diagram for the testbench, and I'm blocked on implementing the testbench until you and Udi say you're ok with the basic design - I don't want to write thousands of lines of code and then have to rework them all.

If you think this DV plan is a reasonable start (and describes a sane framework), maybe the best thing to do is to merge this and hash out details later, in parallel with actually building the thing.

rswarbrick:
New version pushed. The only change is re-instating the line about DV_LIB.

dv/uvm/icache/doc/tb.svg
sriyerg: Its typically how UVM testbenches are depicted. It clearly shows what components are in use and how they interact with each other. A quick glance provides the reader (testbench writer as well as peers who can provide additional guidance if necessary) a very good idea of how the UVM stuff is put together, before it is actually put together.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: Would you at least retain the header with TBD instead of actual links?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: > @sriyerg I understand your note about the testplan, but don't think it really makes sense to write that without a rough block diagram for the testbench, and I'm blocked on implementing the testbench until you and Udi say you're ok with the basic design - I don't want to write thousands of lines of code and then have to rework them all.
> 
> If you think this DV plan is a reasonable start (and describes a sane framework), maybe the best thing to do is to merge this and hash out details later, in parallel with actually building the thing.

A testplan captures what all features are tested. This only requires the design spec to be available as a starting point. You need all of that information first, to know how the testbench needs to be put together. The "what" dictates the "how". The suggestion I was trying to make was to focus on the testplan more. As an example, from the AES testplan review, it was found that it was missing NIST vector testing (which required additional testbench components to be added, and hence DV plan updated). Also, it is reasonable to expect that once you start developing the code, you may realize that how you had originally envisioned the scoreboard will work for example, needs to complete change for whatever reason.

A rough block diagram for the testbench in the DV plan at this stage is highly recommended. However the nitty gritty details on how the scoreboard will work for example, is not required at this stage (but definitely welcome!). The actual implemented code will go through PR reviews anyway. 

I am fine with this landing as-is. We can modify the content as we progress through the development.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: I'd recommend adding shortcuts in you local bashrc instead. The extra layer doesn't seem to be needed at all.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: If you've already implemented it, then we can review the code and see if leveraging SV OOP is a better approach in terms of efficiency and reuse.

udinator: Ah okay, if you already have some implementation of it, then we can review the code when you send out the PR, or we can spin off a side meeting if you want

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Well, the links won't work (because `relref` doesn't work outside of the OT repo). Would you like me to re-instate the auto-generated bullet points?

dv/uvm/icache/doc/tb.svg
rswarbrick: Ok, if this is required for OpenTitan code, I guess it'll have to go in. I will do as you ask.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Ok, thanks for the explanation.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: But then everyone who wants to work with the code needs to maintain their own "how to run dvsim.py for testbench foo" shortcuts in their .bashrc. This is wasted effort and will, of course, bit-rot over time.

Alternatively, we could put shortcut commands for "build and run the tests" and similar in the tree. Much better, because they are in one place and will be maintained if the vendored dvsim.py changes. In what format should those commands be stored? Well, I guess we could have a bunch of shell scripts. Alternatively, a Makefile. Which is what I'm suggesting.

rswarbrick:
The current state of this PR: I've just resolved any conversations that I think have closed (with one person or another going "ok, shall do". There's [one discussion about Makefiles](https://github.com/lowRISC/ibex/pull/754#discussion_r401222969) that I don't want to close in case Sri wants to reply, but I believe think the only remaining sticking point is:

 * [Deleted links](https://github.com/lowRISC/ibex/pull/754#discussion_r401155479). These links won't work (because this isn't in the OT repo). Should we retain bullet items or not?

I have a remaining action item, to fix up (and replace) the `tb.svg` block diagram in the expected format, which I'll try to do this afternoon.

Is there anything I've missed?

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: Hence the TBDs.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
sriyerg: I dont think any shortcut is needed at all. I dont see how or why 
```
dvsim.py ibex_icache_sim_cfg.hjson -i icache_sanity 
```

Is harder / more complicated than
```
make -i TEST_NAME=icache_sanity
```
All OT DV plans and methodology and getting started docs will be updated to reflect the dvsim invocation (when we get to it), which IMO is sufficient.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: Ok. Will do as instructed.

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: @sriyerg Is removing mentions of Make a requirement for merging this PR? If not, let's merge it and you or I can make a later PR to use your preferred command line later. If so, please could you make this explicit? Thanks.

rswarbrick:
I've just pushed up a version with a testbench diagram in the style of the other OpenTitan block diagrams. This also re-instates the "Current Status" section that @sriyerg asked for. I believe the only remaining sticking point is whether we direct users to run dvsim.py directly or do my preferred approach of wrapping it with Make to run builds (which is what I'm going to do when actually developing the DV code).

dv/uvm/icache/doc/ibex_icache_dv_plan.md
rswarbrick: @sriyerg Thinking again, I've updated the text as requested.

rswarbrick:
I believe that I have made all the changes requested. @sriyerg, @udinator: Is there anything else you think needs to change before this can be merged?

udinator:
I think it's good to go, we can make any necessary changes later down the road.

PR #755, opened 2020-03-31
[dv] update Makefile gen step dependency

udinator:
Appends `GEN_DIR` to the list of dependencies for the `gen` step in the simulation flow, as we want instruction generation to also be dependent on the generation source directory (for example to test changes to riscv-dv that have not been vendored into Ibex yet).


udinator:
Switching around GEN_DIR is actually the approach I have been taking thus far, the options you mentioned were ones I considered as well, but I personally find overriding GEN_DIR from the command line is the most natural way to do things (plus it doesn't require the extra couple steps to remove local Ibex commits or add new symlinks or so on), especially since the Makefile allows GEN_DIR to be specified from the command line instead of being hard-coded to the path to the vendored in directory.
Your point is taken though, probably isn't worth explicit dependency support, so I'll close this PR.

PR #756, opened 2020-04-01
[rtl] Add fixed time execution of branches

tomroberts-lowrisc:
- A new parameter and a run-time control bit (SecureIbex and
  secure_mode) enable different behaviour for running security critical
  code sections.
- In the new mode, all branches act as if taken, with not-taken
  branches executing as a branch to the next instruction.
- This should give similar execution time/power characteristics
  regardless of the branch condition.
- Note that with the BranchTargetALU, branches stall an extra cycle in
  secure mode to avoid factoring the branch-taken decision into the
  branch target address mux.

### DIFF
--- a/ibex_core.core
+++ b/ibex_core.core
@@ -100,6 +100,12 @@ parameters:
     default: false
     description: "Enables third pipeline stage (EXPERIMENTAL)"
 
+  SecureIbex:
+    datatype: bool
+    paramtype: vlogparam
+    default: false
+    description: "Enables security hardening features (EXPERIMENTAL)"
+
 targets:
   default:
     filesets:
--- a/lint/verilator_waiver.vlt
+++ b/lint/verilator_waiver.vlt
@@ -18,6 +18,7 @@ lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'RV32M'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'RV32E'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'BranchTargetALU'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'WritebackStage'*"
+lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'SecureIbex'*"
 
 // Filename 'ibex_register_file_ff' does not match MODULE name: ibex_register_file
 // ibex_register_file_ff and ibex_register_file_latch provide two
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -14,6 +14,7 @@
 
 module ibex_cs_registers #(
     parameter bit          DbgTriggerEn     = 0,
+    parameter bit          DataIndTiming    = 1'b0,
     parameter bit          ICache           = 1'b0,
     parameter int unsigned MHPMCounterNum   = 10,
     parameter int unsigned MHPMCounterWidth = 40,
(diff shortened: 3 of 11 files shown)


rtl/ibex_id_stage.sv
```diff
@@ -606,14 +607,45 @@ module ibex_id_stage #(
       end
     end
 
-    assign branch_set = branch_set_q;
+    // Branches always take two cycles in fixed time execution mode, with or without the branch
+    // target ALU.
+    assign branch_set = (BranchTargetALU && !sec_fixed_time_i) ? branch_set_d : branch_set_q;
```
GregAC, line 612: May be worth adding a comment somewhere that for fixed term execution we have to use the non branch target ALU behaviour for timing reasons, otherwise the mux selects for the BT ALU depend upon the ALU comparison result (was initially going to ask why we went to two cycles even with BT ALU before I worked this out).

ibex_core.core
```diff
@@ -100,6 +100,12 @@ parameters:
     default: false
     description: "Enables third pipeline stage (EXPERIMENTAL)"
 
+  SecureIbex:
+    datatype: bool
+    paramtype: vlogparam
+    default: false
+    description: "Enables security hardening features (EXPERIMENTAL) [0/1]"
```
GregAC, line 107: nit: Remove the [0/1] to match the rest of the file.

lint/verilator_waiver.vlt
```diff
@@ -18,6 +18,7 @@ lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'RV32M'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'RV32E'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'BranchTargetALU'*"
 lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'WritebackStage'*"
+lint_off -rule WIDTH -file "*/rtl/ibex_core_tracing.sv" -match "*'SecureIbex'*"
```
imphil, line 21: I wonder if we can actually regex another part of the error message to cover all `bit` parameters?

rtl/ibex_cs_registers.sv
```diff
@@ -158,7 +160,8 @@ module ibex_cs_registers #(
 
   // CPU control register fields
   typedef struct packed {
-    logic [30:0] unused_ctrl;
+    logic [31:2] unused_ctrl;
+    logic        sec_fixed_time;
```
imphil, line 164: Thanks for making this a separate setting. I must say I like the ARM naming "data-independent timing" (DIT), as it covers quite well what happens, and doesn't (in the name) restrict it to the security domain.

lint/verilator_waiver.vlt
imphil: Yes, the regex matches the warning/error message string.

tomroberts-lowrisc: Ah yes - which is what you wrote in the original comment!
It doesn't mention anything about parameters in the message:
``Operator VAR 'RV32M' expects 1 bits on the Initial value, but Initial value's CONST '32'h1' generates 32 bits``
We could possibly end up suppressing width mismatches if we do anything to wide. Maybe matching on "VAR" and "Initial value"?

imphil: Maybe something like that is sufficiently strict but still covers all "assign 32'h0 or 32'h1 to a 1-bit constant":

`-match "Operator VAR '*' expects 1 bits on the Initial value, but Initial value's CONST '32'h?' generates 32 bits"`

Note all of that is strictly optional, feel free to go as far as you'd like to push this topic :smile:

PR #761, opened 2020-04-02
ICache core agent

rswarbrick:
This PR has the bones of the ICache "core" UVM agent (see DV plan at #754). It doesn't currently do any monitoring, but it can drive signals to the cache.

### DIFF
--- a/dv/uvm/icache/dv/Makefile
+++ b/dv/uvm/icache/dv/Makefile
@@ -19,22 +19,22 @@ SEED=1
 
 ibex-top       := ../../../..
 dvsim-py       := $(ibex-top)/vendor/lowrisc_ip/dvsim/dvsim.py
-dvsim-std-args := --skip-ral --scratch-root $(ibex-top)/build -v h
+dvsim-std-args := --skip-ral --scratch-root $(scratch-root)
 
 waves-arg      := $(if $(filter-out 0,$(WAVES)),--waves,)
 verbosity-arg  := $(if $(VERBOSITY),--verbosity $(VERBOSITY),)
 seed-arg       := $(if $(SEED),--fixed-seed $(SEED),)
 
-dvsim-mk-args := $(waves-arg) $(verbosity-arg) $(seed-arg)
-run-dvsim     := $(dvsim-py) $(dvsim-std-args) $(dvsim-mk-args)
+dvsim-mk-args    := $(waves-arg) $(verbosity-arg) $(seed-arg)
+run-icache-dvsim := $(dvsim-py) ibex_icache_sim_cfg.hjson $(dvsim-std-args) $(dvsim-mk-args)
 
 .PHONY: all
 all: run
 
 .PHONY: just-build
 just-build:
-	$(run-dvsim) --build-only ibex_icache_sim_cfg.hjson
+	$(run-icache-dvsim) --build-only
 
 .PHONY: run
 run:
-	$(run-dvsim) ibex_icache_sim_cfg.hjson
+	$(run-icache-dvsim)
--- a/dv/uvm/icache/dv/Makefile
+++ b/dv/uvm/icache/dv/Makefile
@@ -1,0 +1,40 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Simple wrappers around dvsim.py to avoid needing long command lines. This
+# doesn't do any dependency tracking (as you might expect in a Makefile) because
+# the dvsim infrastructure doesn't support it.
+
+# Set WAVES to 1 to get VPD/FSDB dumps
+WAVES=0
+
+# Set VERBOSITY to a nonempty string and it will be passed to dvsim with the
+# --verbosity parameter, overriding whatever was set in the hjson file.
+VERBOSITY=
+
+# Specify the seed for the test to run. If this is empty, dvsim.py will pick
+# random seeds. By default, we make runs reproducible, so force the seed.
+SEED=1
+
+ibex-top       := ../../../..
+dvsim-py       := $(ibex-top)/vendor/lowrisc_ip/dvsim/dvsim.py
+dvsim-std-args := --skip-ral --scratch-root $(ibex-top)/build -v h
+
+waves-arg      := $(if $(filter-out 0,$(WAVES)),--waves,)
+verbosity-arg  := $(if $(VERBOSITY),--verbosity $(VERBOSITY),)
+seed-arg       := $(if $(SEED),--fixed-seed $(SEED),)
+
+dvsim-mk-args := $(waves-arg) $(verbosity-arg) $(seed-arg)
+run-dvsim     := $(dvsim-py) $(dvsim-std-args) $(dvsim-mk-args)
+
+.PHONY: all
+all: run
+
+.PHONY: just-build
+just-build:
+	$(run-dvsim) --build-only ibex_icache_sim_cfg.hjson
+
+.PHONY: run
+run:
+	$(run-dvsim) ibex_icache_sim_cfg.hjson
--- a/dv/uvm/icache/dv/env/ibex_icache_virtual_sequencer.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_virtual_sequencer.sv
@@ -8,8 +8,8 @@ class ibex_icache_virtual_sequencer extends dv_base_virtual_sequencer #(
   );
   `uvm_component_utils(ibex_icache_virtual_sequencer)
 
-  ibex_icache_sequencer ibex_icache_sequencer_h;
-  ibex_mem_intf_slave_sequencer ibex_mem_intf_slave_sequencer_h;
+  ibex_icache_sequencer         core_sequencer_h;
+  ibex_mem_intf_slave_sequencer mem_sequencer_h;
 
   `uvm_component_new
 
--- a/dv/uvm/icache/dv/env/seq_lib/ibex_icache_base_vseq.sv
+++ b/dv/uvm/icache/dv/env/seq_lib/ibex_icache_base_vseq.sv
@@ -8,25 +8,15 @@ class ibex_icache_base_vseq extends dv_base_vseq #(
     .VIRTUAL_SEQUENCER_T (ibex_icache_virtual_sequencer)
   );
   `uvm_object_utils(ibex_icache_base_vseq)
-
-  // various knobs to enable certain routines
-  bit do_ibex_icache_init = 1'b1;
-
   `uvm_object_new
 
   virtual task dut_init(string reset_kind = "HARD");
     super.dut_init();
-    if (do_ibex_icache_init) ibex_icache_init();
   endtask
 
   virtual task dut_shutdown();
     // check for pending ibex_icache operations and wait for them to complete
     // TODO
-  endtask
-
-  // setup basic ibex_icache features
-  virtual task ibex_icache_init();
-    `uvm_error(`gfn, "FIXME")
   endtask
 
 endclass : ibex_icache_base_vseq
--- a/dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
+++ b/dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
@@ -2,14 +2,21 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// basic sanity test vseq
+// Basic sanity test
+
 class ibex_icache_sanity_vseq extends ibex_icache_base_vseq;
   `uvm_object_utils(ibex_icache_sanity_vseq)
 
   `uvm_object_new
 
+  // A sanity sequence for the core agent
+  ibex_icache_sanity_seq core_seq;
+
   task body();
-    `uvm_error(`gfn, "FIXME")
+    // TODO: This currently just drives the core sequence (which clearly isn't going to work!)
+    `uvm_create_on(core_seq, p_sequencer.core_sequencer_h)
+    `DV_CHECK_RANDOMIZE_FATAL(core_seq)
+    core_seq.start(p_sequencer.core_sequencer_h);
   endtask : body
 
 endclass : ibex_icache_sanity_vseq
--- a/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
+++ b/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
@@ -4,34 +4,124 @@
 
 class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_agent_cfg);
   `uvm_component_utils(ibex_icache_driver)
-
-  // the base class provides the following handles for use:
-  // ibex_icache_agent_cfg: cfg
-
   `uvm_component_new
 
-  virtual task run_phase(uvm_phase phase);
-    // base class forks off reset_signals() and get_and_drive() tasks
-    super.run_phase(phase);
-  endtask
-
   // reset signals
-  virtual task reset_signals();
+  virtual task automatic reset_signals();
+    cfg.vif.reset();
   endtask
 
   // drive trans received from sequencer
-  virtual task get_and_drive();
+  virtual task automatic get_and_drive();
     forever begin
       seq_item_port.get_next_item(req);
-      $cast(rsp, req.clone());
-      rsp.set_id_info(req);
       `uvm_info(`gfn, $sformatf("rcvd item:\n%0s", req.sprint()), UVM_HIGH)
-      // TODO: do the driving part
-      //
-      // send rsp back to seq
+
+      case (req.trans_type)
+        ICacheTransTypeBranch: drive_branch_trans(req);
+        ICacheTransTypeReq:    drive_req_trans(req);
+        default:               `uvm_fatal(`gfn, "Unknown transaction type")
+      endcase
+
       `uvm_info(`gfn, "item sent", UVM_HIGH)
-      seq_item_port.item_done(rsp);
+      seq_item_port.item_done();
     end
+  endtask
+
+  // Drive the cache for a "branch" transaction.
+  //
+  // This concurrently asserts branch with a given address for a cycle while doing the usual
+  // (enable/disable, invalidate, read instructions).
+  virtual task automatic drive_branch_trans(ibex_icache_item req);
+    // Make sure that req is enabled (has no effect unless this is the first transaction)
+    cfg.vif.req <= 1'b1;
+
+    fork
+        cfg.vif.branch_to(req.branch_addr);
+        if (req.toggle_enable) toggle_enable();
+        if (req.invalidate) invalidate();
+        read_insns(req.num_insns);
+    join
+  endtask
+
+  // Drive the cache for a "req transaction".
+  //
+  // This lowers req for zero or more cycles, at the same time as setting the enable pin and (maybe)
+  // pulsing the invalidate line. Once that is done, it reads zero or more instructions.
+  virtual task automatic drive_req_trans(ibex_icache_item req);
+    int unsigned req_low_cycles;
+    bit          allow_no_low_cycles;
+
+    // How many cycles should we lower the request line? If there aren't any instructions to read,
+    // this is constrained to be positive (to avoid a confusing zero-time transaction).
+    allow_no_low_cycles = req.num_insns > 0;
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(req_low_cycles,
+                                       req_low_cycles dist {
+                                           0           :/ (allow_no_low_cycles ? 20 : 0),
+                                           [1:33]      :/ 5,
+                                           [100:200]   :/ 2,
+                                           [1000:1200] :/ 1 };)
+
+    fork
+        if (req_low_cycles > 0) lower_req(req_low_cycles);
+        if (req.toggle_enable) toggle_enable();
+        if (req.invalidate) invalidate();
+    join
+    read_insns(req.num_insns);
+  endtask
+
+  // Toggle whether the cache is enabled
+  virtual task automatic toggle_enable();
+    cfg.vif.enable <= ~cfg.vif.enable;
+  endtask
+
+  // Read up to num_insns instructions from the cache, stopping early on an error
+  virtual task automatic read_insns(int num_insns);
+    for (int i = 0; i < num_insns; i++) begin
+      read_insn();
+      // Spot any error and exit early
+      if (cfg.vif.err)
+        break;
+    end
+  endtask
+
+  // Read a single instruction from the cache
+  virtual task automatic read_insn();
+    int unsigned delay;
+
+    // Maybe (1 time in 10) wait for a valid signal before even considering asserting ready.
+    if ($urandom_range(9) == 0)
+      wait (cfg.vif.valid);
+
+    // Then pick how long we wait before asserting that we are ready.
+    //
+    // TODO: Make this configurable and weight 0 more heavily.
+    cfg.vif.wait_clks($urandom_range(3));
+
+    // Assert ready and then wait until valid
+    cfg.vif.ready <= 1'b1;
+    while (1'b1) begin
+      @(posedge cfg.vif.clk);
+      if (cfg.vif.valid)
+        break;
+    end
+
+    cfg.vif.ready <= 1'b0;
+  endtask
+
+  // Lower the req line for the given number of cycles
+  virtual task automatic lower_req(int unsigned num_cycles);
+    cfg.vif.req <= 1'b0;
+    repeat (num_cycles) @(posedge cfg.vif.clk);
+    cfg.vif.req <= 1'b1;
+  endtask
+
+  // Raise the invalidate line for a randomly chosen number of cycles > 0.
+  virtual task automatic invalidate();
+    int unsigned num_cycles;
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(num_cycles,
+                                       num_cycles dist { 0 :/ 499, [1:20] :/ 1 };)
+    cfg.vif.invalidate_pulse(num_cycles);
   endtask
 
 endclass
--- a/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
+++ b/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
@@ -2,10 +2,66 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-interface ibex_icache_if ();
+interface ibex_icache_if (input clk);
 
-  // interface pins
+  // Set when core is enabled (and might request instructions soon)
+  logic         req;
 
-  // debug signals
+  // Branch request
+  logic         branch;
+  logic [31:0]  branch_addr;
+
+  // Passing instructions back to the core
+  logic         ready;
+  logic         valid;
+  logic [31:0]  rdata;
+  logic [31:0]  addr;
+  logic         err;
+  logic         err_plus2;
+
+  // Enable/disable or invalidate the cache
+  logic         enable;
+  logic         invalidate;
+
+  // Busy signal from the cache (either there's a request on the bus or the cache is invalidating
+  // itself)
+  logic         busy;
+
+  // Drive the branch pin for a single cycle, redirecting the cache to the given instruction
+  // address.
+  task automatic branch_to(logic [31:0] addr);
+    branch      <= 1'b1;
+    branch_addr <= addr;
+
+    @(posedge clk);
+
+    branch      <= 1'b0;
+    branch_addr <= 'X;
+  endtask
+
+  // Raise a pulse on the invalidate line for the given number of cycles.
+  //
+  // A one-cycle pulse will start an invalidation, but testing might want a longer pulse (which the
+  // cache should support)
+  task automatic invalidate_pulse(int num_cycles);
+    invalidate <= 1'b1;
+    repeat (num_cycles) @(posedge clk);
+    invalidate <= 1'b0;
+  endtask
+
+  // A task that waits for num_clks posedges on the clk signal
+  task automatic wait_clks(int num_clks);
+    repeat (num_clks) @(posedge clk);
+  endtask
+
+  // Reset all the signals from the core to the cache (the other direction is controlled by the
+  // DUT)
+  task automatic reset();
+    req        <= 1'b0;
+    branch     <= 1'b0;
+    ready      <= 1'b0;
+    enable     <= 1'b0;
+    invalidate <= 1'b0;
+  endtask
 
 endinterface
--- a/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
+++ b/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
@@ -4,9 +4,70 @@
 
 class ibex_icache_item extends uvm_sequence_item;
 
-  // random variables
+  // The type of transaction
+  rand ibex_icache_trans_type_e trans_type;
+
+  // The branch address for a branch transaction (only has effect if trans_type is
+  // ICacheTransTypeBranch)
+  rand bit [31:0]               branch_addr;
+
+  // Whether the cache enable/disable should be toggled
+  rand bit                      toggle_enable;
+
+  // Whether to invalidate the cache
+  rand bit                      invalidate;
+
+  // The number of instructions to read (always non-negative, but may be zero)
+  rand int                      num_insns;
+
+  constraint c_non_branch_trans_addr {
+    // If the transaction type is ICacheTransTypeBranch then branch_addr can be anything. To make
+    // reading debug logs a bit easier, we force it to be zero otherwise.
+    (trans_type != ICacheTransTypeBranch) -> branch_addr == 0;
+
+    // Pick trans_type before the other values. We need to do this because constraining branch_addr
+    // to 0 for non-branch transactions would otherwise mean branch transactions got weighted 2^32
+    // times higher.
+    solve trans_type before branch_addr;
+  }
+
+  constraint c_branch_addr_alignment {
+    // The branch address is always required to be half-word aligned. We don't bother conditioning
+    // this on the type of transaction because branch_addr is forced to be zero in anything but a
+    // branch transaction.
+    !branch_addr[0];
+  }
+
+  constraint c_toggle_enable_dist {
+    // Toggle the cache enable line one time in 50. This should allow us a reasonable amount of time
+    // in each mode (note that each transaction here results in multiple instruction fetches)
+    toggle_enable dist { 0 :/ 49, 1 :/ 1 };
+  }
+
+  constraint c_invalidate_dist {
+    // Poke the cache invalidate line one time in 500. This takes ages and we don't want to
+    // accidentally spend most of the test waiting for invalidation.
+    invalidate dist { 0 :/ 499, 1 :/ 1 };
+  }
+
+  constraint c_num_insns_dist {
+    // For branch transactions, we want to read zero instructions reasonably frequently. For req
+    // transactions, much less so. Also, we don't bother with long sequences for req transactions:
+    // they won't look any different from the tail end of branch transactions from the cache's point
+    // of view.
+    if (trans_type == ICacheTransTypeBranch)
+      num_insns dist { 0 :/ 5, [1:20] :/ 20, [21:100] :/ 1 };
+    else
+      num_insns dist { 0 :/ 1, [1:20] :/ 20 };
+  }
+
 
   `uvm_object_utils_begin(ibex_icache_item)
+    `uvm_field_enum(ibex_icache_trans_type_e, trans_type, UVM_DEFAULT)
+    `uvm_field_int (branch_addr,   UVM_DEFAULT | UVM_HEX)
+    `uvm_field_int (toggle_enable, UVM_DEFAULT)
+    `uvm_field_int (invalidate,    UVM_DEFAULT)
+    `uvm_field_int (num_insns,     UVM_DEFAULT)
   `uvm_object_utils_end
 
   `uvm_object_new
--- a/dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
+++ b/dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
@@ -1,0 +1,31 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Sanity test seq
+//
+// This is unlikely to find many cache hits (since it branches all over the 4GiB address space).
+class ibex_icache_sanity_seq extends ibex_icache_base_seq;
+  `uvm_object_utils(ibex_icache_sanity_seq)
+  `uvm_object_new
+
+  rand int count;
+  constraint c_count { count > 0; count < 100; }
+
+  task body();
+    // Generate a request which is constrained to have trans_type ICacheTransTypeBranch: the core
+    // must start with a branch to tell the cache where to fetch from in the first place.
+    req = ibex_icache_item::type_id::create("req");
+    start_item(req);
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(req, req.trans_type == ICacheTransTypeBranch;)
+    finish_item(req);
+
+    // Generate and run count ibex_icache_item sequence items (with no other constraint) through the
+    // req port.
+    repeat (count - 1) begin
+      start_item(req);
+      `DV_CHECK_RANDOMIZE_FATAL(req)
+      finish_item(req);
+    end
+  endtask
+endclass
--- a/dv/uvm/icache/dv/tb/tb.sv
+++ b/dv/uvm/icache/dv/tb/tb.sv
@@ -17,13 +17,27 @@ module tb;
 
   // interfaces
   clk_rst_if clk_rst_if(.clk(clk), .rst_n(rst_n));
-  ibex_icache_if ibex_icache_if();
-  ibex_mem_intf ibex_mem_intf();
+  ibex_icache_if core_if (.clk(clk));
+  ibex_mem_intf  ibex_mem_intf();
 
   // dut
   ibex_icache dut (
-    .clk_i                (clk        ),
-    .rst_ni               (rst_n      )
+    .clk_i           (clk),
+    .rst_ni          (rst_n),
+
+    // Connect icache <-> core interface
+    .req_i           (core_if.req),
+    .branch_i        (core_if.branch),
+    .addr_i          (core_if.branch_addr),
+    .ready_i         (core_if.ready),
+    .valid_o         (core_if.valid),
+    .rdata_o         (core_if.rdata),
+    .addr_o          (core_if.addr),
+    .err_o           (core_if.err),
+    .err_plus2_o     (core_if.err_plus2),
+    .icache_enable_i (core_if.enable),
+    .icache_inval_i  (core_if.invalidate),
+    .busy_o          (core_if.busy)
 
     // TODO: add remaining IOs and hook them
   );
@@ -32,7 +46,7 @@ module tb;
     // drive clk and rst_n from clk_if
     clk_rst_if.set_active();
     uvm_config_db#(virtual clk_rst_if)::set(null, "*.env", "clk_rst_vif", clk_rst_if);
-    uvm_config_db#(virtual ibex_icache_if)::set(null, "*.env.m_ibex_icache_agent*", "vif", ibex_icache_if);
+    uvm_config_db#(virtual ibex_icache_if)::set(null, "*.env.m_ibex_icache_agent*", "vif", core_if);
     uvm_config_db#(virtual ibex_mem_intf)::set(null, "*.env.m_ibex_mem_intf_slave_agent*", "vif", ibex_mem_intf);
     $timeformat(-12, 0, " ps", 12);
     run_test();
--- a/dv/uvm/icache/dv/env/ibex_icache_env.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_env.sv
@@ -10,7 +10,7 @@ class ibex_icache_env extends dv_base_env #(
   );
   `uvm_component_utils(ibex_icache_env)
 
-  ibex_icache_agent m_ibex_icache_agent;
+  ibex_icache_agent         m_ibex_icache_agent;
   ibex_mem_intf_slave_agent m_ibex_mem_intf_slave_agent;
 
   `uvm_component_new
@@ -31,10 +31,10 @@ class ibex_icache_env extends dv_base_env #(
       m_ibex_mem_intf_slave_agent.monitor.addr_ph_port.connect(scoreboard.ibex_mem_intf_slave_fifo.analysis_export);
     end
     if (cfg.is_active && cfg.m_ibex_icache_agent_cfg.is_active) begin
-      virtual_sequencer.ibex_icache_sequencer_h = m_ibex_icache_agent.sequencer;
+      virtual_sequencer.core_sequencer_h = m_ibex_icache_agent.sequencer;
     end
     if (cfg.is_active && m_ibex_mem_intf_slave_agent.get_is_active()) begin
-      virtual_sequencer.ibex_mem_intf_slave_sequencer_h = m_ibex_mem_intf_slave_agent.sequencer;
+      virtual_sequencer.mem_sequencer_h = m_ibex_mem_intf_slave_agent.sequencer;
     end
   endfunction
 
--- a/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_agent.core
+++ b/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_agent.core
@@ -19,6 +19,7 @@ filesets:
       - ibex_icache_monitor.sv: {is_include_file: true}
       - ibex_icache_agent.sv: {is_include_file: true}
       - seq_lib/ibex_icache_base_seq.sv: {is_include_file: true}
+      - seq_lib/ibex_icache_sanity_seq.sv: {is_include_file: true}
       - seq_lib/ibex_icache_seq_list.sv: {is_include_file: true}
     file_type: systemVerilogSource
 
--- a/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_agent_pkg.sv
+++ b/dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_agent_pkg.sv
@@ -8,6 +8,11 @@ package ibex_icache_agent_pkg;
   import dv_utils_pkg::*;
   import dv_lib_pkg::*;
 
+  typedef enum {
+    ICacheTransTypeBranch,
+    ICacheTransTypeReq
+  } ibex_icache_trans_type_e;
+
   // macro includes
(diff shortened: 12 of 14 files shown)


rswarbrick:
For anyone who's been paying close attention: yes, the names are a bit odd here given that there's going to be a memory agent as well. I've got a followup patch that replaces `ibex_icache_$foo` with `ibex_icache_core_$foo`, but that patch is kind of noisy so I thought it would make more sense to split out implementing stuff and renaming it.

dv/uvm/icache/dv/Makefile
```diff
@@ -0,0 +1,40 @@
...
+ibex-top       := ../../../..
+dvsim-py       := $(ibex-top)/vendor/lowrisc_ip/dvsim/dvsim.py
+dvsim-std-args := --skip-ral --scratch-root $(ibex-top)/build -v h
+
+waves-arg      := $(if $(filter-out 0,$(WAVES)),--waves,)
+verbosity-arg  := $(if $(VERBOSITY),--verbosity $(VERBOSITY),)
+seed-arg       := $(if $(SEED),--fixed-seed $(SEED),)
+
+dvsim-mk-args := $(waves-arg) $(verbosity-arg) $(seed-arg)
+run-dvsim     := $(dvsim-py) $(dvsim-std-args) $(dvsim-mk-args)
```
udinator, line 29: just a heads up, but the very first argument that dvsim will expect is the `ibex_icache_sim_cfg.hjson` file, all of the other cmdline options need to be after it

dv/uvm/icache/dv/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+.PHONY: all
+all: run
+
+.PHONY: just-build
+just-build:
```
udinator, line 35: as an extension to my above comment, the `just-build` command here would have to be something like:
```console
$(run-dvsim) ibex_icache_sim_cfg.hjson --build-only $(dvsim-std-args) $(dvsim-mk-args)
```

udinator: the `--skip-ral` option is also necessary for both building and running dvsim with the Icache

dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
```diff
@@ -2,14 +2,19 @@
...
 
   `uvm_object_new
 
+  // A sanity sequence for the core agent
+  ibex_icache_sanity_seq core_seq;
+
   task body();
-    `uvm_error(`gfn, "FIXME")
+    // TODO: This currently just drives the core sequence (which clearly isn't going to work!)
+    `uvm_do_on(core_seq, p_sequencer.core_sequencer_h)
```
udinator, line 17: our DV style guide mandates that the uvm_do_* sequence macros should not be used, instead you should manually call  `start_item(...)`, `finish_item(...)`, and `get_response(...)`

dv/uvm/icache/dv/env/ibex_icache_virtual_sequencer.sv
```diff
@@ -8,8 +8,8 @@ class ibex_icache_virtual_sequencer extends dv_base_virtual_sequencer #(
   );
   `uvm_component_utils(ibex_icache_virtual_sequencer)
 
-  ibex_icache_sequencer ibex_icache_sequencer_h;
-  ibex_mem_intf_slave_sequencer ibex_mem_intf_slave_sequencer_h;
+  ibex_icache_sequencer         core_sequencer_h;
```
udinator, line 11: Just a question, it seems that you are making heavy use of the `core` term to name UVM components that are in the agent that deals with the connections between the icache and the rest of the core.
Maybe you could re-generate this agent with uvmdvgen and name the entire thing `ibex_icache_core_agent` or something like that, to standardize the naming convention as per the recommendations in the DV style guide?

dv/uvm/icache/dv/env/seq_lib/ibex_icache_base_vseq.sv
```diff
@@ -8,25 +8,15 @@ class ibex_icache_base_vseq extends dv_base_vseq #(
...
-    if (do_ibex_icache_init) ibex_icache_init();
   endtask
 
   virtual task dut_shutdown();
     // check for pending ibex_icache operations and wait for them to complete
     // TODO
   endtask
 
-  // setup basic ibex_icache features
-  virtual task ibex_icache_init();
```
udinator, line 28: will this init task not be needed?

dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
```diff
@@ -2,14 +2,19 @@
...
-// basic sanity test vseq
+// Basic sanity test
+
 class ibex_icache_sanity_vseq extends ibex_icache_base_vseq;
   `uvm_object_utils(ibex_icache_sanity_vseq)
 
   `uvm_object_new
 
+  // A sanity sequence for the core agent
+  ibex_icache_sanity_seq core_seq;
```
udinator, line 13: see my earlier comment about renaming the agent

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -4,19 +4,15 @@
 
 class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_agent_cfg);
   `uvm_component_utils(ibex_icache_driver)
-
-  // the base class provides the following handles for use:
-  // ibex_icache_agent_cfg: cfg
-
   `uvm_component_new
 
-  virtual task run_phase(uvm_phase phase);
```
udinator, line 13: Not sure if I'm missing something here, but it seems that you are missing the `get_and_drive` task called in the run_phase of dv_base_driver

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -4,19 +4,15 @@
...
-    // base class forks off reset_signals() and get_and_drive() tasks
-    super.run_phase(phase);
-  endtask
-
   // reset signals
   virtual task reset_signals();
+    cfg.vif.req        <= 1'b0;
```
udinator, line 11: I'd recommend putting the reset functionality in your interface so here you can just call `cfg.vif.reset()` or whatever you choose to name it

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
       $cast(rsp, req.clone());
       rsp.set_id_info(req);
       `uvm_info(`gfn, $sformatf("rcvd item:\n%0s", req.sprint()), UVM_HIGH)
-      // TODO: do the driving part
-      //
-      // send rsp back to seq
+
+      unique case (req.trans_type)
```
udinator, line 26: This is more of a nit, but for our DV environments, we generally use plain old case statements, and default to a `uvm_fatal` report macro

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+      @(posedge cfg.vif.clk);
+      if (cfg.vif.valid)
+        break;
+    end
+
+    cfg.vif.ready <= 1'b0;
+  endtask
+
+  // Lower the req line for the given number of cycles
+  virtual task lower_req(int num_cycles);
```
udinator, line 129: this can be moved to the interface as well

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+  virtual task toggle_enable();
+    cfg.vif.enable <= ~cfg.vif.enable;
+  endtask
+
+  // Raise a pulse on the invalidate line for the given number of cycles.
+  virtual task invalidate(int num_cycles);
+    cfg.vif.invalidate <= 1'b1;
+    repeat (num_cycles) @(posedge cfg.vif.clk);
+    cfg.vif.invalidate <= 1'b0;
+  endtask
```
udinator, line 90: This functionality can be placed into the interface itself, since a lot of these tasks have pretty simple logic that just wiggles interface pins. See the interfaces of various IPs in https://github.com/lowRISC/opentitan/tree/master/hw/dv/sv for reference, in general it is good to have as many utility tasks as needed in the interface so as to prevent having to dig through the hierarchy to get the signals in other components.
In particular, I am referring to `lower_req`, `branch_to`, and `invalidate`.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+    // Maybe (1 time in 10) wait for a valid signal before even considering asserting ready.
+    if ($urandom_range(9) == 0)
+      wait (cfg.vif.valid);
+
+    // Then pick how long we wait before asserting that we are ready.
+    //
+    // TODO: Make this configurable and weight 0 more heavily.
+    delay = $urandom_range(3);
+
+    repeat (delay) @(posedge cfg.vif.clk);
```
udinator, line 115: It is also good to have a `wait_clks` task in the interface that takes in an argument value representing how many posedge/negedge to wait for, that way you can simply call `cfg.vif.wait_clks($urandom_range(3));`
Alternatively, this functionality already exists in the clk_rst_if, which you can use instead

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+
+    // Assert ready and then wait until valid
```
udinator, line 117: this can probably be simplified to something like
```systemverilog
cfg.vif.ready <= 1'b1;
cfg.vif.wait_valid();
```

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
       seq_item_port.item_done(rsp);
     end
   endtask
 
+  // Drive the cache for a "branch" transaction.
+  //
+  // This concurrently asserts branch with a given address for a cycle while doing the usual
+  // (enable/disable, invalidate, read instructions).
+  virtual task drive_branch_trans(ibex_icache_item req);
+    fork
```
udinator, line 41: nit: could you wrap each concurrent process in begin...end statements for readability?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+      // Make sure that req is enabled (has no effect unless this is the first transaction)
+      cfg.vif.req <= 1'b1;
```
udinator, line 43: could this be placed before the fork?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,78 @@
...
+
+  // The number of cycles to drop req_i (only has effect if trans_type is ICacheTransTypeReq)
+  rand int                      req_low_cycles;
+
+  // Whether the cache enable/disable should be toggled
+  rand bit                      toggle_enable;
+
+  // Whether to invalidate the cache and, if so, how many cycles to hold the invalidate line up for.
+  // A value of zero means not to invalidate.
+  rand int                      invalidate_cycles;
```
udinator, line 22: Instead of doing this, where the sequence item contains the number of delay cycles and various things like that, could you:
1) just assign cfg.vif.enable to the value of req.toggle_enable? This way it is more clear if some transaction is enabling/disabling the cache, and simplifies some of this logic, as you can just drive the enable bit straight from the sequence item to the interface.

2) in a similar vein, can you change the value in the sequence item to just an `invalidate` bit? Like before, that is more clear whether you are invalidating the cache, and you can just drive that value onto the `icache_invalidate_i` line for some number of cycles which can be randomized in the driver itself.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,78 @@
...
-  // random variables
+  // The type of transaction
+  rand ibex_icache_trans_type_e trans_type;
+
+  // The branch address for a branch transaction (only has effect if trans_type is
+  // ICacheTransTypeBranch)
+  rand logic [31:0]             branch_addr;
+
+  // The number of cycles to drop req_i (only has effect if trans_type is ICacheTransTypeReq)
+  rand int                      req_low_cycles;
```
udinator, line 15: My understanding of this value is that you use it to randomize the number of delay cycles before the cache starts honoring your fetch request, if that sounds about right?
I feel that perhaps this can be changed to just `req`, which is then driven straight onto the interface into the cache, and you can randomize the delay in the driver itself.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -4,19 +4,15 @@
 
 class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_agent_cfg);
```
udinator, line 5: Just sort of a general note, but the sequence item should (generally speaking) contain at a minimum the same information that will be driven through the interface to the DUT, as a result the driver should be as "dumb" as possible and pretty much just drive the information it picks up from the sequence item, with minimum calculations and extra logic.
With that said, I have a question/suggestion for the structure of the "transactions" that you are implementing here. If I understand correctly, the two transactions that you are trying to model are as follows (at a very high level):

1) A branch to some arbitrary PC
2) A series of instruction reads from the cache

I feel the current approach of having the driver create these higher level abstractions adds some unnecessary complexity, so maybe it is better to model these transactions at the sequence level instead? With this approach, you could have a utility API sequence to send a specified number of instruction fetch transactions to the driver, and another sequence to send a specified number of branch transactions to the driver; this way the driver does not have so much complexity, and can solely "get and drive" information handed to it from the TLM port (no pun intended) , and these API sequences can be leaned upon in main test sequences.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,78 @@
...
+  // Whether to invalidate the cache and, if so, how many cycles to hold the invalidate line up for.
+  // A value of zero means not to invalidate.
+  rand int                      invalidate_cycles;
+
+  // The number of instructions to read (always non-negative, but may be zero)
+  rand int                      num_insns;
+
+
+  // Constraints
+  constraint c_data {
```
udinator, line 29: Another general note, it is recommended to not include all necessary constraints in a single constraint block; the DV style guide recommends to put similar/related constraints in one constraint block, and give it a descriptive name pertaining to the thing it is constraining.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,114 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+      read_insns(req.num_insns);
+    join
+  endtask
+
+  // Drive the cache for a "req transaction".
+  //
+  // This lowers req for req_low_cycles, at the same time as setting the enable pin and (maybe)
+  // pulsing the invalidate line. Once that is done, it reads instructions.
+  virtual task drive_req_trans(ibex_icache_item req);
+    fork
```
udinator, line 58: same thing here about wrapping concurrent processes in begin...end blocks

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,78 @@
...
+    // reading debug logs a bit easier, we force it to be zero otherwise.
+    (trans_type != ICacheTransTypeBranch) -> branch_addr == 0;
+
+    // If the transaction type is not ICacheTransTypeReq, force req_low_cycles to zero to make
+    // reading debug logs a bit easier. Otherwise, we pick a number of cycles to drop req_i for.
+    // Interesting values are zero (don't actually drop req_i), then short numbers, then long enough
+    // for the bus to become quiescent (maybe a hundred cycles or so), then ages: time for an
+    // invalidation to finish.
+    if (trans_type == ICacheTransTypeReq)
+      req_low_cycles dist { 0 :/ 20, [1:33] :/ 5, [100:200] :/ 2, [1000:1200] :/ 1 };
```
udinator, line 44: another style nit, for `dist` blocks, we generally have each range:distribution pairing on a new line for readability, as follows:
```systemverilog
req_low_cycles dist {  0           :/ 20,
                       [1:33]      :/ 5,
                       [100:200]   :/ 2,
                       [1000:1200] :/ 1
};
```

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
```diff
@@ -0,0 +1,24 @@
...
+  `uvm_object_utils(ibex_icache_sanity_seq)
+  `uvm_object_new
+
+  rand int count;
+  constraint c_count { count > 0; count < 100; }
+
+  task body();
+    // Generate a request which is constrained to have trans_type ICacheTransTypeBranch: the core
+    // must start with a branch to tell the cache where to fetch from in the first place.
+    `uvm_do_with(req, { req.trans_type == ICacheTransTypeBranch; });
```
udinator, line 18: as I mentioned earlier, we usually disallow the `uvm_do_*` macros

dv/uvm/icache/dv/tb/tb.sv
```diff
@@ -17,13 +17,28 @@ module tb;
 
   // interfaces
   clk_rst_if clk_rst_if(.clk(clk), .rst_n(rst_n));
-  ibex_icache_if ibex_icache_if();
-  ibex_mem_intf ibex_mem_intf();
+  ibex_icache_if core_if ();
```
udinator, line 20: Re: my previous comments about renaming this agent

udinator:
> For anyone who's been paying close attention: yes, the names are a bit odd here given that there's going to be a memory agent as well. I've got a followup patch that replaces `ibex_icache_$foo` with `ibex_icache_core_$foo`, but that patch is kind of noisy so I thought it would make more sense to split out implementing stuff and renaming it.

Does this patch re-generate the agent with the `uvmdvgen` tool? We want to keep the naming scheme that the tool uses.

udinator:
@rswarbrick - thanks for getting this PR opened! 
I've left several comments throughout, mostly pertaining to the driver/sequence_item with regards to the two types of high level transactions that you propose to drive into the icache.
A few other notes, since this PR is for the `ibex_icache_core_agent` (currently named `ibex_icache_agent`):
- You need to `::get()` the `ibex_icache_if` from the configuration database inside the `build_phase` of the agent and then assign it to a handle inside the agent's configuration object, as the dv_base_agent does not do that for you already.
- Would it be possible to implement the monitor and add it to this PR? I think it would make sense to include it in this PR to get all of the code for the ibex_icache_core_agent at once. If we do this, we can either leave this PR open or mark it as draft for the time being.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
```diff
@@ -4,8 +4,25 @@
 
 interface ibex_icache_if ();
```
udinator, line 5: Is `busy_o` not going to be in this interface? I thought it propagates through the rest of the Ibex core before being output at the top level, it shouldn't be an output of the memory interface since it relates to the invalidation signals.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
rswarbrick: I want to make sure I understand what you mean here. Are you suggesting that the sequence items should be simplified to "get an instruction", "drop req", "invalidate" and "branch"? (And possibly some others - I'm not sure I got everything) I agree that you could build this, but the problem is that there are relations that must hold between them.

For example, the core is not allowed to try to fetch instructions when the `req_i` line is low. Similarly, as soon as an instruction fetch returns an error, the core *must* issue a branch (hmm, actually I'm not sure the code gets that right at the moment: I'll add that to my TODO list).

So there's going to have to be code somewhere that knows how to glue together the very low level "pin wiggles" into transactions that are allowed by the cache interface. My understanding of UVM is that sequences generally contain logical transactions (which are "allowed") and then the driver is expected to turn these transactions into pin wiggles.

Note also that there are "timing issues": I want to be able to assert a branch at the same time as requesting an instruction, or before doing so. While I agree that it's possible to do this, I feel like it's a bit confusing when sequence items take zero time.

I can sort of see how you could do this with a virtual sequence. The virtual sequence would contain the items that I've got at the moment. It would then drive one or more sequences (I'm not quite sure) with the lower-level transactions, which would then get driven onto the DUT pins by the driver code. But that seems way more complicated and confusing than what's currently there.

Am I missing something?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
rswarbrick: Coming back to toggling the cache enable line: I want to make sure that the cache doesn't switch between enabled and disabled too often (I think that would make it quite hard to get in to some of the corners of the implementation). So I think the test *has* to work like "do some stuff.... enable... do some stuff.... disable..." where each "do some stuff" takes quite a while.

The code currently tries to do this by making a sequence item only occasionally change between enable/disable. At the moment this essentially works by storing a single bit of state on the interface itself, so that a "toggle me now" item can switch it.

Alternatively, I could change the code so that the sequence stores the state and the constraint that generates a new item encodes the "don't change too often" rule. Then each sequence item would contain the enable flag, rather than its edges. Is that what you meant? I'm not sure whether that's clearer, but I'd happily change to that if you think it's better.

rswarbrick:
> > For anyone who's been paying close attention: yes, the names are a bit odd here given that there's going to be a memory agent as well. I've got a followup patch that replaces `ibex_icache_$foo` with `ibex_icache_core_$foo`, but that patch is kind of noisy so I thought it would make more sense to split out implementing stuff and renaming it.
> 
> Does this patch re-generate the agent with the `uvmdvgen` tool? We want to keep the naming scheme that the tool uses.

No, because it's editing the existing code. But I think the naming is consistent - I made the patch with some `sed -i` calls! Let's come back to that when I push up that PR.

rswarbrick:
...
>     * You need to `::get()` the `ibex_icache_if` from the configuration database inside the `build_phase` of the agent and then assign it to a handle inside the agent's configuration object, as the dv_base_agent does not do that for you already.

Can you explain what that's for? I assumed that the driver and monitor needed to see the interface and no-one else (since that's the whole point of transaction level modelling!). Is there some other way it needs to be exposed?

>     * Would it be possible to implement the monitor and add it to this PR? I think it would make sense to include it in this PR to get all of the code for the ibex_icache_core_agent at once. If we do this, we can either leave this PR open or mark it as draft for the time being.

I'd rather not. Since I realise that you guys are under time pressure, I'm trying to avoid wasting your time with unnecessary or impossibly large review requests. For example, I've already applied some of your review comments from this PR to the code that I've got for a memory agent, so you'll never have to see the version that didn't conform to the style guide! That way, you shouldn't have to repeat yourself too much (which should avoid you wasting your time). Similarly, small PRs mean that I can get feedback as I go to avoid me writing stuff that I have to throw away (avoiding me wasting my time).

Also I want to actually land PRs in a finite time, so think it probably makes more sense to push up stuff in smallish chunks. So, I'd rather implement the monitor in a separate PR.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -26,12 +22,126 @@ class ibex_icache_driver extends dv_base_driver #(ibex_icache_item, ibex_icache_
...
+    join
+  endtask
+
+  // Drive the cache for a "req transaction".
+  //
+  // This lowers req for zero or more cycles, at the same time as setting the enable pin and (maybe)
+  // pulsing the invalidate line. Once that is done, it reads zero or more instructions.
+  virtual task drive_req_trans(ibex_icache_item req);
+    int unsigned req_low_cycles;
+    logic        allow_no_low_cycles;
```
udinator, line 67: nit: change to a `bit` type

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: you might also be able to just get away with `wait(cfg.vif.valid == 1'b1)`, which functionally does the same thing you are going for here

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
```diff
@@ -4,8 +4,56 @@
...
+  logic         enable;
+  logic         invalidate;
+
+  // Busy signal from the cache (either there's a request on the bus or the cache is invalidating
+  // itself)
+  logic         busy;
+
+  // Drive the branch pin for a single cycle, redirecting the cache to the given instruction
+  // address.
+  task branch_to(logic [31:0] addr);
```
udinator, line 34: these tasks need to be `automatic` as per the style guide.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
```diff
@@ -4,8 +4,56 @@
...
+
+  // Enable/disable or invalidate the cache
+  logic         enable;
+  logic         invalidate;
+
+  // Busy signal from the cache (either there's a request on the bus or the cache is invalidating
+  // itself)
+  logic         busy;
+
+  // Drive the branch pin for a single cycle, redirecting the cache to the given instruction
```
udinator, line 32: Will you ever want to pulse this for more than one cycle?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,78 @@
 
 class ibex_icache_item extends uvm_sequence_item;
 
-  // random variables
+  // The type of transaction
+  rand ibex_icache_trans_type_e trans_type;
+
+  // The branch address for a branch transaction (only has effect if trans_type is
+  // ICacheTransTypeBranch)
+  rand logic [31:0]             branch_addr;
```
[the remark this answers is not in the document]
udinator: This ties in to my (fairly long) comment about implementing the high level transactions themselves in the sequence level, so we can continue this discussion in that thread.

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
```diff
@@ -0,0 +1,31 @@
...
+    req = ibex_icache_item::type_id::create("req");
+    start_item(req);
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(req, req.trans_type == ICacheTransTypeBranch;)
+    finish_item(req);
+
+    // Generate and run count ibex_icache_item sequence items (with no other constraint) through the
+    // req port.
+    repeat (count - 1) begin
+      start_item(req);
+      `DV_CHECK_RANDOMIZE_WITH_FATAL(req, req.trans_type == ICacheTransTypeBranch;)
```
udinator, line 27: I might be missing something here, but why is every transaction (except the initial one) in the sanity test a branch transaction?

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
```diff
@@ -0,0 +1,31 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Sanity test seq
+//
+// This is unlikely to find many cache hits (since it branches all over the 4GiB address space).
+class ibex_icache_sanity_seq extends ibex_icache_base_seq;
```
udinator, line 8: This comment also ties into the above discussion about implementing an API at the sequence level, but the sanity test should be a virtual sequence `ibex_icache_sanity_vseq`, that starts one (or more) non-virtual sequences to send transactions to the DUT.

udinator:
> Can you explain what that's for? I assumed that the driver and monitor needed to see the interface and no-one else (since that's the whole point of transaction level modelling!). Is there some other way it needs to be exposed?

Sorry, you can disregard this comment, I misread one of the variable names...
It's fine as is, it already pulls the interface out of the configuration database and passes it to the agent's config object.

> I'd rather not. Since I realise that you guys are under time pressure, I'm trying to avoid wasting your time with unnecessary or impossibly large review requests. For example, I've already applied some of your review comments from this PR to the code that I've got for a memory agent, so you'll never have to see the version that didn't conform to the style guide! That way, you shouldn't have to repeat yourself too much (which should avoid you wasting your time). Similarly, small PRs mean that I can get feedback as I go to avoid me writing stuff that I have to throw away (avoiding me wasting my time).
> 
> Also I want to actually land PRs in a finite time, so think it probably makes more sense to push up stuff in smallish chunks. So, I'd rather implement the monitor in a separate PR.

Sure, that's fine with me, some discussions on this PR might go for a little while and don't really relate to the monitor, so it's fine in a later PR.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: It doesn't necessarily just have to be based off of reusability, it's also good to have as little clutter in the driver as possible since its job is so straightforward, so I'd lean towards putting that in the interface as well for style reasons, but I'll leave the final decision to you.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: Hmm , I see your point. Maybe we could get rid of the begin...end but keep the `if` statements on a single line?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
udinator: Sure, I understand what you are trying to do here, I agree that the cache should not switch between enabled/disabled too often as well. My concern is that by using a `toggle_enable` signal  to invert the current toggle signal, you won't know what state the cache is in other than by actively checking some internal signals for this status.
My feeling is that by explicitly specifying whether you want the cache enabled or disabled with an `enable` bus in the interface instead of `toggle_enable`, you can have better control of your test scenarios and can test error scenarios where the cache switches back and forth between enabled and disabled. I think that this is very similar to the alternate idea you proposed of storing the current state of the cache at a higher level of abstraction and having constraints control the enable flag instead of its edges; I'd lean towards an implementation like this to make things cleaner down the road when you're developing a library of test sequences

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: > I want to make sure I understand what you mean here. Are you suggesting that the sequence items should be simplified to "get an instruction", "drop req", "invalidate" and "branch"? (And possibly some others - I'm not sure I got everything) I agree that you could build this, but the problem is that there are relations that must hold between them.

Partially yes. Essentially what I'm suggesting is that the sequence items be simplified to contain (at a minimum) the inputs that will be driven into the icache (`req_i, branch_i, addr_i, icache_enable_i, icache_inval_i`), along with some corresponding delay values (if necessary for some test scenarios).  As in your code, `ready_i` would not be in the sequence item.
This seq_item would then be randomized appropriately in a sequence to reflect the desired transaction to send to the core (e.g. if the sequence wanted to send a branch request but disable the cache, it would randomize the seq_item, keeping `branch_i` at 1'b1, `icache_enable_i` at 0, and `addr_i` at whatever address the cache should jump to).

> For example, the core is not allowed to try to fetch instructions when the `req_i` line is low. Similarly, as soon as an instruction fetch returns an error, the core _must_ issue a branch (hmm, actually I'm not sure the code gets that right at the moment: I'll add that to my TODO list).

Correct me if I'm wrong, but the relations you mentioned sound like either they should be implemented with constraints in the sequence item or sequences. In regard to the `req_i` relation you describe, that sounds more like a condition for functional correctness, in that if a sequence sends a transaction where `req_i`is constrained to be low, then it should not expect any response from the cache.
The relation between an instruction fetch error and the cache having to issue a branch immediately is something that should be coordinated at a higher level (since the agent in this PR does not have visibility of any errors from the memory interface). It's definitely a good thing to think about from the perspective of stimulus generation though.

> So there's going to have to be code somewhere that knows how to glue together the very low level "pin wiggles" into transactions that are allowed by the cache interface. My understanding of UVM is that sequences generally contain logical transactions (which are "allowed") and then the driver is expected to turn these transactions into pin wiggles.

You're exactly right that somewhere there will be code that puts together a series of low level pin wiggles to form transactions, and that the uvm sequences create these transactions and send them to the driver after arbitration in the sequencer.

> Note also that there are "timing issues": I want to be able to assert a branch at the same time as requesting an instruction, or before doing so. While I agree that it's possible to do this, I feel like it's a bit confusing when sequence items take zero time.

Can you explain a bit more what you mean by sequence items taking zero time?
The timing issues you mention are very possible with this standard UVM approach, for example to send a branch at the same time as sending an instruction, the seq_item could simply be randomized to have both `branch_i` and `req_i` as 1'b1, and the same approach can be followed for asserting a branch before requesting an instruction, and so on.

> I can sort of see how you could do this with a virtual sequence. The virtual sequence would contain the items that I've got at the moment. It would then drive one or more sequences (I'm not quite sure) with the lower-level transactions, which would then get driven onto the DUT pins by the driver code. But that seems way more complicated and confusing than what's currently there.

I think you have the general idea of the virtual sequence usage, you essentially want it to coordinate between multiple nonvirtual sequences to send a series of transactions into the DUT. Each virtual sequence would effectively represent a different test scenario, which would then be modeled by a potentially different series of physical sequences that it starts, with various constraints to accurately model each testcase.
This is the standard UVM test methodology, and it is how we write our tests for all IPs except Ibex (it's a very nonstandard testbench, so I wouldn't use that as a reference here): the regular sequences are very small and contain a small set of constraints on the sequence item object to model a single (or a series of) transactions, and then the virtual sequences contain most of the complex testing logic and select one or more physical sequences to start to stress different parts of the design. You can look at some of the other IP testbenches to see active examples, such as UART and the alert handler.
Would you prefer that we set up a video call to discuss further? It might save much more time than daily updates to this PR thread.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
rswarbrick: Ah, right. Now I understand. Yep, I can do that.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
rswarbrick: Ok, I think I'm going to leave it as-is for now, because I think it's easier to understand. I don't have massively strong feelings though: I'm sure it can move easily if necessary in the future!

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
rswarbrick: Hi @udinator 

As you suggest, I think a video call sounds like a good idea. Beforehand, could you have a quick read through the icache documentation (specifically, the "detailed behaviour" section). The interface is rather weird and that document is the result of countless "oh, but what if..." conversations I had with Tom. In particular, I'm not sure you've understood what the super-confusingly-named `req_i` does (neither did I at first!).

As a general point, I *do* understand how constrained random test generation works: this is my first time writing UVM code, not my first time writing CR testbenches. About the zero time question, I thought you were suggesting that "branch" and "get an instruction" were separate sequence items. The problem is that they can both happen at once and we definitely need to test this. You can do this with separate items, but only if "branch" takes zero time (eugh).

About the high-level structure of the sequence items, this is kind of frustrating. The whole reason that I pushed up the DV plan documentation before pushing up code was to get reviews on the design. That document explicitly lists the different types of sequence items in the core agent, which I did to make sure that we would agree about the general thrust of how things worked.

Rather than you waking up to an invitation for a video call in 5 minutes, which seems like it would be a bit unreasonable(!), would you invite me to a meeting? I generally work until 17:30 or 18:00 UK time.

Thanks!

sjgitty: @rswarbrick @udinator @sriyerg @asb how about Tuesday 30 min
before the Silicon meeting, or Wednesday at 8:00am PDT 16:00pm
UK time?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: Ah, I see what you mean, if you are trying to only check for `cfg.vif.valid` on a `posedge`, then I'm ok with this.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: @rswarbrick - Sure, I will take a look at the Icache docs before the meeting.

@sjgitty - My vote is for Tuesday before the Silicon meeting.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
udinator: Understood.

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
udinator: I think it's fine for now, we can discuss this also in the meeting with Sri and the others.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
asb: 30 min before the silicon meeting seems like a plan to me.

dv/uvm/icache/dv/Makefile
sriyerg: Not anymore (if you can vendor in again from OT master).

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
udinator: marking as resolved

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
udinator: This looks good to me, I'll mark this thread as resolved.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_driver.sv
```diff
@@ -4,34 +4,126 @@
...
+
+    // How many cycles should we lower the request line? If there aren't any instructions to read,
+    // this is constrained to be positive (to avoid a confusing zero-time transaction).
+    allow_no_low_cycles = req.num_insns > 0;
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(req_low_cycles,
+                                       req_low_cycles dist {
+                                           0           :/ (allow_no_low_cycles ? 20 : 0),
+                                           [1:33]      :/ 5,
+                                           [100:200]   :/ 2,
+                                           [1000:1200] :/ 1 };)
```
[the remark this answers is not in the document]
rswarbrick: That makes sense, thanks. I think I'll leave it as-is for now and revisit if we find we're missing some coverage somewhere.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
```diff
@@ -4,9 +4,65 @@
...
+  rand bit                      invalidate;
+
+  // The number of instructions to read (always non-negative, but may be zero)
+  rand int                      num_insns;
+
+  // Constraints
+  constraint c_trans_type_dist {
+    // Force the distribution of trans_type (to avoid it getting skewed by the enormous space of
+    // different branch addresses)
+    trans_type dist { ICacheTransTypeBranch :/ 1, ICacheTransTypeReq :/ 1 };
```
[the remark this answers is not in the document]
rswarbrick: Ah, I'm obviously confused about SV randomization. I thought that if you had this:

    logic         foo;
    logic [9:0]   bar;

    constraint xxx {
        foo -> bar == 0;
    }

Then the 1025 possible assignments to foo and bar would each happen with the same frequency (see first paragraph of section 12.4.9 "Variable ordering" in the LRM version 3.1a). As I understood it, if you wanted `foo` to be true half of the time (rather than one time in 1025), you needed to either give an ordering constraint (`solve foo before bar;`) or explicitly set the distribution for `foo`.

Have I got this wrong, or is there something else that I've missed about this case?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
```diff
@@ -4,8 +4,66 @@
 
 interface ibex_icache_if ();
 
-  // interface pins
+  logic         clk;
```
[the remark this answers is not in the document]
rswarbrick: Yes, happily. Is that the preferred style? (Some interfaces have it as a member rather than an input port, including `ibex_mem_intf`, which I think is the one I copied).

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
```diff
@@ -4,8 +4,66 @@
...
+  logic         branch;
+  logic [31:0]  branch_addr;
+
+  // Passing instructions back to the core
+  logic         ready;
+  logic         valid;
+  logic [31:0]  rdata;
+  logic [31:0]  addr;
+  logic         err;
+  logic         err_plus2;
```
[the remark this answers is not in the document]
rswarbrick: Are there any examples in OpenTitan that do this already?

I've tried adding a clocking block and waiting on that instead of the explicit posedge, but it adds lots of lines of code and doesn't seem any clearer than what was there before, so I suspect I'm missing the point. Maybe there's existing code I could emulate?

dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
```diff
@@ -2,14 +2,21 @@
...
+
   task body();
-    `uvm_error(`gfn, "FIXME")
+    // TODO: This currently just drives the core sequence (which clearly isn't going to work!)
+    `uvm_create_on(core_seq, p_sequencer.core_sequencer_h)
+    `DV_CHECK_RANDOMIZE_FATAL(core_seq)
+    `uvm_send(core_seq)
```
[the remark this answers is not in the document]
rswarbrick: Done!

rswarbrick:
The latest version rebases on origin/master and switches the interface to taking a clock as an input. Other than that, no change.

rswarbrick:
The latest version adds a constraint that branch addresses must be halfword aligned (this isn't currently in the icache spec, but should be - I'll add that in a separate PR once I've finished the SVA module that I'm currently writing for this interface).

rswarbrick:
The latest version gets rid of the unnecessary responses that were being generated by the copy-paste initial code in `ibex_icache_driver.sv`. These were harmlessly getting generated and discarded when we used `` `uvm_do``, but I forgot to discard them explicitly when switching the sequence to use `start_item` / `finish_item`. I guess this is now a bit cleaner.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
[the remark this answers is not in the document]
udinator: Yep, actually most of the other IPs use clocking blocks in their interfaces, a good example would be the [i2c interface](https://github.com/lowRISC/opentitan/blob/master/hw/dv/sv/i2c_agent/i2c_if.sv).

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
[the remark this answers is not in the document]
udinator: If you're going for a 50/50 split between the two transaction types, can you use a `solve trans_type before branch_addr` clause in the `c_non_branch_trans_addr` constraint below instead of this distribution?
That should make things a bit more clear as to why this distribution is being enforced (implicitly or otherwise).

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
```diff
@@ -0,0 +1,31 @@
...
+    // Generate a request which is constrained to have trans_type ICacheTransTypeBranch: the core
+    // must start with a branch to tell the cache where to fetch from in the first place.
+    req = ibex_icache_item::type_id::create("req");
+    start_item(req);
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(req, req.trans_type == ICacheTransTypeBranch;)
+    finish_item(req);
+
+    // Generate and run count ibex_icache_item sequence items (with no other constraint) through the
+    // req port.
+    repeat (count - 1) begin
```
udinator, line 25: is the branch transaction included in the `count` amount?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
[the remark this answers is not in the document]
rswarbrick: If you think this would be clearer, I'll make the change.

What I wrote seems more "declarative" (in that it says what the results should be, rather than telling SV's randomization engine what order to do something in, in order to get the same result). It even explains why (which I'd expect to make sense to anyone who knew how SystemVerilog constraints worked). But if you believe that the other approach is better, I'll make that change and add a long comment explaining what it does. Please let me know if you think the change is needed.

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
rswarbrick: `count` is the number of sequence items. The first is forced to start with a branch (that's lines 19-21 above). The other `count-1` sequence items don't have the constraint.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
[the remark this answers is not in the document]
rswarbrick: Ok, just to make sure I understand what you're asking: I should add something like this?
```
  // An explicit clocking block. Port directions are from the environment's point of view.
  clocking cb @(posedge clk);
    output req;
    output branch;
    output branch_addr;
    output ready;
    input  valid;
    input  rdata;
    input  addr;
    input  err;
    input  err_plus2;
    output enable;
    output invalidate;
    input  busy;
  endclocking
```
And then replace `@(posedge cfg.vif.clk)` in the driver/monitor etc. with `@(cfg.vif.cb)`? This doesn't seem to have any informational content to me, but I guess I don't have strong feelings...

dv/uvm/icache/dv/ibex_icache_agent/seq_lib/ibex_icache_sanity_seq.sv
udinator: got it, thanks

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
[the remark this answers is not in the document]
udinator: Makes sense. I'm personally biased a little in favor of the `solve...before` syntax, but both should be equivalent here so I'm perfectly fine with either, just wanted to bring up an alternate option.
I'll leave the final decision up to you though as to whichever you think is the clearest to understand.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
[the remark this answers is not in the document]
udinator: It would look something like this:

```systemverilog
clocking driver_cb @(posedge clk_i);
  output req;
  output branch;
  output branch_addr;
  output ready;
  output enable;
  output invalidate;
endclocking
// Optional modport
modport drv_mp(clocking drv_cb);

clocking monitor_cb @(posedge clk_i);
  input valid;
  input rdata;
  input addr;
  input err;
  input err_plus_2;
  input busy;
endclocking
// Optional modport
modport mon_mp(clocking mon_cb);
```

Note the optional use of the `modport`, it's not necessary but you can add them if you choose to.

Like you said, you can then wait on `cfg.vif.[driver/monitor]_cb` in your driver and monitor components instead of referencing the `clk` directly to have a clear separation between signals you are driving and signals you are simply monitoring.
I think the clocking blocks would be a good idea, but don't feel like modports are necessary.

@sriyerg - what do you think?

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_item.sv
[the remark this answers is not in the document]
sriyerg: Ohh, I now realize why you did this. The `solve..before` construct is provided exactly for this reason. It is a lot clearer and a more widely used construct, so I don't think a long comment explaining it is necessary. Plus, I would imagine adding ordering reduces constraint solver complexity by reducing the state space for the rest of the variables, which could in theory, improve sim time (just a conjecture, not sure if this is actually true). dvsim has a `--sim-profile` switch that works with VCS. If you have the time, you can quickly run this experiment and check if this is true.

In any case, I am fine with existing approach as well, considering it achieves the same result.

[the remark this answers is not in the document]
rswarbrick: I've switched this across, folding the constraint into `c_non_branch_trans_addr` as Udi suggested.

dv/uvm/icache/dv/ibex_icache_agent/ibex_icache_if.sv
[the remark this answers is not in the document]
rswarbrick: I'm going to wait on this and save it for another PR. I think I understand what you're getting at (and I've just read the relevant bits of Cummings's paper). However, I'm still a bit stuck on how to structure things like the driver where it waits for a signal, then drives another (does that go through the two different clocking blocks?). I'm going to open an issue to remind me that this needs sorting. Maybe we can discuss further details there or in a call.

PR #762, opened 2020-04-03
[dv] Remove IUS support

GregAC:
IUS lacks support for certain language features used in Ibex (such as
use of $clog2 in localparam definitions) so remove it as a simulator
that can be used by dv.

I did a quick grep through dv/uvm and I think I've removed anything that refers to IUS.


dv/uvm/core_ibex/sim.py
```diff
@@ -537,13 +537,7 @@ def main():
     # Run RTL simulation
     if steps['sim']:
         check_return_code = True
-        # Don't check return code for IUS sims, as a failure will short circuit
-        # the entire simulation flow
         check_return_code = True
-        if args.simulator == "ius":
-            check_return_code = False
-            logging.debug("Disable return code checking for %s simulator"
-                          % args.simulator)
```
rswarbrick, line 546: I think this means we can strip out the `check_return_code` stuff entirely. (And from `rtl_sim` and `run_sim_commands`).

rswarbrick: Uh, I just re-read this and my comment wasn't very clear. I meant to say something like "Could you rip out all the `check_return_code` stuff from sim.py? Then I reckon the patch is good to go."

rswarbrick:
Looks good to me!

PR #765, opened 2020-04-06
Initial icache testplan

rswarbrick:
This doesn't list all that many entries, possibly because the icache
doesn't really have many modes of operation (enabled/disabled;
invalidating).

### DIFF
--- a/dv/uvm/icache/data/ibex_icache_testplan.hjson
+++ b/dv/uvm/icache/data/ibex_icache_testplan.hjson
@@ -7,24 +7,95 @@
   entries: [
     {
       name: sanity
-      desc: '''**Goal**: Basic sanity test acessing a major datapath in IBEX_ICACHE.
+      desc: '''Basic sanity test for icache
 
-            **Stimulus**: Describe the stimulus procedure.
+            Make a series of requests with unconstrained addresses, branching
+            occasionally and disabling the request line occasionally (modelling
+            the core going to sleep). Toggle whether the cache is enabled.
+            Invalidate the cache very occasionally.
 
-            **Checks**": Describe the self-check procedure.
-            - add bullets as needed
-            - second bullet<br>
-              describe second bullet
-
-            Start a new paragraph.'''
+            Self-checking performed as described in the DV plan (the same
+            self-check tests are run for all the tests described below, as well
+            as any extra tests they describe).'''
       milestone: V1
       tests: ["ibex_icache_sanity"]
     }
+
     {
-      name: feature1
-      desc: '''Add more test entries here like above.'''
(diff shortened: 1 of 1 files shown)


rswarbrick:
This feels a little thin on the ground, especially compared to the [AES testplan](https://docs.opentitan.org/hw/ip/aes/doc/dv_plan/#testplan), but I'm struggling to think of other behaviours to test (which aren't already tested by the sanity test and similar). I'd love to hear of other ideas!

udinator:
@rswarbrick - Can you set up a testplan review meeting with Tom, Scott, and some other DV folks (Sri, Cindy, Weicai)? We'll get this PR merged after the meeting, after a wider audience has a chance to provide any input they might have.

rswarbrick:
@udinator Yep, can do. I'm not sure what times are best for you guys - since we're going to talk tomorrow afternoon anyway, let's figure out a good time then.

udinator:
@rswarbrick Sure, that sounds good to me!

dv/uvm/icache/data/ibex_icache_testplan.hjson
```diff
@@ -7,24 +7,95 @@
   entries: [
```
udinator, line 7: How is error handling going to be tested (both PMP errors and "IMEM" errors)? Will there be explicit tests for the error lines, or will those be toggled during all of these tests?

In addition, how will resets be tested?

rswarbrick: In all tests, some memory regions will return errors (PMP or actual bus errors). The scoreboard will be checking they are handled correctly. So I think this comes "for free".

About resets: Good point, I haven't thought about that. Do we have any documentation about what we're actually testing for with them? (After all, the world comes up with all signals as 'X', so in some sense we do a reset test every time!)

udinator: Re error testing: Got it. Just a suggestion, what do you think about disabling errors in the sanity test (to make it as "sane" as possible) and creating a new error stressing test where all errors are asserted much more frequently? I'm assuming that in the rest of the tests errors won't be asserted super frequently (but correct me if I'm wrong).
Maybe another interesting scenario would be having an error asserted during invalidation.

Re reset testing: It sort of varies per IP, for example in Ibex when I test resets I just make sure the core can reset, jump to the start address, and finish the test without much other checking. For the Icache there would be some more security-related implications (we don't want old data being accessible after a reset clears/invalidates everything), so just going off of the existing documentation my guess is just that you would want to reset the cache and then guarantee that everything gets invalidated afterwards.

PR #766, opened 2020-04-06
Ibex xprop fixes

GregAC:


rtl/ibex_cs_registers.sv
```diff
@@ -1077,13 +1074,7 @@ module ibex_cs_registers #(
...
-  // Selectors must be known/valid.
-  `ASSERT(IbexCsrOpValid, csr_op_i inside {
-      CSR_OP_READ,
-      CSR_OP_WRITE,
-      CSR_OP_SET,
-      CSR_OP_CLEAR
-      })
-  `ASSERT_KNOWN(IbexCsrWdataIntKnown, csr_wdata_int)
+  `ASSERT(IbexCsrOpEnRequiresAccess,
+    csr_op_en_i |-> csr_access_i)
```
rswarbrick, line 1078: Nit: can this be a single line?

rtl/ibex_id_stage.sv
```diff
@@ -893,21 +893,21 @@ module ibex_id_stage #(
   ////////////////
 
   // Selectors must be known/valid.
-  `ASSERT_KNOWN(IbexAluOpMuxSelKnown, alu_op_a_mux_sel, clk_i, !rst_ni)
-  `ASSERT(IbexAluAOpMuxSelValid, alu_op_a_mux_sel inside {
+  `ASSERT_KNOWN_IF(IbexAluOpMuxSelKnown, alu_op_a_mux_sel, instr_valid_i)
+  `ASSERT(IbexAluAOpMuxSelValid, instr_valid_i |-> alu_op_a_mux_sel inside {
```
rswarbrick, line 897: There are a lot of these. Would it make sense to combine them into an "`ASSERT_INSIDE_IF" macro?

shared/rtl/prim_assert.sv
```diff
@@ -53,36 +53,46 @@
...
-`define ASSERT_I(__name, __prop)              \
-`ifdef INC_ASSERT                             \
-  __name: assert (__prop)                     \
-    else `ASSERT_RPT(`PRIM_STRINGIFY(__name)) \
+`define ASSERT_I(__name, __prop)           \
+`ifdef INC_ASSERT                          \
+  __name: assert (__prop)                  \
+    else begin                             \
+      `ASSERT_RPT(`PRIM_STRINGIFY(__name)) \
+    end                                    \
```
rswarbrick, line 61: Possibly a stupid question, but why do we need `begin` / `end` here now?

rtl/ibex_id_stage.sv
GregAC: I considered it, one issue is due to the nature of macros you'd need to use { } in the macro usage, e.g.

```systemverilog
`ASSERT_INSIDE_IF(alu_op_a_mux_sel, 
      {OP_A_REG_A,
      OP_A_FWD,
      OP_A_CURRPC,
      OP_A_IMM},
      instr_valid_i
)
```

Which makes it a bit more awkward (or more likely to produce confusing errors if you get this wrong).

I also felt the current way looks pretty clean and understandable so didn't see much value in the extra macro. The primary reason I like wrapped xchecks up in macros is the `!$isunknown` call looks a bit awkward so a specific xcheck macro looks cleaner.

Another reason to like xcheck macros is you can quickly find all the xcheck assertions. Which we loose for anything using inside currently. So a macro may be a good idea but I didn't go with it this time round.

shared/rtl/prim_assert.sv
GregAC: I'm not sure we do, but as updating prim_assert is sufficiently annoying (as it's shared with OT) I'd prefer to leave this as is.

rtl/ibex_cs_registers.sv
GregAC: Sure

shared/rtl/prim_assert.sv
rswarbrick: Oh, that's already landed in OT? Ok.

imphil: It's a safety measure, see https://github.com/lowRISC/opentitan/pull/1762

PR #767, opened 2020-04-06
Dv error changes

GregAC:
Before I added the set_max_quit_count call I had a failure (xprop related that caused several assertions to fire each cycle) that caused a ~2GB log file and a failure to dump waves (the simulation timed out which seemed to mess up the wave dump).

I set the default to 10 rather than 1 as it can be useful to let the simulation run a little past the first error if only to get another cycle or two in the waves.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -89,9 +89,12 @@ def process_ibex_sim_log(ibex_log, csv, full_trace=1):
     log and save to a standard CSV format.
     """
     logging.info("Processing ibex log : %s" % ibex_log)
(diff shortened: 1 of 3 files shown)


GregAC:
Thanks for the rapid review @rswarbrick @udinator, update I've pushed just alters a comment in sim.py to remove reference to raising exceptions as it was now incorrect/confusing.

sriyerg:
> Before I added the set_max_quit_count call I had a failure (xprop related that caused several assertions to fire each cycle) that caused a ~2GB log file and a failure to dump waves (the simulation timed out which seemed to mess up the wave dump).
> 
> I set the default to 10 rather than 1 as it can be useful to let the simulation run a little past the first error if only to get another cycle or two in the waves.

Why not set the default to 2? Most typically, after the first error, the sim is done for.

GregAC:
I just went for 1 for the quit count in the end, it is easy enough to adjust via the plus arg or switching the default with a new PR if we decide a higher number is called for.

PR #768, opened 2020-04-08
[dv] Fix running on LSF

udinator:
Minor fixes to fix running commands on LSF:
1) Adding the LSF_CMD makefile option to the compile and rtl_sim steps to actually pass in the 'bsub' command
2) Fixing a syntax error in one of the `sim.py` functions that ran whenever lsf is enabled


udinator:
@rswarbrick - regarding mypy, could you create an issue for this, and we can discuss there? Just to avoid cluttering this PR (I also need to read up on mypy, never used it before).

PR #772, opened 2020-04-14
Switch from 'bool' to 'int' parameters in fusesoc core files

rswarbrick:
Fusesoc has an unfortunate bug[1] where a boolean parameter which has
default true can't be disabled. For now, just make all our boolean
parameters back into integers again. In the future, when that's fixed,
maybe we should switch things back.

[1] https://github.com/olofk/fusesoc/issues/392

### DIFF
--- a/dv/cs_registers/tb_cs_registers.core
+++ b/dv/cs_registers/tb_cs_registers.core
@@ -59,10 +59,10 @@ scripts:
 
 parameters:
   PMPEnable:
-    datatype: bool
+    datatype: int
     paramtype: vlogparam
-    default: true
-    description: PMP enabled [true/false]
+    default: 1
+    description: PMP enabled [1/0]
   PMPNumRegions:
     datatype: int
     paramtype: vlogparam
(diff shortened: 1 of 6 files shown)


GregAC:
I also note that whilst CI is passing the ibex_config.py is failing for some reason (though of course works fine when it's run standalone as the first step...). I should look into what's happening.

rswarbrick:
@GregAC Is this as simple as changing the True/False constants in `ibex_configs.yaml` to 1/0, or is there something cleverer I need to do?

GregAC:
just switch them to 0/1

GregAC:
Could also remove support for boolean parameters altogether from ibex_config.py but I think it's fine to just avoid using them for now.

rswarbrick:
@GregAC Done!

rswarbrick:
> Could also remove support for boolean parameters altogether from ibex_config.py but I think it's fine to just avoid using them for now.

I agree, let's leave them in. Eventually, we'll hopefully be able to use bools again with fusesoc.

rswarbrick:
Cool. If I forget, feel free to rebase it yourself or ping me a reminder!

PR #773, opened 2020-04-14
[ci] Fix multi-config CI

GregAC:
This PR includes some testing bits and pieces so shouldn't be merged yet, will resubmit a version without them once I'm happy everything is working

Multi-config CI wasn't actually trying multiple configurations. This
fixes that issue and uses a less fragile method of producing fusesoc
options. They are generated once and stored in a variable so we cannot
accidentally break one or more steps by using an incorrect
ibex_config.py command in one step whilst using a correct
ibex_config.py in the display step (which is also intended to check the
ibex_config.py command is correct).


GregAC:
This looks to be working so is ready to merge.

PR #774, opened 2020-04-14
Teach check_tool_requirements to check for edalize versions

rswarbrick:
We need this specific edalize version because recent verilators have
got pickier about string parameter passing, breaking the
"MultiplierImplementation" parameter.

As well as teaching check_tool_requirements.py to get the edalize
version from pip3, this patch also does a bit of tidying up, coping
better if tool_requirements.py is missing or malformed.

### DIFF
--- a/util/check_tool_requirements.py
+++ b/util/check_tool_requirements.py
@@ -6,15 +6,50 @@
 from distutils.version import StrictVersion
 import logging as log
 import os
+import re
 import subprocess
 import sys
 
 # Display INFO log messages and up.
 log.basicConfig(level=log.INFO, format="%(levelname)s: %(message)s")
 
-# Populate __TOOL_REQUIREMENTS__
-topsrcdir = os.path.join(os.path.dirname(__file__), '..')
-exec(open(os.path.join(topsrcdir, 'tool_requirements.py')).read())
+
+def get_tool_requirements_path():
+    '''Return the path to tool_requirements.py, at the top of the Ibex repo'''
+    # top_src_dir is the top of the repository
+    top_src_dir = os.path.normpath(os.path.join(os.path.dirname(__file__),
+                                                '..'))
+
+    return os.path.join(top_src_dir, 'tool_requirements.py')
+
+
+def read_tool_requirements(path=None):
+    '''Read tool requirements from a Python file'''
+    if path is None:
+        path = get_tool_requirements_path()
+
+    with open(path, 'r') as pyfile:
+        globs = {}
+        exec(pyfile.read(), globs)
+
+        # We expect the exec call to have populated globs with a
+        # __TOOL_REQUIREMENTS__ dictionary.
+        reqs = globs.get('__TOOL_REQUIREMENTS__')
+        if reqs is None:
+            log.error('The Python file at {} did not define '
+                      '__TOOL_REQUIREMENTS__.'
+                      .format(path))
+            return None
+
+        # reqs should be a dictionary (mapping tool name to minimum version)
+        if not isinstance(reqs, dict):
+            log.error('The Python file at {} defined '
+                      '__TOOL_REQUIREMENTS__, but it is not a dict.'
+                      .format(path))
+            return None
+
+        return reqs
+
 
 def get_verilator_version():
     try:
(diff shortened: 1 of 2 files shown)


util/check_tool_requirements.py
```diff
@@ -32,8 +56,39 @@ def get_verilator_version():
...
-def check_version(tool_name, required_version, actual_version):
-    if required_version is None or actual_version is None:
+
+def pip3_get_version(tool):
+    '''Run pip3 to find the version of an installed module'''
+    cmd = ['pip3', 'show', tool]
+    proc = subprocess.run(cmd,
+                          stderr=subprocess.DEVNULL,
+                          stdout=subprocess.PIPE,
+                          universal_newlines=True)
```
imphil, line 66: Use `check=True` and handle the exception

util/check_tool_requirements.py
```diff
@@ -32,8 +56,39 @@ def get_verilator_version():
...
+    if proc.returncode:
+        log.error('Failed to get version of {} with pip3: is it installed?'
+                  .format(tool))
+        return None
+
+    version_re = 'Version: (.*)'
+    for line in proc.stdout.split('\n'):
```
imphil, line 73:
```suggestion
    for line in proc.stdout.splitlines():
```

util/check_tool_requirements.py
```diff
@@ -32,8 +56,39 @@ def get_verilator_version():
...
         return None
 
-def check_version(tool_name, required_version, actual_version):
-    if required_version is None or actual_version is None:
+
+def pip3_get_version(tool):
+    '''Run pip3 to find the version of an installed module'''
+    cmd = ['pip3', 'show', tool]
+    proc = subprocess.run(cmd,
+                          stderr=subprocess.DEVNULL,
```
imphil, line 64: Remove this line to capture stderr with the option to print it if processing fails.

util/check_tool_requirements.py
```diff
@@ -6,15 +6,39 @@
...
 import os
+import re
 import subprocess
 import sys
 
 # Display INFO log messages and up.
 log.basicConfig(level=log.INFO, format="%(levelname)s: %(message)s")
 
-# Populate __TOOL_REQUIREMENTS__
-topsrcdir = os.path.join(os.path.dirname(__file__), '..')
```
imphil, line 16: Please keep the root of the repository as a separate variable somewhere to better understand the hierarchy and make it easier to adjust as the file is moved around.

util/check_tool_requirements.py
```diff
@@ -47,17 +102,28 @@ def check_version(tool_name, required_version, actual_version):
...
     return 0
 
+
 if __name__ == "__main__":
-    sys.exit(main())
+    try:
+        sys.exit(main())
+    except (RuntimeError, ValueError) as err:
+        sys.stderr.write('Error: {}\n'.format(err))
+        sys.exit(1)
```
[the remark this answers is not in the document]
rswarbrick: But they aren't internal program errors: they are things that can happen if you edit a different file (`tool_requirements.py`). If you believe we shouldn't propagate the "aargh! stop now" logic using an exception, I can change the logic to do it C-style with return values.

util/check_tool_requirements.py
rswarbrick: Done.

util/check_tool_requirements.py
rswarbrick: Ok.

util/check_tool_requirements.py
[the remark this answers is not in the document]
imphil: I see tool_requirements.py as an internal file to the tool. It's intentionally not a full config file, it's just a couple constants which happen to live in another file. Otherwise you end up with with more config file parsing/validation logic than you have "business logic."

util/check_tool_requirements.py
rswarbrick: Ok.

util/check_tool_requirements.py
rswarbrick: Thanks. Done.

rswarbrick:
@imphil I understand your explanation and fundamentally disagree. My take is that any file that gets loaded and parsed by a program needs at least basic sanity checking. I believe that `tool_requirements.py` is a config file, implemented a fragment of Python because it makes the implementation easier for us.

The changes in this patch mean that we no longer splice arbitrary variables into `globals()` (which I wasted several hours debugging in the past: don't do this!). They also check that the magic variable has been defined by the config file and that it has a sane type, so that looking stuff up in it will work.

Before this patch, an typo in `tool_requirements.py` will probably result in a failure in a different part of the code from the loading. It also defeats any static analysis tools (try running flake8 on the code before this patch: it moans that `__TOOL_REQUIREMENTS__` is used without being defined).

rswarbrick:
The most recent patch gets rid of any explicit exception throwing from the code: it now signals errors with return values and prints messages with `log.error`.

PR #775, opened 2020-04-14
Reorder statements to prevent latch inference

jrrk2:
This change prevents latches from being inferred by Vivado for unused counters for variable mhpmcounter_incr

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -860,6 +860,11 @@ module ibex_cs_registers #(
   // event selection (hardwired) & control
   always_comb begin : gen_mhpmcounter_incr
 
+    // Assign inactive counters (first to prevent latch inference)
+    for (int unsigned i=0; i<32; i++) begin : gen_mhpmcounter_incr_inactive
+      mhpmcounter_incr[i] = 1'b0;
+    end
+
     // When adding or altering performance counter meanings and default
     // mappings please update dv/verilator/pcount/cpp/ibex_pcounts.cc
     // appropriately.
@@ -878,11 +883,6 @@ module ibex_cs_registers #(
     mhpmcounter_incr[10] = instr_ret_compressed_i; // num of compressed instr
     mhpmcounter_incr[11] = mul_wait_i;             // cycles waiting for multiply
     mhpmcounter_incr[12] = div_wait_i;             // cycles waiting for divide
-
-    // inactive counters
-    for (int unsigned i=3+MHPMCounterNum; i<32; i++) begin : gen_mhpmcounter_incr_inactive
-      mhpmcounter_incr[i] = 1'b0;
-    end
   end
 
   // event selector (hardwired, 0 means no event)


rtl/ibex_cs_registers.sv
```diff
@@ -864,6 +864,10 @@ module ibex_cs_registers #(
     // mappings please update dv/verilator/pcount/cpp/ibex_pcounts.cc
     // appropriately.
     //
+    // inactive counters (first to prevent latch inference)
+    for (int unsigned i=0; i<32; i++) begin : gen_mhpmcounter_incr_inactive
+      mhpmcounter_incr[i] = 1'b0;
+    end
```
GregAC, line 870: nit: Add a newline here to seperate the for loop and the active counters below

rtl/ibex_cs_registers.sv
```diff
@@ -879,10 +883,6 @@ module ibex_cs_registers #(
     mhpmcounter_incr[11] = mul_wait_i;             // cycles waiting for multiply
     mhpmcounter_incr[12] = div_wait_i;             // cycles waiting for divide
```
GregAC, line 885: nit: remove this newline

rtl/ibex_cs_registers.sv
```diff
@@ -864,6 +864,10 @@ module ibex_cs_registers #(
     // mappings please update dv/verilator/pcount/cpp/ibex_pcounts.cc
     // appropriately.
```
GregAC, line 865: nit: This comment would be better above the active counter assignments.

rtl/ibex_cs_registers.sv
rswarbrick: Could you expand the comment here? Something like "Note: it's important that this comes first, otherwise Vivado 2019.2 infers a latch". Without something like this, a future editor might change the order for whatever reason, check all is well in their non-Vivado simulator/synthesis tool, and break stuff again.

jrrk: I would have thought it was obvious where the latches arise. But feel free to add or delete as much white space as you feel like.

rswarbrick: @jrrk: I'm sorry, I don't understand your comment. As far as I can tell, this PR is about a tool-specific coding style. At least one experienced RTL designer (Greg, not me!) has already said he's surprised it's needed. So we need a comment in the code explaining what's going on.

jrrk: I used Vivado 2018.3. I don’t have other tools available to try. It seems to me obvious that to avoid latch inference you must make sure a vector is always fully defined in every circumstance (which is easily done by giving it a default value).

GregAC: Certainly with the default parameter set the `mhpmcounter_incr` vector is fully defined (first 13 have explicit assignments then the rest are assigned in the for loop). Were you to use an `MHPMCounterNum` value greater than 10 then you'll end up with some `mhpmcounter_incr` elements never getting assigned. I think by Verilog semantics they end up as constant `x` in that case. You're into the realms of undefined/poorly defined behaviour for synthesis there so creating a latch may not be unreasonable if a bit odd.

You're correct that fully defining the vector as done in this PR is a sensible thing to do.

Out of interest do you know what `MHPMCounterNum` was set to in the run that inferred the latch? If it was greater than 10 (e.g. set to 32) then the latch inference is easier to understand. If it was 10 or less than every element of `mhpmcounter_incr` has been assigned a value combinationaly so a latch inference looks like a synthesis bug.

We should check to see if there's any other dodgy bits of RTL that don't work well if you have an `MHPMCounterNum` greater than 10.

jrrk: It was set to 29 for some unknown reason.

rtl/ibex_cs_registers.sv
tomroberts-lowrisc: Fixed

rtl/ibex_cs_registers.sv
tomroberts-lowrisc: Fixed

rtl/ibex_cs_registers.sv
tomroberts-lowrisc: Fixed

PR #779, opened 2020-04-15
Rename ibex_icache_agent -> ibex_icache_core_agent

rswarbrick:
There is more than one icache-specific agent that we need for the
icache testbench, so "ibex_icache_agent" isn't a very helpful name.
This commit was pretty much automated, except for a few spacing
cleanups, with commands like:

    git grep -l ibex_icache_agent | \
      xargs sed -i 's!ibex_icache_agent!ibex_icache_core_agent!g'

(and then rename the directory and files).


dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -11,11 +11,11 @@ class ibex_icache_scoreboard extends dv_base_scoreboard #(
   // local variables
 
   // TLM agent fifos
-  uvm_tlm_analysis_fifo #(ibex_icache_item) ibex_icache_fifo;
+  uvm_tlm_analysis_fifo #(ibex_icache_core_item) ibex_icache_fifo;
```
udinator, line 14: Should the instance name `ibex_icache_fifo` also be renamed to `ibex_icache_core_fifo`?
I believe this is also a relic of the old naming scheme.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -11,11 +11,11 @@ class ibex_icache_scoreboard extends dv_base_scoreboard #(
...
   uvm_tlm_analysis_fifo #(ibex_mem_intf_seq_item) ibex_mem_intf_slave_fifo;
 
   // local queues to hold incoming packets pending comparison
-  ibex_icache_item ibex_icache_q[$];
+  ibex_icache_core_item ibex_icache_q[$];
```
udinator, line 18: same with `ibex_icache_q`

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Can do, although the prefixes are getting ridiculous, so I'll just rename it `core_fifo` (we have namespacing in the language; let's use it!)

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: I've actually got rid of these queues entirely: I've got some prototype scoreboard code I'm working on now, and they are not needed. I'll re-instate them if necessary later.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: SGTM

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: Got it, sounds reasonable to me.

PR #782, opened 2020-04-15
[bitmanip] Bugfix: Erroneous Illegal Instructions for RV32B instructions and Decoder Bit-Slicing

ganoam:
This commit fixes a flaw in the decoder, which results in wrongly
generating an illegal instruction upon an orc_b instruction.

* Fix decoding of orc_b in illegal_insn generation.

* Correct bit-slicing for bitmanip immediate instructions, according
        to specification.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -335,14 +335,14 @@ module ibex_decoder #(
           3'b111: illegal_insn = 1'b0;
 
           3'b001: begin
-            unique case (instr[31:25])
-              7'b000_0000: illegal_insn = 1'b0;                 // slli
-              7'b001_0000: illegal_insn = RV32B ? 1'b0 : 1'b1;  // sloi
-              7'b011_0000: begin
-                unique case(instr[24:20])
-                  5'b00000,                                     // clz
-                  5'b00001,                                     // ctz
-                  5'b00010: illegal_insn = RV32B ? 1'b0 : 1'b1; // pcnt
+            unique case (instr[31:27])
+              5'b0_0000: illegal_insn = 1'b0;                      // slli
+              5'b0_0100: illegal_insn = RV32B ? 1'b0 : 1'b1;       // sloi
+              5'b0_1100: begin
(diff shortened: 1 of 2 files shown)


vogelpi:
Thanks for fixing this @ganoam . As discussed in the meeting earlier, the bit-slicing should be done according to the spec (what this PR ensures), even if the spec might change at some point.

I think this PR needs to be rebased as the Zbt PR has been merged now.

ganoam:
Ok, I have rebased and updated the PR.

The illegal instructions for orc.b are fixed. 

Bit slicing is now according to spec, ignoring bits `insn[26:25]` for decoding.

Also, I fixed another small inconsistency, see comment in code.

rtl/ibex_decoder.sv
```diff
@@ -383,11 +388,11 @@ module ibex_decoder #(
         rf_ren_a_o      = 1'b1;
         rf_ren_b_o      = 1'b1;
         rf_we           = 1'b1;
```
ganoam, line 390: Bit `insn[31]` high is no longer a guaranteed illegal instruction. For Ternary bitmanip instructions, this is part of the adress for rs3.

ganoam:
According to our private discussion, I have changed the decoder to no longer explicitly check bit `Insn[31]` to generate illegal instructions. I adjustied  ALU instruction decoding, such that the complete relevant bit ranges are checked in the case statements to assign operator signal and generate illegal instruction signals. 

This commit now does the following:

* Fix decoding of orc_b in illegal_insn generation.

* Insn[31] is no longer checked separately for generation of illegal instructions:
        This bit is part of the rs3 register adress for ternary
        bitmanipulation instructions (zbt). 



* Correct bit-slicing for ALU reg-immediate instructions according
        to specification: immediates are encoded in the range
        insn[26:20] in all cases. Where a shift-amount is encoded, bits
        [26:25] will have no effect, but will no longer generate
        illegal instructions.

ganoam:
Adjusted formating of bitstrings

Issue #778, opened 2020-04-15
[dv] Add clocking blocks to icache UVM code


rswarbrick:
See discussion on PR #761, in particular [these comments](https://github.com/lowRISC/ibex/pull/761#discussion_r405777417).

Sri and Udi have suggested that the code would be better if the interface included clocking blocks. Suggested code:
```systemverilog
clocking driver_cb @(posedge clk_i);
  output req;
  output branch;
  output branch_addr;
  output ready;
  output enable;
  output invalidate;
endclocking

clocking monitor_cb @(posedge clk_i);
  input valid;
  input rdata;
  input addr;
  input err;
  input err_plus_2;
  input busy;
endclocking
```

rswarbrick:
For example, some code in the driver currently looks like this:
``` systemverilog
    // Assert ready and then wait until valid
    cfg.vif.ready <= 1'b1;
    while (1'b1) begin
      @(posedge cfg.vif.clk);
      if (cfg.vif.valid)
        break;
    end
```
This reads from `valid` (driven by the DUT, so in `monitor_cb` above) and writes to `ready`. They share a clock. Is two clocking blocks the right way to structure this? I'd appreciate some guidance!

rswarbrick:
@udinator, @sriyerg: Could you give me some advice about how this should look?

rswarbrick:
Reading [Cummings's paper](http://www.sunburst-design.com/papers/CummingsSNUG2006Boston_SystemVerilog_Events.pdf) more carefully, I think I have an idea of how this would work.

Use a single clocking block, specifying `default output negedge` to make it more obvious what's going on in dumped waves. The code above would be:

``` systemverilog
cfg.vif.cb.ready = 1'b1;
while (1'b1) begin
  @(cfg.vif.cb);
  if (cfg.vif.cb.valid)
    break;
end
```

Suppose that there's a clock with a posedge every 100ns and let's pretend that `#1step` is 1ns. Then I think a sample execution would look like

Time (ns) | Event
---|--
0 | `cfg.vif.cb.ready = 1'b1` executed
51 | ready = 1 actually happens
... | ...
700 | valid = 1 (set by an `@(posedge clk)` block in DUT)
799 | valid sampled by cb (`#1step` before time step at 800ns)
800 | break

Does that make sense? If so, I propose to use a single clocking block and I'll push up a PR tomorrow. If not, can you let me know what would be better?

rswarbrick:
[...]

PR #786, opened 2020-04-16
Add clocking blocks to icache core agent interface

rswarbrick:
As discussed in issue #778.

The other two patches in the queue are embarrassing: it seems my "edit a long patch queue and push something near the bottom" flow missed a couple of typos which were fixed further down the list. I've checked, and you can actually run the code after this.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
@@ -27,16 +27,52 @@ interface ibex_icache_core_if (input clk);
   // itself)
   logic         busy;
 
+  // A clocking block for test code that drives the interface
+  clocking driver_cb @(posedge clk);
+
+    // Drive signals on the following negedge: this isn't needed by the design, but makes it
+    // slightly easier to read dumped waves.
+    default output negedge;
+
+    output req;
+
+    output branch;
+    output branch_addr;
+
+    output ready;
+    input  valid;
+    input  err;
+
+    output enable;
+    output invalidate;
+  endclocking
+
+  // A clocking block for test code that needs to monitor the interface
+  clocking monitor_cb @(posedge clk);
+    input  req;
+    input  branch;
+    input  branch_addr;
+    input  ready;
+    input  valid;
+    input  rdata;
+    input  addr;
+    input  err;
+    input  err_plus2;
+    input  enable;
+    input  invalidate;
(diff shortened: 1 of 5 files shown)


dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
```diff
@@ -27,41 +27,77 @@ interface ibex_icache_core_if (input clk);
...
   // Drive the branch pin for a single cycle, redirecting the cache to the given instruction
   // address.
   task automatic branch_to(logic [31:0] addr);
-    branch      <= 1'b1;
-    branch_addr <= addr;
+    driver_cb.branch      <= 1'b1;
+    driver_cb.branch_addr <= addr;
 
-    @(posedge clk);
+    ##1;
```
udinator, line 72: Why this delay statement instead of waiting on the clocking block, or something similar?

rswarbrick: This *is* waiting on the clocking block (note the double `#`) and the default clocking statement above means the clocking block on which it's waiting is `driver_cb`. I thought it was neater doing this than writing `@(driver_cb)`, but don't have strong feelings. Do you think the latter would be clearer?

rswarbrick:
Yes, the testbench compiles properly with these changes (indeed, the first two patches are fixing silly bugs that I didn't notice because I wasn't compile-testing what I thought I was - oops!).

Let me know what you think about the delay statements, then I'll update if necessary and merge.

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
udinator: Sorry, I was referring more to the syntax of it, rather than what it meant.
Agreed that both have the same effect; I feel that maybe it should be written as `@(driver_cb)` instead to be a bit more explicit (this syntax is also what we generally tend to use in other DV environments).

rswarbrick: Ok, I'll do that, thanks.

udinator:
It should be good to get merged in after the change!

rswarbrick:
The force-pushes switch `##1;` to `@(driver_cb);` as suggested and get rid of the default clocking, which is no longer needed.

PR #791, opened 2020-04-17
[memutils] Add support for > 32b memories

### LINKED ISSUE #790: Verilator memutil: Support 64 bit wide memories
Currently memutil expects the input file (ELF/BIN) to be 32 bit words, and the memory also to be 32 bit words. We need to support the case where the input file stays 32 bit wide, but the memory is a multiple of that, e.g. 64 bit.

https://github.com/lowRISC/ibex/blob/master/dv/verilator/memutil/cpp/verilator_memutil.cc#L426

https://github.com/lowRISC/ibex/blob/master/shared/rtl/prim_generic_ram_1p.sv#L93

This change is also needed in OT. First, do a PR in Ibex. Then do a PR in OT, updating Ibex through the vendor tool, and the primitive separately.

tomroberts-lowrisc:
Pack > 32bit memory loads into wider RAM instances. Fixes #790

### DIFF
--- a/dv/verilator/memutil/cpp/verilator_memutil.cc
+++ b/dv/verilator/memutil/cpp/verilator_memutil.cc
@@ -31,12 +31,22 @@ extern void simutil_verilator_memload(const char *file);
  *
  * @return 1 if successful, 0 otherwise
  */
-extern int simutil_verilator_set_mem(int index, const svLogicVecVal *val);
+extern int simutil_verilator_set_mem(int index, const svBitVecVal *val);
 }
 
 bool VerilatorMemUtil::RegisterMemoryArea(const std::string name,
                                           const std::string location) {
-  MemArea mem = {.name = name, .location = location};
+  // Default to 32bit width
+  return RegisterMemoryArea(name, location, 32);
+}
+
+bool VerilatorMemUtil::RegisterMemoryArea(const std::string name,
+                                          const std::string location,
+                                          size_t width_bit) {
+  MemArea mem = {.name = name, .location = location, .width_bit = width_bit};
+
+  assert((width_bit == 32 || width_bit == 64) &&
+         "TODO: Check if width other than 32 and 64 works as expected.");
 
   auto ret = mem_register_.emplace(name, mem);
   if (ret.second == false) {
@@ -130,8 +140,9 @@ bool VerilatorMemUtil::ParseCLIArguments(int argc, char **argv,
 void VerilatorMemUtil::PrintMemRegions() const {
   std::cout << "Registered memory regions:" << std::endl;
   for (const auto &m : mem_register_) {
-    std::cout << "\t'" << m.second.name << "' at location: '"
-              << m.second.location << "'" << std::endl;
+    std::cout << "\t'" << m.second.name << "' (" << m.second.width_bit
+              << "bits) at location: '" << m.second.location << "'"
+              << std::endl;
   }
 }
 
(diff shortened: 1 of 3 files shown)


tomroberts-lowrisc:
@imphil - I didn't end up making any changes on the C++ side, but might have misunderstood what was required, let me know if so!

imphil:
Looks like the prim_generic_ram_1p.sv in Ibex already diverged from the one in OT, the 32 bit generate block isn't in OT. But we'll get that sorted next.

Your approach should do the trick for now. However, it changes the original meaning of the `simutil_verilator_set_mem()` function from "load the `mem` block in whatever width (but we haven't implemented anything other than 32 bit width yet)" to "this is a special case function to always load 32 bit words into the memory and pack them neatly next to each other". 

Can you:
- Either retain the original meaning by widening the argument of `simutil_verilator_set_mem()` to whatever `Width` the memory has, and doing the packing on the C++ side, or
- Rename this function (or add another one) along the lines of `simutil_verilator_set_mem_32b()`.

The first approach gives us more flexibility in the future if we want to add other parity or checksum data in a pre-computed form into memories with "odd" width. The second approach is easier and sufficient for now.

tomroberts-lowrisc:
I had a go at fixing it on the C++ side, allowing an extra parameter to be registered for each memory type.
I'm assuming we can pass arbitrary data widths for packed arrays through the DPI interface, but I haven't tested it - @imphil any experience with that? I moved from svLogicVecVal to svBitVecVal since that's just a uin32_t pointer rather than a struct, and I don't think we need the 4-state stuff here.

dv/verilator/memutil/cpp/verilator_memutil.h
```diff
@@ -21,6 +21,7 @@ enum MemImageType {
 struct MemArea {
   std::string name;      // Unique identifier
   std::string location;  // Design scope location
+  int width;             // Memory width
```
imphil, line 24: How about making that size_t as well and remove the cast in `VerilatorMemUtil::MemWrite()`?

dv/verilator/memutil/cpp/verilator_memutil.h
```diff
@@ -41,10 +42,18 @@ class VerilatorMemUtil : public SimCtrlExtension {
...
    * 'simutil_verilator_memload' and 'simutil_verilator_set_mem' used for
    * 'vmem' and 'elf' files, respectively.
+   * The |width| argument specifies the with in bits of the target memory
+   * instance (used for packing data).
    *
    * Memories must be registered before command arguments are parsed by
    * ParseCommandArgs() in order for them to be known.
    */
+  bool RegisterMemoryArea(const std::string name, const std::string location,
+                          const int width);
```
imphil, line 52: suggestion: width -> width_bit

dv/verilator/memutil/cpp/verilator_memutil.cc
```diff
@@ -410,7 +426,8 @@ bool VerilatorMemUtil::MemWrite(const MemArea &m, const std::string &filepath,
 }
 
 bool VerilatorMemUtil::WriteElfToMem(const svScope &scope,
-                                     const std::string &filepath) {
+                                     const std::string &filepath,
+                                     const size_t size_bytes) {
```
imphil, line 430:
```suggestion
                                     const size_t size_byte) {
```

To match `VerilatorSimCtrl::FileSize()`

And I'd remove the `const` for pass-by-value arguments, it doesn't add a lot of meaning, but typing.

shared/rtl/prim_generic_ram_1p.sv
```diff
@@ -71,16 +71,16 @@ module prim_generic_ram_1p #(
       $readmemh(file, mem);
     endtask
 
-    // TODO: Allow 'val' to have other widths than 32 bit
+    // Width must be a multiple of 32bit for this function to work
     // Note that the DPI export and function definition must both be in the same generate
     // context to get the correct name.
-    if (Width == 32) begin : gen_32bit
+    if ((Width % 32) == 0) begin : gen_set_mem
       // Function for setting a specific 32 bit element in |mem|
```
imphil, line 78:
```suggestion
      // Function for setting a specific element in |mem|
```

tomroberts-lowrisc:
All comments addressed - thanks for the review!

PR #792, opened 2020-04-17
[rtl] Fix icache PMP error handling

tomroberts-lowrisc:
- Requests receiving a PMP error need to output a valid indicator, even
  though they will not have received any beats of data

### DIFF
--- a/rtl/ibex_icache.sv
+++ b/rtl/ibex_icache.sv
@@ -600,10 +600,11 @@ module ibex_icache #(
     // data output, and have data available to send.
     // Data is available if:
     // - The request hit in the cache
+    // - The request received an error (since a PMP error might not actually receive any data)
     // - Buffered data is available (fill_rvd_cnt_q is ahead of fill_out_cnt_q)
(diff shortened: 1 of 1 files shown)


rtl/ibex_icache.sv
```diff
@@ -603,7 +603,7 @@ module ibex_icache #(
     // - Buffered data is available (fill_rvd_cnt_q is ahead of fill_out_cnt_q)
     // - Data is available from the bus this cycle (fill_rvd_arb)
     assign fill_out_req[fb]    = fill_busy_q[fb] & ~fill_stale_q[fb] & ~fill_out_done[fb] &
-                                 (fill_hit_ic1[fb] | fill_hit_q[fb] |
+                                 (fill_hit_ic1[fb] | fill_hit_q[fb] | (|fill_err_q[fb]) |
```
rswarbrick, line 606: Can you update the comment above this line? At the moment, it implies that `fill_out_req[fb]` is true if the fill buffer has data that it can send to the core. If there's been an error (especially a PMP error), it doesn't actually have anything useful to give the core except for an error.

PR #793, opened 2020-04-17
Initial memory agent for ICache UVM tests

rswarbrick:
The main patch is the first in the sequence: the second just disables PMP/memory errors to work around a bug in some different code, so we can land this PR independently of that fix.

Here's the commit message for the first patch:

> **Initial memory agent for ICache UVM testbench**
> 
> As with the core agent, this doesn't yet have any proper monitoring,
> and the testbench has no scoreboard, so we're not actually checking
> anything.
> 
> However, it *does* implement a slave agent which can respond to
> instruction fetches from the icache. This runs a test to completion,
> try:
> 
>    `make -C dv/uvm/icache/dv run WAVES=1`
> 
> Note that the timeout in the hjson file was too small, so this bumps
> it to a larger value. A later patch will replace the timeout with a
> heartbeat monitor, which should be much cleaner.
>
> The exact dance for the UVM reactive slave is complicated. See
> [dv/uvm/icache/dv/ibex_icache_mem_agent/README.md](https://github.com/rswarbrick/ibex/blob/mem-agent/dv/uvm/icache/dv/ibex_icache_mem_agent/README.md) for the details.


dv/uvm/icache/dv/env/seq_lib/ibex_icache_common_vseq.sv
```diff
@@ -1,18 +0,0 @@
-// Copyright lowRISC contributors.
-// Licensed under the Apache License, Version 2.0, see LICENSE for details.
-// SPDX-License-Identifier: Apache-2.0
-
-class ibex_icache_common_vseq extends ibex_icache_base_vseq;
```
udinator, line 5: Will this common vseq not be used?

dv/uvm/icache/dv/ibex_icache_mem_agent/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The basic idea is that instruction memory is modelled with a single 32-bit *seed*.
+A simple hash function takes this seed and calculates 32 bits of memory for each (aligned) base address.
+The seed also determines a memory region which should result in PMP errors and another memory region that should result in memory errors.
+
+The agent exposes a slave interface to the core using the usual UVM architecture for a reactive slave, but it looks a little complicated because we need to spot PMP errors immediately (rather than on the next clock edge).
+The entire dance is as follows:
+
+  1. The monitor spots a new request (either because of a posedge on the `req` line or a change in the `addr` line).
+  1. It generates an `ibex_icache_mem_req_item` with `is_grant` field false and writes it to the sequencer's request port.
```
udinator, line 14: Are these all meant to be `1.`?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
```diff
@@ -0,0 +1,133 @@
...
+    join
+  endtask
+
+  // Drive the grant line. This toggles independently of any other signal on the bus.
+  task automatic drive_grant();
+    int gnt_delay;
+
+    forever begin
+      // Pick a number of cycles for the grant line to be low
+      if (!std::randomize(gnt_delay) with {
```
udinator, line 51: Use the `DV_CHECK_...` randomization macros found in `vendor/lowrisc_ip/dv_utils/dv_macros.svh` for any randomization like this.

dv/uvm/icache/dv/env/ibex_icache_env.sv
```diff
@@ -2,39 +2,40 @@
...
+    .COV_T              (ibex_icache_env_cov           #(BusWidth)),
+    .VIRTUAL_SEQUENCER_T(ibex_icache_virtual_sequencer #(BusWidth)),
+    .SCOREBOARD_T       (ibex_icache_scoreboard        #(BusWidth))
   );
   `uvm_component_utils(ibex_icache_env)
 
-  ibex_icache_core_agent    m_ibex_icache_core_agent;
-  ibex_mem_intf_slave_agent m_ibex_mem_intf_slave_agent;
+  ibex_icache_core_agent            core_agent;
+  ibex_icache_mem_agent #(BusWidth) mem_agent;
```
sriyerg, line 14: If it is known that this version of icache env will only cater to a 32-bit mem agent, then I suggest fixing the parameter value here than propagating it everywhere in the env code.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
```diff
@@ -0,0 +1,133 @@
...
+  // Drive the PMP line.
+  //
+  // This task gets spawned by take_requests each time a new address comes in. The driver should
+  // have at most one instance running at once.
+  //
+  // If err is false, there is nothing to do. If err is true, it sets the PMP error flag and then
+  // waits until either req is de-asserted or the address changes, at which point it clears the flag
+  // and exits.
+  task static drive_pmp(bit err);
+    if (pmp_driven) begin
```
udinator, line 115: What is `pmp_driven` used for?

dv/uvm/icache/dv/ibex_icache_mem_agent/README.md
sjgitty: > Are these all meant to be `1.`?

@udinator .md does the right thing here, and is the easiest
way to maintain a numbered list so you can add / remove
easily.

udinator: Ah okay, thanks for letting me know; I'll resolve this conversation.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_if.sv
```diff
@@ -0,0 +1,76 @@
...
+    repeat (num_clks) @driver_cb;
+  endtask
+
+  // Drive a response with the given rdata and possible error signals for a single cycle
+  task automatic send_response(logic rsp_err, logic [BusWidth-1:0] rsp_rdata);
+    driver_cb.rvalid <= 1'b1;
+    driver_cb.err    <= rsp_err;
+    driver_cb.rdata  <= rsp_rdata;
+
+    @driver_cb;
```
udinator, line 69: nit: `@(driver_cb)` to stay consistent with the formatting in the rest of these files

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
```diff
@@ -0,0 +1,125 @@
...
+// address causing an error. We should probably do something more sensible, but for now let's XOR
+// the address with something else (32'hf00dbeef) and pass it to the PMP calculation.
+//
+// For the memory content hash function, we use the first example from [1] (explicitly
+// public-domain), except that we have 30+32 = 62 bits of input, which we just XOR together
+// (aligning LSBs).
+//
+// [1] https://burtleburtle.net/bob/hash/integer.html
+
+class ibex_icache_mem_model #(parameter int unsigned BusWidth = 32);
```
udinator, line 24: Could you have this class extend from `uvm_object` and then instantiate with `::create()`?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_monitor.sv
```diff
@@ -0,0 +1,93 @@
...
+  // This must collect the "I have a new address request" immediately, rather than at the next clock
+  // edge, to make sure that a PMP error can be signalled in the same cycle.
+  //
+  // Note that it's possible the underlying memory seed will change after the request has been seen.
+  // We don't try to spot this, and instead report the error state based on the seed when the
+  // request appeared.
+  task automatic collect_requests();
+    forever begin
+      // Wait for either a posedge on req or any change to the address
+      @(posedge cfg.vif.req or cfg.vif.addr);
```
udinator, line 51: Does the change in `cfg.vif.addr` have to be qualified with a check that `cfg.vif.req` is high e.g. can the address change when `instr_req_o` is low?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_agent.sv
```diff
@@ -0,0 +1,32 @@
...
+    if (!uvm_config_db#(virtual ibex_icache_mem_if)::get(this, "", "vif", cfg.vif)) begin
+      `uvm_fatal(`gfn, "failed to get ibex_icache_mem_if handle from uvm_config_db")
+    end
+  endfunction
+
+  function void connect_phase(uvm_phase phase);
+    super.connect_phase(phase);
+    if (cfg.is_active) begin
+      // Pass snooped requests from the monitor to the sequencer
+      monitor.request_port.connect(sequencer.request_export);
```
udinator, line 28: You can actually directly do this:
```systemverilog
monitor.request_port.connect(sequencer.request_fifo.analysis_export);
```
with no need for the additional `uvm_analysis_export` handle in the sequencer.

dv/uvm/icache/dv/env/ibex_icache_env.sv
rswarbrick: Well, the ICache is can be parameterised by BusWidth and I think it's likely we'll build systems with different width instruction buses in the future (e.g. what if this goes straight to SDRAM? the bus could well be 64- or 128-bit). Since it's not difficult to plumb the parameter through, I just did it.

The virtual sequence has the parameter hardcoded (as has `tb.sv`) because I couldn't figure out how to hoist stuff all the way to top-level. I thought this was probably a reasonable half-way house: if we need to get things working for other widths in the future, there might be some fiddling, but we shouldn't need a very intrusive patch.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_agent.sv
rswarbrick: Ah, good point. Will fix this this morning.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
rswarbrick: Gah, I thought I'd caught all of these. Thanks!

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
rswarbrick: It's a "semaphore" to check that `drive_pmp` is not used re-entrantly (as the error message below says). This would be a programming bug in the monitor, but it would probably be annoying to debug if it happened (there would be more than one thing driving the pmp_err line). So this checks it doesn't happen.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_if.sv
rswarbrick: Ah, thanks. Hmm, I guess I must have got this from a tutorial: I didn't even realise it was valid syntax! Will change.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
rswarbrick: Can you explain the benefit of doing so? I don't believe this code needs any of the UVM machinery.

dv/uvm/icache/dv/env/seq_lib/ibex_icache_common_vseq.sv
rswarbrick: I'm not sure. When would you expect to use it?

udinator: Actually you're correct that this is not necessarily needed, this common vseq is generally used to start various common tests, including interrupt test sequences, TileLink error sequences, CSR tests and so on, none of which particularly apply to the Icache (but generally apply to other IPs extending from the cip_lib library).

Something that Sri and I had been considering briefly is to add a TileLink RAL adapter to allow integration with the existing memory test suite used by other IPs within OpenTitan, @sriyerg if we were to do this would the common_vseq class be needed?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
udinator: Ah got it. Would you consider using the built in semaphore class to perhaps make this behavior more explicit?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
udinator: A main benefit would be the ability to use UVM reporting macros (`uvm_info`, etc...)

udinator:
Thanks for the fixes @rswarbrick! I think in general this agent looks good, considering the extra complexity of immediate responses to PMP errors.
Let's wait for Sri to take a look at the code, in the meantime I'll give this another read-through to make sure that everything looks okay.

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_base_seq.sv
```diff
@@ -0,0 +1,19 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class ibex_icache_mem_base_seq #(parameter int unsigned BusWidth = 32)
```
sriyerg, line 5: IMO, the only 2 places where the parameter value should be set is the testbench and the test class. It is generally not a good practice to combine generic / template classes with actual implementation classes in the same package. The environment package is seen as a vertically/horizontally reusable entity. This implementation makes reuse counterintuitive and hard. 

Either:
1. Go with my previous suggestion of making the width runtime and remove the parameterizations (easier and flexible)
2. Move 'actual' implementation sequences (that set the width) to a separate package. I think it would be better to also indicate the width in the class name, since it won't work for any other widths; perhaps `ibex_32b_icache_mem_base_seq`? (easiest but inflexible - need additional possibly duplicative work to support width != 32)
3. Make all classes including sequences parameterized, customize the `uvm_object_param_utils` invocation below to support it (slightly difficult, but flexible and most ideal)

I can help with option 3 if you decide to take it. I have never done it so it maybe a bit of trial and error.

sriyerg: I am fine with making this change in a later PR to keep it simple.

dv/uvm/icache/dv/env/seq_lib/ibex_icache_common_vseq.sv
sriyerg: Is the mem within icache addressable like a traditional SRAM (example, as a TCM)? If not, then the automated mem tests won't work, right?

udinator: @rswarbrick - I think you are in a better position to answer this

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
udinator: Whoops - I didn't refresh this PR again before adding in my comments so didn't see Sri's comments until just now, sorry for the repetition of points here and below!

dv/uvm/icache/dv/env/seq_lib/ibex_icache_common_vseq.sv
rswarbrick: The memory within the icache is not addressable (well, except by internal logic in the cache, but it's not exposed on any ports).

dv/uvm/icache/dv/env/ibex_icache_env.sv
rswarbrick: Ok, understood. I'll rip this out and make it 32-bit only. If we need a different bus width in the future, I agree that it won't be difficult to retro-fit.

dv/uvm/icache/dv/env/seq_lib/ibex_icache_sanity_vseq.sv
```diff
@@ -4,19 +4,38 @@
...
-  `uvm_object_utils(ibex_icache_sanity_vseq)
+class ibex_icache_sanity_vseq extends ibex_icache_base_vseq #(32);
 
+  `uvm_object_utils(ibex_icache_sanity_vseq)
   `uvm_object_new
 
-  // A sanity sequence for the core agent
+  // A sanity sequence for the core agent and a basic slave sequence for the memory agent
   ibex_icache_core_sanity_seq core_seq;
+  ibex_icache_mem_resp_seq    mem_seq;
```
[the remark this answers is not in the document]
rswarbrick: Good catch, thanks. Since I'm getting rid of the parameter, I'll claim this was foresight rather than an oversight :-)

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
rswarbrick: Ok, I can do that I guess. Seems a little overkill, but can't hurt.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
rswarbrick: Ok, I'll do this.

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_base_seq.sv
rswarbrick: As above, I'm going to rip out the parametrisation, so I guess this doesn't matter. But I don't completely understand your comment (and I want to!).

This base sequence doesn't require a concrete width (I made the default value for the parameter 32, but only because you have to supply *something*). I think that the only places where I specified the width to be 32 are the testbench and the virtual sequence: everything else should have picked up the parameter from there. Have I misunderstood what you're getting at?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
rswarbrick: Oh, wait a sec. I've just looked again and I don't think I want a semaphore here (maybe I was using the wrong terminology).

A semaphore would let a re-entrant `drive_pmp` call wait until the previous one was done, but that's not what I want. I want the simulation to explode in a ball of flames, saying "Congratulations, you have found a programming error in the monitor code. Go and shout at Rupert".

I believe that the current code will do that. I'll add a comment above the check making it clear what's going on. Let me know if this isn't ok.

rswarbrick:
This latest version mostly removes the `BusWidth` parameterisation. I've left it in the memory model code, because I think the logic in `read_data` is correct and it was a little tricky to get right with different widths, so seemed silly to throw away. Otherwise, the parameters were basically plumbing, and they've all been stripped out.

I've also added a comment explaining how the re-entrancy check in `drive_pmp` works.

rswarbrick:
Oops, and I forgot to inherit from `uvm_object` in the memory model as Udi asked. This latest version does that too.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
```diff
@@ -0,0 +1,132 @@
...
+
+class ibex_icache_mem_model #(parameter int unsigned BusWidth = 32)
+  extends uvm_object;
+
+  protected bit [31:0] seed;
+  protected bit        no_pmp_errs = 0;
+  protected bit        no_mem_errs = 0;
+
+  function new(string name="", bit disable_pmp_errs=0, bit disable_mem_errs=0);
+    seed = 32'd0;
```
udinator, line 32: nit: is this explicit assignment to 0 needed?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_driver.sv
udinator: Ah understood, in that case this looks fine to me.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_req_item.sv
```diff
@@ -0,0 +1,37 @@
...
+
+class ibex_icache_mem_req_item extends uvm_sequence_item;
+
+  bit               is_grant;
+  logic [31:0]      address;
+
+  rand bit [31:0]   seed;
+
+  // Change the memory seed roughly one time in 500 reads
+  constraint c_seed_dist {
```
udinator, line 23: Just to clarify, will this distribution cause the seed to be changed too infrequently?

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_model.sv
rswarbrick: Huh, no: I should have initialised it on the line it was declared. Will fix.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_req_item.sv
rswarbrick: I don't think so: we should be able to see bugs with only quite occasional seed changes, since each change gives lots of tests (each time you read before and after the change). If we have seen lots of seeds since the last invalidation, the scoreboard will run rather slower because it has to check lots of possible answers. I'm imagining that a test might do 10k fetches in maybe 50k cycles (because of all the stalling we add to the memory and core), so this would be roughly 20 seeds, which seems reasonably sane.

Obviously, we can come back to this if it doesn't look like it's changing enough.

rswarbrick:
Most recent version rebases onto master and assigns the `seed` field in the memory model to zero in a different place.

rswarbrick:
Most recent version strips out a couple of stray parameters when instantiating things. If you make that mistake, VCS just leaves a warning deep in a log file and ignores the parameter, so I only spotted the problem when debugging a build failure with later code. Aargh!

dv/uvm/icache/dv/env/seq_lib/ibex_icache_common_vseq.sv
sriyerg: In that case this sequence is not necessary.

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_base_seq.sv
sriyerg: I might have put this comment at the wrong place. I was probably talking about this:
```systemverilog
class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq #(.BusWidth(32));
```

rswarbrick: Ah, right, that makes sense. Thanks.

dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_req_item.sv
udinator: Sounds good to me, just wanted to make sure that the test would do enough memory reads to make this distribution viable.

PR #796, opened 2020-04-17
[dv] Fix Xcelium compilation

weicaiyang:
Need these fixes to pass Xcelium compile in OpenTitan

Signed-off-by: Weicai Yang <weicai@google.com>

### DIFF
--- a/shared/rtl/bus.sv
+++ b/shared/rtl/bus.sv
@@ -15,8 +15,8 @@
  * - Host (master) arbitration is strictly priority based.
  */
 module bus #(
-  parameter NrDevices,
-  parameter NrHosts,
+  parameter NrDevices    = 1,
+  parameter NrHosts      = 1,
   parameter DataWidth    = 32,
   parameter AddressWidth = 32
 ) (
--- a/shared/rtl/ram_1p.sv
+++ b/shared/rtl/ram_1p.sv
@@ -5,6 +5,9 @@
 /**
  * Single-port RAM with 1 cycle read/write delay, 32 bit words
  */
+
+`include "prim_assert.sv"
+
 module ram_1p #(
     parameter int Depth = 128
 ) (
@@ -75,7 +78,7 @@ module ram_1p #(
   `endif
 
   `ifdef SRAM_INIT_FILE
-    localparam MEM_FILE = `"`SRAM_INIT_FILE`";
+    localparam MEM_FILE = `PRIM_STRINGIFY(`SRAM_INIT_FILE);
     initial begin
       $display("Initializing SRAM from %s", MEM_FILE);
       $readmemh(MEM_FILE, mem);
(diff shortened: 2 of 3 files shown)


weicaiyang:
@imphil @GregAC can you take a look at this PR?

imphil:
How are these files even used in OpenTitan? They are only relevant for the example systems we have in the Ibex repository. Looks like this PR shouldn't be needed, and the fix is actually somewhere else.

Side: Either your git email address is wrong, of you manually edited it in the sign-off line (don't do that, use `git commit -s`): "Weicai <weicai@google.ocm>"

On the change itself: if you use PRIM_STRINGIFY, you need to include prim_assert.sv into these files and make sure appropriate dependencies exist in the relevant core files.

weicaiyang:
@imphil 
> How are these files even used in OpenTitan? They are only relevant for the example systems we have in the Ibex repository. Looks like this PR shouldn't be needed, and the fix is actually somewhere else.

there are duplicated prim in from OT and Ibex in our OT simulation: https://github.com/lowRISC/opentitan/issues/1231
This fix will also make ibex easy to run with the other simulator

> Side: Either your git email address is wrong, of you manually edited it in the sign-off line (don't do that, use `git commit -s`): "Weicai [weicai@google.ocm](mailto:weicai@google.ocm)"
There is a typo in my git email address. Fixed. Thanks

> On the change itself: if you use PRIM_STRINGIFY, you need to include prim_assert.sv into these files and make sure appropriate dependencies exist in the relevant core files.

I already have the dependency in core file.
```
    depend:
      - lowrisc:prim:assert
    files:
      - ./rtl/prim_clock_gating.sv
      - ./rtl/prim_generic_ram_1p.sv
      - ./rtl/prim_secded_28_22_enc.sv
      - ./rtl/prim_secded_28_22_dec.sv
      - ./rtl/prim_secded_72_64_enc.sv
      - ./rtl/prim_secded_72_64_dec.sv
      - ./rtl/ram_1p.sv
      - ./rtl/ram_2p.sv
```

imphil:
For OpenTitan: The ram_1p and ram_2p primitives are named differently to the ones in OpenTitan (where they're called *prim_*ram_1p), so no code in OpenTitan should pick up these files, and even less include them (and I don't see any code including them there either).

For this PR: Can you add an `include "prim_assert.sv" to the files where you use STRINGIFY? With that this PR should be good to go.

weicaiyang:
> For OpenTitan: The ram_1p and ram_2p primitives are named differently to the ones in OpenTitan (where they're called *prim_*ram_1p), so no code in OpenTitan should pick up these files, and even less include them (and I don't see any code including them there either).

My bad. duplicated name issue isn't related to these primitives. These are just included in the top-level simulation when ibex is involved. And Xcelium doesn't like the syntax.

> For this PR: Can you add an `include "prim_assert.sv" to the files where you use STRINGIFY? With that this PR should be good to go.

Added. PTAL. Thanks

weicaiyang:
@imphil is this ready to land?

imphil:
Sorry @weicaiyang, I lost track of this one. Merged.

weicaiyang:
Thanks @imphil

PR #800, opened 2020-04-21
Use Shared Intermediate Value Register with Multdiv Slow

ganoam:
This commit adds support for the shared immediate value register in the
id_stage for the slow implementation of the multdiv module.

Register accum_window_q is now stored in the intermediate value
register.

### DIFF
--- a/rtl/ibex_multdiv_slow.sv
+++ b/rtl/ibex_multdiv_slow.sv
@@ -28,6 +28,10 @@ module ibex_multdiv_slow
     output logic [32:0]      alu_operand_a_o,
     output logic [32:0]      alu_operand_b_o,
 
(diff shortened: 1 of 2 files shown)


rtl/ibex_multdiv_slow.sv
```diff
@@ -28,6 +28,10 @@ module ibex_multdiv_slow
     output logic [32:0]      alu_operand_a_o,
     output logic [32:0]      alu_operand_b_o,
 
+    input  logic [33:0]     imd_val_q_i,
```
vogelpi, line 31: Nit: alignment

ganoam: Very good. Done, thanks!

PR #801, opened 2020-04-21
CI: Use packaged version of Verilator

imphil:
Reduce CI times by using a packaged version of Verilator. The packaging
is done by me on Open Build Service, which provides Verilator packages
for various distributions. The packages are versioned to help choose the
exact version in CI.

This is experimental for now, let's see if that's a suitable path
forward in the longer term.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,8 +6,7 @@
 # Documentation at https://aka.ms/yaml
 
 variables:
-  VERILATOR_VERSION: 4.028
-  VERILATOR_PATH: /opt/buildcache/verilator/$(VERILATOR_VERSION)
+  VERILATOR_VERSION: 4.032
   RISCV_TOOLCHAIN_TAR_VERSION: 20200323-1
   RISCV_COMPLIANCE_GIT_VERSION: 844c6660ef3f0d9b96957991109dfd80cc4938e2
   VERIBLE_VERSION: v0.0-266-g9e55307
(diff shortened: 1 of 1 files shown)


imphil:
OK, going from roughly 9 min to 3:30 min CI latency sounds worth it :smile:

imphil:
> Think we can get this into OpenTitan CI too?

That was actually the main motivation for all of this packaging :smile:

asb:
Minor concern: it's not very good hygiene to change both the packaging and the verilator version at the same time.

tomroberts-lowrisc:
Sounds great! Is it possible to host the packages on a lowRISC server?

imphil:
Sorry Alex and Tom, things were overlapping.

- Alex: I agree, however I didn't want to build the old version as well, and the update seemed safe. So saving myself and the builders a bit of time here for a shortcut.
- Tom: We could (in various ways), but it's not giving us much other than cleaner download URLs (which would certainly be nice). We could go in various directions, e.g. hosting a own OBS instance, mirroring repositories, etc. My hope was to give that effort a try on "mini scale" for now and see how much work is involved in maintaining those packages before we announce it more widely, when questions like hosting will come up again.

imphil:
Thinking about @asb's comment again, there obviously is a way to avoid combining the Verilator update and the binary change in the same PR by just getting the order of updates right. Done that for OT in https://github.com/lowRISC/opentitan/pull/2016.

PR #803, opened 2020-04-21
The ibex_tracer rewritten to be synthesizable

jacekmw8:
I have rewritten the ibex_tracer module to be synthesizable and produce the trace when running on emulators. 
The trace format has been adapted to be used by and offline debug tools, I got rid of tabs, cycle number in favor of $time (some tools still generate cycles as a result of $time).
The trace has been split in 3 different groups:
- instruction trace with disassembly closely similar to objdump format
- register trace, logging register value changes
- memory trace, logging memory reads and writes
All these can be enabled individually though +arguments, which allows user to trace only the changes he is interested in.

The synthesis requirements implied eliminating all dynamic data structures, all string types and $sformats had to go. They have been replaced by bit vectors and their concatenations. It is not as pretty as $sformat, as register names and values have to use the same, constant width, but it's clear and well readable (with a fixed width font)
        310 ns PC=0x1a0003d4,     0x35c9, jal     0x1a000296             
        310 ns REG= x1, 0x1a0003d6
        350 ns PC=0x1a000296, 0x1a104737, lui     x14,0x1a104            
        350 ns REG=x14, 0x1a104000
        410 ns PC=0x1a00029a,     0x5b7c, lw      x15,0x74(x14)          
        410 ns MR=0x1a104074, 0x00000000
        410 ns REG=x15, 0x00000000
I have made an effort to rewrite the tracer with compliance to the SV guidelines,

As a requirement to generate a trace on the hardware, the trace log needs to be opened before the run, as it includes the hartid it needed to be passed as a parameter to the ibex_core and ibex_tracer, not as an input signal. This is the only visible change at the core level (which goes down to crs registers). But I think it is actually beneficial overall, as the hartid is not supposed to change.

There is few other changes, 
- added optional CSRs, as per debug specifications, like TINFO
This register is read out by the Lauterbach Trace32 debugger during the connection. 

I have integrated the latest ibex with the Pulpissimo platform:
https://github.com/pulp-platform/pulpissimo
and have it running with Verilator, Cadence Xcelium, Palladium and Protium emulators with Trace32 connected through a virtual debug interface. 

I hope these changes are useful for the community and can be integrated in the project


rswarbrick:
Hi there,

Thank you very much for your proposed contribution. I don't think we can take this code in its current form, because it's quite an invasive change and actually drops some information we use (such as cycle counts).

If I understand correctly, your use-case is that you need to get an execution trace when running on an FPGA-based emulator. Maybe it would be easier to write a simple wrapper around the core (to use instead of `core_tracer`), which just dumped the various `rvfi` signals. Then you could post-process these "off-chip" to get the trace you need.

For identical results to the existing flow, I suspect you could then write a testbench module that pretended to be the core and loaded `rvfi` signals from the dump. If you attached that testbench to an `ibex_tracer` module and ran the result with a simulator like Verilator, I think it would spit out the right thing.

Alternatively, you might well find that a simpler post-processing step was enough (maybe you even just need an address trace, which is immediately visible in the `rvfi` signals).

PR #804, opened 2020-04-21
[dv] fix spurious LSF enabling

udinator:
This fixes a small but annoying bug - when `run_sim_commands()` is called, the `lsf_cmd` argument that it takes in is always set to `True` by default even when we don't want to run on LSF, as the default value of `lsf_cmd` passed in from the top level Makefile is always `''`, which when compared to `None` will return `True`, causing seemingly random regression timeouts.
This fixes this issue by defaulting `args.lsf_cmd` to an empty string and comparing the `lsf_cmd` argument value in the `run_sim_commands` call to an empty string instead of `None`.

### DIFF
--- a/dv/uvm/core_ibex/sim.py
+++ b/dv/uvm/core_ibex/sim.py
@@ -499,6 +499,13 @@ def main():
     setup_logging(args.verbose)
     parser.set_defaults(verbose=False)
 
+    # If args.lsf_cmd is an empty string return an error message and exit from
+    # the script, as doing nothing will result in arbitrary simulation timeouts
+    # and errors later on in the run flow.
+    if args.lsf_cmd == "":
+        logging.error("The LSF command passed in is an empty string.")
+        return RET_FAIL
+
     # Create the output directory
     output_dir = ("%s/rtl_sim" % args.o)
     bin_dir = ("%s/instr_gen/asm_tests" % args.o)
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -325,10 +325,10 @@ $(OUT)/rtl_sim/.compile.stamp: \
 	@./sim.py \
 	 --o=${OUT} \
 	 --steps=compile \
-	 --lsf_cmd="${LSF_CMD}" \
 	 ${COMMON_OPTS} \
 	 --simulator="${SIMULATOR}" --simulator_yaml=yaml/rtl_simulation.yaml \
-	 $(cov-arg) $(wave-arg) --cmp_opts="${COMPILE_OPTS}"
+	 $(cov-arg) $(wave-arg) $(lsf-arg) \
+	 --cmp_opts="${COMPILE_OPTS}"
 	$(call dump-vars,$(OUT)/rtl_sim/.compile-vars.mk,comp,$(compile-var-deps))
 	@touch $@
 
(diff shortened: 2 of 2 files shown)


udinator:
1. Sure, that approach sounds good as well, I'll just add a check after parsing and set `lsf_cmd` to `None` if it is an empty string, that way the existing check in `rtl_sim` can stay unchanged.
2. Whoops(!), I'm not sure how I didn't see that `$(lsf-arg)` was already present in the compile and rtl_sim targets...if memory serves me correctly I added in the hardcoded `--lsf_cmd` argument in both due to the command seemingly not being passed in correctly (if at all) when I tried to run through LSF. Let me take a look tomorrow at this to clear it up.
3. Yup, sgtm.

rswarbrick:
Oh, I've just re-read what I wrote. For (1), I meant that if you call `sim.py --lsf_cmd=""`, you should see an error message. I think that specifying `--lsf_cmd=foo` means "Please run things using LSF. To do so, prefix each command with `foo `". Thus `--lsf_cmd=""` has a meaning, it's just always going to be the wrong thing to do! I don't believe that silently dropping the argument is the right thing to do in this case, since that makes `sim.py` less predictable in its behaviour.

udinator:
Ah, understood, that's probably a cleaner way of doing things.

udinator:
The update contains an equality check against an empty string for the `lsf_cmd` argument, we exit the sim flow if this check is true, and adds the `$(lsf-arg)` variable to the `compile` target in the Makefile and removes the explicit `--lsf_cmd="${LSF_CMD}"` argument in both the `compile` and `rtl_sim` targets.

dv/uvm/core_ibex/sim.py
```diff
@@ -499,6 +499,14 @@ def main():
     setup_logging(args.verbose)
     parser.set_defaults(verbose=False)
 
+    # If args.lsf_cmd is an empty string return an error message and exit from
+    # the script, as doing nothing will result in arbitrary simulation timeouts
+    # and errors later on in the run flow.
+    if args.lsf_cmd == "":
+      logging.error("""The LSF command passed in is an empty string.
+                       Please pass in a valid LSF command.""")
```
rswarbrick, line 507: If you run this, you'll see lots of spaces in the error message before "Please". Because Python concatenates adjacent string literals like C, the normal Python idiom is something like this:

``` python
        logging.error('The LSF command passed in is an empty string.\n'
                      'Please pass in a valid LSF command.')
```

However, I don't think you want two lines (it makes the error message look weird). I'd just delete the stuff after the period. (And I strongly believe that software shouldn't plead with the user! :-))

dv/uvm/core_ibex/sim.py
```diff
@@ -499,6 +499,14 @@ def main():
...
+      sys.exit(RET_FAIL)
```
rswarbrick, line 508: This is in the main function, whose return value is passed to `sys.exit` already. Just use `return`.

rswarbrick: Oh, sorry. Also, you used a 2-space indent here, rather than the 4-indent in the rest of the file (and specified by PEP8). Can you fix that too?

dv/uvm/core_ibex/sim.py
udinator: Done.

dv/uvm/core_ibex/sim.py
udinator: whoops, you're absolutely right! fixed.

Issue #703, opened 2020-03-13
[DV] Prerequisites for bitmanip DV


imphil:
We are making progress on implementing the bitmanip (B) extension in Ibex in #89. This issue tracks things we need to have before we can start DV work on it.

- [x] A compiler toolchain with bitmanip support in its assembler (at least). Tracking in https://github.com/lowRISC/lowrisc-toolchains/issues/8, https://github.com/lowRISC/ibex/pull/724
- [x] Bitmanip support in spike and ovpsim (that exists AFAIK, we might need to adjust the build instructions)
- [x] Bitmanip support in riscv-dv: https://github.com/google/riscv-dv/issues/520

imphil:
@taoliug @udinator what are the plans for bitmanip in riscv-dv?

udinator:
In riscv-dv, we've had a PR (https://github.com/google/riscv-dv/pull/498) opened by an external contributor to add most of the support for B-extension instructions. @weicaiyang is going to help with merging this PR into the codebase soon, and then I will take over B-extension verification for Ibex from there.

imphil:
An updated compiler toolchain is available in CI after https://github.com/lowRISC/ibex/pull/724 is merged. If you want to manually get one, download a release 20200323-1 or newer from https://github.com/lowRISC/lowrisc-toolchains/releases.

vogelpi:
FYI: I have tested the Bitmanip support in Spike and RISCV-DV. The linked PR enables it and adds a tests for Ibex.

What is missing is support for functional coverage for the B instructions. But people are working on that upstream (RISCV-DV).

vogelpi:
Spike/OVPSim support for bitmanip can be enabled and we documented how to do this. RISCV-DV has support for bitmanip support.

I think we are done here, WDYT @imphil ?

PR #809, opened 2020-04-23
[dv] Enable verification of the Bitmanip Extension with OVPsim and Spike

vogelpi:
This PR enables verification of the Bitmanip Extension with OVPsim and Spike and adds a new test that can be updated in the future to also include newly added instruction groups. For Spike,
a different branch has to be used. This PR updates the documentation accordingly.

What is still missing is support for coverage. This will be later vendored in from upstream RISCV-DV.

This is related to lowRISC/ibex#703.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
+++ b/dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
@@ -1,5 +1,6 @@
 # riscOVPsim configuration file converted from YAML
 --variant RV32IMC
+--override riscvOVPsim/cpu/add_Extensions=B
 --override riscvOVPsim/cpu/misa_MXL=1
 --override riscvOVPsim/cpu/misa_MXL_mask=0x0 # 0
 --override riscvOVPsim/cpu/misa_Extensions_mask=0x0 # 0
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -587,3 +587,13 @@
     +pmp_max_offset=00021000
     +boot_mode=u
   rtl_test: core_ibex_base_test
+
+- test: riscv_bitmanip_test
+  desc: >
+    Random instruction test with supported B extension instructions
+  iterations: 10
+  gen_test: riscv_rand_instr_test
+  gen_opts: >
+    +enable_b_extension=1
+    +enable_bitmanip_groups=zbb,zbt,zbs
+  rtl_test: core_ibex_base_test
--- a/doc/verification.rst
+++ b/doc/verification.rst
@@ -95,12 +95,13 @@ In order to run the co-simulation flow, you'll need:
   ``--enable-misaligned`` tells Spike to simulate a core that
   handles misaligned accesses in hardware (rather than jumping to a
   trap handler).
+  In addition, Spike does not support the `RISC-V Bit Manipulation Extension <bitmanip_>`_  (Bitmanip) by default.
+  To support this draft extension implemented in Ibex, the `riscv-bitmanip branch <Spike_>`_ of Spike needs to be used.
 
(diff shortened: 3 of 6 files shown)


vogelpi:
I guess I will also have to update CI (to use the Bitmanip-enabled version of Spike) + I would need someone to actually check OVPsim.

vogelpi:
Pinging also @weicaiyang as he is working on the coverage.

vogelpi:
Ah yes, we don't the UVM tests in CI here. I mixed that up. I also thought the instructions could be cleaned up.

The Zbb and Zbt tests passed, but I just ran a couple of iterations. Let me check if also the other tests are still passing.

udinator:
Thanks for doing this Pirmin! I'll try this out with OVPsim on my end today to check that it works.

weicaiyang:
Thanks Pirmin. Coverage will be completed soon.

vogelpi:
@ganoam noted that only the first option `enable_bitmanip_groups` in `testlist.yaml` is taken into account. To enable multiple groups, they have to be provided as a comma separated list:
```
+enable_bitmanip_groups=zbb,zbt
``` 
I have now adapted the testlist accordingly. We get some failures in Zbt now. I suggest to debug these once this PR is in.

ganoam:
Oh. thanks. I am already on it.

vogelpi:
As for the other tests, we get failures in:
- riscv_mmu_stress_test
- riscv_debug_branch_jump_test

udinator:
> As for the other tests, we get failures in:
> 
> * riscv_mmu_stress_test

There are two underlying issues causing failures here, one was an RTL bug that Tom has pushed in a fix for, the second is a more subtle but annoying problem that I have a large patch in riscv-dv to fix, waiting for a review on that end.

> * riscv_debug_branch_jump_test

This is an issue with the debug mode sub-program generation in riscv-dv as well that I need to root cause.

dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
```diff
@@ -1,5 +1,5 @@
 # riscOVPsim configuration file converted from YAML
---variant RV32IMC
+--variant RV32IMCB
```
udinator, line 2:
```suggestion
--variant RV32I
--override riscvOVPsim/cpu/add_Extensions=MCB
```

Structuring the supported extensions in this way is required for OVPsim to recognize bitmanip instructions and successfully run, otherwise it will still fail.

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -587,3 +587,13 @@
     +pmp_max_offset=00021000
     +boot_mode=u
   rtl_test: core_ibex_base_test
+
+- test: riscv_bitmanip_test
+  desc: >
+    Random instruction test with supported B extension instructions
+  iterations: 1
```
udinator, line 594: nit: can you change this to 10 iterations?

vogelpi: Yep, done.

dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
vogelpi: :+1: thanks for the suggestion

vogelpi:
Thanks @udinator for checking out OVPsim, providing the required fix, and for clarifying the situation around current test failures.

I have implemented your suggestions, rebased on current master and enabled verification of Zbs for which the RTL got added in the meantime.

udinator:
@vogelpi - LGTM, approved!
Can you run one iteration of the new `riscv_bitmanip_test` with the new ZBS extension enabled just to make sure the flow works smoothly before you merge?

vogelpi:
Yes, it is working. Thanks for your approval @udinator !

PR #810, opened 2020-04-23
Add heartbeat tracking to ICache testbench

rswarbrick:
This pretty much follows various UVM tutorials, except that we use the
sequencers (ibex_icache_core_sequencer, ibex_icache_mem_sequencer) to
poke the heartbeat objection object if there is one.

Doing this means that we can dispense with the explicit test timeout
in the hjson file: the test will fail if there is no activity on
either interface for 2000 cycles (this number is chosen because the
longest "wait a bit" time is 1200 cycles in the core driver).

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_sequencer.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_sequencer.sv
@@ -2,24 +2,13 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// A sequencer class for the icache memory agent.
-
-class ibex_icache_mem_sequencer
-  extends dv_base_sequencer #(.ITEM_T (ibex_icache_mem_resp_item),
-                              .CFG_T  (ibex_icache_mem_agent_cfg));
-
-  `uvm_component_utils(ibex_icache_mem_sequencer)
+class ibex_icache_core_sequencer extends dv_base_sequencer #(.ITEM_T (ibex_icache_core_item),
+                                                             .CFG_T  (ibex_icache_core_agent_cfg));
+  `uvm_component_utils(ibex_icache_core_sequencer)
   `uvm_component_new
-
-  uvm_tlm_analysis_fifo #(ibex_icache_mem_req_item) request_fifo;
 
   // An objection used for heartbeat tracking. Set with register_hb.
   uvm_callbacks_objection hb_objection;
-
-  function void build_phase(uvm_phase phase);
-    super.build_phase(phase);
-    request_fifo   = new("request_fifo", this);
-  endfunction
 
   function void register_hb (uvm_callbacks_objection obj);
     hb_objection = obj;
@@ -28,8 +17,9 @@ class ibex_icache_mem_sequencer
   virtual function void send_request(uvm_sequence_base sequence_ptr,
                                      uvm_sequence_item t,
                                      bit rerandomize = 0);
-    if (hb_objection != null)
+    if (hb_objection != null) begin
       hb_objection.raise_objection(this);
+    end
 
     super.send_request(sequence_ptr, t, rerandomize);
   endfunction
--- a/dv/uvm/icache/dv/env/ibex_icache_env.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_env.sv
@@ -13,6 +13,11 @@ class ibex_icache_env extends dv_base_env #(
   ibex_icache_core_agent core_agent;
   ibex_icache_mem_agent  mem_agent;
 
+  // Heartbeat tracking
+  uvm_callbacks_objection           hb_objection;
+  uvm_heartbeat                     heartbeat;
+  uvm_event                         hb_event;
+
   `uvm_component_new
 
   function void build_phase(uvm_phase phase);
@@ -23,6 +28,10 @@ class ibex_icache_env extends dv_base_env #(
 
     mem_agent = ibex_icache_mem_agent::type_id::create("mem_agent", this);
     uvm_config_db#(ibex_icache_mem_agent_cfg)::set(this, "mem_agent*", "cfg", cfg.mem_agent_cfg);
+
+    hb_objection = new("hb_objection");
+    heartbeat    = new("heartbeat", this, hb_objection);
+    hb_event     = new("hb_event");
   endfunction
 
   function void connect_phase(uvm_phase phase);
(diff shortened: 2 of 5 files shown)


udinator:
@rswarbrick - This heartbeating approach is currently not used in our UVM environments, we just raise and drop objections from the scoreboard to signal EOT.
I've personally never seen this `uvm_heartbeat` mechanism before, so I am adding the rest of the Google DV team for any input they might have; maybe this is worth evaluating to see if we should use this approach in the other testbenches or stick with the current approach in all testbenches.

udinator:
@rswarbrick - We currently don't use the heartbeat mechanism at all in our existing testbenches; I'm adding the rest of the Google DV team to this PR to provide any input.
This might be interesting to evaluate to see whether we want to use it in all of our testbenches or forgo use of it altogether.

weicaiyang:
I haven't used uvm_heartbeat before. I feel it's useful. It serves as a common timer to check if any activity occurs during certain period, which may help us reduce some watchdog. Since it monitors the update for transaction request, we should check max delay used in sequence rather than driver, right? @rswarbrick 

Since we have some timeout mechanism in our base lib, this may be not a must-have

rswarbrick:
@weicaiyang Yes, you're right that the relevant time is governed by the sequence, not the driver. But then the driver is the thing that wiggles the pins that cause the transactions that poke the heartbeat monitor, so I guess you can say it either way. [*]

The problem with the global test timeout is that you have to adjust it whenever the test changes. It's also a bit "probabilistic": if your sequence items take different amounts of time, how long should a test be allowed to take? `num_items * max_time`, maybe? But maybe that's 100 times as long as tests normally take, which is painful if you're waiting on a stuck simulation. So in practice, you pick a plausible time based on the simulations you've seen and hope that you don't see a timing outlier, which the guessed timeout would wrongly kill. With a per-item heartbeat, you spot the hang `max_time` after things get stuck and there's no guesswork required.

[*] I'm not sure whether "I know an old lady who swallowed a fly" is well-known in the states. If it is, it's a good backing tune to this sentence!

weicaiyang:
> @weicaiyang Yes, you're right that the relevant time is governed by the sequence, not the driver. But then the driver is the thing that wiggles the pins that cause the transactions that poke the heartbeat monitor, so I guess you can say it either way. [*]
 
actually need to consider delay in both driver and seq. Since `hb_objection.raise_objection`is called at  `send_request `, need delay > driver_delay + seq_delay

for ... begin
  `uvm_do() // delay in driver
  #dly; // delay in seq
end

> The problem with the global test timeout is that you have to adjust it whenever the test changes. It's also a bit "probabilistic": if your sequence items take different amounts of time, how long should a test be allowed to take? `num_items * max_time`, maybe? But maybe that's 100 times as long as tests normally take, which is painful if you're waiting on a stuck simulation. So in practice, you pick a plausible time based on the simulations you've seen and hope that you don't see a timing outlier, which the guessed timeout would wrongly kill. With a per-item heartbeat, you spot the hang `max_time` after things get stuck and there's no guesswork required.

I got the idea, uvm_heartbeat is helpful, but global test timeout is still needed.
In this case, heartbeat could never timeout if CSR ctrl is always 0. We still need a very big global timer.
```
forever begin
  csr_rd(ctrl, val);
  if (val) break;
end
```

udinator:
I agree with Weicai, I think a global timeout is still needed.

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_sequencer.sv
```diff
@@ -0,0 +1,26 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class ibex_icache_core_sequencer extends dv_base_sequencer #(.ITEM_T (ibex_icache_core_item),
```
weicaiyang, line 5: FYI, [This paper](http://www.synapse-da.com/Uploads/PDFFiles/04_UVM-Heartbeat.pdf) (I saw it in Udi's email) provides an easy way to monitor all components without adding any code in sequencer, which is cleaner. And it also checks activity in monitor as well.

But perhaps we can address it later. IMO, it doesn't have to be prefect now.

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_sequencer.sv
```diff
@@ -0,0 +1,26 @@
...
+
+  function void register_hb (uvm_callbacks_objection obj);
+    hb_objection = obj;
+  endfunction
+
+  virtual function void send_request(uvm_sequence_base sequence_ptr,
+                                     uvm_sequence_item t,
+                                     bit rerandomize = 0);
+    if (hb_objection != null)
+      hb_objection.raise_objection(this);
```
weicaiyang, line 21: nit: use begin...end when `if` statement is more than 1 line

rswarbrick: Thanks for spotting this. I'll fix this on Monday before merging anything.

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_sequencer.sv
rswarbrick: That's funny - I was actually following that paper when I implemented this! :-) You're right that his version has less code, but I'm not convinced it's cleaner: I had to think quite hard to understand what he was doing (and regexes to select components? Eugh!), so I thought I'd go for the "stupid and obviously correct" approach, at least at first.

This approach (putting the logic in the sequencer, rather than the monitor) also means that you only poke the heartbeat monitor when making forward progress through the test. I guess that this isn't quite true for a reactive slave (like the memory below), but I guess it would be better than tracking every bus transaction if you just had masters. Hmm, thinking about that, maybe this should remove the monitoring from the memory entirely. I'll have a more careful think on Monday.

rswarbrick:
> I agree with Weicai, I think a global timeout is still needed.

Ok. I guess that will catch programming errors if nothing else. I'll put it back in before pushing anything up on Monday.

rswarbrick:
The new push is rebased onto master. It also adds a global timeout back in (although it's 100 times what it was: hopefully will never be hit) and adds the missing begin/end that Weicai spotted.

I thought a bit about whether the heartbeat is useful on the memory side and realised that it probably is for this environment. A sequence item for the core agent can last quite a while (wait a bit with the request line low, then fetch 100 instructions), so without the memory heartbeat, we'd need quite a long timeout. Barring spectacular design bugs, the memory side shouldn't generate much traffic after the core has stopped requesting data. (It will actually just fill up the fill buffers. With the default configuration, that's 8 loads.)

I'm going to merge this version, but I'd love to hear any follow-up comments (either here or elsewhere).

weicaiyang:
> The new push is rebased onto master. It also adds a global timeout back in (although it's 100 times what it was: hopefully will never be hit) and adds the missing begin/end that Weicai spotted.
> 
> I thought a bit about whether the heartbeat is useful on the memory side and realised that it probably is for this environment. A sequence item for the core agent can last quite a while (wait a bit with the request line low, then fetch 100 instructions), so without the memory heartbeat, we'd need quite a long timeout. Barring spectacular design bugs, the memory side shouldn't generate much traffic after the core has stopped requesting data. (It will actually just fill up the fill buffers. With the default configuration, that's 8 loads.)
> 
> I'm going to merge this version, but I'd love to hear any follow-up comments (either here or elsewhere).

Agree. Sometimes one item in driver may cause many activities in the reactive side, which driver could be idle for a while and only monitor sees some activity. Checking activity in all monitors will be easier to figure out the maximum possible idle time.
Hence, I feel the approach in that paper may be a bit complicated, but it can quickly enable activity check for all components and is more extensible to the other TBs.

PR #807, opened 2020-04-22
Monitor and SVA assertions for core agent in icache UVM testbench

rswarbrick:
Since the memory agent PR (#793) is slightly stalled, I've split out the follow-up work that adds monitors and SVA assertions. This is the code that applies to the core agent.

You should be able to build it, but it won't run to completion because there is no memory agent. I have (of course) checked that rebasing the patches from #793 on top of this patchset results in something that works.

PR #811, opened 2020-04-23
Add monitor and SVA checks to icache <-> memory interface in UVM testbench

rswarbrick:
The equivalent PR for the core side is PR #807. (In fact, these patches have been split, so the commit messages look rather similar).

I don't believe there are any remaining review items open on that PR that apply here. These were relevant, but are now resolved:

  1. ~~Naming of the interface that holds the assertions (see [here](https://github.com/lowRISC/ibex/pull/807#discussion_r413213740)).~~ (this naming seems ok)
  1. ~~Naming of the object that reports bus transactions (see [here](https://github.com/lowRISC/ibex/pull/807#discussion_r413201343)).~~ (renamed as requested to `ibex_icache_mem_bus_item`)

### DIFF
--- a/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
@@ -11,7 +11,7 @@ class ibex_icache_scoreboard
 
   // TLM agent fifos
   uvm_tlm_analysis_fifo #(ibex_icache_core_bus_item) core_fifo;
-  uvm_tlm_analysis_fifo #(ibex_icache_mem_resp_item) mem_fifo;
+  uvm_tlm_analysis_fifo #(ibex_icache_mem_bus_item)  mem_fifo;
 
   `uvm_component_new
 
@@ -42,10 +42,10 @@ class ibex_icache_scoreboard
   endtask
 
   virtual task process_mem_fifo();
-    ibex_icache_mem_resp_item item;
+    ibex_icache_mem_bus_item item;
     forever begin
       mem_fifo.get(item);
-      `uvm_info(`gfn, $sformatf("received ibex_mem_intf_seq item:\n%0s", item.sprint()), UVM_HIGH)
+      `uvm_info(`gfn, $sformatf("received mem transaction:\n%0s", item.sprint()), UVM_HIGH)
     end
   endtask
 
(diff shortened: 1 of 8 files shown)


dv/uvm/icache/dv/ibex_icache_mem_agent/ibex_icache_mem_bus_transaction.sv
```diff
@@ -0,0 +1,32 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// A transaction item that represents something happening on the memory bus. A 'request' is
+// initiated by the cache and comes with an address. A 'response' comes from the memory system
+// (including possibly the PMP checker). It comes with data and error flags.
+
+class ibex_icache_mem_bus_transaction extends uvm_sequence_item;
```
udinator, line 9: Does this mean the icache testbench will have 3 "types" of sequence items in total?

rswarbrick: 5, I think. The memory agent has

1. a request item which gets generated when the monitor sees a request on the interface
2. a response item which it generates and contains data to drive onto the interface with the driver
3. a bus transaction item which the monitor generates and passes to the scoreboard when something happens.

The core agent currently has

1. a sequence item to tell the driver what to do
2. (when #807 lands) a bus transaction item that the monitor generates and passes to the scoreboard when something happens.

udinator: Got it.
I think this question also applies to the other SVA monitor PR as well, but why is a new class of sequence item (both here and in the core_agent) solely for monitor<->scoreboard interactions necessary over using a single transaction class for both the driver and monitor? I'm not sure if I'm missing something, but it seems like more complexity over just using a single transaction class.

udinator: Ah, just saw that Sri commented something very similar on the other SVA PR about the new sequence items.  Just to reduce unnecessary noise, we can move all discussion to that comment thread.
I think I agree with him - I feel it would be simpler to use the same sequence item cater to both driver and monitor, but since you already various components implemented, I'll leave the final decision up to you.

rswarbrick: Ok, let's talk there: I'll resolve this conversation.

rswarbrick:
Latest push renames the bus transaction object to `ibex_icache_mem_bus_item` and rebases. No other change.

rswarbrick:
(The force-push is just a rebase, fixing the conflicts with the next-door lines from the core-monitor PR that just landed)

PR #815, opened 2020-04-24
[rtl] Rework multdiv_en/sel

### LINKED ISSUE #781: ID/EX: Use `multdiv_sel` instead of `multdiv_en` and variants where possible
While reviewing #746 I noted two things in ID/EX:
- In `ibex_ex_stage.sv` we are generating `multdiv_en` based on `mult_en_*` and `div_en_*` signals to control the result mux.
- Inside `ibex_multdiv_fast/slow.sv` we are using the same `mult_en_*` and `div_en_*` to control both FSMs and muxes. The `multdiv_sel` signal is not available (only inside the ALU).

The main difference between these signals is that `multdiv_sel` directly comes out of the decoder where as `mult_en_*` and `div_en_*` are qualified with `instr_executing` inside the ID stage (which itself depends on various other signals from ID stage and the controller).

As far as I understand, this is needed to control the FSMs inside `ibex_multdiv_fast/slow.sv` but not for most of the muxes. These muxes or at least some of them could be controlled depending on the decoder output only. Since all those muxes are critical, I would suggest to investigate which of the muxes can be controlled by the less critical `multdiv_sel` inside the ID/EX stage.

Do you @GregAC , @tomroberts-lowrisc , @rswarbrick , @imphil think this makes sense? Do you have any thoughts on this?

vogelpi:
This PR reworks the generation and usage of the `mult/div_en/sel` signals.

Previously, the `multdiv_sel` was generated in the ALU decoder and used inside the ALU primarily. The `mult/div_en` signals were generated in the regular decoder, qualified with `illegal_insn` and `instr_executing` and used for controlling the FSMs inside multdiv as well as a couple of data muxes. 
This means that, If the 3rd pipeline stage is enabled, also LSU exception and hazard signals are factored into the control signals of these data muxes (some of which are on the critical path).

With this PR, these signals are reworked as follows:
- The unqualified `multdiv_sel` signal is split into `mult_sel` and `div_sel` both generated inside the ALU decoder. Only these unqualified signals are used to control data muxes in the EX stage.
- The qualified `mult_en` and `div_en` siganls are dervied from `mult_sel` and `div_sel` and qualified as before. They are no longer used for controlling data muxes.
- The regular decoder is only responsible for detecting illegal instructions related to `multdiv`. Everything else is moved to the ALU decoder.

This resolves #781 .

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -78,9 +78,9 @@ module ibex_decoder #(
 
     // MULT & DIV
     output logic                 mult_en_o,             // perform integer multiplication
-    output logic                 div_en_o,              // perform integer division or
-                                                        // remainder
-    output logic                 multdiv_sel_o,
+    output logic                 div_en_o,              // perform integer division or remainder
+    output logic                 mult_sel_o,            // as above but static, for data muxes
+    output logic                 div_sel_o,             // as above but static, for data muxes
 
     output ibex_pkg::md_op_e     multdiv_operator_o,
     output logic [1:0]           multdiv_signed_mode_o,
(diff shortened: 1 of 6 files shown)


vogelpi:
According to @ganoam (thanks for checking), the impact on timing is negligible.

rtl/ibex_ex_block.sv
```diff
@@ -204,7 +211,8 @@ module ibex_ex_block #(
     );
   end
 
-  // ALU output valid in same cycle, multiplier/divider may require multiple cycles
-  assign ex_valid_o = multdiv_en ? multdiv_valid : !alu_imd_val_we;
+  // Multiplier/divider may require multiple cycles. ALU output valid in same cycle, unless
+  // intermediate result register is written (indicates non-final cycle of ALU operation).
```
GregAC, line 215:
```suggestion
  // Multiplier/divider may require multiple cycles. The ALU output is valid in the same cycle
  // unless the intermediate result register is being written (which indicates this isn't the
  // final cycle of ALU operation).
```

rtl/ibex_decoder.sv
```diff
@@ -926,17 +883,54 @@ module ibex_decoder #(
...
-            {7'b000_0001, 3'b011}, // mulhu
-            {7'b000_0001, 3'b100}, // div
-            {7'b000_0001, 3'b101}, // divu
-            {7'b000_0001, 3'b110}, // rem
+            // RV32M instructions, all use the same ALU operation
+            {7'b000_0001, 3'b000}: begin // mul
+              alu_operator_o        = ALU_ADD;
+              mult_sel_o            = RV32M ? 1'b1 : 1'b0;
+              multdiv_operator_o    = MD_OP_MULL;
+              multdiv_signed_mode_o = 2'b00;
```
GregAC, line 891: I had intentionally left the multiplier control in the original decode block rather than moving it to the ALU decode block in an attempt to balance fan-out between the two instruction registers. Would be good to get some fan-out statistics on the two registers to see where the balance really lies.

For now could we leave it as is? So multdiv_operator_o and multdiv_signed_mode_o assigned in the first decode block with alu_operator_o, mul_sel_o and div_sel_o assigned in this block?

vogelpi: Yep, I can definitely change that.

vogelpi:
Thanks for reviewing this @GregAC . I will discuss the fanout thing with Noam and implement your suggestions.

ganoam:
@GregAC, no max_fanout constraint has been set. For the 2 stage configuration, The nets with a fanout of > 100 are:
``` 
Net                 Fanout   Attributes   Capacitance   Driver
 --------------------------------------------------------------------------------
 rst_ni                 692                    2.79      rst_ni
 cs_registers_i/net6655
                        121                    0.31      cs_registers_i/clk_gate_mstatus_q_reg_mprv_/latch/ECK
 ex_block_i/gen_multdiv_fast_multdiv_i/net7061
                        104                    0.27      ex_block_i/gen_multdiv_fast_multdiv_i/clk_gate_op_quotient_q_reg/latch/ECK
```

For the three stage configuration:
```
 Net                 Fanout   Attributes   Capacitance   Driver
 --------------------------------------------------------------------------------
 rst_ni                 694                    2.82      rst_ni
 cs_registers_i/net6514
                        121                    0.31      cs_registers_i/clk_gate_mstatus_q_reg_mprv_/latch/ECK
 ex_block_i/gen_multdiv_fast_multdiv_i/net6938
                        104                    0.27      ex_block_i/gen_multdiv_fast_multdiv_i/clk_gate_op_quotient_q_reg/latch/ECK
```

vogelpi:
Thanks for providing the reports @ganoam. It seems that the decoder outputs are not amongst the most critical fanout nets. But maybe a limit of 100 is too high.

Anyway, I reworked the decoder as suggested by @GregAC . The ALU decoder only generates the `mult/div_sel/en` signals. The rest is generated as before by the regular decoder.

vogelpi:
Thanks @GregAC  for approving this. I will merge once CI has completed.

PR #816, opened 2020-04-24
Add RV32B parameter option

towoe:


towoe:
Thanks @vogelpi. I only quickly checked the PR titles...

vogelpi:
I was too lazy to factor it out into a separate PR, sorry. :-/

PR #818, opened 2020-04-24
ICache version for Riviera-PRO

dawidzim:
Hey!
In this pull request I hjson file to run icache verification on Riviera-Pro and small change in `sim.mk` to detect Riviera-Pro

### DIFF
--- a/dv/uvm/data/sim.mk
+++ b/dv/uvm/data/sim.mk
@@ -33,7 +33,12 @@ gen_sv_flist: pre_compile ral
 
 compile: gen_sv_flist
 	@echo "[make]: compile"
+ifeq (${build_cmd},vlog)
+	cd ${sv_flist_gen_dir} && vlib work && ${build_cmd} ${build_opts}
+else
+	$(info sim: ${vlog})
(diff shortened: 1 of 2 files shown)


dv/uvm/data/sim.mk
```diff
@@ -33,7 +33,12 @@ gen_sv_flist: pre_compile ral
 
 compile: gen_sv_flist
 	@echo "[make]: compile"
+ifeq (${build_cmd},vlog)
+	cd ${sv_flist_gen_dir} && vlib work && ${build_cmd} ${build_opts}
```
rswarbrick, line 37: I'm happy either way, but I think a slightly cleaner way to do this would be something like:

``` make
compile-prep := true
ifeq ($(build_cmd},vlog)
  compile-prep := vlib work
endif

compile: gen_sv_flist
	@echo "[make]: compile"
	cd ${sv_flist_gen_dir} && $(compile-prep) && ${build_cmd} ${build_opts}
```
which avoids the weird indentation in the rule definition.

dawidzim:
I will create new PR with more changes for Icache

PR #820, opened 2020-04-27
Assorted bugfixes in the ICache UVM test code

rswarbrick:
These fix various bugs that I ran into when trying to get the scoreboard working properly. The most interesting patch is probably the second: it turns out that my previous plan to spot fetch requests is fragile and doesn't take the non-determinacy of delta-cycle scheduling into account. Oops!


rswarbrick:
Well, it's a less impressive catch if you know that I spent half an hour staring at uvm_info traces going "huh?!", but at least it works now :-)

PR #822, opened 2020-04-28
Update dv_lib to lowRISC/opentitan@1d17b122

rswarbrick:
Update code from subdir hw/dv/sv/dv_lib in upstream repository
https://github.com/lowRISC/opentitan to revision
1d17b1225d324c81da522c69317335a83edd5ddb

* [dv] Add excl for rstmgr, pwrmgr and fix top-level csr test (Weicai
  Yang)
* [dv] Allow dv_lib-based sequences to have different RSP/REQ types
  (Rupert Swarbrick)
* [dv] Support WO, RO type for mem (Weicai Yang)
* [dv,sw] SW -> DV tb self-checking mechanism - SV (Srikrishna Iyer)
* [dv/top] Fix csr rw test (Cindy Chen)

Signed-off-by: Rupert Swarbrick <rswarbrick@lowrisc.org>

### DIFF
--- a/vendor/lowrisc_dv_lib.lock.hjson
+++ b/vendor/lowrisc_dv_lib.lock.hjson
@@ -9,7 +9,7 @@
   upstream:
   {
     url: https://github.com/lowRISC/opentitan
-    rev: 0d7f7ac755d4e00811257027dd814edb2afca050
+    rev: 1d17b1225d324c81da522c69317335a83edd5ddb
     only_subdir: hw/dv/sv/dv_lib
   }
 }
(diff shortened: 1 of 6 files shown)


rswarbrick:
This PR is to allow sequences with different RSP/REQ types: I've got a bugfix in the core agent which requires the driver to respond back to the sequence and it feels much cleaner to respond with a "was this an error" bit than adding a field to the request sequence item.

rswarbrick:
Wow, that was quick. Thanks, Udi!

PR #823, opened 2020-04-28
Branch after error

rswarbrick:
This fixes a bug in the core agent where it could start trying to fetch data again without a branch after it had seen an error from the cache.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_rsp_item.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_rsp_item.sv
@@ -1,0 +1,19 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class ibex_icache_core_rsp_item extends uvm_sequence_item;
+
+  // A sequence item that is used for responses from the driver back to the sequence. This is needed
+  // so that we can spot when there was an error signalled by the core (which means the next
+  // sequence item will need to be a branch).
+
+  bit saw_error;
+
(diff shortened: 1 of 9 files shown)


rswarbrick:
(Force-push fixes a comment and gets rid of an unnecessary tidy-up in an unrelated file)

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_rsp_item.sv
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class ibex_icache_core_rsp_item extends uvm_sequence_item;
+
+  // A sequence item that is used for responses from the driver back to the sequence. This is needed
+  // so that we can spot when there was an error signalled by the core (which means the next
+  // sequence item will need to be a branch).
+
+  bit saw_error;
```
udinator, line 11: I feel like having a whole new sequence item just to contain a single `bit` might be a tad overkill here, what do you think about adding this field to the existing `core_req_item` and using a soft constraint to keep it at 0 unless overridden?

rswarbrick: I think I probably prefer having different types for different stuff. I agree that a full sequence item to hold a single bit is a little crazy, but it feels less bad than adding an "out parameter" to the sequence item going the other way. Let's go with this and, if we find it's killing simulation performance, we can always revisit.

udinator: sounds good to me.

PR #825, opened 2020-04-29
Minor tweaks to make icache test sequences more interesting

rswarbrick:
The first commit makes them longer; the second makes sure we actually see the invalidate pin toggle occasionally.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
@@ -84,7 +84,7 @@ interface ibex_icache_core_if (input clk, input rst_n);
   //
   // A one-cycle pulse will start an invalidation, but testing might want a longer pulse (which the
   // cache should support)
-  task automatic invalidate_pulse(int num_cycles);
+  task automatic invalidate_pulse(int unsigned num_cycles);
     driver_cb.invalidate <= 1'b1;
     wait_clks(num_cycles);
     driver_cb.invalidate <= 1'b0;
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/seq_lib/ibex_icache_core_sanity_seq.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/seq_lib/ibex_icache_core_sanity_seq.sv
@@ -10,7 +10,7 @@ class ibex_icache_core_sanity_seq extends ibex_icache_core_base_seq;
   `uvm_object_new
 
   rand int count;
-  constraint c_count { count > 0; count < 100; }
+  constraint c_count { count inside {[800:1000]}; }
 
   task body();
     // If this is set, the next request will be constrained to have trans_type
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_driver.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_driver.sv
@@ -143,7 +143,7 @@ class ibex_icache_core_driver
   virtual task automatic invalidate();
     int unsigned num_cycles;
     `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(num_cycles,
-                                       num_cycles dist { 0 :/ 499, [1:20] :/ 1 };)
+                                       num_cycles dist { 1 :/ 10, [2:20] :/ 1 };)
     cfg.vif.invalidate_pulse(num_cycles);
   endtask
 
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_req_item.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_req_item.sv
@@ -45,9 +45,9 @@ class ibex_icache_core_req_item extends uvm_sequence_item;
   }
 
   constraint c_invalidate_dist {
-    // Poke the cache invalidate line one time in 500. This takes ages and we don't want to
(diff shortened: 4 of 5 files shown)


dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
```diff
@@ -85,6 +85,8 @@ interface ibex_icache_core_if (input clk, input rst_n);
   // A one-cycle pulse will start an invalidation, but testing might want a longer pulse (which the
   // cache should support)
   task automatic invalidate_pulse(int num_cycles);
+    assert(num_cycles > 0);
```
udinator, line 88: I think this is unnecessary, as the randomization of `num_cycles` in the `invalidate()` task will throw an error if something funky happens in the constraint solver and the number goes to 0 somehow.

dv/uvm/icache/dv/ibex_icache_core_agent/seq_lib/ibex_icache_core_sanity_seq.sv
```diff
@@ -10,7 +10,7 @@ class ibex_icache_core_sanity_seq extends ibex_icache_core_base_seq;
   `uvm_object_new
 
   rand int count;
-  constraint c_count { count > 0; count < 100; }
+  constraint c_count { count > 800; count < 1000; }
```
udinator, line 13: nit : could you use an `inside` constraint to improve the readability?

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
rswarbrick: This is to catch programming errors. Yes, if I've got things right, it should never trigger. But I didn't before! :-)

udinator: Got it. Then can you use one of the `DV_CHECK...` macros here instead? Or an `ASSERT` macro if you want a "true" assertion behavior.

rswarbrick: Done.

dv/uvm/icache/dv/ibex_icache_core_agent/seq_lib/ibex_icache_core_sanity_seq.sv
rswarbrick: Done.

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_if.sv
```diff
@@ -84,7 +86,10 @@ interface ibex_icache_core_if (input clk, input rst_n);
   //
   // A one-cycle pulse will start an invalidation, but testing might want a longer pulse (which the
   // cache should support)
-  task automatic invalidate_pulse(int num_cycles);
+  task automatic invalidate_pulse(int unsigned num_cycles);
+    import uvm_pkg::*;
```
udinator, line 90: Hmm, looks like my previous suggestion ended up making things messier due to the necessary imports of the uvm package and the dv_macros file to access the DV_CHECK macro here.
If this check should still remain in this task, would moving this task to the driver present any issues? That way the extra imports won't be needed.

rswarbrick: This is getting a little silly! :-)

This task is in the interface because you suggested it should be factored out from the driver in a previous PR. That seems sensible enough to me. There's a way to mis-use the task (calling with a zero cycle count), and I think it would be good to put in a check to spot such a programming error.

I agree that a dependency on UVM in the interface isn't great. Which leaves us with... an assertion, which is what my code originally had. Is there a problem with using one of these?

udinator: We disallow using `assert` in favor of the `DV_CHECK*` macros to check conditions (such as this) in non-assertion-monitor components due to us running some simulations with `$assertoff` and other related system functions which allows us to turn off any assertion monitor (or fpv) components; this also has the side effect of disabling any extra assertions within the testbench like the ones you have here.
This is also a UVM coding style requirement - to use `DV_CHECK*` macros instead of `assert`, especially since that is the whole purpose behind having these macros to begin with.

rswarbrick: Ok, so if I understand correctly:
  1. The only way that DV code is allowed to check for errors is via the UVM machinery
  2. We don't want to add a dependency on UVM code in interfaces
  3. Despite being an interface, this code lives under the `dv` directory, so is considered DV code.

Therefore it is impossible to do any sort of error checking in `ibex_icache_core_if.sv`.

This seems like a strange state of affairs, but I want to land this change rather than argue forever, so I'm going to just delete the check: it was only to catch programming errors anyway. I guess I'll just have to be careful not to make mistakes :-)

Finally, the DV style guide doesn't mention assertions (except in the context of randomization). Please could you raise an issue or a PR adding the rule above? Probably that's the right place to discuss such things, rather than here.

udinator: Sounds good.

With respect to assertions and the style guide, I agree this is something that should be added as well, so it's going into some upcoming changes to the style guide to add some new topics and update existing ones with useful information like this.

rswarbrick:
Force-push rebases and removes the disallowed error check. Otherwise, no change.

PR #828, opened 2020-04-29
Initial icache scoreboard

rswarbrick:
This correctly tracks fetch addresses and fetched data. It understands
changing memory seeds, errors, invalidations and enable/disable.

Most of the complexity is in checking whether a fetch got the right
answer, given the set of memory seeds that it might have used. This
isn't conceptually hard (use a local memory model; check each seed and
see whether it matches), but is a bit fiddly in practice. In
particular, a misaligned 4-byte load might actually correspond to two
different seeds: note the nested loops in `check_compatible_2`.

The general flow of these checks is:

```
     check_compatible
  -> check_compatible_<i>    (loop over plausible seeds)
  -> is_seed_compatible_<i>  (ask the memory what data to expect)
  -> is_fetch_compatible_<i> (compare seen/expected data)
```

Note that the `check_compatible_<i>` functions have a "chatty"
parameter. This is to help with debugging when something goes wrong:
if the check fails then the `check_compatible` function calls it again
with `chatty=1`, which dumps a list of the seeds we checked and why they
didn't match.

Other than the scoreboard itself, this patch also adds a "seed" field
to `ibex_icache_mem_bus_item`. This is used by the monitor to inform the
scoreboard when a new memory seed has been set. Obviously, this
doesn't correspond to any actual monitored signals, but we need some
sort of back channel, and this looked like a sensible way to do it.

The patch also stops reporting PMP responses to the scoreboard: it
turns out the scoreboard doesn't need to care about them, so we can
simplify things slightly this way.

At the moment, the scoreboard doesn't check that fetched data isn't
stored when the cache is disabled. You could see this by disabling the
cache, fetching from an address, enabling the cache and changing the
memory seed, and fetching from the address again. I think it would be
reasonably easy to make an imprecise version of the check, where a
seed gets discarded from the queue if its live period is completely
within a period where the cache was disabled, but I want to wait until
we've got some tests that actually get cache hits before I implement
this.

There's also a slight imprecision in the busy line check that needs
tightening up.

Both of these to-do items have TODO comments in the code.

### DIFF
--- a/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
@@ -7,22 +7,58 @@ class ibex_icache_scoreboard
 
   `uvm_component_utils(ibex_icache_scoreboard)
 
-  // local variables
-
   // TLM agent fifos
   uvm_tlm_analysis_fifo #(ibex_icache_core_bus_item) core_fifo;
   uvm_tlm_analysis_fifo #(ibex_icache_mem_bus_item)  mem_fifo;
+
+  localparam BusWidth = 32;
+
+  // A memory model. Note that the model is stateful (because of the seed). Since we need to try out
+  // different seeds to check whether a fetch was correct, we need our own instance, rather than a
+  // handle to the model used by the agent. The actual seed inside mem_model is ephemeral: we keep
+  // track of state in mem_seeds, below.
+  ibex_icache_mem_model #(BusWidth) mem_model;
+
+  // A queue of memory seeds.
+  bit [31:0]   mem_seeds[$];
+
+  // Tracks the next address we expect to see on a fetch. This gets reset to 'X, then is set to an
+  // address after each branch transaction.
+  logic [31:0] next_addr;
+
+  // This counter is used for tracking invalidations. When we see an invalidation happen, we set
+  // this to the index of the last seed in mem_seeds. When the next branch happens, we clear out
+  // memory seeds up to that index.
+  int unsigned invalidate_seed = 0;
+
+  // This counter points to the index (in mem_seeds) of the seed that was in use when the last
+  // branch happened. If the cache is supposed to be disabled, a fetch is only allowed to return
+  // data corresponding to that seed or later.
+  int unsigned last_branch_seed = 0;
+
+  // A counter of outstanding transactions on the memory bus. The busy flag should never go to zero
+  // when this is non-zero.
+  int unsigned mem_trans_count = 0;
+
+  // Track the current enabled state. The enabled flag tracks whether the cache is currently
+  // enabled. The no_cache flag is high if the cache has been disabled since before the last branch.
+  // If this is set, we can be stricter about the data that can be fetched.
+  //
+  // TODO: Our enable tracking checks that we don't get cache hits when the cache is disabled. It
+  //       doesn't yet check that the cache doesn't store fetches when disabled.
+  bit          enabled = 0;
+  bit          no_cache = 1'b1;
+
+  // Track the current busy state
+  bit          busy = 0;
 
   `uvm_component_new
 
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
     core_fifo = new("core_fifo", this);
-    mem_fifo = new("mem_fifo", this);
-  endfunction
-
-  function void connect_phase(uvm_phase phase);
-    super.connect_phase(phase);
+    mem_fifo  = new("mem_fifo",  this);
+    mem_model = new("mem_model");
   endfunction
 
   task run_phase(uvm_phase phase);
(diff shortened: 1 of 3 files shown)


rswarbrick:
Force-push: I'd left the busy check as a `uvm_warning` and forgot to turn it into an error.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -7,22 +7,61 @@ class ibex_icache_scoreboard
...
+  // TODO: Our enable tracking checks that we don't get cache hits when the cache is disabled. It
+  //       doesn't yet check that the cache doesn't store fetches when disabled.
+  bit          enabled = 0;
+  bit          no_cache = 1'b1;
+
+  // Track the current busy state
+  bit          busy = 0;
+
+
+  function new(string name="", uvm_component parent=null);
```
udinator, line 56: nit: use `uvm_component_new` macro here

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
+    next_addr = item.address;
+
+    if (invalidate_seed > 0) begin
+      // We've seen an invalidate signal recently. Clear out any expired seeds.
+      assert(invalidate_seed < mem_seeds.size);
+      mem_seeds = mem_seeds[invalidate_seed:$];
+      invalidate_seed = 0;
+    end
+
+    assert(mem_seeds.size > 0);
```
udinator, line 101: this applies here and in the other locations where `assert` is used - you should be using the various `DV_CHECK...` macros to do all of the various checks instead.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
+    last_branch_seed = mem_seeds.size - 1;
+
+    if (!enabled) no_cache = 1'b1;
+  endtask
+
+  task process_fetch(ibex_icache_core_bus_item item);
+    // Check that the address is what we expect
+    if (item.address !== next_addr) begin
```
udinator, line 109: similar to above, use `DV_CHECK_EQ` here:
```systemverilog
`DV_CHECK_EQ(item.address, next_addr,
             $sformatf("Fetch got data from address 0x%08h, but we expected address 0x%08h.",
                        item.address, next_addr))
```

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
+  // We'll miss cases where the busy line is slow to go high, but still manages to do so before the
+  // response comes back.
+  //
+  // TODO: Is this ok (because sometimes the memory will come back quick enough to catch us) or do
+  //       we need to do something cleverer? For example, we could have an SVA property (bound into
+  //       the design, or at testbench level) that checked the busy pin was high the cycle after a
+  //       request went out.
+
+  task automatic busy_check();
+    if ((mem_trans_count > 0) && !busy) begin
```
udinator, line 490: you can use a `DV_CHECK(...)` macro invocation here.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
+                                                 bit [31:0]   exp_insn_data,
+
+                                                 bit [31:0]   seed_lo,
+                                                 bit [31:0]   seed_hi,
+                                                 bit          chatty);
+
+    // We should only be called when the seen instruction data for the lower word is uncompressed
+    assert(seen_insn_data[1:0] == 2'b11);
+
+    if (exp_err_lo) begin
```
udinator, line 252: For all of the `if begin...end` blocks that return 0 (here and below), do you want these blocks to issue `uvm_error`s?

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
+      // The easier case: either the fetch was aligned, the lower word seems to have caused an error
+      // or the bits in the lower word are a compressed instruction (so we don't care about the
+      // upper 16 bits anyway).
+      if (!check_compatible_1(item.address, item.insn_data, item.err, 1'b0)) begin
+        void'(check_compatible_1(item.address, item.insn_data, item.err, 1'b1));
+        `uvm_error(`gfn,
+                   $sformatf("Fetch at address 0x%08h got data incompatible with available seeds.",
+                             item.address));
+      end
+    end
```
udinator, line 462: This is sort of a general comment regarding most of the `*_[i]` helper functions that are being used here, but I feel like the `chatty` argument is somewhat unnecessary.
I'd propose another option - instead of having every `uvm_info` gated by `chatty`, why not just change the verbosity argument of these macros to `UVM_DEBUG`? This way when you run a "normal" simulation or regression all of this debug information is hidden, but when you are developing new tests or hunting down any RTL/testbench bugs you can run sims with the `+UVM_VERBOSITY=UVM_DEBUG` command line plusarg to enable all of this debug information to be printed.
This would also have the side effect of preventing you from having to potentially call `check_compatible_[i]` twice just to flip a verbosity switch the second time around, and instead you can now just structure it like this:
```systemverilog
if (misaligned && good_bottom_work && uncompressed) begin
  // ...
  `DV_CHECK(check_compatible_2(item.address, item.insn_data, ...),
            $sformatf("Fetch at address 0x%08h got data incompatible with available seeds.",
                      item.address))
end else begin
  // ...
  `DV_CHECK(check_compatible_1(item.address, ...),
            $sformatf("Fetch at address 0x%08h got data incompatible with available seeds.",
                      item.address))
end
```

udinator:
@rswarbrick thanks for opening this PR! I think the logic here makes sense to me, but will give it a second look tomorrow to make sure I haven't missed anything.
I added some nit-ish comments, generally regarding the syntax of how you perform the checking; with one not-as-nitty comment regarding the `chatty` parameter for controlling the verbosity of your functions, let me know what you think.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Oh, good point. I think there was something meaningful in here at some point, but apparently not any more!

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -33,30 +72,426 @@ class ibex_icache_scoreboard
...
     forever begin
       core_fifo.get(item);
       `uvm_info(`gfn, $sformatf("received core transaction:\n%0s", item.sprint()), UVM_HIGH)
+      case (item.trans_type)
+        ICacheCoreBusTransTypeBranch:     process_branch(item);
+        ICacheCoreBusTransTypeFetch:      process_fetch(item);
+        ICacheCoreBusTransTypeInvalidate: process_invalidate(item);
+        ICacheCoreBusTransTypeEnable:     process_enable(item);
+        ICacheCoreBusTransTypeBusy:       process_busy(item);
+        default: assert(0);
```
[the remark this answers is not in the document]
rswarbrick: Ok.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Well, this is checking for a bug in the scoreboard code (rather than a bug in the RTL). Personally, I'd like the simulation to stop instantly in that case: there's no point in going on because Rupert screwed up! :-)

Do we have a style requirement that DV code must always UVM messages instead of asserts? If not, I'd rather leave it as-is.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Done (with `DV_CHECK_EQ`: I couldn't find a `DV_CHECK_CASE_EQ` in the source tree).

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: No. There are several seeds to check, so it's perfectly fine to fail some of them, so long as one matches. See the loops in `check_compatible_2` and `check_compatible_1`.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: If this happens then we've found a bug in the DUT, which should raise the busy line if there are any outstanding memory transactions. This isn't fatal to the simulation (but is definitely a bug in the design).

I've switched the code to use the `DV_CHECK` wrapper. It's now of the form `if (mem_trans_count > 0) DV_CHECK(busy, ...)`, because I think that makes the implication clearer than `DV_CHECK(busy || (mem_trans_count == 0), ...)`.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: The problem is that a failure in the checking functions (`is_fetch_compatible_*`) can be absolutely fine: it might just mean that the fetch matches a different seed in our list of possibilities. The current design is intended to let you see why an error comes out: it only goes around and prints messages if something hasn't matched any seed.

I'd quite like to keep that behaviour. Indeed, since I put this in, I've been able to debug a couple of RTL bugs for Tom already with no hacking of the test code required - it seems to work quite well!

I'm going to push up a version with the same structure for now, but switching all the `UVM_LOW` messages to `uvm_warning`. That way, the user doesn't even need to re-run the test on an error. Does this make sense to you? If you still think this is the wrong way to do it, can you suggest a different approach that will keep the "only print messages when something has gone awry" property?

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Oh, sorry, I misunderstood the last bit of your message. I think I definitely want either an assertion or `DV_CHECK_*FATAL`. Is there a reason to prefer one or the other?

udinator: Yep, we require that the `DV_CHECK` family of macros be used for this sort of checking -they tie much more closely into the UVM infrastructure and will not be affected by use of `$assertoff` or anything like that when we want to disable assertion monitors.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: I believe the case equality macro is also in the vendored in `dv_utils/dv_macros.svh` file - but here I don't think it matters either way whether 4-state checking is used or not.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: sgtm

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: Ah okay. If the behavior that you want is to only print these messages if no seeds match, then I think the current approach is okay.
In this case, can you change the `uvm_warning` to `uvm_info` with a verbosity of `UVM_MEDIUM`? We disallow use of `uvm_warning` macro, and `UVM_MEDIUM` verbosity better matches your intent of these debug messages only printing when no matches are seen (`UVM_LOW` verbosity is more meant to mark major testbench events, such as component instantiation, beginning/end of stimulus generation, etc...).

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Please add this to the DV style guide. You and Sri might know about these rules, but it's kind of difficult for new engineers to get involved if they only find out about them in PR reviews.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
rswarbrick: Ok, can do.

rswarbrick:
Force-push replaces the calls to `uvm_warning` with `uvm_info` (medium).

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
udinator: I totally agree with you, there are a fair amount of small (but important) things like this that didn't make it into the first go round of the style guide; I'm currently working on a PR to add several new topics (and update existing ones with things like this) to it.

PR #829, opened 2020-04-29
[rtl] icache error signalling fix

tomroberts-lowrisc:
- Data valid should only be signalled when the current beat is
  signalling an error
- PMP errors for future beats can sneak in while waiting for the
  current beat

### DIFF
--- a/rtl/ibex_icache.sv
+++ b/rtl/ibex_icache.sv
@@ -600,11 +600,12 @@ module ibex_icache #(
     // data output, and have data available to send.
     // Data is available if:
     // - The request hit in the cache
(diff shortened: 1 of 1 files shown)


tomroberts-lowrisc:
> Looks good to me, correctness-wise. I assume the foo[bar[baz]] lookups won't be a problem for timing?
Should be ok for the two bit signals used here.
Removed the ready-valid stuff to pull into it's own PR

rswarbrick:
> Removed the ready-valid stuff to pull into it's own PR

Cool, sounds sensible.

PR #834, opened 2020-05-01
Rework choosing new seeds in icache UVM memory model

rswarbrick:
The flow for a memory fetch is:

  1. Cache requests data for a memory address
  2. Agent spots the request, maybe signalling a PMP error
  3. Grant line goes high, at which point the request is granted.
  4. Sometime later (in-order pipeline), agent sends a response

Occasionally, we need to pick a new seed for the backing memory.
Before this patch, we picked these seeds at point (3).

Unfortunately this was wrong in the following case:

  1. We're switching from seed S0 to seed S1.
  2. The request is spotted with seed S0 and doesn't signal a PMP error
  3. The request is granted and we switch to seed S1.
  4. We respond with data from memory based on S1, with no memory
     error either

If S1 would have caused a PMP error, the resulting fetch (no error,
but data from S1) doesn't match any possible seed and the scoreboard
gets confused.

This patch changes to picking new seeds at (2) to solve the problem.
This isn't quite enough by itself, because if a request is granted on
a clock-edge, a new request address might appear and there isn't a
guaranteed ordering in the simulation between the new request and the
old grant (both things happen at the same time). To fix this, the
response sequence now maintains a queue of requests and their
corresponding seeds to make sure that all the checks for a fetch are
done with a single seed.

The patch also gets rid of the seed state in the memory model: it
turns out that this didn't really help: the scoreboard is always
asking "what would I get with this seed?" and now the sequence is
doing something similar.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
+++ b/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
@@ -17,34 +17,70 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
   endtask
 
   task body();
+    // We pick new seeds when we spot a request (rather than when we spot a grant) to ensure that
+    // any given fetch corresponds to exactly one seed. Unfortunately, there's a race if these two
+    // things happen at one clock edge:
+    //
+    //    - Request for new address which gets a new seed
+    //    - Grant of previous request
+    //
+    // If that happens, and the event scheduler happened to schedule the monitor to spot the request
+    // before the grant, we might accidentally handle the grant with the new seed rather than the
+    // old one, which would cause confusion in the scoreboard.
+    //
+    // To avoid this problem, we keep a FIFO of pending addresses along with their corresponding
+    // seeds. When a grant message comes in, we know that we can look up the correct seed there.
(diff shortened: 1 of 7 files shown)


dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
```diff
@@ -17,34 +17,70 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
...
+    //    - Request for new address which gets a new seed
+    //    - Grant of previous request
+    //
+    // If that happens, and the event scheduler happened to schedule the monitor to spot the request
+    // before the grant, we might accidentally handle the grant with the new seed rather than the
+    // old one, which would cause confusion in the scoreboard.
+    //
+    // To avoid this problem, we keep a FIFO of pending addresses along with their corresponding
+    // seeds. When a grant message comes in, we know that we can look up the correct seed there.
+    mailbox pending_grants = new("pending_grants");
```
udinator, line 33: nit: can you parameterize this?

rswarbrick: Hah, I saw the original comment on Friday evening and was going to come back saying "huh". Thanks for clarifying :-) Yep, can do.

rswarbrick:
Force-push rebases and parametrises the mailbox type.

PR #835, opened 2020-05-01
[rtl] Stop regfile writeback for load errors

### LINKED ISSUE #832: PMP load error qualifies register writeback
Hi,
I encountered this situation when running some pmp tests on Ibex:
Screenshot from 2020-04-30 16-52-47
The instruction at `PC=0x80002250` is an `lb` instruction that tries to access the address `0x800148e4`. One cycle after this instruction reaches the decode stage, it causes a PMP fault and `pmp_err_q` goes high for two cycles, but `lsu_rdata_valid_o` still goes high to `1'b1` even though the DMEM interface has not sent out a request at all, triggering an incorrect register writeback of of `XXXX_XXXX`.

tomroberts-lowrisc:
- A data or PMP error should stop the register file from being updated
- Fixes #832

### DIFF
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -476,7 +476,7 @@ module ibex_load_store_unit
 
   assign data_or_pmp_err    = lsu_err_q | data_err_i | pmp_err_q;
   assign lsu_resp_valid_o   = (data_rvalid_i | pmp_err_q) & (ls_fsm_cs == IDLE);
-  assign lsu_rdata_valid_o  = lsu_resp_valid_o & ~data_we_q;
+  assign lsu_rdata_valid_o  = (ls_fsm_cs == IDLE) & data_rvalid_i & ~data_or_pmp_err & ~data_we_q;
 
   // output to register file
   assign lsu_rdata_o = data_rdata_ext;


rtl/ibex_load_store_unit.sv
```diff
@@ -476,7 +476,7 @@ module ibex_load_store_unit
 
   assign data_or_pmp_err    = lsu_err_q | data_err_i | pmp_err_q;
   assign lsu_resp_valid_o   = (data_rvalid_i | pmp_err_q) & (ls_fsm_cs == IDLE);
-  assign lsu_rdata_valid_o  = lsu_resp_valid_o & ~data_we_q;
+  assign lsu_rdata_valid_o  = data_rvalid_i & ~data_or_pmp_err & ~data_we_q;
```
GregAC, line 479: I think we need to factor `(ls_fsm_cs == IDLE)` in here too (which we previously got from `lsu_resp_valid_o`. When you have an unaligned access the first data beat will raise `data_rvalid_i` but shouldn't signal `lsu_rdata_valid_o` (which indicates all of the load data is available). We won't be in IDLE for the first data beat only for the second.

tomroberts-lowrisc: Good point - now fixed

PR #838, opened 2020-05-01
[dv/fcov] Fix Ibex log parsing script

udinator:
Currently the riscv-dv functional coverage flow errors out when run on Ibex trace logs. This is because the operand string that this coverage flow analyzes is assumed to contain ABI register names, but Ibex does not display registers in ABI, thus the coverage test decides that it doesn't understand the Ibex CSV trace and exits out.
This PR fixes this issue by converting all register names in this operand string to ABI format, allowing the functional coverage flow to proceed as normal.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -69,7 +69,10 @@ def _process_ibex_sim_log_fd(log_fd, csv_fd, full_trace=True):
             trace_entry.pc = m.group("pc")
             trace_entry.binary = m.group("bin")
             if full_trace:
-                trace_entry.operand = m.group("instr").split()[1]
+                # Convert the operands into ABI format for
+                # the functional coverage flow
+                operands = m.group("instr").split()[1]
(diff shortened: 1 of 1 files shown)


rswarbrick:
This looks sensible enough, but I'm a bit confused because I think you've only edited one conversion script. Don't we compare the results of this conversion in some test that isn't the riscv-dv functional flow? If so, (and this is a genuine question!) why doesn't this change break things for that comparison?

udinator:
@rswarbrick good question!
After we get our Ibex trace logs from RTL simulation, the first step of the post-comparison stage is to create a CSV file that contains several bits of information - the instruction string itself, the operands of the instruction, and any register writes that may have occurred, among a few others.
The post-comparison with the ISS log mostly just looks at the register write information (since we base our comparison off of the assumption that Ibex and the ISS should both see the exact same sequence of register writes). On the other hand, the functional coverage flow does not need this information about register writes, as it is more concerned with the operands to each instruction, and thus will only look at that piece of each CSV entry (which is the bit I'm modifying with this PR).
As a result, while these changes are still present in the CSV file, the post-comparison step does not need this information, and will not be affected by this change.

rswarbrick:
Ok, that makes sense.

PR #839, opened 2020-05-04
[prim] Split out primitives used by icache

tomroberts-lowrisc:
- All primitives the icache uses are specified in distinct core files
  with names that match those existing (or about to exist) in OpenTitan
- When vendoring-in Ibex, none of those primitives need to be copied
  across, since OpenTitan will use its own versions
- Relates to lowRISC/opentitan/#1231


tomroberts-lowrisc:
Good point - maybe ``lowrisc::prim::secded`` since that's all it contains?

rswarbrick:
Sounds sensible to me.

rswarbrick:
Sounds sensible to me.

tomroberts-lowrisc:
Matching OT changes are on their way. Had forgotten about the target-dependent primitives - dummy core file added as suggested.

Issue #840, opened 2020-05-04
The contents of dv/uvm/data are not vendored properly from OpenTitan


rswarbrick:
These were copied from OpenTitan as part of the mega-commit at PR #711. Since they are copied into the main Ibex repo, rather than vendored in, they don't stay in-sync with changes there.

Recently (30th March), some of the keys for coverage generation changed (see OpenTitan commit 39ffebd). These don't now match and you can't run `dvsim.py` to generate coverage without hacking `dv/uvm/data/vcs/vcs.hjson`, at least.

I think we should probably be vendoring this in properly. Is it part of dvsim? In which case, maybe it needs moving around in the OT repo and getting pulled in with the existing rules. If not, we need (yet) another vendor rule and either way this code needs to live under `/vendor`.

udinator:
The files in `dv/uvm/data` are needed for dvsim, to specify project-wide variables and default values; same with the contents of `dv/uvm/data/vcs` as well.
It is definitely possible to vendor these into the Ibex repository - this will result in 3 additional vendor rules:
1) The `dv/uvm/data/common_project_cfg.hjson` in the Ibex repo lives at `hw/data/common_project_cfg.hjson` in OpenTitan.
2) The 4 `.hjson` files, `sim.mk`, `vcs/vcs.hjson`, and `xcelium/xcelium.hjson` that are all currently under Ibex's `dv/uvm/data` directory all live at `hw/dv/data` in OpenTitan.
3) `cover.cfg`, `vcs.fsdb`, and `xprop.cfg` that are currently under Ibex's `dv/uvm/data/vcs` directory all live at `hw/dv/tools/vcs` in OpenTitan.

It is also likely that by vendoring in all of these files, we will need to patch them (and update the patches) somewhat regularly to fix directory paths and things like this that will keep being added onto these configuration files.
WDYT?

imphil:
Vendoring is probably only half of a solution here.

Ideally, the dvsim tool would work with less repeated config -- all default config should be part of the tool itself (which we vendor in), and only deviations from that should be in our config files. That's especially true for all files @udinator mentioned in (2) above (e.g. `sim.mk`).

I'd take that as a medium-term goal, and for now simply keep the two configurations manually in sync. A git diff between the two data directories in OT and Ibex shows some deviations which seem to be intentional, and others which aren't. Maybe @udinator you can align them again as much as possible, and open a tracking issue to get this sorted on the dvsim side?

udinator:
[...]

PR #841, opened 2020-05-05
[dv] Manually update dvsim config files

udinator:
As per the discussion in #840, I've aligned the dvsim config files in `dv/uvm/data` as much as possible with their updated counterparts in the main OpenTitan repository.

### DIFF
--- a/dv/uvm/data/common_sim_cfg.hjson
+++ b/dv/uvm/data/common_sim_cfg.hjson
@@ -23,10 +23,11 @@
 
   // pass and fail patterns
   build_pass_patterns: []
-  build_fail_patterns: []
+  build_fail_patterns: ["^ERROR:.*$"] // fusesoc build error
   run_pass_patterns:   ["^TEST PASSED (UVM_)?CHECKS$"]
   run_fail_patterns:   ["^UVM_ERROR\\s[^:].*$",
                         "^UVM_FATAL\\s[^:].*$",
+                        "^UVM_WARNING\\s[^:].*$",
                         "^Assert failed: ",
                         "^\\s*Offending '.*'",
                         "^TEST FAILED (UVM_)?CHECKS$"]
@@ -42,6 +43,11 @@
   m: UVM_MEDIUM
   h: UVM_HIGH
   d: UVM_DEBUG
+
+  // Path to the dut instance (this is used in a couple of places such as coverage cfg
+  // file, xprop cfg file etc. If this is different for your block, then override it with
+  // the 'overrides' directive.
+  dut_instance: "{tb}.dut"
 
   // Default waves dump settings
   dump_base:  waves
@@ -69,6 +75,7 @@
     WAVES: {waves}
     DUT_TOP: {dut}
     TB_TOP: {tb}
+    dut_instance: {dut_instance}
   ]
 
   // Regressions are tests that can be grouped together and run in one shot
--- a/dv/uvm/data/vcs/vcs.hjson
+++ b/dv/uvm/data/vcs/vcs.hjson
@@ -37,7 +37,7 @@
                "+UVM_TEST_SEQ={uvm_test_seq}"]
 
   // Coverage related.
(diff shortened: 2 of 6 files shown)


udinator:
This force push removes the `sw_build` target from `sim.mk`, as we do not need this for icache simulations.

dv/uvm/data/xcelium/xcelium.hjson
```diff
@@ -7,49 +7,83 @@
 
   // Indicate the tool specific helper sources - these are copied over to the
   // {tool_srcs_dir} before running the simulation.
-  tool_srcs:  ["{proj_root}/dv/uvm/data/xcelium/*"]
+  tool_srcs:  ["{proj_root}/hw/dv/tools/xcelium/*"]
```
rswarbrick, line 10: This looks like it might not be quite right?

dv/uvm/data/common_sim_cfg.hjson
```diff
@@ -43,8 +44,13 @@
...
   d: UVM_DEBUG
 
+  // Path to the dut instance (this is used in a couple of places such as coverage cfg
+  // file, xprop cfg file etc. If this is different for your block, then override it with
+  // the 'overrides' directive.
+  dut_instance: "{tb}.dut"
+
   // Default waves dump settings
-  dump_base:  waves
+  dump_file:  waves.{dump}
```
rswarbrick, line 53: I'd rather keep the dump_base/dump_file stuff as was (see notes at vcs_fsdb.tcl).

rswarbrick:
For what it's worth, I've pushed up a PR with the dumping tidy-ups as https://github.com/lowRISC/opentitan/pull/2128.

dv/uvm/data/xcelium/xcelium.hjson
udinator: whoops you're right, good catch

udinator:
@rswarbrick - The missed xcelium directory path is an easy fix, I'll get that pushed up today.
It looks like the remaining comments are regarding the general wave dumping mechanism - since your PR to reflect the code in Ibex is currently under review in OpenTitan, should we wait until lowRISC/opentitan#2128 is merged to reflect those changes here? Otherwise we'll probably have to open a second PR to manually update these files.

rswarbrick:
Let's leave those files unchanged in Ibex until that PR has been merged. I'm also working on some code at the moment that will get rid of lots of the OpenTitan paths from the hjson files. I was hoping it would be done by today, but looks like it will extend into tomorrow. Once that's merged, I'm hoping that we'll genuinely be able to vendor in the code from (OpenTitan's) hw/dv/data directory. We'll have some path changes on the Ibex side, obviously, but that should all come out in the wash, I think.

udinator:
Sounds fine to me; I agree, hopefully we can properly vendor things in soon.
I'll fix the xcelium path and revert the vcs TCL files back to their original state later today (I'm currently attempting to remotely reboot my machine).

udinator:
The latest force pushes fix the Xcelium pathname issue, delete vcs_vpd.tcl, and revert vcs_fsdb.tcl back to its original state.

dv/uvm/data/vcs/vcs.hjson
```diff
@@ -31,13 +31,13 @@
                "-LDFLAGS \"-Wl,--no-as-needed\""]
 
   run_opts:   ["-licqueue",
-               "-ucli -do {tool_srcs_dir}/vcs_fsdb.tcl",
+               "-ucli -do {tool_srcs_dir}/vcs_{dump}.tcl",
```
rswarbrick, line 34: This (confusingly) needs to be `vcs_fsdb.tcl` at the moment.

udinator: ah yes good point

udinator:
Latest force push fixes the `vcs_fsdb.tcl` naming inside of `vcs/vcs.hjson`.

PR #842, opened 2020-05-05
[dv] enable writeback stage and branch ALU

udinator:
Pretty straightforward PR, all this does is to enable the writeback stage and the branch ALU in the Ibex testbench.


GregAC:
Yes I have been meaning to extend my multiple config work to the DV environment. I'll get that done this week.

In the meantime switching to BT ALU with Writeback stage as the single default config here seems sensible.

PR #844, opened 2020-05-06
[memutils] Make memutils work with wider RAMs

tomroberts-lowrisc:
- Using the RAM width in the DPI function definition causes Verilator
  compilation failures where there are multiple memory instances with
  different widths.
- With this patch, the width of the function is fixed at 128bits, and
  the memory loader can load any width of data (+ ECC bits) that it
  likes (up to 128bits total). The RAM on the SystemVerilog side will
  discard any extra unused bits.

### DIFF
--- a/dv/verilator/memutil/cpp/verilator_memutil.cc
+++ b/dv/verilator/memutil/cpp/verilator_memutil.cc
@@ -45,8 +45,8 @@ bool VerilatorMemUtil::RegisterMemoryArea(const std::string name,
                                           size_t width_bit) {
   MemArea mem = {.name = name, .location = location, .width_bit = width_bit};
 
-  assert((width_bit == 32 || width_bit == 64) &&
-         "TODO: Check if width other than 32 and 64 works as expected.");
+  assert((width_bit <= 128) &&
+         "TODO: Memory loading only supported up to 128 bits.");
 
   auto ret = mem_register_.emplace(name, mem);
   if (ret.second == false) {
@@ -397,9 +397,9 @@ bool VerilatorMemUtil::MemWrite(const MemArea &m, const std::string &filepath,
     return false;
   }
 
-  if ((m.width_bit % 32) != 0) {
+  if ((m.width_bit % 8) != 0) {
     std::cerr << "ERROR: width for: " << m.name
-              << "must be a multiple of 32 (was : " << m.width_bit << ")"
+              << "must be a multiple of 8 (was : " << m.width_bit << ")"
               << std::endl;
     return false;
   }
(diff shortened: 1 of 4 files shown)


tjaychen:
this lgtm, i'm anxious to try it out! thanks!

tjaychen:
i'm trying this locally right now, and the compilation breaks at the moment because rom also overloads the same function. Do you guys think it makes sense, since the interface has to be fixed, to refactor the common verilator code out into its own module? Then the other memories can just instantiate the common module?

tomroberts-lowrisc:
Hi Tim
If we go with this option, we'll update the ROM to use the same interface when the new changes are vendored in from Ibex. Agreed, pulling this stuff into it's own common module might be a good plan.

imphil:
Yes, a common module was something I was looking into as well. But first I want to be really sure that we can't make it work without the 128 bit width limitation; it currently looks that there might be a way; I'll get back to you soonish.

imphil:
The 128 bit change looks good for now, I have an idea how we could get around that, but that's for another day. But we need to get out of this rabbit hole of code duplication and half-vendored files between OT and Ibex. The duplication of the same code in prim_ram, prim_rom, ram_1p and ram_2p is really killing us to make improvements to this code. 

* To get rid of that duplication, I've started in the OT repository. Let's get that code merged there first and then copy over the resulting primitives to Ibex.
  * https://github.com/lowRISC/opentitan/pull/2205 factors out the memloading functionality.
  * https://github.com/lowRISC/opentitan/pull/2202 adds a write mask to the 2p RAM, which we need if we want to replace the ram_2p in Ibex.
* Adjust Ibex to remove the ram_1p and ram_2p modules. I have code around that makes them a simple wrapper:
  * https://github.com/lowRISC/ibex/pull/851 is doing that for the 1p RAM
  * Another PR will follow for the 2p RAM as soon as we have wmask support in OT.
* Once that all is fixed, we can take this PR without fear of running into issues where we have diverging function signatures for the `sim_memload` etc.


Slightly orthogonal improvements (help to get rid of the ram_1p/ram_2p files completely in Ibex):
* https://github.com/lowRISC/opentitan/issues/2201
* https://github.com/lowRISC/opentitan/issues/2181

imphil:
And of course there's one circular dependency: Making prim_ram_2p loadable (which we need to port over the ram_2p to use that) requires the Verilator memutil changes which are in this PR first, since OT uses 2p memories with odd widths (for ECC). So these things need done at the same time. The code is at https://github.com/imphil/opentitan/commit/24fcf016d40ac35a13a23a2975e0a617b0e8d018

tjaychen:
hey all, is there any plan to merge this PR in its current form?

imphil:
Yes, as soon as we got the dependencies sorted as outlined in https://github.com/lowRISC/ibex/pull/844#issuecomment-627368445. It should be almost there with the last batch of changes which went into OT, I'll have a look tomorrow.

tjaychen:
thanks @imphil !

imphil:
We took this change through https://github.com/lowRISC/opentitan/pull/2311.

PR #848, opened 2020-05-07
[dv] Remove duplicate lowrisc:constants:top_pkg

wallento:
The core file and SV package were copied in here, which breaks the
OpenTitan build. The constants file is provided there too, but a
recent builds will pick up this one here. This is incorrect.

As the core file seems to be unused anyways, remove it from here along
with the package.

Added the, as far as I can see, only used constant here for
simplicity, alternatively it is an option to properly depend on the
other core.

Not tested because of tool availability.


rswarbrick:
Can we do this a different way? Add a vendoring rule on the import into OpenTitan that throws away these files (and possibly a much bigger subtree?)

In the reasonably near future, I hope that this code (which was copied from OpenTitan) will be vendored properly. In which case, it will be in a vendor subdirectory in Ibex, which will be easy to exclude when pulling things in in the other direction.

See issue #840 for me grumbling and https://github.com/lowRISC/opentitan/pull/2141 for the work on the OpenTitan side needed for this to happen. Modulo stalled reviews, I'd hope this will be sorted in the next week.

imphil:
I guess we can close this one; OpenTitan doesn't vendor this file in any more, so there's only one copy around in OT.

PR #851, opened 2020-05-08
Use prim_generic_ram_1p in ram_1p

imphil:
ram_1p is almost a copy of the single-port RAM primitive we have in
OpenTitan, called prim_ram_1p, with its generic implementation
prim_generic_ram_1p. Instead of having a copy of that file in Ibex,
consistently use the OpenTitan one.

Unfortunately, ram_1p has slightly different semantics around some
signals, especially rvalid. This commit adjusts the meanings of the
signals for now, since I don't have a way to test the Arty board
which also uses this primitive (together with the compliance test
suite). With the testing in the compliance suite I'm reasonably certain
that the Arty board will work as well.

### DIFF
--- a/dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
+++ b/dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
@@ -39,8 +39,10 @@ module ibex_riscv_compliance (
     TestUtilDevice
   } bus_device_e;
 
-  localparam NrDevices = 2;
-  localparam NrHosts = 3;
+  localparam int unsigned NrDevices = 2;
+  localparam int unsigned NrHosts = 3;
+  // 64 kB RAM. Must be a power of 2. Check bus configuration below when changing.
+  localparam int unsigned RamSizeWords = 64*1024/4;
 
   // host and device signals
   logic           host_req    [NrHosts];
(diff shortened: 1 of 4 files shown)


imphil:
This is in preparation to get rid of all RAM primitives other than the prim_* ones, together with their slightly customized copies of the Verilator memory loading functionality. The ram_2p is next in a separate PR.

dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
```diff
@@ -39,8 +39,9 @@ module ibex_riscv_compliance (
     TestUtilDevice
   } bus_device_e;
 
-  localparam NrDevices = 2;
-  localparam NrHosts = 3;
+  localparam int unsigned NrDevices = 2;
+  localparam int unsigned NrHosts = 3;
+  localparam int unsigned RamSizeWords = 64*1024/4; // 64 kB
```
rswarbrick, line 44: Nit: Does this need a comment mentioning that it needs to be a power of 2? (In order for `cfg_device_addr_mask` to work below)

imphil: No, the address space will always be a power of two, but the RAM itself can be smaller.

rswarbrick: In which case, I think you need a $clog2 in the definition of `cfg_device_addr_mask`?

imphil: Oh indeed, good catch. Added a comment to that effect, doing it all dynamically is probably too fragile for little gain.

imphil:
Merged this as a reviewed starting point, it will change again with an upcoming PR.

PR #854, opened 2020-05-11
[rtl] Fix instr_valid_i exception issue

### LINKED ISSUE #849: PMP store fault does not jump to exception handler
Hi Tom,
I ran into this situation when debugging a PMP test (with icache disabled):
Screenshot from 2020-05-07 15-46-15
The instruction at `PC=0x8001fb0a` (under the cursor) is a store instruction to address `0x80031034`, which the PMP detection logic correctly flags as an error for not matching any pmp_cfg CSR in user-mode and prevents a request from being sent to the memory.
During this, the controller is in the `DECODE` state, but it does not actually detect `special_req_all` going high because `instr_valid_i` has already gone low by the time it reflects the rising `exc_req_lsu` wire - as a result the core pretends everything is okay and goes on to execute the instruction at the next PC.
I suspect this has to do with how `instr_valid_id_d/q` are set in the IF stage, as they appear to be dropping one cycle too early - additionally, when I enabled the icache and reran this test, this issue did not appear, e.g. `instr_valid_i` in the controller went high at the same time as `special_req_all` to qualify it due to `instr_valid_id_d/q` staying high for an additional cycle:
Screenshot from 2020-05-07 16-10-02
We can see that with the icache, `instr_valid_id_d/q` stay high for extra cycles, allowing the controller to correctly qualify the PMP exception and then start fetching the address of the exception handler (`0x80001400`).

tomroberts-lowrisc:
- The controller state machine could only progress to FLUSH to handle an
  exception if instr_valid_i was set
- When the exception comes from a load/store in the Writeback stage, and
  no new instruction has been driven into the ID stage, this could cause
  exception to be missed
- The instr_valid_i qualification is therefore removed from the state
  machine as all relevant signals inside that if block are already
  qualified by instr_valid_i anyway
- Fixes #849

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -182,12 +182,14 @@ module ibex_controller #(
   // it is needed to break the path from ibex_cs_registers/illegal_csr_insn_o
   // to pc_set_o.  Clear when controller is in FLUSH so it won't remain set
   // once illegal instruction is handled.
+  // All terms in this expression are qualified by instr_valid_i
   assign illegal_insn_d = (illegal_insn_i | illegal_dret | illegal_umode) & (ctrl_fsm_cs != FLUSH);
 
   // exception requests
   // requests are flopped in exc_req_q.  This is cleared when controller is in
   // the FLUSH state so the cycle following exc_req_q won't remain set for an
   // exception request that has just been handled.
+  // All terms in this expression are qualified by instr_valid_i
   assign exc_req_d = (ecall_insn | ebrk_insn | illegal_insn_d | instr_fetch_err) &
                      (ctrl_fsm_cs != FLUSH);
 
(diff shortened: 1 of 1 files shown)


udinator:
Confirming that this change fixes the issue I reported in #849.

rtl/ibex_controller.sv
```diff
@@ -399,31 +404,29 @@ module ibex_controller #(
...
+          // Halt IF but don't flush ID. This leaves a valid instruction in
+          // ID so controller can determine appropriate action in the
+          // FLUSH state.
+          ctrl_fsm_ns = FLUSH;
+          halt_if     = 1'b1;
+        end
 
-          if ((branch_set_i || jump_set_i) && ~special_req_branch) begin
-            pc_set_o       = 1'b1;
+        if ((branch_set_i || jump_set_i) && ~special_req_branch) begin
```
vogelpi, line 417: I think `branch_set_i` and `jump_set_i` are not qualified by `instr_valid_i`. I am not sure if it is a problem for the branch, but for jumps, it is possible that `pc_set_o` is now set for multiple cycles (if the BT ALU is enabled).

vogelpi: Sorry, my comment above is wrong. The FSM setting `jump_set_i` based on `jump_set_dec` is qualified with `instr_executing` and thus with `instr_valid_i`. So this should be fine.

PR #858, opened 2020-05-11
Fix two bugs in the icache <-> core protocol checker

rswarbrick:
Tom, could you have a look at the 2nd patch in particular, to check that it matches what you think should be going on?

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
@@ -88,7 +88,7 @@ interface ibex_icache_core_protocol_checker (
     end
   end
 
-  `ASSERT(NoFetchWithoutAddr, req |-> (branch | has_addr), clk, !rst_n)
+  `ASSERT(NoFetchWithoutAddr, ready |-> (branch | has_addr), clk, !rst_n)
 
   // The err_plus2 signal means "this error was caused by the upper two bytes" and is only read when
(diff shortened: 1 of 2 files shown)


dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
```diff
@@ -88,7 +67,35 @@ interface ibex_icache_core_protocol_checker (
...
+  `ASSERT(ValidUntilReady, unfinished_valid |=> valid,                clk, !rst_n);
+  `ASSERT(AddrStable,      unfinished_valid |=> $stable(addr),        clk, !rst_n);
+  `ASSERT(ErrStable,       unfinished_valid |=> $stable(err),         clk, !rst_n);
+  `ASSERT(Err2Stable,      unfinished_valid |=> $stable(err_plus2),   clk, !rst_n);
+
+  `ASSERT(LoRDataStable,
+          unfinished_valid |=> $stable(rdata[15:0]),
+          clk, !rst_n);
+  `ASSERT(HiRDataStable,
+          unfinished_valid & (rdata[1:0] == 2'b11) |=> $stable(rdata[15:0]),
```
tomroberts-lowrisc, line 97: Should this be looking at bits [32:16] instead?

rswarbrick: Erk. Yes! I'll push up a new version with that fix in a sec.

Issue #859, opened 2020-05-11
[dv/makefile] Same seed used for each iteration of Ibex regressions


udinator:
Hi,
I've been seeing an issue with Ibex regressions where if I run a test (say riscv_debug_basic_test) 10 times, each of the ten iterations of this test all use the exact same seed since it is now generated in the top level makefile and passed down to every simulation iteration.
This is totally fine for sanity regressions where each test is run exactly once, but for large regressions (nightly, etc...), having every iteration of a test use the exact same seed runs counter to the point of having large numbers of tests in regressions to begin with.
Ideally, if I specify `ITERATIONS=10` on the command line, I expect that each of the ten iterations is using different seeds - I'm not quite sure if having the Makefile generate a top level seed used for every iteration would be the best approach here.
Could you please take a look?
Thanks!

rswarbrick:
Hmm. You're right this doesn't quite make sense. I've pushed a fix at instruction generation time here: https://github.com/google/riscv-dv/pull/570. The sim-time fix is in PR #861.

rswarbrick:
Tao wasn't keen on defining the combination of `--seed` and `--iterations` > 0 in riscv-dv, so I've amended that patch to use a new flag, `--seed_start`. This will work for us, but needs a minor tweak to how we call riscv-dv when the new code is vendored in. Assuming that https://github.com/google/riscv-dv/pull/570 gets merged in something like it's present form, we'll want a patch like the attached (renamed to .txt because GitHub doesn't allow files called .patch, sigh).

[0001-Pass-start_seed-rather-than-seed-to-riscv-dv-s-run.p.txt](https://github.com/lowRISC/ibex/files/4633553/0001-Pass-start_seed-rather-than-seed-to-riscv-dv-s-run.p.txt)

rswarbrick:
Oh, and looking again at what you originally wrote, @udinator, my idea is that we'll pick a starting seed (either at random, or given by the SEED argument to Make). We then run tests with seeds `{SEED, SEED+1, ..., SEED+num_iters}`. Assuming the PRNG isn't completely terrible, this has the same "randomness" as running `num_iters` arbitrary tests, but it lets us make sure the seeds we use in riscv-dv match the seeds we use in the simulator.

Modulo the problem with tests within a batch (see note at bottom of the riscv-dv PR), I think the PRs I've pushed should get everything working for the "Generate 1000 tests with random seeds" nightly use-case.

udinator:
[...]

PR #861, opened 2020-05-12
Make --seed argument more helpful in core_ibex/sim.py

rswarbrick:
Now, successive test iterations use successive seeds. So if you pass
`--seed 123` and run ten iterations, they will run with seeds 123, 124,
... through 133.


udinator:
Adding @taoliug to take a look at this PR, since it pretty closely relates to google/riscv-dv#570

udinator:
I think Tao mentioned this in the riscv-dv PR, but I alos think it's better to randomize the seeds instead of fixing them at top level (and incrementing by 1 for every iteration) - could you implement something similar to how riscv-dv creates a `seed.yaml` file to allow rerunning full regressions?
This would allow full randomization of the seeds used to not limit the size of the explored state space while also allowing regressions to be rerun.

rswarbrick:
@udinator I don't think I understand this comment. Assuming you have a half-decent PRNG, reading 100 seeds from a `seed.yaml` has the same "randomness" as picking a starting seed and running 100 seeds starting from there. The latter option means I can specify my regression run with a single number. The former requires me to generate (and check in) a YAML file.

As I assume you understand, whether I specify seeds or not is basically the difference between soak testing and regression testing. Clearly, we need to support both, and we do (don't specify a seed for soak testing; generate a YAML file for regression testing).

However, our tools currently do something silly if you ask for a seed and more than one test (which is what your issue #859 reports). We should either throw an error ("Hey, this doesn't make sense. Please write a list of numbers in a file for me.") or, better, do something sensible. That's what these patches do.

rswarbrick:
Right. I've put the new behaviour in a differently named argument and this PR now contains the vendor patch that pulls in the new riscv-dv code.

PR #869, opened 2020-05-13
[dv] Enable use of ibex configs in DV

GregAC:
This allows you to set an Ibex config to run DV with. There is still one issue in that there's tests that only work/make sense with certain configs, e.g. you don't want to run the bitmanip tests if you haven't got a config with bitmanip on. We need to consider how to alter the test list and general DV parameters (i.e. don't allow bitmanip instruction gen) based on the config which this PR doesn't solve.

So whilst in abstract it allows you to run DV on multiple configs currently if you try any of the other configs (which don't have bitmanip) it'll fail due to the bitmanip test failing and maybe others if they can also produce bitmanip instructions.

The slightly odd python script to generate a makefile fragment purely for simulator options exists so we can get a nice clean failure if there's an issue with the config. Otherwise you're using `$(shell ...)` inside the makefile to get the result of running `ibex_config.py` and there's no simple way to get that to fail nicely. You just get a silent lack of config setting and will run with the default parameters without knowing.


GregAC:
Currently CI failure is because I need to include the PMP options in the various .core files so fusesoc can understand them, will sort that out tomorrow.

dv/uvm/core_ibex/Makefile
```diff
@@ -132,6 +124,25 @@ gen-dirs := $(OUT) $(OUT-SEED) $(metadata) $(OUT)/rtl_sim
 $(gen-dirs): %:
 	mkdir -p $@
 
+# sim-cfg-mk is a makefile fragment that sets-up anything simulator specific, it
+# is generated by sim_makefrag_gen.py
+sim-cfg-mk = $(OUT)/.sim-cfg.mk
+
+# The FORCE pre-requisite causes sim-cfg-mk to always be built, however this
+# also triggers a makefile restart. The MAKE_RESTARTS guard around this target
+# ensures this doesn't result in an infinite recursion.
```
rswarbrick, line 133: For the next person trying to get this working in their Makefile, maybe we should be a bit more helpful (it took us quite a few goes!). Maybe something like this?

```make
# The include of $(sim-cfg-mk) below tells Make that it should ensure the file
# exists. This rule tells Make how to build it. We also want to ensure it's
# built on every run. Declaring the rule to be .PHONY doesn't work because it
# causes Make to no longer think the rule actually generates the file. If you
# do that, the file gets re-built, but Make doesn't read the new contents. So
# instead we depend on FORCE (a phony target). This ensures a rebuild, but also
# causes an infinite recursion: when we re-read this file to include the new
# contents of $(sim-cfg-mk), we run the rule again. To avoid that, we check for
# MAKE_RESTARTS, which is defined on re-runs. Phew!
```

dv/uvm/core_ibex/sim_makefrag_gen.py
```diff
@@ -0,0 +1,108 @@
...
+
+    args = argparser.parse_args()
+
+    global ibex_config_path
+    global ibex_config_name
+    global ibex_config_filename
+
+    ibex_config_path = os.path.join(args.ibex_top, 'util/ibex_config.py')
+    ibex_config_name = args.config
+    ibex_config_filename = os.path.join(args.ibex_top, 'ibex_configs.yaml')
```
rswarbrick, line 96: Rather than using globals, wouldn't it be cleaner to pass these to `sim_fns[...]` as arguments?

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -26,42 +26,63 @@ module core_ibex_tb_top;
...
+  parameter     MultiplierImplementation = `"`IBEX_CFG_MultiplierImplementation`";
+
   ibex_core_tracing #(
-    .DmHaltAddr(`BOOT_ADDR + 'h0),
-    .DmExceptionAddr(`BOOT_ADDR + 'h4),
-    .PMPEnable(1'b1),
-    .BranchTargetALU(1'b1),
-    .WritebackStage(1'b1),
-    .RV32B(1'b1)
+    .DmHaltAddr               ( `BOOT_ADDR + 'h0         ),
```
rswarbrick, line 46: Does our style guide allow the leading space here? (I don't have strong feelings, but just trying to avoid style-guide-fix code churn)

util/ibex_config.py
```diff
@@ -152,28 +215,40 @@ def get_config_file_location():
 
 
 def main():
-    config_outputs = {'fusesoc_opts': _config_dict_to_fusesoc_opts}
+    outputters = [
```
rswarbrick, line 218: This seems reasonable enough, but have you considered doing it "the other way", so the command line would specify a tool (vcs, riviera or fusesoc) and an action (build, simulate)? That feels like it might be cleaner for command line handling, and would probably be nicer to extend with new tools in the future.

If I've missed something and this is a horrible idea, no worries!

util/ibex_config.py
```diff
@@ -127,19 +127,82 @@ def get_config_dicts(config_file):
...
+            'fusesoc_opts', help=('Outputs options for fusesoc'))
+        output_argparser.set_defaults(output_fn=self.output)
+
+    def output(self, config_dict, args):
+        fusesoc_cmd = []
+        for parameter, value in config_dict.items():
+            if isinstance(value, bool):
+                # For fusesoc boolean parameter are set to true if given on the
+                # command line otherwise false, it doesn't support an explicit
+                # --param=True style
```
rswarbrick, line 142: Nit: "parameters", then the comma before "it" should be a full stop or semicolon.

util/ibex_config.py
```diff
@@ -127,19 +127,82 @@ def get_config_dicts(config_file):
...
+            ins_hier_path = ''
+
+        sim_opts = []
+
+        for parameter, value in config_dict.items():
+            if isinstance(value, str):
+                parameter_define = args.string_define_prefix + parameter
+                define_set_str = self.define_set_fn(parameter_define, value)
+
+                if define_set_str:
```
rswarbrick, line 191: I think this is always true with the code as it stands, so the `else` branch is dead code. Remove the check?

util/ibex_config.py
```diff
@@ -152,28 +215,40 @@ def get_config_file_location():
...
-    config_outputs = {'fusesoc_opts': _config_dict_to_fusesoc_opts}
+    outputters = [
+        FusesocOpts(),
+        SimOpts('vcs_opts', 'VCS compile',
+                lambda p, v: '-pvalue+' + p + '=' + v,
+                lambda d, v: '+define+' + d + '=' + v, '.'),
+        SimOpts('riviera_sim_opts', 'Riviera simulate',
+                lambda p, v: '-g/' + p + '=' + v, lambda d, v: None, '/'),
+        SimOpts('riviera_compile_opts', 'Riviera compile', lambda p, v: None,
+                lambda d, v: '+define+' + d + '=' + v, '/')
```
rswarbrick, line 226: This is quite hard to parse by eye because of the commas in the lambda syntax. Probably better with newlines between arguments.

dv/uvm/core_ibex/sim_makefrag_gen.py
GregAC: I went this way because you otherwise end up passing three ibex_config_X arguments down to the sim_fns which in turn pass them to run_ibex_config. The sim_fns themselves shouldn't care about the arguments, they're just opaque things to hand to run_ibex_config so it'd be good to avoid passing them to the sim_fns altogether. I could wrap them up in a class/tuple so it's one argument to pass through rather than one but felt this was just a cleaner way which avoids extra arguments. Given it's a small utility script the ugly use of globals for this seemed fine.

util/ibex_config.py
GregAC: I shied away from that mostly because for fusesoc and vcs there are no multiple actions, it's only riviera that needs the two different option sets. I felt it would just complicate matters without really adding anything.

util/ibex_config.py
GregAC: riviera_sim_opts gives None for the define string and riviera_compile_opts gives None for the parameter string so this is used.

Initially I just returned an empty string to avoid this handling but sadly the `shlex.quote` turns the empty string into a literal ''.

util/ibex_config.py
GregAC: I blame the pylint auto fix

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
GregAC: I had intially thought it specified that you had to use this style, looks like it doesn't and indeed doesn't specify a particular style for port/parameter lists in module instantiation, however this is the style used in the Ibex RTL.

I certainly prefer some vertical alignment to port names and port connections are nicely split and aligned, don't really mind about the spacing but I'd prefer to leave this as is now I've done it if there's no strong feeling either way.

util/ibex_config.py
rswarbrick: I'm not convinced this is simpler, but if you prefer it, no problem.

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
rswarbrick: Ok cool. (In Emacs, there are rectangle commands for tidying this sort of thing up really easily: `C-x r k` here; I guess your editor probably has something similar).

GregAC:
Ah sim_makefrag_gen.py should not have disappeared, failed to add it again when I split this up into two commits, should be back now.

GregAC:
@udinator any comments?

udinator:
sorry for the delayed review! 

LGTM, I think this looks good as an initial step.

One concern that I see with the current state of this PR is that I can still modify the `PMP_GRANULARITY` and `PMP_REGIONS` parameters in the Makefile, and while it currently will not affect the RTL as sims now read the config values from `ibex_config.yaml`, the riscv-dv generator will still pick up the modified PMP config parameters and result in failures.
However, as you mention in your very first comment, I agree that this should wait for a later PR just to get basic functionality merged.

GregAC:
> One concern that I see with the current state of this PR is that I can still modify the PMP_GRANULARITY and PMP_REGIONS parameters in the Makefile

Ah yes good point, it should be easy enough to set those from the config instead, but I'll await working on how we're going to deal with the config altering the RISCV DV tests run as a whole first.

Issue #866, opened 2020-05-13
Deal with new `sext.b/h` Zbb instructions


vogelpi:
As mentioned in #863 , we recently vendored in a riscv-dv version from upstream that features two new instructions added in the [latest working draft version of the bitmanip extension](https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-draft.pdf) (no version number yet).

These two instructions are not part of the version 0.92 draft implemented in Ibex. As a consequence, the Zbb tests started to fail on master. To work around this, we can either:
1. Implement the new instructions, i.e., go beyond the version 0.92 draft. We might need to do this at some point anyway.
2. Patch the vendored in RISCV-DV files to not emit those instructions, i.e., stay at the version 0.92 draft.

I would prefer Option 2 at least until the latest working draft gets a version number.

ganoam:
The spec version 0.92 does include the sext instructions. They are even explicitly mentioned in the change history. But they are not categorized in the table on page 3. I think this is an oversight, and they should really have been part of zbb already. 

I think, if riscv-dv chose to include it already, it would make more sense for us to do it too. They are pretty simple to implement.

Sidenote: it's not the Zbb tests that fail, but Zbe, as RISCV-DV have categorized them in Zbe for now. However they should probably be moved to Zbb as in the latest working draft.

vogelpi:
Ah, I see thanks for clarifying that. In this case, it indeed makes sense to implement them at some point.

I will get in touch with DV folks to move them to Zbb. I don't see why they should be part of Zbe.

vogelpi:
Update: this will be solved upstream by creating a new Zb_tmp instruction group for the two instructions, see google/riscv-dv#573.

ganoam:
This has been merged in #870

PR #863, opened 2020-05-12
[bitmanip] Add ZBE Instruction Group

ganoam:
This commit implements the Bit Manipulation Extension ZBE instruction
group: bext (bit extract) and bdep (bit deposit).

Architectural details:

        * bext/bdep: A new butterfly and inverse butterfly network is
        implemented. The generation of its controlbits depend on a
        parallel prefix bitcount of the deposit / extract mask.

        * bitcounter: The path for bext / bdep instructions traverses
        the bit counter and the butterfly network, resulting in both a
        larger delay and area. To mitigate the bitcounter has been
        changed from a serial bit counter to a radix-2 tree structure.

        * grev/gorc: Zbp instructions general reverse and general
        or-combine have as of yet shared the shifters reversal
        structure. It has proven benefitial to area and timing to reuse
        the novel butterfly network instead

The butterfly network itself consumes ~3.5kGE and ~1.1kGE for synthesis
with tight and relaxed timing constraints respectively. Including the
optimizations of the bitcounter and grev/gorc, the overall change in
area consumption is +4.6kGE (+1.2kGE) and +3.3kGE (+1.1kGE) for
synthesis with tight (relaxed) timing constraints for 2- and 3-stage
configurations respectively. For tight timing constraints that is a
growth by around ~10%, for relaxed ~5%.

The impact on the maximum frequency is negligable.

Signed-off-by: ganoam <gnoam@live.com>

PR #870, opened 2020-05-13
[bitmanip] Add sext.b/h instructions

ganoam:
This commit implements the Bit Manipulation Extension sign-extend
instructions: sext.b (sign-extend byte) and sext.h (sign-extend half
word).

The implementation is basically a one-liner, duplicating the msb of the
byte / half-word into the msb of the output register.

### DIFF
--- a/rtl/ibex_alu.sv
+++ b/rtl/ibex_alu.sv
@@ -376,6 +376,7 @@ module ibex_alu #(
   logic [31:0] minmax_result;
   logic [5:0]  bitcnt_result;
   logic [31:0] pack_result;
+  logic [31:0] sext_result;
   logic [31:0] multicycle_result;
   logic [31:0] singlebit_result;
 
@@ -860,11 +861,20 @@ module ibex_alu #(
         default: pack_result = {operand_b_i[15:0], operand_a_i[15:0]};
       endcase
     end
+
+    //////////
+    // Sext //
+    //////////
+
+    assign sext_result = (operator_i == ALU_SEXTB) ?
+        { {24{operand_a_i[7]}}, operand_a_i[7:0]} : { {16{operand_a_i[15]}}, operand_a_i[15:0]};
+
   end else begin : g_no_alu_rvb
     // RV32B result signals
     assign minmax_result       = '0;
     assign bitcnt_result       = '0;
     assign pack_result         = '0;
+    assign sext_result         = '0;
     assign multicycle_result   = '0;
     assign singlebit_result    = '0;
     assign shuffle_result      = '0;
@@ -894,7 +904,7 @@ module ibex_alu #(
       // Shift Operations
       ALU_SLL,  ALU_SRL,
       ALU_SRA,
-      // RV32B Ops
+      // RV32B
       ALU_SLO,  ALU_SRO: result_o = shift_result;
 
       // Shuffle Operations (RV32B)
@@ -918,6 +928,9 @@ module ibex_alu #(
       ALU_PACK, ALU_PACKH,
       ALU_PACKU: result_o = pack_result;
 
+      // Sign-Extend (RV32B)
+      ALU_SEXTB, ALU_SEXTH: result_o = sext_result;
+
       // Ternary Bitmanip Operations (RV32B)
       ALU_CMIX, ALU_CMOV,
       ALU_FSL,  ALU_FSR,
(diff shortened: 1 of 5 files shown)


ganoam:
`sext.b` and `sext.h` are part of the bitmanip spec version 0.92, but are not categorized. This is most likely an oversight. The implementation of this instruction has been discussed in #866 

I am leaving this as a draft until #863 is merged: Since riscv-dv categorizes those instructions in zbe, I have so far not been able to test it without `bext` / `bdep`.

rtl/ibex_tracer_pkg.sv
```diff
@@ -74,12 +74,19 @@ parameter logic [31:0] INSN_PMULHU  = { 7'b0000001, 10'b?, 3'b011, 5'b?, {OPCODE
...
+parameter logic [31:0] INSN_SLOI  = { 5'b00100        , 12'b?, 3'b001, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_SROI  = { 5'b00100        , 12'b?, 3'b101, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_RORI  = { 5'b01100        , 12'b?, 3'b101, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_CLZ   = { 12'b011000000000, 5'b? , 3'b001, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_CTZ   = { 12'b011000000001, 5'b? , 3'b001, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_PCNT  = { 12'b011000000010, 5'b? , 3'b001, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_SEXTB = { 12'b011000000100, 5'b? , 3'b001, 5'b?, {OPCODE_OP_IMM} };
+parameter logic [31:0] INSN_SEXTH = { 12'b011000000101, 5'b? , 3'b001, 5'b?, {OPCODE_OP_IMM} };
+// sext -- pseudoinstruction: andi rd, rs 255
+parameter logic [31:0] INSN_ZEXTB   = { 4'b0000, 8'b11111111, 5'b?, 3'b111, 5'b?, {OPCODE_OP_IMM} };
```
vogelpi, line 86: Nit: check alignment around equal operator.

rtl/ibex_tracer.sv
```diff
@@ -832,7 +832,12 @@ module ibex_tracer (
         INSN_SLTIU:      decode_i_insn("sltiu");
         INSN_XORI:       decode_i_insn("xori");
         INSN_ORI:        decode_i_insn("ori");
-        INSN_ANDI:       decode_i_insn("andi");
+        INSN_ANDI:begin
+          casez (rvfi_insn)
+            INSN_ZEXTB:  decode_r1_insn("zext.b");
```
vogelpi, line 837: I have a question regarding these pseudo-ops zext.b/h: with this change, the tracer log will contain a zext.b/h but will also the assembler (to what we compare the trace log) contain the zext.b/h or will the assembler contain the actual instruction andi/pack?

ganoam: That is a very good question. The gorc-, grev- and shfl pseudo instructions are defined [here](https://github.com/riscv/riscv-bitmanip/blob/master/cproofs/rvintrin.h). I have previously used this file to emit them. I can not find the zext instructions there. Also, spike logs them as `pack` and `andi`, respectively when executing the DV bitmanip-test.
I assume that at least for now the assembler code will not contain zext instructions. Shall I remove them from the tracer?

vogelpi: Since the things are already in the tracer and we might just activate it at some point, I would leave the previous version in the diff (just the andi), comment out the newly added part in the diff and put a comment above explaining that at the moment instruction simulators don't output these pseudo-instructions.

rtl/ibex_tracer_pkg.sv
ganoam: done, thanks.

rtl/ibex_tracer.sv
ganoam: alright. Done, thank you.

rtl/ibex_alu.sv
```diff
@@ -918,7 +928,10 @@ module ibex_alu #(
       ALU_PACK, ALU_PACKH,
       ALU_PACKU: result_o = pack_result;
 
-      // Ternary Bitmanip Operations (RV32B)
+      // Sign-Extend (RV32B)
+      ALU_SEXTB, ALU_SEXTH: result_o = sext_result;
+
+      // Ternary Bitmanip Operations (RV32B Ops)
```
vogelpi, line 934: Nit: (RV32B Ops) -> (RV32B)
to be inline with the recent changes you did.

ganoam: oups... done, thanks.

PR #874, opened 2020-05-15
[bitmanip] Add ZBC instruction group

ganoam:
This commit implements the Bit Manipulation Extension ZBC instruction
group: clmul[rh] (carry-less multiply [reverse][high])

Carry-less multiplication can be understood as multiplication based on
the addition interpreted as the bit-wise xor operation.

Example: 1101 X 1011 = 1111111:

      1011 X 1101
      -----------
             1101
        xor 1101
        ---------
            10111
       xor 0000
       ----------
           010111
      xor 1101
      -----------
          1111111

Architectural details:
        
        A 32 x 32-bit array
        `[ operand_b[i] ? (operand_a << i) : '0 for i in 0 ... 31 ]`
        is generated. The entries of the array are pairwise 'xor-ed'
        together in a 5-stage binary tree.

The area increase when synthesized with relaxed timing constraints is
1.6-1.7kGE.

Timing figures are improve by 0.1 ns for the 3-stage configuration and
worsen by 0.04ns for the 2-stage implementation. This suggests
fluctuations due to the heuristic nature of the synthesis tools.

### DIFF
--- a/rtl/ibex_pkg.sv
+++ b/rtl/ibex_pkg.sv
@@ -115,7 +115,13 @@ typedef enum logic [5:0] {
 
   // Bit Field Place
   // RV32B
-  ALU_BFP
+  ALU_BFP,
+
+  // Carry-less Multiply
+  // RV32B
+  ALU_CLMUL,
+  ALU_CLMULR,
+  ALU_CLMULH
 } alu_op_e;
 
 typedef enum logic [1:0] {
--- a/doc/instruction_decode_execute.rst
+++ b/doc/instruction_decode_execute.rst
@@ -66,7 +66,7 @@ Other blocks use the ALU for the following tasks:
 
 Support for the RISC-V Bitmanipulation Extension (Document Version 0.92, November 8, 2019) is enabled via the parameter ``RV32B``.
 This feature is *EXPERIMENTAL* and the details of its impact are not yet documented here.
-Currently the Zbb, Zbs, Zbp, Zbe, Zbf and Zbt sub-extensions are implemented.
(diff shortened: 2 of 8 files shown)


ganoam:
Synthesis results are as always avaliable [here](https://docs.google.com/spreadsheets/d/1dDDtzMgS5quMXtv-8z2uD--jBIHMSnaWrqj1EIDp1Nw/edit#gid=1192706963)  

The clmul instruction naturally consumes a lot of logic area: the first and-stage comprises about 32 x 32 / 2 and-gates. The subsequent xor-stages comprise another 31 x 32 xor-gates. However, a multi-cycle implementation could possibly mitigate this. I will look into that in my final round of optimization.

rtl/ibex_pkg.sv
```diff
@@ -115,7 +115,14 @@ typedef enum logic [5:0] {
...
   // RV32B
-  ALU_BFP
+  ALU_BFP,
+
+  // Carry-less Multiply
+  // RV32B
+  ALU_CLMUL,
+  ALU_CLMULR,
+  ALU_CLMULH
+
```
vogelpi, line 125: Nit: remove this newline.

ganoam:
Thanks @vogelpi. Well, I think the main thing is to move things away from the critical path. If we look at the synthesis results, we see that the difference in area consumption is at about 5.4 kGE for relaxed timing constraints. But for tight timing constraints, we are at 8.2. - 12.3 kGE respectively for 3- and 2 stage configurations. I think the first step will be to identify where the bottleneck is that leads to this sharp increase. The easiest will probably be to make some instructions multi-cycle, or skip them for the minimal implementation. But we'll have to check benchmarks to make a decision.

PR #876, opened 2020-05-15
Update ICache testplan after review meeting

rswarbrick:
I think these represent the test cases we discussed. I've also removed
non-existent entries from the "tests" keys: I didn't really understand
how dvsim.py worked when I wrote the original version and they just
cause irritating warnings.


rswarbrick:
Hi @weicaiyang: I've renamed the tests as you suggest, thanks.

Issue #864, opened 2020-05-13
PMP fetch exception causing `X` propagation through IF stage


udinator:
Hi Tom,
I ran into 2 related (at least in this situation) issues when running some PMP tests, the waves and a code snippet are shown below:
![Screenshot from 2020-05-12 17-29-07](https://user-images.githubusercontent.com/16736281/81758278-2241a280-9476-11ea-9fe0-e66195aed79a.png)
![Screenshot from 2020-05-12 15-54-53](https://user-images.githubusercontent.com/16736281/81758072-a9424b00-9475-11ea-8cef-b799ea441c49.png)
In this situation, all PMP regions are using TOR address matching, `pmpaddr0` is set to the address of `<main>` (which gets word-aligned to `0x80001e14`) with all permissions enabled, and `pmpaddr1` is set to `0x8000447c` with all permissions initially disabled. With regards to instruction execution order, Ibex has just executed a jump instruction located at `0x80000314` and has jumped to `<main>`.
Just before the M3 marker in the waves, Ibex is executing an instruction fetch from `80001e14` to handle the first half of the misaligned instruction located at `<main>`, we can see that the `request` line goes high successfully and the testbench responds by asserting the `grant` line which goes high right at the M3 marker.
The first issue I see here is that given the PMP configuration this fetch to `80001e14` should raise an exception immediately as this address is controlled by the `pmpaddr1` region, but the core allows this fetch to succeed as it has somehow matched this to region0. I think this error is due to how `region_match_high[c][r]` is assigned within the PMP matching logic - it seems like it should be a strict `<` inequality instead of a `<=` to check equality of the upper bound as per the PMP specification of TOR matching.
Going past this error, what happens next is also fairly interesting; first just after M3, the core is about to send a second fetch request to `0x80001e18` to fulfill the second part of the misaligned instruction at `<main>`, but correctly flags it as a PMP fault and does not raise the top level `request` line to the testbench to complete the misaligned instruction fetch.
[...]

PR #877, opened 2020-05-17
Add Xcelium in (IUS compatibility mode) support

tudortimi:
This PR contains initial support for Xcelium (in Incisive compatibility mode).

It doesn't pass any tests. There seem to be issues with the multiplication logic, which lead to `X` values on signals coming from the ALU. I created a new dummy config that doesn't contain the `M` extensions and with it I managed to get an RTL simulation to run without any assertion errors:

```
small_no_mul:
  RV32E                    : 0
  RV32M                    : 0
  RV32B                    : 0
  BranchTargetALU          : 0
  WritebackStage           : 0
  MultiplierImplementation : "fast"
  PMPEnable                : 0
  PMPGranularity           : 0
  PMPNumRegions            : 4
```

Even with this configuration, I can't get a full test to pass. I generated a `riscv_arithmetic_basic_test` with only 10 instructions. This test reports 210 instructions in OVPSim, but 190 instructions in the RTL simulation. There also be other issues, but I didn't debug it any further at this time. These issues could either be simulator/SV code related. They could also be due to the fact that the DUT doesn't contain the `M` extension, but the ISG, and hence the ISS, might think that it does.


tudortimi:
There are a few other things that would be needed to help with debugging the issues further. First would be to have a configuration without the `M` extension available. As I mentioned above, I think that for such a configuration the ISS also needs to be informed somehow (via `riscv-dv`).

tudortimi:
Second would be to have an easy way to generate tests with less randomness in them. Running only 10 instructions is fine, but better would be to be able to configure to only run only `ADD`, for example. This would simplify the debug of issues related to the ISS vs. RTL log compare.

(It was also a bit painful to debug that the ISG expects a minimum of 10 instructions. By selecting less than 10, the ISG gen run fails, but the fact that something went wrong is flagged by the ISS run not finding a binary. This means that some fixes in `riscv-dv` might be helpful as well.)

tudortimi:
I didn't add any coverage options yet, as I felt this isn't essential at this time. More important is to get a passing test suite.

tudortimi:
The `X` values I encountered might be caused by #864.

tomroberts-lowrisc:
Hi @tudortimi 
Thanks for these changes - I'll leave @udinator to comment on the code and the generation issues. 

Are you able to make a wave dump available of the X issues (vcd ideally)? There might be a simple thing we can fix quickly to get that working.

udinator:
Hi @tudortimi , thanks for these changes.

I tried rebasing your changes in this PR on top of a clean branch, but I'm unable to even compile the RTL testbench due to the use of `$clog2()` to define a `localparam` in the `shared/rtl/prim_generic_ram_1p.sv` design - a feature that IUS does not seem to support.
Do you have some workaround to bypass this issue?

In regards to your question about informing the riscv-dv and ISS that the M-extension is disabled, you can simply pass in `ISA=rv32ic` from the command line when running your Makefile command.

tudortimi:
@udinator I can't really say about IUS. I used Xcelium, but the tool commands should be the same. I didn't add a new simulator called `xlm`, because that would also have meant changes to `riscv-dv`, which already provides `ius` as a simulator.

My plan was to do that after I got the entire setup going using the dummy IUS compatibility mode.

tudortimi:
@tomroberts-lowrisc Here's the VCD dump of the failing simulation (zipped with `zip`): 

[waves.vcd.zip](https://github.com/lowRISC/ibex/files/4646256/waves.vcd.zip)

The log file contains the following errors (configured to stop at first error time):

```
xmsim: *E,ASRTST (/home/tudor/projects/ibex/rtl/ibex_load_store_unit.sv,516): (time 32810 NS) Assertion core_ibex_tb_top.dut.u_ibex_core.load_store_unit_i.IbexDataOffsetKno
wn has failed
[ASSERT FAILED] [core_ibex_tb_top.dut.u_ibex_core.load_store_unit_i] IbexDataOffsetKnown (/home/tudor/projects/ibex/rtl/ibex_load_store_unit.sv:516)
xmsim: *E,ASRTST (/home/tudor/projects/ibex/rtl/ibex_id_stage.sv,1003): (time 32810 NS) Assertion core_ibex_tb_top.dut.u_ibex_core.id_stage_i.IbexBranchDecisionValid has fa
iled
[ASSERT FAILED] [core_ibex_tb_top.dut.u_ibex_core.id_stage_i] IbexBranchDecisionValid (/home/tudor/projects/ibex/rtl/ibex_id_stage.sv:1003)
```

imphil:
IUS is not supported in Ibex or opentitan, so remove that part and focus on xcelium only.

udinator:
To add to @imphil's comment, you may need to also add explicit Xcelium support to riscv-dv as we only support IUS there as of right now.

tomroberts-lowrisc:
Had a look at the waves, the X's are indeed coming from the multiplier - primarily because there isn't a multiplier!

The way the multiplier is instantiated uses string matching on the ``MultiplierImplementation`` parameter. If the parameter doesn't match any of the specified options, you get nothing at all. I guess that string parameters are not supported by Xcelium?

tudortimi:
@tomroberts-lowrisc LOL, that's one way to generate an X. I'll have a look, maybe I botched the parameter passing somehow.

tudortimi:
Any reason why the scripts aren't passing a string directly for the multiplier implementation? I.e. `"fast"` instead of `fast`.

GregAC:
The current method of supplying via a `define is due to VCS, you can only set integer parameters from the command line (or at least I had no luck getting string parameters to work after trying multiple different ways of doing things). Something in VCS appeared to be eating the `"` from the command line so passing the string with `"` as a `define then using that to produce the actual string in the .sv file was the way around that.

tudortimi:
@tomroberts-lowrisc Fixed it. I made a mistake when computing the value of the parameter and it didn't expand the value of the `IBEX_CFG_MultiplierImplementation` define. It was running with `MultiplierImplementation = "IBEX_CFG_MultiplierImplementation"`.

tudortimi:
Any reason why you're not using enums instead of strings? Synthesis tools don't like them?

tudortimi:
I've added first class for Xcelium to riscv-dv and I'll redo this PR once the version is updated here.

tudortimi:
I've added first class Xcelium support with #889. We can drop this PR.

PR #878, opened 2020-05-18
Two bug-fixes in icache memory model

rswarbrick:
These patches fix two bugs that cropped up when I started branching in a constrained way as part of writing the "passthru" test.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
+++ b/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
@@ -8,6 +8,22 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
 
   ibex_icache_mem_model #(.BusWidth (32)) mem_model;
 
+  // We pick new seeds when we spot a request (rather than when we spot a grant) to ensure that
+  // any given fetch corresponds to exactly one seed. Unfortunately, there's a race if these two
+  // things happen at one clock edge:
+  //
+  //    - Request for new address which gets a new seed
+  //    - Grant of previous request
+  //
+  // If that happens, and the event scheduler happened to schedule the monitor to spot the request
+  // before the grant, we might accidentally handle the grant with the new seed rather than the
+  // old one, which would cause confusion in the scoreboard.
+  //
+  // To avoid this problem, we keep a FIFO of pending addresses along with their corresponding
+  // seeds. When a grant message comes in, we know that we can look up the correct seed there.
+  protected mailbox #(bit[63:0]) pending_grants;
+  protected bit [31:0]           cur_seed = 0;
+
   `uvm_object_utils(ibex_icache_mem_resp_seq)
   `uvm_object_new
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
For any reviewer, note that the commit messages are quite long and (I hope!) informative.

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
```diff
@@ -17,79 +33,88 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
...
+    while (pending_grants.num() > 0) begin
+      pending_grants.get(gnt_data);
+      {gnt_addr, gnt_seed} = gnt_data;
+      if (gnt_addr == req_item.address) break;
     end
-  endtask
+
+    // If nothing went wrong, we should have found the right item and gnt_addr should now
+    // equal req_item.address
+    `DV_CHECK_FATAL(gnt_addr == req_item.address,
```
udinator: nit - use the `DV_CHECK_EQ_FATAL` macro here

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
```diff
@@ -91,19 +90,27 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
...
+      {gnt_addr, gnt_seed} = pending_grants[N - 1 - i];
+      if (gnt_addr == req_item.address)
+        break;
     end
 
-    // If nothing went wrong, we should have found the right item and gnt_addr should now
-    // equal req_item.address
-    `DV_CHECK_FATAL(gnt_addr == req_item.address,
+    // If i == pending_grants.size(), we didn't find a hit. That's not supposed to happen.
+    `DV_CHECK_FATAL(i < N,
```
udinator, line 105: nit - use the `DV_CHECK_LT_FATAL` macro here

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
rswarbrick: I don't think this would be right: the only difference with this macro is that it would print `gnt_addr` and `req_item.address` on failure. But `req_item.address` appears in the error message already and `gnt_addr` is irrelevant (it's the last address that happened to be in the queue: definitely more confusing than useful in an error message!).

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
rswarbrick: I'm not keen on this: the error message on failure would then contain fluff about `i` and `N`, which doesn't really have anything to do with "There is no pending grant for this address". Thinking about it, there would be some information that might be interesting for debug (the value of `N`), so I'll add that to the message manually.

udinator:
I think your comment on the macro usage makes sense - it does seem like these situations are meant to indicate information other than a strict equality or inequality. 
LGTM as is, feel free to merge!

rswarbrick:
Great, thanks!

PR #881, opened 2020-05-20
Syn additions

GregAC:
Used the synthesis flow recently which resulted in me adding these additions.

The flow is actually broken out of the box currently, there's some RTL issues to investigate which I haven't had time to do but could be quickly hacked around for getting the flow going.


GregAC:
Fixing the issues so the flow does work out of the box is something I'll hopefully look at soon, along with getting some kind of yosys CI integration to avoid it breaking without us realising in the future.

PR #882, opened 2020-05-20
[rtl] Various small icache bugfixes

tomroberts-lowrisc:
- Remove any ready -> valid dependency by allowing the skid buffer to
  accept data when the core is not ready
- Tighten-up behaviour around invalidations and cache enable/disable
- Remove xprop through output_compressed from invalid data when driving errors
- Make behaviour more consistent where speculative requests return
  different data/error conditions to existing cache hit

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
@@ -91,10 +91,10 @@ interface ibex_icache_core_protocol_checker (
   `ASSERT(Err2Stable,      unfinished_valid |=> $stable(err_plus2),   clk, !rst_n);
 
   `ASSERT(LoRDataStable,
-          unfinished_valid |=> $stable(rdata[15:0]),
+          unfinished_valid & ~err |=> $stable(rdata[15:0]),
           clk, !rst_n);
   `ASSERT(HiRDataStable,
-          unfinished_valid & (rdata[1:0] == 2'b11) |=> $stable(rdata[31:16]),
+          unfinished_valid & ~err & (rdata[1:0] == 2'b11) |=> $stable(rdata[31:16]),
           clk, !rst_n);
 
   // The err_plus2 signal means "this error was caused by the upper two bytes" and is only read when
--- a/rtl/ibex_icache.sv
+++ b/rtl/ibex_icache.sv
@@ -125,7 +125,8 @@ module ibex_icache #(
   logic                                gnt_or_pmp_err, gnt_not_pmp_err;
   logic [$clog2(NUM_FB)-1:0]           fb_fill_level;
   logic                                fill_cache_new;
-  logic                                fill_new_alloc, fill_spec_done, fill_spec_hold;
+  logic                                fill_new_alloc;
+  logic                                fill_spec_req, fill_spec_done, fill_spec_hold;
   logic [NUM_FB-1:0][NUM_FB-1:0]       fill_older_d, fill_older_q;
   logic [NUM_FB-1:0]                   fill_alloc_sel, fill_alloc;
   logic [NUM_FB-1:0]                   fill_busy_d, fill_busy_q;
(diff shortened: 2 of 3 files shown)


dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
```diff
@@ -94,7 +94,7 @@ interface ibex_icache_core_protocol_checker (
           unfinished_valid |=> $stable(rdata[15:0]),
           clk, !rst_n);
   `ASSERT(HiRDataStable,
-          unfinished_valid & (rdata[1:0] == 2'b11) |=> $stable(rdata[31:16]),
+          unfinished_valid & (rdata[1:0] == 2'b11) & ~err |=> $stable(rdata[31:16]),
```
rswarbrick, line 97: You're right about this ("If the err signal is high, the contents of rdata don't matter"). Could we split it into a separate commit with a commit message explaining what's going on? And probably some of the other checks should have the same "& ~err". LoRDataStable, at least, I think?

rtl/ibex_icache.sv
```diff
@@ -865,20 +876,26 @@ module ibex_icache #(
     end
   end
 
-  // The data in the skid buffer is a complete compressed instruction
-  assign skid_complete_instr = skid_valid_q & (skid_data_q[1:0] != 2'b11);
+  // The data in the skid buffer is a complete compressed instruction or if there's an error
```
rswarbrick, line 879: Maybe "The data in the skid buffer is ready if ..."?

rtl/ibex_icache.sv
```diff
@@ -865,20 +876,26 @@ module ibex_icache #(
...
       // Once valid, the skid buffer stays valid until a compressed instruction realigns the stream
-      (skid_valid_q ? ~(ready_i & (skid_data_q[1:0] != 2'b11)) :
-      // The skid buffer becomes valid when a compressed instruction misaligns the stream
-                      ((output_addr_q[1] ^ output_compressed) & data_valid & ready_i));
+      (skid_valid_q ? ~(ready_i & ((skid_data_q[1:0] != 2'b11) | skid_err_q)) :
+      // The skid buffer becomes valid when:
+                        // - we branch to an unaligned uncompressed instruction
+                      (((output_addr_q[1] & (~output_compressed | output_err)) |
+                        // - a compressed instruction misaligns the stream
+                        (~output_addr_q[1] & output_compressed & ~output_err & ready_i)) & data_valid));
```
rswarbrick, line 898: What does the `_ | output_err` bit do here? If we branch to an unaligned uncompressed instruction and see an error, it doesn't really matter what the skid buffer contains, since the core is going to have to branch anyway, no?

rtl/ibex_icache.sv
```diff
@@ -902,7 +919,7 @@ module ibex_icache #(
   assign output_addr_en = branch_i | (ready_i & valid_o);
 
   // Increment the address by two every time a compressed instruction is popped
-  assign addr_incr_two = output_compressed;
+  assign addr_incr_two = output_compressed & ~err_o;
```
rswarbrick, line 922: I don't think it's wrong, but why do we need this? If `err_o` is true, I would have thought our internal PC can perfectly well be 'X: it'll need resetting on the next branch anyway.

rswarbrick:
Oh sorry, I forgot to add this. Could you also append a patch that reverts `ff5c0c5`?

dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_protocol_checker.sv
tomroberts-lowrisc: Yep will do

rtl/ibex_icache.sv
tomroberts-lowrisc: Will do

rtl/ibex_icache.sv
tomroberts-lowrisc: True it doesn't matter whether ``skid_valid_q`` gets set or not, but if we get an error, ``output_compressed`` will be X, which makes ``skid_valid_q`` and then various other things problematically X.

rtl/ibex_icache.sv
tomroberts-lowrisc: Same as the other comment - true it'll get tidied up when the branch comes, but the X's in the address tend to propagate out problematically before then so I just mask it here.

PR #883, opened 2020-05-20
[rtl] Add speculative branch signal

tomroberts-lowrisc:
- Drive a speculative version of the branch signal into the IF stage to
  drive address muxing
- The speculative signal is the same as the regular branch signal but
  assumes all conditional branches are taken
- This breaks the timing path from branch condition calculation into
  address muxing (and therefore PMP error calculation)
- When the branch is not taken, any external request we might otherwise
  have made is suppressed
- This has a minor performance cost (0.8% without I$, ~0% with I$)

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -45,6 +45,7 @@ module ibex_controller #(
     // to prefetcher
     output logic                  instr_req_o,             // start fetching instructions
     output logic                  pc_set_o,                // jump to address set by pc_mux
+    output logic                  pc_set_spec_o,           // speculative branch
     output ibex_pkg::pc_sel_e     pc_mux_o,                // IF stage fetch address selector
                                                            // (boot, normal, exception...)
     output ibex_pkg::exc_pc_sel_e exc_pc_mux_o,            // IF stage selector for exception PC
@@ -58,6 +59,7 @@ module ibex_controller #(
 
     // jump/branch signals
     input  logic                  branch_set_i,            // branch taken set signal
+    input  logic                  branch_set_spec_i,       // speculative branch signal
     input  logic                  jump_set_i,              // jump taken set signal
 
     // interrupt signals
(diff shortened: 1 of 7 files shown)


rtl/ibex_controller.sv
```diff
@@ -421,6 +426,11 @@ module ibex_controller #(
           perf_jump_o    = jump_set_i;
         end
 
+        // pc_set signal excluding branch taken condition
+        if ((branch_set_spec_i || jump_set_i) && ~special_req_branch) begin
```
imphil, line 430:
```suggestion
        if ((branch_set_spec_i || jump_set_i) && !special_req_branch) begin
```

tomroberts-lowrisc:
The performance cost comes when ``branch_spec_i`` is high, and ``branch_i`` is low (not-taken branch). Without these changes, the IF stage can make an ``instr_req_o`` request in that cycle (for the next instruction after the not-taken branch, or the instruction after that, say). With these changes, the core can't do anything in that cycle and will make the next external request one cycle later.

It's a low performance cost since most of the time we execute a not-taken branch, we already have the next instruction buffered and can absorb the loss of one cycle.

rtl/ibex_controller.sv
```diff
@@ -721,4 +738,7 @@ module ibex_controller #(
       RESET, BOOT_SET, WAIT_SLEEP, SLEEP, FIRST_FETCH, DECODE, FLUSH,
       IRQ_TAKEN, DBG_TAKEN_IF, DBG_TAKEN_ID})
 
+  // The speculative branch signal should be set whenever the actual branch signal is set
+  `ASSERT(IbexSpecImpliesSet, pc_set_o |-> pc_set_spec_o)
```
GregAC, line 742: Nit: Could this name mention the PC also? E.g. `IbexSpecImpliesSetPC`

vogelpi:
Thanks for the clarification @tomroberts-lowrisc , I did not realize that previously we would just continue to fetch from `fetch_addr_q` if the branch was not taken.

rtl/ibex_controller.sv
tomroberts-lowrisc: Fixed

rtl/ibex_controller.sv
tomroberts-lowrisc: Fixed

PR #889, opened 2020-05-21
Add Xcelium support

tudortimi:


tudortimi:
Using enums instead of strings would avoid the need for the workaround in 6f8d1c5. Are there limitations here with any of the other tools?

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -40,7 +40,12 @@ module core_ibex_tb_top;
   parameter bit RV32B                    = 1'b0;
   parameter bit BranchTargetALU          = 1'b0;
   parameter bit WritebackStage           = 1'b0;
-  parameter     MultiplierImplementation = `"`IBEX_CFG_MultiplierImplementation`";
+
+  // VCS has issues taking a string as a define, so we have to build up the string via the
+  // pre-processor. The LRM only allows '`"` to appear in macro expansions, so we have to create
+  // a dummy macro to do the string expansion.
```
imphil, line 46: Can you elaborate a bit more what's going on here? The current code works in VCS apparently, so it's either VCS being too lenient here, or Xcelium not supporting something?

Anyway, instead of providing a own macro, you can include "prim_assert.sv" and use "PRIM_STRINGIFY". You'll probably need to add `shared/rtl` as include dir to `ibex_dv.f` for that.

tudortimi: VCS is too lenient. IEEE 1800-2012, Section 22.5.1:

> The `define macro text can also include `", `\`", and ``.
>
> An `" overrides the usual lexical meaning of " and indicates that the expansion shall include the quotation
mark, substitution of actual arguments, and expansions of embedded macros. This allows string literals to be
constructed from macro arguments.

tudortimi: Do you prefer a rebase to replace this commit with usage of `PRIM_STRINGIFY` or an extra commit?

tudortimi:
I didn't include `prim_assert.sv` in the TB top, as it was already in `ibex_dv.f`. 

(Comment: This looks fishy, as the file contains both macros and a package. This can cause issues when compiling using multiple compilation units.)

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
imphil: A rebase please, leaving only the actual change, not the history leading to it.

PR #890, opened 2020-05-21
[dv] Change performance counter access to DPI

wallento:
DPI access is suggested and more generic than Verilator direct signal
access. This changes the access to the performance counters from the
Verilator testbench to use DPI instead of directly accessing the
array.

### DIFF
--- a/examples/simple_system/ibex_simple_system.cc
+++ b/examples/simple_system/ibex_simple_system.cc
@@ -28,17 +28,19 @@ int main(int argc, char **argv) {
     return 1;
   }
 
+  // Set the scope to the root scope, the ibex_pcount_string function otherwise
+  // doesn't know the scope itself. Could be moved to ibex_pcount_string, but
+  // would require a way to set the scope name from here, similar to MemUtil.
+  svSetScope(svGetScopeFromName("TOP.ibex_simple_system"));
   // TODO: Exec can return with "true" (e.g. with `-h`), but that does not mean
   // `RunSimulation()` was executed. The folllowing values will not be useful
   // in this case.
   std::cout << "\nPerformance Counters" << std::endl
             << "====================" << std::endl;
-  std::cout << ibex_pcount_string(top.ibex_simple_system__DOT__mhpmcounter_vals,
-                                  false);
+  std::cout << ibex_pcount_string(false);
 
   std::ofstream pcount_csv("ibex_simple_system_pcount.csv");
-  pcount_csv << ibex_pcount_string(
-      top.ibex_simple_system__DOT__mhpmcounter_vals, true);
+  pcount_csv << ibex_pcount_string(true);
 
   return 0;
 }
--- a/examples/simple_system/rtl/ibex_simple_system.sv
+++ b/examples/simple_system/rtl/ibex_simple_system.sv
@@ -249,12 +249,10 @@ module ibex_simple_system (
       .timer_intr_o   (timer_irq)
     );
 
-  // Expose the performance counter array so it's easy to access in
-  // a verilator siumulation
-  logic [63:0] mhpmcounter_vals [32] /*verilator public_flat*/;
+  export "DPI-C" function mhpmcounter_get;
 
-  for(genvar i = 0;i < 32; i = i + 1) begin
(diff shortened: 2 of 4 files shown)


examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -249,12 +249,10 @@ module ibex_simple_system (
...
-  // a verilator siumulation
-  logic [63:0] mhpmcounter_vals [32] /*verilator public_flat*/;
+  export "DPI-C" function mhpmcounter_get;
 
-  for(genvar i = 0;i < 32; i = i + 1) begin
-      assign mhpmcounter_vals[i] = u_core.u_ibex_core.cs_registers_i.mhpmcounter[i];
-  end
-endmodule
+  function longint mhpmcounter_get(int index);
+    mhpmcounter_get = u_core.u_ibex_core.cs_registers_i.mhpmcounter[index];
```
imphil, line 255:
```suggestion
    return u_core.u_ibex_core.cs_registers_i.mhpmcounter[index];
```

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -249,12 +249,10 @@ module ibex_simple_system (
...
-  // Expose the performance counter array so it's easy to access in
-  // a verilator siumulation
-  logic [63:0] mhpmcounter_vals [32] /*verilator public_flat*/;
+  export "DPI-C" function mhpmcounter_get;
 
-  for(genvar i = 0;i < 32; i = i + 1) begin
-      assign mhpmcounter_vals[i] = u_core.u_ibex_core.cs_registers_i.mhpmcounter[i];
-  end
-endmodule
+  function longint mhpmcounter_get(int index);
```
imphil, line 254: If possible (I'm not sure how Verilator likes that with DPI exports):

```suggestion
  automatic function longint mhpmcounter_get(int index);
```

(To align with https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#functions-and-tasks)

examples/simple_system/ibex_simple_system.cc
```diff
@@ -17,6 +17,7 @@ int main(int argc, char **argv) {
   simctrl.SetTop(&top, &top.IO_CLK, &top.IO_RST_N,
                  VerilatorSimCtrlFlags::ResetPolarityNegative);
 
+  svSetScope(svGetScopeFromName("TOP.ibex_simple_system"));
```
imphil, line 20: Can you move that into the `ibex_pcount_string()` routine? (Ideally, restore the previous scope there as well after the function is done to reduce side effects, but that's probably overkill and could impact performance. Maybe add a comment to the `ibex_pcount_string() doxygen comment instead.)

wallento: Ah, I wanted to comment there why I didn't do it, but forgot: I think that this would reduce the flexibility of usage in systems where the toplevel is not `ibex_simple_system`. Would you suggest to:

1. have an init function pcount_init (or even better, put it into a class with constructor), or
2. get the toplevel string from `verilated_top.h`?

The reason I put it here was my thinking that we maybe want to use the root scope as default (contrary to no scope). This was the third option that I chose to reduce the number of changes and still come up with a reasonable solution, forgot to mention there.

What do you prefer?

imphil: I see, let's not go overboard then. Can you move the `svSetScope()` function bit further down to the reading of the perf counters, and add a comment why it is needed? That should match what we have today roughly, so good enough.

wallento: Option 1 and 2 would allow to not hardcode the toplevel name into the generic files obviously.

imphil: (We do the "full generic" solution in `VerilatorMemUtil`, but I wouldn't put that much effort into reading performance counters from a test system.)

wallento: yeah, right, that's what I thought too

wallento:
Amended as suggested/discussed

PR #891, opened 2020-05-21
Change simple system core file to provide default files

wallento:
To allow other modules to reference the simple system, it must provide
default files. In particular this is useful in DV settings where bind
is used.

Signed-off-by: Stefan Wallentowitz <stefan.wallentowitz@hm.edu>

### DIFF
--- a/examples/simple_system/ibex_simple_system.core
+++ b/examples/simple_system/ibex_simple_system.core
@@ -83,11 +83,15 @@ parameters:
     description: "Number of PMP regions"
 
 targets:
-  sim:
-    default_tool: verilator
+  default: &default_target
     filesets:
       - tool_verilator ? (files_verilator_waiver)
       - files_sim_verilator
+    toplevel: ibex_simple_system
+
+  sim:
+    <<: *default_target
+    default_tool: verilator
     parameters:
       - RV32M
       - RV32E
@@ -99,7 +103,6 @@ targets:
       - PMPGranularity
       - PMPNumRegions
       - SRAM_INIT_FILE
-    toplevel: ibex_simple_system
     tools:
       vcs:
         vcs_options:


examples/simple_system/ibex_simple_system.core
```diff
@@ -83,6 +83,10 @@ parameters:
     description: "Number of PMP regions"
 
 targets:
+  default:
```
imphil, line 86:
```suggestion
  default: &default_target
```

examples/simple_system/ibex_simple_system.core
```diff
@@ -83,6 +83,10 @@ parameters:
...
+    filesets:
+      - files_sim_verilator
+
   sim:
     default_tool: verilator
     filesets:
```
imphil, line 92: Replace the whole `filesets` entry and its children with

```
<<: *default_target
```

examples/simple_system/ibex_simple_system.core
```diff
@@ -83,6 +83,10 @@ parameters:
     description: "Number of PMP regions"
 
 targets:
+  default:
+    filesets:
+      - files_sim_verilator
+
```
imphil, line 89: Move the `toplevel` also in here.

wallento:
Thanks, I don't understand this part though: https://github.com/lowRISC/opentitan/blob/master/hw/ip/flash_ctrl/flash_ctrl.core#L74
Why isn't that using the default set?

wallento:
Thanks, I think I may have done it properly now ;)

imphil:
> Why isn't that using the default set?

Oversight aka "bug" :smile: It should be used there as well.

wallento:
> > Why isn't that using the default set?
> 
> Oversight aka "bug" smile It should be used there as well.

Ah, okay, think I understand it then, yaml magic ;)

Issue #893, opened 2020-05-22
DV: Use FuseSoC instead of hardcoding file list in ibex_dv.f


imphil:
Similar to what we do in OpenTitan, we must move away from a hard-coded file list to make full use of our primitives system and other functionality enabled by FuseSoC and its `.core` files.

PR #894, opened 2020-05-25
Use vendored-in primitives from OpenTitan

imphil:
Instead of using copies of primitives from OpenTitan, vendor the files
in directly from OpenTitan, and use them.

Benefits:

- Less potential for diverging code between OpenTitan and Ibex, causing
  problems when importing Ibex into OT.

- Use of the abstract primitives instead of the generic ones. The
  abstract primitives are replaced during synthesis time with
  target-dependent implementations. For simulation, nothing changes. For
  synthesis for a given target technology (e.g. a specific ASIC or FPGA
  technology), the primitives system can be instructed to choose
  optimized versions (if available).

  This is most relevant for the icache, which hard-coded the generic
  SRAM primitive before. This primitive is always implemented as
  registers. By using the abstract primitive (prim_ram_1p) instead, the
  RAMs can be replaced with memory-compiler-generated ones if necessary.

There are no real draw-backs, but a couple points to be aware of:

- Our ram_1p and ram_2p implementations are kept as wrapper around the
  primitives, since their interface deviates slightly from the one in
  prim_ram*. This also includes a rather unfortunate naming confusion
  around rvalid, which means "read data valid" in the OpenTitan advanced
  RAM primitives (prim_ram_1p_adv for example), but means "ack" in
  PULP-derived IP and in our bus implementation.

- The core_ibex UVM DV doesn't use FuseSoC to generate its file list,
  but uses a hard-coded list in `ibex_files.f` instead. Since the
  dynamic primitives system requires the use of FuseSoC we need to
  provide a stop-gap until this file is removed. Issue #893 tracks
  progress on that.

- Dynamic primitives depend no a not-yet-merged feature of FuseSoC
  (https://github.com/olofk/fusesoc/pull/391). We depend on the same
  functionality in OpenTitan and have instructed users to use a patched
  branch of FuseSoC for a long time through `python-requirements.txt`,
  so no action is needed for users which are either successfully
  interacting with the OpenTitan source code, or have followed our
  instructions. All other users will see a reasonably descriptive error
  message during a FuseSoC run.

- This commit is massive, but there are no good ways to split it into
  bisectable, yet small, chunks. I'm sorry. Reviewers can safely ignore
  all code in `vendor/lowrisc_ip`, it's an import from OpenTitan.

- The check_tool_requirements tooling isn't easily vendor-able from
  OpenTitan at the moment. I've filed
  https://github.com/lowRISC/opentitan/issues/2309 to get that sorted.

- The LFSR primitive doesn't have a own core file, forcing us to include
  the catch-all `lowrisc:prim:all` core. I've filed
  https://github.com/lowRISC/opentitan/issues/2310 to get that sorted.


imphil:
Here we go, the final battle to bring (more) order into mess of copied files between OT and Ibex. Please have a look at the individual commits for slightly more details.

There should be no functional change after this PR is applied.

imphil:
In terms of testing we should be good as well.

- CI tests are green, which includes
  - Ibex compliance
  - Lint & Co.
- ibex_core UVM tests pass
- The simple system example can execute code.
- The Arty A7-35 example synthesizes (I don't have a board to test)
- icache UVM tests pass

shared/rtl/ram_2p.sv
```diff
@@ -4,13 +4,6 @@
 
 /**
  * Dual-port RAM with 1 cycle read/write delay, 32 bit words.
- *
- * The two ports are in Read-First Mode: when reading from and writing to the same address
- * simultaneously, the old content is returned, before the new content is written. New content
- * is made available to both ports with one cycle delay.
- *
- * Simultaneous write operations by both ports to the same address are to be avoided: The data
- * written to memory is not determined.
```
vogelpi, line 13: I usually find it useful to have this information somewhere but agree that now that this `ram_2p.sv` is just a wrapper for the actual prim, it is no longer the right place.

I think we should add this information in the actual primitive upstream.

imphil: I removed it because we cannot make those guarantees in the wrapper, and the non-wrapper doesn't actually have any such guarantees at the moment. (https://github.com/lowRISC/opentitan/issues/2267)

python-requirements.txt
```diff
@@ -7,3 +7,6 @@ git+https://github.com/lowRISC/edalize.git@ot
 
 # Development version with OT-specific changes
 git+https://github.com/lowRISC/fusesoc.git@ot
+
+pyyaml
+mako
```
rswarbrick, line 12: Is this for `primgen`? Does that now get run as part of an Ibex build?

util/check_tool_requirements.py
```diff
@@ -6,50 +6,15 @@
...
 
 # Display INFO log messages and up.
 log.basicConfig(level=log.INFO, format="%(levelname)s: %(message)s")
 
-
-def get_tool_requirements_path():
-    '''Return the path to tool_requirements.py, at the top of the Ibex repo'''
-    # top_src_dir is the top of the repository
-    top_src_dir = os.path.normpath(os.path.join(os.path.dirname(__file__),
-                                                '..'))
```
rswarbrick, line 21: This patch seems to revert commit `db6f8f0`. Is that intentional?

python-requirements.txt
imphil: Yes. (pyyaml is also used for ibex_config.py but hasn't been included in the requirements file so far and was installed on CI runners by "luck" [not real luck, it's a dependency of fusesoc and edalize])

util/check_tool_requirements.py
imphil: No, that's something I missed. These changes need to go into OT first now, I'll open a PR there.

rswarbrick: I'm confused. Is `util/check_tool_requirements.py` vendored from the OT repo?

imphil: You have every reason to be confused.

Before, `util/check_tool_requirements.py` was a diverged copy from OT with a different core name. 

We now need to make sure we have the same core name for this functionality between OT and Ibex, as there are dependencies on it in other core files. With the same core name, I also wanted to make sure that we actually have the same functionality.

So what this PR does is align the core file and the contents of the script with what we have in OT, but we cannot easily vendor it in until https://github.com/lowRISC/opentitan/issues/2309 is fixed.

rswarbrick: In which case, isn't it better to keep the existing functionality and add it to the OT version once that's possible? At the moment, this patch is deleting code when it's not technically needed, I think.

imphil: We need to have the core name of the OT version, since that's used in dependencies. And I don't want to use the same core name for cores with different implementations between OT and Ibex -- doing stuff like that got us into this mess in the first place.

I'll update this PR to keep the original functionality once https://github.com/lowRISC/opentitan/pull/2319 goes in, so we don't loose anything.

imphil: Fixed now.

imphil:
I've updated this PR to use the new vendor.py script, took the updated check_requirements.py, and found a small change which wasn't related to this PR (#901).

imphil:
Good catch @rswarbrick, fixed.

PR #896, opened 2020-05-26
[DV] Tie test_en_i to zero

tomroberts-lowrisc:
- test_en_i is a DFT feature that shouldn't be enabled for normal
  runtime testing
- Only really affects the clock gate in the design, but is needed for
  running tests with the latch-based register file


tomroberts-lowrisc:
Thanks - I ran some wfi related tests as they seemed like the only ones that would be affected by clock gate changes and all seemed fine.

Issue #902, opened 2020-05-27
Incorrect pmp NA4/NAPOT address matching


udinator:
#875 seems to have introduced an issue with `NA4/NAPOT` address matching modes - the scenario is as follows.

In my test, `csr_pmp_addr[3]` is `0x8000_3f54` with its corresponding configuration bits `csr_pmp_cfg[3]` set to `{0, PMP_MODE_NA4, 0, 1, 1}`, and the core requests an instruction fetch from `pmp_req_addr_i[0] = 0x8000_3f54`. Note that since we are using NA4 mode, `region_start_addr[3] = 0x8000_3f54` and `region_addr_mask[3] = 0xffff_ffff`. We are also using a granularity of 0.

With this setup, we should expect that the requested instruction fetch should match `pmpaddr3` but then then fail due to a fetch error since the corresponding access bit is disabled.

However, `region_match_low[0][3]` goes high as expected since
`pmp_req_addr_i[0][33:2] >= region_start_addr[3][33:2]`,
but `region_match_high[0][3]` stays low since the check
`pmp_req_addr_i[0][33:2] < csr_pmp_addr[3][33:2]` will fail due to lack of equality checking - this causes an altogether mismatch with this PMP region.

I think we still need to include equality checking to cover NA4 and NAPOT matching (as the error I describe above can also easily occur in NAPOT mode with any granularity), but these equality checks should only be enabled when we are in one of those two modes.

tomroberts-lowrisc:
Thanks @udinator - I was trying to be (too) clever and avoid the extra equality matching by using address masking. I think this still works for NA4 if the mask is set up correctly. I'll push a proposed fix as a PR and hopefully you can run some testing on it to confirm.

PR #903, opened 2020-05-27
[rtl] Fix PMP NA4 address matching

tomroberts-lowrisc:
- Apply address masking to the bottom address bit for NA4 as well as
  NAPOT
- Relates to #902

### DIFF
--- a/rtl/ibex_pmp.sv
+++ b/rtl/ibex_pmp.sv
@@ -32,9 +32,10 @@ module ibex_pmp #(
   // Access Checking Signals
   logic [33:0]                                region_start_addr [PMPNumRegions];
   logic [33:PMPGranularity+2]                 region_addr_mask  [PMPNumRegions];
-  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_high;
-  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_low;
-  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_both;
+  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_gt;
+  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_lt;
+  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_eq;
+  logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_match_all;
   logic [PMPNumChan-1:0][PMPNumRegions-1:0]   region_perm_check;
   logic [PMPNumChan-1:0]                      access_fault;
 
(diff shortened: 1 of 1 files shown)


tomroberts-lowrisc:
@udinator to run some PMP testing on this

udinator:
Hmm, I'm not quite sure if this change is correct from a spec-compliant way (functionally it works but technically shouldn't).
This causes bit [2] of region_addr_mask to be set to 0 when we're in NA4 mode, making this equivalent to a NAPOT check with granularity 1 and results in additional (incorrect) matches with other addresses in the 8-byte space.

Screenshot from 2020-05-27 14-26-17
Just going off of the spec, I think the original mask of `0xffff_ffff` was already correct for NA4.

tomroberts-lowrisc:
Thanks @udinator you're right - not sure what I was thinking there!
I have pushed a new version which does more straightforward address matching. I'd appreciate a bit of testing on this new version if you have time.
Cheers
Tom

rtl/ibex_pmp.sv
```diff
@@ -71,16 +70,30 @@ module ibex_pmp #(
...
+          PMP_MODE_NA4   : region_match_all[c][r] = region_match_eq[c][r];
+          PMP_MODE_NAPOT : region_match_all[c][r] = region_match_eq[c][r];
+          PMP_MODE_TOR   : begin
+            region_match_all[c][r] = (region_match_eq[c][r] | region_match_gt[c][r]) &
+                                     region_match_lt[c][r];
+          end
+          default        : region_match_all[c][r] = 1'b0;
+        endcase
+      end
+
```
udinator, line 96: I think this is almost there, but in this implementation TOR matching is incorrect, as it takes into consideration equality  checking with `pmpaddr[r]` instead of with `pmpaddr[r-1]`.
Maybe something like the suggestion below could work to take into account the necessary equality checking of the lower bound.

```suggestion
      assign region_match_eq_hi[c][r] = (pmp_req_addr_i[c][33:PMPGranularity+2] &
                                         region_addr_mask[r]) ==
                                        (csr_pmp_addr_i[r][33:PMPGranularity+2] &
                                         region_addr_mask[r]);
      assign region_match_eq_lo = (pmp_req_addr_i[c][33:PMPGranularity+2] &
                                   region_addr_mask[r]) ==
                                  (region_start_addr[r][33:PMPGranularity+2] &
                                   region_addr_mask[r]);
      assign region_match_gt[c][r] = pmp_req_addr_i[c][33:PMPGranularity+2] >
                                     region_start_addr[r][33:PMPGranularity+2];
      assign region_match_lt[c][r] = pmp_req_addr_i[c][33:PMPGranularity+2] <
                                     csr_pmp_addr_i[r][33:PMPGranularity+2];

      always_comb begin
        region_match_all[c][r] = 1'b0;
        unique case (csr_pmp_cfg_i[r].mode)
          PMP_MODE_OFF   : region_match_all[c][r] = 1'b0;
          PMP_MODE_NA4   : region_match_all[c][r] = region_match_eq_hi[c][r];
          PMP_MODE_NAPOT : region_match_all[c][r] = region_match_eq_hi[c][r];
          PMP_MODE_TOR   : begin
            region_match_all[c][r] = (region_match_eq_lo[c][r] | region_match_gt[c][r]) &
                                     region_match_lt[c][r];
          end
          default        : region_match_all[c][r] = 1'b0;
        endcase
      end

```

We can't actually use `region_match_eq[c][r-1]` in TOR equality checks because that will use an incorrect mask, which is why I suggested a new array `region_match_eq_lo` to handle equality of the lower bounds of the TOR region (we still need equality checks with the `csr_pmp_addr_i[r]` for NA4 and NAPOT checks).

tomroberts-lowrisc: Good spot - thanks @udinator 
I added back the `region_start_addr` muxing which was there before - should give the correct match address and mask for TOR mode now.

udinator: awesome, seems like re-adding that muxing has fixed up this PR.

tomroberts-lowrisc: thanks @udinator !

Issue #904, opened 2020-05-27
Failing UVM sanity with commit "[rtl] Add speculative branch signal"


imphil:
Commit 12b39476c0c6bb16dc0d43057730f263836dac95 (`[rtl] Add speculative branch signal`) broke the UVM sanity tests.

Running `./vendor/lowrisc_ip/dvsim/dvsim.py dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson -t vcs --waves` I get:

## List of Failures

**TEST:** ibex_icache_sanity, **SEED:** 3278749286<br>
**LOG:** $scratch_path/0.ibex_icache_sanity/out/run.log<br>
```
UVM_FATAL @    189350 ps: (ibex_icache_mem_resp_seq.sv:113) uvm_test_top.env.mem_agent.sequencer [uvm_test_top.env.virtual_sequencer.ibex_icache_sanity_vseq.mem_seq] Check failed (i < N) No pending grant for address 0xxxxxxxxX (1 items in queue). 
UVM_INFO @    189350 ps: (uvm_report_server.svh:894) [UVM/REPORT/SERVER] 
--- UVM Report Summary ---

Quit count :     0 of     1
```

**TEST:** ibex_icache_passthru, **SEED:** 1046335328<br>
**LOG:** $scratch_path/0.ibex_icache_passthru/out/run.log<br>
```
UVM_FATAL @    688601 ps: (ibex_icache_mem_resp_seq.sv:113) uvm_test_top.env.mem_agent.sequencer [uvm_test_top.env.virtual_sequencer.ibex_icache_passthru_vseq.mem_seq] Check failed (i < N) No pending grant for address 0xxxxxxxxX (1 items in queue). 
UVM_INFO @    688601 ps: (uvm_report_server.svh:894) [UVM/REPORT/SERVER] 
--- UVM Report Summary ---

Quit count :     0 of     1
```

@tomroberts-lowrisc @rswarbrick

PR #905, opened 2020-05-27
Optimize use of RAM primitive in icache

imphil:
The RAM primitive provides a way to specify the granularity of the write
mask (wmask) signal, which can be used to select an appropriate
implementation (e.g. a SRAM with only byte selects, or no subselects at
all).

---
Even though this change should be safe, I've seen X asserts fire on this one previously in icache UVM DV. I cannot test right now until #904 is fixed.


imphil:
Re-ran DV with #904 fixed and it looks good.

PR #911, opened 2020-05-28
[ibex/dv] Add several PMP tests

udinator:
This PR adds 3 new PMP tests, and updates the existing PMP test in the Ibex testlist.

1) `riscv_pmp_basic_test` - this updates the existing Ibex-specific PMP sanity test, where every PMP region is given full x/w/r permissions and operates in TOR address matching mode, as a result the core should be able to successfully execute every instruction fetch and memory load/store without any faults being raised.

2) `riscv_pmp_disable_all_regions_test` - In this test, all x/w/r permissions are stripped from every PMP region and TOR address matching is enabled by default for every region. We should see every fetch from IMEM and every load/store from DMEM raise an exception to trap into the custom exception handler.

3) `riscv_pmp_out_of_bounds_test` - In this test, default PMP settings are used (full permissions granted to every region), but the random assembly program is injected with directed instructions that access memory locations that are not contained by any PMP region. If the core is in M-mode, we ensure that the core successfully executes these instructions, but if the core is in U-mode, we ensure that the core raises an exception.

4) `riscv_pmp_full_random_test` - In this test, we completely randomize all PMP regions (their corresponding addresses, address matching modes, and all access privileges). This is extremely useful due to the huge size of the state space created by all PMP-related parameters and can help efficiently uncover any hidden edge case scenarios, hence the large number of iterations of this test.

A few notes:
- We use the standard post-comparison flow where Ibex trace logs are fully compared to OVPsim trace logs for correctness checking, no new reference model needs to be created since OVPsim already fully supports PMP functionality.
- No new functionality in the testbench needs to be added to support PMP testing, pretty much all of the work lies within the riscv-dv domain, hence no new parameter values under the `sim_opts` section of each YAML target.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -575,17 +575,72 @@
   sim_opts: >
     +require_signature_addr=1
 
-- test: riscv_pmp_test
+- test: riscv_pmp_basic_test
   desc: >
-    Random PMP settings
-  iterations: 1
+    Basic PMP test - all PMP regions will be configured to default setting,
+    enabling all forms of accesses, expect that no exception will be thrown.
+    Randomize mstatus.mprv.
+  iterations: 10
   gen_test: riscv_rand_instr_test
   gen_opts: >
     +instr_cnt=6000
-    +pmp_randomize=0
-    +pmp_allow_addr_overlap=0
-    +pmp_max_offset=00021000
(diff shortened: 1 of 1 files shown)


udinator:
@rswarbrick - ah sorry, I referred specifically to OVPsim just because I mostly use that ISS.
These tests do all run on Spike as well.

udinator:
@tomroberts-lowrisc - Do these tests make sense to you, and do you think any other PMP tests need to be added?

PR #917, opened 2020-06-01
[rtl] Make speculative branch optional

tomroberts-lowrisc:
- The speculative branch behaviour causes a performance degradation of
  around 3% in the max config. This change enables that behaviour only
  the maximum PMP config, which is where it is most needed for timing
  closure.

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -108,6 +108,11 @@ module ibex_core #(
   localparam int unsigned PMP_NUM_CHAN      = 2;
   localparam bit          DataIndTiming     = SecureIbex;
   localparam bit          DummyInstructions = SecureIbex;
+  // Speculative branch option, trades-off performance against timing.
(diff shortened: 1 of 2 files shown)


rtl/ibex_core.sv
```diff
@@ -108,6 +108,9 @@ module ibex_core #(
   localparam int unsigned PMP_NUM_CHAN      = 2;
   localparam bit          DataIndTiming     = SecureIbex;
   localparam bit          DummyInstructions = SecureIbex;
+  // Speculative branch option, trades off performance against timing (eases critical path but
+  // reduces performance). Only really makes sense on critical paths in a big PMP config.
```
imphil, line 112: I get what you  mean, but we could potentially make it slightly stronger which setting actually flips the coin to which side. For example: `Setting SpecBranch to 1 eases timing, but decreases CPU performance.`

tomroberts-lowrisc: Makes sense - will do

Issue #912, opened 2020-05-29
Back to back exceptions


udinator:
Hi all,
I found an interesting issue when running a PMP test, the waves are shown below:
![Screenshot from 2020-05-28 20-29-01](https://user-images.githubusercontent.com/16736281/83218111-2e8a5880-a122-11ea-8110-808a8ff98168.png)

The situation is as follows.
The instruction at `PC=0x8000921c` raises a PMP exception trying to store data to address `0x80014d04`, which is correctly propagated through the LSU and to the controller, which sees its `store_err_i` and `store_err_q` go high at the yellow marker.
The very next instruction at `PC=80009220` raises a PMP fetch exception in the IF stage, which again gets propagated to the controller correctly, as its `instr_fetch_err_i` and `instr_fetch_err` signals both go high.
The problem here arises due to the fact that `instr_fetch_err` and `store_err_q` are both high in the controller during the cycle when it transitions to the `FLUSH` state to handle an exception, causing the core to re spond to the instruction fetch exception and ignore the store fault completely due to exception prioritization.

I'm pretty certain that this can be recreated with non-PMP exceptions as it just requires exceptions raised by two sequential instructions (as long as one is a memory-related exception) to be seen by the controller on the same cycle.

GregAC:
I'll take a look at this

GregAC:
This is an issue introduced by the 3rd pipeline stage. Whilst for the same instruction an instruction fault clearly take priority over a store fault (as you cannot execute the instruction to produce a store fault if there's an instruction fault) when you have an instruction in the writeback stage any load/store fault it sees takes priority over the instruction fault seen in the instruction following. The controller fault prioritization doesn't currently understand this so the instruction fault ends up taking priority.

GregAC:
Whilst investigating this found we have another bug where if there was an outstanding load or store in writeback that had yet to get a response (so we don't know if it will cause an exception) but an instruction in ID/EX sees something that would cause an exception (e.g. a PMP failure for the instruction or an invalid instruction) we'd immediately trigger the exception for ID/EX.

Instead we need to ensure that when triggering an exception for ID/EX that anything in WB has finished so any exception it raises is correctly dealt with.

#919 fixes both the issue here and the one I've just mentioned

PR #919, opened 2020-06-01
Ibex wb exception fix

GregAC:


udinator:
Confirming that this fixes the issue I brought up in #912, thanks @GregAC for the quick fix!

PR #921, opened 2020-06-03
Teach the ICache testbench to avoid bad multi-way hits

rswarbrick:
There was a missing bit of spec that I hadn't realised about, which means the core isn't allowed to change instructions and then branch to them without an invalidation in the meantime.

This set of patches fixes things by moving the "new seed" decision to the core agent and using a FIFO to pass the information from there to the memory agent. The second patch ("*Make sure we don't see multi-way hits in icache testbench*") contains that change.

The following two patches fix bugs in the memory monitor and driver that this change uncovers. This is all to do with what happens if a seed changes, changing whether to signal a PMP error, while the cache has the request line high. To trigger the bugs, you need back-to-back branches with the same address and a seed update: it turns out that the passthru tests do this a lot!

The first patch is just a minor tidy-up, making an error message slightly more helpful.

@tomroberts-lowrisc: Could you check that my change to `icache.rst` matches what you think the interface should be?


doc/icache.rst
```diff
@@ -269,3 +269,9 @@ While doing this, the cache behaves as if ``icache_enable_i`` is false and will
 
 The ``busy_o`` signal is guaranteed to be high while the cache is invalidating its internal memories or whenever it has a pending fetch on the instruction bus.
 When the ``busy_o`` signal is low, it is safe to clock gate the cache.
+
+The cache doesn't have circuitry to avoid inconsistent multi-way hits.
+As such, the core must never fetch from an address with the cache enabled after modifying the data at that address, without first starting a cache invalidation.
+
+.. note::
+   This is a constraint on *software*, not just on the core.
```
tomroberts-lowrisc, line 277: Yep looks good to me - could reference the "Zifencei" of the unpriv spec for background if you wanted.

rswarbrick: I've just looked at that chapter and can't find the bit that says "valid code must issue a fence.i instruction before executing from code memory that might have been touched". It sounds like something that might not belong in the ISA (since it's only required by some cache architectures). Am I looking in the wrong place?

rswarbrick:
(force-push puts back a constraint on `invalidate` that I accidentally removed in the previous version of *Make sure we don't see multi-way hits in icache testbench*)

doc/icache.rst
tomroberts-lowrisc: I think the closest thing is the sentence:
" RISC-V does not guarantee that stores to instruction memory will be made visible to instruction fetches on a RISC-V hart until that hart executes a FENCE.I instruction."

rswarbrick: Yep, ok. Of course, that doesn't say "and you might get random garbage rather than *either* version of the instruction" :-)

I guess I'll just leave it as is for now. Presumably we'll have some sort of platform document at some point, which will say this sort of thing (in OT- and/or Ibex-land).

PR #923, opened 2020-06-04
Regfile parity

tomroberts-lowrisc:


doc/integration.rst
```diff
@@ -160,3 +162,13 @@ Interfaces
 |                         |                         |     | if an external event (interrupt or     |
 |                         |                         |     | debug req) wakes the core up           |
 +-------------------------+-------------------------+-----+----------------------------------------+
+| ``alert_minor_o``       | 1                       | out | Core has detected a fault which it can |
+|                         |                         |     | safely recover from. Can be used by a  |
+|                         |                         |     | system to log errors over time and     |
+|                         |                         |     | detect tampering / attack.             |
++-------------------------+-------------------------+-----+----------------------------------------+
+| ``alert_major_o``       | 1                       | out | Core has detected a fault which cannot |
```
vogelpi, line 170: Should we document if this signal is a pulse or if it remains high?

rtl/ibex_core.sv
```diff
@@ -739,29 +745,102 @@ module ibex_core #(
...
+        .d_o        (),
+        .syndrome_o (),
+        .err_o      (rf_ecc_err_a)
+      );
+      prim_secded_39_32_dec regfile_ecc_dec_b (
+        .in         (rf_rdata_b_ecc),
+        .d_o        (),
+        .syndrome_o (),
+        .err_o      (rf_ecc_err_b)
+      );
```
vogelpi, line 780: Nit: indentation

rtl/ibex_core.sv
```diff
@@ -739,29 +745,102 @@ module ibex_core #(
...
+    // Assign read outputs - no error correction, just trigger an alert
+    assign rf_rdata_a = rf_rdata_a_ecc[31:0];
+    assign rf_rdata_b = rf_rdata_b_ecc[31:0];
+
+    // Calculate errors - qualify with WB forwarding to avoid xprop into the alert signal
+    assign rf_ecc_err_a_id = |rf_ecc_err_a & rf_ren_a & ~rf_rd_a_wb_match;
+    assign rf_ecc_err_b_id = |rf_ecc_err_b & rf_ren_b & ~rf_rd_b_wb_match;
+
+    // Combined error
+    assign rf_ecc_err_comb = instr_first_cycle_id & (rf_ecc_err_a_id | rf_ecc_err_b_id);
```
vogelpi, line 791: I think the use of `instr_first_cycle_id` here is not sufficient. This means we can only signal an error during the first cycle of an instruction. But there are instructions that need the regfile output during multiple cycles:
- Some bitmanip instructions, e.g., the ternary ones read rs1, rs2 in the first cycle, and rs3 in the second cycle.
- (Misaligned) load/stores (I think but didn't check these) also rely on the address to remain correct in further cycles, maybe the request is not granted immediately, or to compute the incremented address for the second half.

Thus, I think we should qualify the error signals with something like `instr_executing` (from ID stage) or not at all. Question: Is it problematic to always generate an alert in case of an ECC error, even the core does not use the value?

rtl/ibex_id_stage.sv
```diff
@@ -234,10 +234,8 @@ module ibex_id_stage #(
   logic        rf_we_dec, rf_we_raw;
   logic        rf_ren_a, rf_ren_b;
 
-`ifdef RVFI
```
vogelpi, line 237: I noted that we have `unused_rf_ren_a/b` signals inside `ibex_id_stage.sv` (if the writeback stage is disabled). Now that `rf_ren_a/b` are always forwarded, we could remove those.

doc/integration.rst
tomroberts-lowrisc: Yep will do - they both should be pulses (although the major alert pulse might end up being multiple cycles)

rtl/ibex_core.sv
tomroberts-lowrisc: Will fix

rtl/ibex_core.sv
tomroberts-lowrisc: Good point. Most of the qualification is there for xprop reasons. The register file can be X after reset, so the qualifications are there to stop those Xs getting into the alert signals.
Beyond stopping the xprop, I don't think it matters if we signal the alert when the core isn't necessarily using the read value. A straightforward instr_valid_i qualification should be fine. It means the alert might be high for a few cycles rather than one, but that doesn't matter since the alert is fatal.

rtl/ibex_id_stage.sv
tomroberts-lowrisc: Good spot - will remove

doc/integration.rst
vogelpi: Thanks.

rtl/ibex_core.sv
vogelpi: Thanks @tomroberts-lowrisc for explaining the situation and for changing the implementation. Using `instr_valid_i` seems to be the correct thing to do.

rtl/ibex_core.sv
vogelpi: Sorry Tom, just the indentation of the module name needed correction. Now also the port lists have changed. The indentation for those should be 2 spaces (instead of 4).

tomroberts-lowrisc: thanks - will fix

PR #925, opened 2020-06-04
Fix Writeback stage lint errors

eunchan:
- unpacked array decending order
- unspecified generate block


imphil:
Thanks Eunchan!

PR #927, opened 2020-06-05
[doc] Clarify fetch_enable_i meaning

tomroberts-lowrisc:


doc/integration.rst
```diff
@@ -153,7 +153,11 @@ Interfaces
 +-------------------------+------------------------------------------------------------------------+
 | ``debug_*``             | Debug interface, see :ref:`debug-support`                              |
 +-------------------------+-------------------------+-----+----------------------------------------+
-| ``fetch_enable_i``      | 1                       | in  | Enable the core, won't fetch when 0    |
+| ``fetch_enable_i``      | 1                       | in  | Trigger the core to start fetching and |
+|                         |                         |     | executing instructions after reset.    |
+|                         |                         |     | Once triggered, the core will continue |
+|                         |                         |     | executing until the next reset,        |
+|                         |                         |     | regardless of the value of this pin.   |
```
rswarbrick, line 160: I'm not completely sure about the wording, because it sounds a bit like this might be edge triggered ("To start the core, I have to set `rst_ni` true and then do a posedge on `fetch_enable_i`"). Maybe something like this? "When it comes out of reset, the core won't start fetching until it sees `fetch_enable_i` true. Once started, the will continue executing until the next reset, regardless of the value of this pin."

tomroberts-lowrisc: Sounds good - I'll reword to that effect

PR #928, opened 2020-06-05
Fix some interrupt related issues found in OT chip tests

tomroberts-lowrisc:


udinator:
thanks for getting a fix out for this issue!
I will test these changes on Monday and see if I can reproduce the issue in lowrisc/opentitan#2405.

rtl/ibex_id_stage.sv
```diff
@@ -882,12 +863,15 @@ module ibex_id_stage #(
       instr_valid_i & ~instr_kill & ~instr_executing |-> stall_id)
 
     // Stall for reasons related to memory:
-    // * Requested by LSU (load/store in ID/EX needs to be held in ID/EX whilst a data request or
```
GregAC, line 885: May need re-writing but I think this point should be left in the comment as it's distinct from needing to stall for exceptions to resolve and needing to stall because we can't send a request yet.

rtl/ibex_id_stage.sv
```diff
@@ -882,12 +863,15 @@ module ibex_id_stage #(
...
-    //   granted or the second half of a misaligned access is sent out)
-    // * LSU access required but data requests aren't currently allowed
     // * There is an outstanding memory access that won't resolve this cycle (need to wait to allow
     //   precise exceptions)
-    assign stall_mem = instr_valid_i & ((lsu_req_dec & (~data_req_allowed | (~data_req_complete_q & ~lsu_req_done_i))) | outstanding_memory_access);
+    // * There is a load/store request not being granted (needs to stay in ID for the address
+    //   calculation)
+    assign stall_mem = instr_valid_i & (outstanding_memory_access | (lsu_req_dec & ~lsu_req_done_i));
```
GregAC, line 870: So I think this works though I'm hesitant about leaving out `data_req_complete_q`. `lsu_req_done_i` will only be asserted for the cycle the load/store is clear to proceed to WB. If it doesn't `lsu_req_done_i` will drop and the instruction will stall, forever I think.

As far as I can tell we won't actually get this scenario as the only reason for it not to enter WB is another load/store is in WB and awaiting a result so when we get `lsu_req_done_i` we must be clear to enter WB.

Though there's always the possibility of future stalling behaviour that breaks this.

Perhaps we don't need to factor in `data_req_complete_q` for now to cover this (currently non-existent) case but an assertion may be prudent? Something that states if a load/store sees `lsu_req_done_i` then it must be proceeding to WB the same cycle.

rtl/ibex_controller.sv
```diff
@@ -483,7 +486,7 @@ module ibex_controller #(
           // Halt IF but don't flush ID. This leaves a valid instruction in
           // ID so controller can determine appropriate action in the
           // FLUSH state.
-          halt_if = 1'b1;
+          retain_id = 1'b1;
```
GregAC, line 489: Looking at the other places `halt_if` is used I think we might instead want `retain_id`, e.g. in the SLEEP state the next instruction is already in the ID instruction flop so we don't want to clear it out. Are there actually any scenarios we want to clear the ID instruction flop but don't want IF to continue (i.e. `retain_id` == 0 and `halt_if` == 1)? The existing `flush_id` exists to deal with this cases which ignores `halt_if` for the purposes of clearing the ID instruction register.

So I think this ends up being a straight rename from `halt_if` -> `retain_id` and I prefer the former naming.

rtl/ibex_id_stage.sv
```diff
@@ -882,12 +863,15 @@ module ibex_id_stage #(
...
+
+    // Indicate to the controller that an lsu req is in ID stage - we cannot handle interrupts or
+    // debug requests until the load/store completes
+    assign stall_ld  = instr_valid_i & lsu_req_dec;
```
GregAC, line 874: As this isn't a generic stall signal, but rather used for a specific case could we rename it, maybe something like `lsu_req_in_id` or `stall_ld_st_on_irq_or_dbg`? (In particular this isn't just loads so `stall_ld` doesn't suffice).

rtl/ibex_controller.sv
GregAC: Thanks for the explanation, all makes sense. I had misread the SLEEP state RTL (comment there refers to instruction being in IF not ID).

Looking again at the controller FSM I wonder if the `FIRST_FETCH` state is really required, in particular I doubt we test it properly right now (in terms of interrupt or debug request when we're in that state).

We should also look at some sleep/wakeup tests and how that all works.

PR #936, opened 2020-06-08
Add interface-level functional coverage to icache tests

rswarbrick:
This doesn't define points for interesting behaviours that we're trying to hit with special tests in the testplan (I'll do that in a follow-up). It also doesn't define points for uarch coverage, which will use slightly different machinery.

This PR depends on PR #935 for the first patch: let's review that commit there, rather than here, if necessary.

@tomroberts-lowrisc: Can you think of any other sensible interface-level coverage points that we should define? I'd have thought that most "weird timing" coverage points should be uarch-level, because it will be easier to describe them there, but maybe I've missed something?


tomroberts-lowrisc:
Yep those look good to me.
Slightly unrelated note on the branch_spec change - we lost the TODO comment, but there is still a TODO to drive that signal at some point.

rswarbrick:
Good point about the branch_spec signal. I'll probably work on that this afternoon, so the missing TODO blip shouldn't last too long. :-/

PR #937, opened 2020-06-08
[doc] Add top-level status table

tomroberts-lowrisc:
- Some basic info about a few possible design points and their
  area/perf/dv status

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,38 @@ the RV32IMC instruction set architecture.
 
 <p align="center"><img src="doc/images/blockdiagram.svg" width="650"></p>
 
-Ibex offers several configuration parameters to meet the needs of various application scenarios.
-The options include two different choices for the architecture of the multiplier and divider unit,
-as well as the possibility to drop the support for the "M" extension completely. In addition, the
-"E" extension can be enabled when opting for a minimum-area configuration.
-
 This core was initially developed as part of the [PULP platform](https://www.pulp-platform.org)
 under the name "Zero-riscy" \[[1](https://doi.org/10.1109/PATMOS.2017.8106976)\], and has been
 contributed to [lowRISC](https://www.lowrisc.org) who maintains it and develops it further. It is
 under active development, with further code cleanups, feature additions, and test and verification
 planned for the future.
+
+## Configuration
+
+Ibex offers several configuration parameters to meet the needs of various application scenarios.
+The options include different choices for the architecture of the multiplier unit, as well as a range of performance and security features.
+The table below indicates performance, area and verification status for a few selected configurations.
+These are configurations on which lowRISC is focusing for performance evaluation and design verification (see [supported configs](ibex_configs.yaml)).
+
+| Config | "small" | "maxperf" | "maxperf-pmp-bm" |
+| ------ | ------- | --------- | ---------------- |
+| Features | RV32IMC, 3 cycle mult | RV32IMC, 1 cycle mult, Branch target ALU, Writeback stage | RV32IMCB, 1 cycle mult, Branch target ALU, Writeback stage, 16 PMP regions |
+| Performance (Coremark/MHz) | 2.44 | 3.09 | 3.09 |
+| Area - Yosys (kGE) | 33.15 | 39.03 | 63.32 |
+| Area - Commercial (estimated kGE) | ~27 | ~31 | ~50 |
+| Verification status | Green | Amber | Amber |
+
+Notes:
+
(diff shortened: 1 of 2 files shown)


tomroberts-lowrisc:
The table is pretty horrible - if anyone knows a better way of formatting it in github markdown, let me know!

imphil:
No strong opinion one way or the other. We'll want more information about the configurations in the documentation in the medium term, but having the basic numbers in the README doesn't hurt either. 

As for markdown tables: no, there's no better way (except for aligning things slightly more). I don't think inline HTML in Markdown is supported in GitHub READMEs.

asb:
I quite like having it in the README and it doesn't look too ugly to me (at least viewing the rich diff in GitHub).

tomroberts-lowrisc:
Yeah I wanted something right on the "front page" so to speak. Agree though that it makes sense to also create something a bit more in-depth to go in the main documentation at some point.

GregAC:
> I quite like having it in the README

I agree, I think it's quite important such a table does appear in the README, it's the first thing someone will see when checking out Ibex and begins to answer some of the questions anyone thinking of using the core will have without them having to go into the documentation proper. We can then expand on things in the documentation (potentially replicating the table, which is a bit of faff to keep them in sync but I think justifiable).

README.md
```diff
@@ -7,17 +7,34 @@ the RV32IMC instruction set architecture.
...
+Ibex offers several configuration parameters to meet the needs of various application scenarios.
+The options include different choices for the architecture of the multiplier unit, as well as a range of performance and security features.
+
+| Config | "small" | "maxperf" | "maxperf-pmp-bm"  |
+| ------ | ----------------- | ----------------- | ----------------- |
+| Features | RV32IMC, 3 cycle mult | RV32IMC, 1 cycle mult, Branch target ALU, Writeback stage | RV32IMCB, 1 cycle mult, Branch target ALU, Writeback stage, 16 PMP regions |
+| Performance (Coremark/MHz) | 2.44 | 3.09 | 3.09 |
+| Area - Yosys (kGE) | 33.15 | 39.03 | 63.32 |
+| Area - Commercial (estimated kGE) | 24.00 | 29.00 | 46.00 |
+| Verification status | Green | Amber | Amber |
```
GregAC, line 27: We should have some description of what we mean by 'Green' and 'Amber' (and presumably 'Red'), maybe better off in the documentation rather than the READMe

README.md
```diff
@@ -7,17 +7,34 @@ the RV32IMC instruction set architecture.
...
-
 This core was initially developed as part of the [PULP platform](https://www.pulp-platform.org)
 under the name "Zero-riscy" \[[1](https://doi.org/10.1109/PATMOS.2017.8106976)\], and has been
 contributed to [lowRISC](https://www.lowrisc.org) who maintains it and develops it further. It is
 under active development, with further code cleanups, feature additions, and test and verification
 planned for the future.
 
+## Configuration
+
+Ibex offers several configuration parameters to meet the needs of various application scenarios.
```
GregAC, line 18: We should make it clear there are many possible configs it's just the three listed ones are 'supported' (which we also need to define!)

README.md
```diff
@@ -7,17 +7,39 @@ the RV32IMC instruction set architecture.
...
+The options include different choices for the architecture of the multiplier unit, as well as a range of performance and security features.
+The table below indicates performance, area and verification status for a few selected configurations.
+These are configurations on which lowRISC is focusing for performance evaluation and design verification (see [supported configs](ibex_configs.yaml)).
+
+| Config | "small" | "maxperf" | "maxperf-pmp-bm" |
+| ------ | ------- | --------- | ---------------- |
+| Features | RV32IMC, 3 cycle mult | RV32IMC, 1 cycle mult, Branch target ALU, Writeback stage | RV32IMCB, 1 cycle mult, Branch target ALU, Writeback stage, 16 PMP regions |
+| Performance (Coremark/MHz) | 2.44 | 3.09 | 3.09 |
+| Area - Yosys (kGE) | 33.15 | 39.03 | 63.32 |
+| Area - Commercial (estimated kGE) | 24.00 | 29.00 | 46.00 |
```
imphil, line 28: Nit: potentially remove the ".00" part from these numbers to avoid giving the impression of accuracy where there is none.

rswarbrick: I agree. If the result looks weird when left-aligned, you could also try prefixing with "~" or something.

tomroberts-lowrisc: Good point - will do. I might bump them up a bit too given the feedback from Bert.

PR #938, opened 2020-06-08
Add missing dependency on python3-bitstring to python-requirements

### LINKED ISSUE #930: Missing dependency on python3-bitstring
The riscv-dv code uses it and I don't think we have it in python-requirements.

rswarbrick:
Closes #930.

### DIFF
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -13,3 +13,6 @@ mako
 
 # Needed by dvsim.py (not actually used in Ibex)
 premailer
+
+# Recurse to get any requirements from riscv-dv
+-r vendor/google_riscv-dv/requirements.txt


udinator:
We already include the `bitstring` package in the requirements.txt file in the `riscv-dv` repository, should it be duplicated here also?

python-requirements.txt
```diff
@@ -13,3 +13,6 @@ mako
 
 # Needed by dvsim.py (not actually used in Ibex)
 premailer
+
+# Needed by riscv-dv
+bitstring
```
imphil, line 18: Following @udinator's sugggestion, we can do that:

```suggestion
-r vendor/google_riscv-dv/requirements.txt
```

rswarbrick: Oh, neat! I didn't know that pip did that. I'll push up an amended version.

rswarbrick:
Yeah. For now, the requirements won't be vendored in. Maybe we should change the OT top-level to include a `util/dvsim/python-requirements.txt`, which would let us do that?

imphil:
Something along those lines would be nice one day, but there's no immediate need to do that right now.

PR #940, opened 2020-06-09
Added Gift cipher for pointer authentication

flaviens:
We have shown that the Gift cipher is among the most suitable ciphers for lightweight pointer authentication. This is an iterative decomposition of the cipher with 5 cycles latency (including the input cycle and the output cycle).


vogelpi:
Hi @flaviens,
thanks for the PR! Could please add the following license header to all files?
```
// Copyright lowRISC contributors.
```

vogelpi:
Thanks @flaviens for adding the license headers. Would you mind to squash/fixup that third commit into the first one such that the headers are there right from the beginning?

flaviens:
Alright, it's done now. All my apologies for the mistake and late response.

PR #942, opened 2020-06-11
Add ibex_rv_trace module template to implement risc5_trace_specs

timeprints:
This is initial trace module interface (to be compatible with riscv-trace-spec doc at https://github.com/riscv/riscv-trace-spec/blob/master/riscv-trace-spec.pdf ). 
The aim is to check IO's and agree about them. 
A core module which includes this tracer is also added (ibex_core_rv_tr.sv).
Signed-off-by: Rehab Massoud <massoud@uni-bremen.de>


timeprints:
The trace_signals specified by the riscv-trace-spec should be implemented in ibex_core. 
Files here should be added only later as the encoder and a top module containing it and the core with its new added trace-signals. So I'll close this pull-request and initiate another one with what is actually required in the first step.

PR #943, opened 2020-06-12
[rtl] Use generate loop in FF register file

GregAC:
By giving each register its own always_ff block clock gating is more
obvious to synthesis tools.

This also includes some minor naming tweaks to make use of the _q
convention for flops.


imphil:
Would it make sense to do the same change for the latch-based one, they seem to be rather similar.

GregAC:
If nothing else it's good for consistency so sure.

PR #945, opened 2020-06-12
Simplify timestamps in ICache tests

rswarbrick:
This patch updates the imported OpenTitan code to get support and then modifies the ICache testbench to dump waves which are aligned to 10ns boundaries.

@imphil: I had to squash two commits to get the vendoring right here (add a mapping item, then run `vendor.py -U`). Is there something cleverer I could have done?


imphil:
> I had to squash two commits to get the vendoring right here (add a mapping item, then run vendor.py -U). Is there something cleverer I could have done?

No, that's fine.

PR #946, opened 2020-06-12
[rtl] Use outer generate loop for latch RF

GregAC:
This matches the style used in the FF based register file. It gives each
register its own always block with a single enable rather than having
multiple registers with enables in a single always block.


GregAC:
I test this with a UVM DV run and all looks well

PR #947, opened 2020-06-12
[rtl] Simplify I$ ECC error handling

tomroberts-lowrisc:
- Remove the timing optimisations that delay the factoring-in of ecc
  errors into valid_o.
- Optimisations are probably unnecessary here due to the minimal logic
  hanging off valid_o, and the optimisations cause protocol checker
  violations.

### DIFF
--- a/rtl/ibex_icache.sv
+++ b/rtl/ibex_icache.sv
@@ -562,9 +562,8 @@ module ibex_icache #(
                                  (fill_cache_q[fb] & fill_busy_q[fb] &
                                   icache_enable_i & ~icache_inval_i);
     // Record whether the request hit in the cache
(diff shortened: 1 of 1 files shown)


rtl/ibex_icache.sv
```diff
@@ -866,7 +865,7 @@ module ibex_icache #(
   // must be used here rather than fill_out_req because data can become valid out of order
   // (e.g. cache hit data can become available ahead of an older outstanding miss).
   // Any ECC error suppresses the output that cycle.
```
rswarbrick, line 867: Ooh, does this comment need changing?

tomroberts-lowrisc: Yep - will fix

PR #948, opened 2020-06-12
Add ECC UVM tests for the ICache

rswarbrick:
There are three commits in this PR. The first just turns ECC on. This is a little tricky to do in `dvsim.py`, not sure whether there's a neater approach.

The other commits add ECC testing by allowing the SRAMs inside the icache to sometimes respond with single-bit errors. Commit 2 does the "design-side" work: it defines a RAM wrapper that can be asked to stomp on bits and instantiates it between the icache and the underlying RAMs. Commit 3 does the "UVM-side" work. We make an agent for each RAM, binding its interface into the design to allow it to change the internal signals. These agents' sequencers only get given sequences if ECC errors are enabled, which we currently just do in a new 'ibex_icache_ecc' test.

### DIFF
--- a/dv/uvm/icache/dv/env/ibex_icache_env.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_env.sv
@@ -13,6 +13,9 @@ class ibex_icache_env extends dv_base_env #(
   ibex_icache_core_agent core_agent;
   ibex_icache_mem_agent  mem_agent;
 
+  ibex_icache_ecc_agent  ecc_tag_agents[];
+  ibex_icache_ecc_agent  ecc_data_agents[];
+
   // Heartbeat tracking
   uvm_callbacks_objection           hb_objection;
   uvm_heartbeat                     heartbeat;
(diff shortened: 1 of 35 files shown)


rswarbrick:
Hmm. This ain't gonna work until I fix how the "fake techmap" gets used. At the moment, completely unrelated tests (such as `fusesoc --cores-root=. run --target=sim --tool=verilator lowrisc:ibex:tb_cs_registers`) are picking it up and failing.

Tom has also pointed out to me that this should generate 2-bit errors too. Easy enough, and I'll amend this PR accordingly.

I'll "un-draft" once this looks like it might work.

rswarbrick:
Force-push adds 2-bit errors. The fusesoc problem is not fixed.

rswarbrick:
I've done some digging. The problem is that if you define something as a "techlib" like patch no. 2 does, you'll pick it up with primgen on any fusesoc run in the tree. That's not necessarily a problem, but my code used `wor` (wired-or), which isn't supported by Verilator. Now when you try to build a completely unrelated test, Verilator will be asked to parse all the code you added in your techlib. In this case... boom.

Short-term fix: Figure out how to do the "bound interface controls outside signal" trick without using `wor`. Longer-term fix: Is there a way for fusesoc/primgen to figure out (or be told) whether to include a techlib?

udinator:
In my first pass through this PR I mainly just looked at the UVM code, in general everything looks good!
One question I have is does the scoreboard not need to know about transactions from the ECC agents?

rswarbrick:
Hi Udi, If the icache gains an alert interface, we'll need to test it (so the scoreboard will need to know when we've done something naughty). Until then, a bad ECC check basically just looks like a cache miss, so it's fine for it to be invisible to the scoreboard. The only slight wrinkle is that failing lots of ECC checks means that the cache would sometimes fail the "am I actually caching enough" check, so we turn that check off for the ECC test.

rswarbrick:
Latest version surrounds the guts of the "badbits" module with an `ifndef VERILATOR` (as suggested by Philipp). @imphil: Does this look sensible to you?

dv/uvm/icache/dv/env/ibex_icache_env.sv
```diff
@@ -29,6 +32,23 @@ class ibex_icache_env extends dv_base_env #(
...
 
+    // If ECC is enabled, create ECC agents for the RAMs. We have already created config objects for
+    // them (the test called create_ecc_agent_cfgs in its build_phase method), so we just have to
+    // make an agent for each config. In practice, there will be the same number of tag and data
+    // agents, but this code doesn't really care.
+    ecc_tag_agents  = new[cfg.ecc_tag_agent_cfgs.size()];
+    for (int unsigned i = 0; i < cfg.ecc_tag_agent_cfgs.size(); i++) begin
+      string tname = $sformatf("ecc_tag_agents[%0d]", i);
+      ecc_tag_agents[i] = ibex_icache_ecc_agent::type_id::create(tname, this);
+      uvm_config_db#(ibex_icache_ecc_agent_cfg)::set(this, {tname, "*"}, "cfg", cfg.ecc_tag_agent_cfgs[i]);
```
udinator, line 43: Here and below, please format `uvm_config_db` calls like this:
```systemverilog
if (!(uvm_config_db#(...)::set(...))) begin
  `uvm_fatal(`gfn, ...)
end

rswarbrick: `uvm_config_db#(...)::set(..)` is a function returning `void`, so I don't think that's right. There are calls to `get` in `ibex_icache_ecc_agent.sv` and `ibex_icache_base_test.sv`, both of which do the "fatal dance". I *think* this was probably just a slip-up, so I'm going to merge. Let me know (here or in an issue) if I've missed something, and I'll fix it with a follow-up commit.

udinator: Yep, you're absolutely right! That's a silly slip up on my end.

PR #952, opened 2020-06-14
[ibex/dv] add Questa support

bp-ouhaha:
my trial to add Mentor QuestaSim support for ibex_core regression


bp-ouhaha:
adding Questa support

PR #955, opened 2020-06-15
Drive the branch_spec line in ICache UVM tests

rswarbrick:
This signal already got driven (to 1) when signalling a branch with
the interface's branch_to task. This patch now drives the `branch_spec`
line occasionally even if we don't actually do a branch. (One cycle in
64, for now).

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_driver.sv
+++ b/dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_driver.sv
@@ -59,12 +59,24 @@ class ibex_icache_core_driver
     // Drive the enable state
     cfg.vif.driver_cb.enable <= req.enable;
 
+    // Branch, invalidate and set new seed immediately. When the branch has finished, read num_insns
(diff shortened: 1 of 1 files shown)


dv/uvm/icache/dv/ibex_icache_core_agent/ibex_icache_core_driver.sv
```diff
@@ -110,6 +133,14 @@ class ibex_icache_core_driver
     end
   endtask
 
+  // Randomly drive the branch_spec line one cycle in 64. Never returns.
+  task automatic drive_branch_spec();
+    forever begin
+      cfg.vif.driver_cb.branch_spec <= $urandom_range(64) == 0;
```
tomroberts-lowrisc, line 139: Does this get driven somewhere else to satisfy the constraint `branch |-> branch_spec`?

tomroberts-lowrisc: Ok thanks - looks good to me. I'll leave the code for Udi to comment.

PR #958, opened 2020-06-16
Tweak ICache tests to fill functional coverage defined so far

rswarbrick:
There are 4 patches in the series.

1. Make it more likely that a branch will land at the edge of the address space (to check wrapping)
1. Allow the ready signal to be high when a branch hits (this has no effect, but is allowed by the spec, so we may as well do it sometimes).
1. Make the "cancelled_valid" sequence spot what it was supposed to spot
1. Fix the wiring downstream of the "cancelled_valid" sequence (I was turning on the monitor after the run phase had completed, which meant it didn't see very much!).

### DIFF
--- a/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
@@ -604,7 +604,7 @@ class ibex_icache_scoreboard
 
   // Register an instruction fetch with the caching tracking window
   function automatic void window_take_insn(bit [31:0] addr, bit err);
-    bit [31:0]   window_width;
+    bit [32:0]   window_width;
     int unsigned fetch_ratio_pc;
 
     // Ignore instructions if this check is disabled in the configuration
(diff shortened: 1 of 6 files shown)


rswarbrick:
Force-push adds an extra constraint to base addresses in the first patch.

rswarbrick:
Force-push rebases and adds a fix for an integer overflow.

dv/uvm/icache/dv/env/ibex_icache_scoreboard.sv
```diff
@@ -604,7 +604,7 @@ class ibex_icache_scoreboard
 
   // Register an instruction fetch with the caching tracking window
   function automatic void window_take_insn(bit [31:0] addr, bit err);
-    bit [31:0]   window_width;
+    bit [32:0]   window_width;
```
udinator, line 607: I'm not quite sure I understand this, how does expanding this to 33 bits help?

rswarbrick: The overflow that was happening was with `window_range_hi = 32'hfffffffe` and `window_range_lo = 32'h00000000` (you see this when something "only just wraps"). Then:

    (32'hfffffffe - 32'h00000000 + 3) / 4 = 31/4

which is a very small number, rather than ~1 billion which is what we should have seen.

Adding an extra bit at the top means that `{1'b0, 32'hfffffffe} + 33'd3` doesn't overflow.

udinator: Makes sense to me.

PR #960, opened 2020-06-16
Linter action with LibreCores CI

wallento:
This adds a beta Github action that is run on pushes and pull
requests. It automatically annotates the sources with linter issues.

Nothing extra is needed beyond merging.

As a demo you can see how new PRs will fail: https://github.com/wallento/ibex/pull/4
They will then show the annotation: https://github.com/wallento/ibex/pull/4/commits/bbf8e108bb313655544ebb490ff9816ffb981b87

Removing the current waiver looks like this as a test: https://github.com/wallento/ibex/pull/5/commits/966649b759d146e088b20112503524adc8c8c6fc
So, you can see that it will also annotate files outside the changeset.

There will be robustness extensions, but I think it is usable and I would love ibex as beta testers. It can be deactivated or removed as easily as it was added :)


wallento:
Links are broken by github autolinking, the second one is accessible via https://github.com/wallento/ibex/pull/5

imphil (commented):
Uh, turns out I completely forgot about this; sorry @wallento! 

To clarify, this is using Verilator as linter and posts comments as code annotations, while https://github.com/lowRISC/ibex/pull/1427 uses Verible lint and posts lint output as comments (which we can then reply to). 

Effectively the two PRs do two different things in two different ways. Since it is not really clear a the moment which approach to present linter output to the user is better, I'm happy to give both approaches a try and we can then improve as we move along.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
+on: [push, pull_request]
```
imphil, line 1: Can you add a license header here?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    name: Linter
```
imphil, line 6:
```suggestion
    name: RTL lint with Verilator
```

PR #962, opened 2020-06-16
Install requirements for documentation builds

imphil:
A couple more Python requirements are needed to build documentation, but
we don't include them in the main requirements.txt yet. Keep them
separate for RTD to use, but also include them in the main requirements
for our developers to use.


imphil:
Argh, of course pip cannot merge requirements ... Let's see what other solutions we have.

imphil:
This approach clearly doesn't work, so let's keep it as-is until I find a better solution.

PR #968, opened 2020-06-17
Add ibex_icache_stress_all* tests

rswarbrick:
EDIT: This PR grew somewhat. I've updated the message to match.

This PR contains 12 commits and ends by adding the `ibex_icache_stress_all` and `ibex_icache_stress_all_with_reset` tests.

The first 8 commits are general clean-ups and bug fixes.

Following them is the first commit to add a test (`ibex_icache_stress_all`). This turned up a couple of corner cases which are fixed in the following two commits: I've done it in this order because those changes don't really make much sense if you do them without back-to-back sequences.

The final commit adds the `stress_all_with_reset` test. This is almost the same as the stress_all test, but sometimes resets the DUT unexpectedly in the middle of what it was doing.


rswarbrick:
Force-push adds 8th patch to allow resets.

dv/uvm/icache/dv/env/seq_lib/ibex_icache_combo_vseq.sv
```diff
@@ -68,8 +68,7 @@ class ibex_icache_combo_vseq
       // little.
       trans_now = $urandom_range(50, 100);
 
-      // TODO: Inject some resets
-      should_reset = (seqs_so_far == 0);
+      should_reset = $urandom_range(0, 1);
```
udinator, line 71: These resets will occur at the start of each vseq, correct?

rswarbrick: We reset between sequences with 50% probability, so that we test the "start doing something else without resetting" case as well as "reset and start again". The latter is probably much less good at provoking bugs, but will give some missing toggle coverage.

udinator: Got it.

rswarbrick:
Cool, thanks for the review. I've actually found a couple of minor bugs that I'll try to finish tracking down today (with how we reset and/or chain information between sequences). I'll re-request a review when that's sorted.

Re injecting resets into the middle of sequences: yep, I've got that locally and it will be the ibex_icache_stress_all_with_reset test.

rswarbrick:
Hi @udinator! I've pushed up the rest of this chain of patches to this PR because they are all sort of the same thread of development. Sorry for the extra work! The patches that you haven't already seen are:

1. 7ba2078: Use dv_base_vseq's num_trans field rather than making our own
1. 516183f: Remove empty tasks from ibex_icache_base_vseq.sv
1. 1e43e9f: Tidy up properly after overriding class in ICache back_line_seq
1. 0411fec: Fix ICache caching window test with combination sequences
1. 2dc21f0: Invalidate in an ICache sequence after a change to mem_err_shift
1. 38d1975: Add a stress_all_with_reset ICache test

rswarbrick:
Great, thanks @udinator.

Issue #30, opened 2019-05-08
Cleanup: Remove REG_* defines from ibex_id_stage


imphil:
Can we get rid of those defines and replace them by parameters or structs?

wallento:
I was wondering about the same topic when I looked at the muldiv_fast yesterday.

`ibex_defines` is supposedly the best way to store them.

All occurrences of defines in the RTL code:

```
ibex_cs_registers.sv:  `define MSTATUS_UIE_BITS        0
ibex_cs_registers.sv:  `define MSTATUS_SIE_BITS        1
ibex_cs_registers.sv:  `define MSTATUS_MIE_BITS        3
ibex_cs_registers.sv:  `define MSTATUS_UPIE_BITS       4
ibex_cs_registers.sv:  `define MSTATUS_SPIE_BITS       5
ibex_cs_registers.sv:  `define MSTATUS_MPIE_BITS       7
ibex_cs_registers.sv:  `define MSTATUS_SPP_BITS        8
ibex_cs_registers.sv:  `define MSTATUS_MPP_BITS    12:11
ibex_id_stage.sv:`define REG_S1 19:15
ibex_id_stage.sv:`define REG_S2 24:20
ibex_id_stage.sv:`define REG_D  11:07
ibex_multdiv_fast.sv:`define OP_L 15:0
ibex_multdiv_fast.sv:`define OP_H 31:16
ibex_tracer.sv:`define REG_S1 19:15
ibex_tracer.sv:`define REG_S2 24:20
ibex_tracer.sv:`define REG_S3 29:25
ibex_tracer.sv:`define REG_D  11:07
```

imphil:
Good catch @wallento. Yes, those MSTATUS_* defines should go as well and be replaced by an mstatus enum most likely, making these bit defines obsolete.

imphil:
The MSTATUS_* defines are gone in 5b63ae9.

raulbehl:
How about those REG_* defines? Would it be worth adding an `ibex_defines.svh` file for such instruction index defines?

bp-ouhaha:
If I'm correct, I only find REG_* in the ibex_decoder.sv

We could just define this (permanent assign):
  // Source/Destination register instruction index
  logic [4:0] instr_rs1;
  assign instr_rs1 = instr[19:15];
  logic [4:0] instr_rs2;
  assign instr_rs2 = instr[24:20];
  logic [4:0] instr_rs3;
[...]

PR #969, opened 2020-06-17
fix for #30

bp-ouhaha:

for fox issue https://github.com/lowRISC/ibex/issues/30

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -3,11 +3,6 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// Source/Destination register instruction index
-`define REG_S1 19:15
-`define REG_S2 24:20
-`define REG_S3 31:27
-`define REG_D  11:07
 
 /**
  * Instruction decoder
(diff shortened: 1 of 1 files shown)


rtl/ibex_decoder.sv
```diff
@@ -111,6 +106,15 @@ module ibex_decoder #(
...
   logic [31:0] instr_alu;
+  // Source/Destination register instruction index
+  logic [4:0] instr_rs1;
+  assign instr_rs1 = instr[19:15];
+  logic [4:0] instr_rs2;
+  assign instr_rs2 = instr[24:20];
+  logic [4:0] instr_rs3;
+  assign instr_rs3 = instr[31:27];
+  logic [4:0] instr_rd;
+  assign instr_rd = instr[11:07];
```
GregAC, line 117: You can drop the leading 0 in `07`

rtl/ibex_decoder.sv
```diff
@@ -111,6 +106,15 @@ module ibex_decoder #(
 
   logic [31:0] instr;
   logic [31:0] instr_alu;
+  // Source/Destination register instruction index
+  logic [4:0] instr_rs1;
```
GregAC, line 110: Rather than intermix the declarations and assigns here can we leave the declarations here and move the assigns further down to an appropriate point. I think they'd best fit around line 135 with the 'immediate extraction and sign extraction' assigns.

rtl/ibex_decoder.sv
bp-ouhaha: inherited from the past, removed now.

updated commit message too.

rtl/ibex_decoder.sv
bp-ouhaha: move it to line around line 140.

PR #974, opened 2020-06-19
[rtl] disable clock between reset and fetch_enable_i

bp-ouhaha:
Fixes lowRISC#957


bp-ouhaha:
I was wondering the same. will remove the line.

vogelpi:
Thanks @bp-ouhaha , I ran a first regression run and there are a couple of failures. Not sure if is related to this change though. I am currently running the same tests without your commit to see if the failures were there before already.

PR #975, opened 2020-06-19
Add some basic protocol checking to the icache's RAM interface

rswarbrick:
Since we are binding in an interface anyway, we can add some SV
assertions to make sure nothing too strange is happening.

Note that they aren't as strong as you might expect: we don't check
that rdata isn't X, for example. This is because the cache makes
speculative reads, which it (hopefully) ignores if the data is
invalid.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_ecc_agent/ibex_icache_ecc_protocol_checker.sv
+++ b/dv/uvm/icache/dv/ibex_icache_ecc_agent/ibex_icache_ecc_protocol_checker.sv
@@ -1,0 +1,37 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// An interface that contains SV assertions to check that the icache <-> SRAM interface is being
+// driven correctly.
+//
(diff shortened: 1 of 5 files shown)


dv/uvm/icache/dv/ibex_icache_ecc_agent/ibex_icache_ecc_protocol_checker.sv
```diff
@@ -0,0 +1,37 @@
...
+   input logic [31:0]  addr,
+   input logic [127:0] wdata,
+   input logic [127:0] wmask,
+   input logic [127:0] rdata);
+
+  // The req line should never be unknown. Note that we have no reset signal here, so we assert this
+  // is true at all times (fortunately, it seems that the DUT holds req low when in reset).
+  `ASSERT_KNOWN(ReqKnown, req_i, clk_i, 0)
+
+  // If there is a request, the write signal and address should be known
```
tomroberts-lowrisc, line 27: The comment implies we check the address is known for all req_i cases, but the check below only exists for req_i & write_i
I think the check is probably correct and just the comment needs updating

rswarbrick: Ah, yes. Looks like I fixed the code and not the comment. Thanks!

Issue #964, opened 2020-06-17
Is parameter ICacheECC need to be passed by FuseSoC?


dawidzim:
Hey!
I have question about `ICacheECC `, is this parameter need to be passed by FuseSoC? I've problem with this, because when I generate files for Riviera and Xcelium by command `./vendor/lowrisc_ip/dvsim/dvsim.py dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson -t [riviera/xcelium]` FuseSoC create: ` -pvalue+tb.ICacheECC=1` in file list.

rswarbrick:
The parameter is so that block-level testing of the ICache can choose whether to turn ECC on or not. At the moment, it's "hard-coded" at that point, but we'll eventually hook it up to Ibex's configuration machinery.

What's breaking here? Is it that the parameter value is getting included in a file that's supposed to be a list of SystemVerilog files? (If so, that sounds like it might be a problem with edalizer).

dawidzim:
When I run command: `./vendor/lowrisc_ip/dvsim/dvsim.py dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson -t [riviera/xcelium]` FuseSoC generate files like for VCS, then in compilation we use `-f path/to/file.scr` and in this `file.scr` is `-pvalue+tb.ICacheECC=1`. 
So in Riviera compilation command going like this:
`vlog (...) -f path/to/file.scr` and we get error about unknown switch
In Riviera to pass value to this param we need to use `-g/tb.ICacheECC=1` in `vsim`, like is done in UVM core_ibex.
So to resolve this problem we need to change something in edalizer<?> or try to pass this argument like is done in UVM core_ibex.

imphil:
dvsim currently uses fusesoc/edalize to generate a VCS-specific file list (*.f) file, which happens to have parameters, defines, include paths, and files in it in VCS syntax. Dvsim then passes this file into its own flow, which supports multiple tools. 

This can work, but isn't guaranteed to work, as you found out: neither is the .f file produced by edalize a stable artifact of edalize, nor is the output format tool-agnostic enough to be used in all tools.

I don't think there's an easy solution at the moment.

- dvsim shouldn't rely on the *.f file; either it should use the stable *.eda.yml file, and generate a tool-specific file list itself, or use edalize completely to call the tool.
- I don't know if dvsim currently has a way to pass parameters/defines to tools in a tool-agnostic way. This would effectively require replicating the respective functionality from edalize.

[...]

PR #977, opened 2020-06-19
Enable ICache ECC in the way dvsim.py requires

rswarbrick:
It seems that dvsim.py doesn't actually use fusesoc to do things like
pass parameters. Instead, we have to set the tool-specific options in
the hjson file by hand.

Fixes issue #964.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_sim.core
+++ b/dv/uvm/icache/dv/ibex_icache_sim.core
@@ -19,7 +19,6 @@ filesets:
 targets:
   sim:
     parameters:
-      - ICacheECC
       - PRIM_DEFAULT_IMPL=prim_pkg::ImplBadbit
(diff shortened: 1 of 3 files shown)


rswarbrick:
@dawidzim: Could you check this patch? I think it does what Sri said.

It seems a pity that we're encoding tool-specific flags in our hjson files when we are already using edalizer, which allows you to avoid exactly that thing. :-( Anyway, I guess this will have to do for now.

dawidzim:
@rswarbrick 
LGTM

rswarbrick:
> Oh wait, one small change is required :
> `run_opts: ["-g/tb.ICacheECC=1"]` to `run_opts: ["-gICacheECC=1"]`

@dawidzim: Thanks, fixed.

Issue #954, opened 2020-06-15
UVM DV: Unsynchronized DUT probe signals


tomroberts-lowrisc:
In PR #953 I corrected some paths that were causing spurious triggering of internal TB events. Even with that fix, the spurious triggers remain due to glitchy sampling of those signals.

I hacked around it by replacing all instances of `wait (signal);` with `while (!signal) @(posedge clk);` but I guess the correct fix is a clocking block.

udinator:
Thanks @tomroberts-lowrisc for filing this issue!
You're absolutely right, it's better in general to synchronize interface signal sampling with clocking blocks. I'll work on this when I have some time and get a PR open to add clocking blocks in the Ibex testbench.

udinator:
#978 has been merged, which synchronizes all Ibex interfaces with clocking blocks.
@tomroberts-lowrisc do you think you could rerun the failing seeds with the top-of-tree? From what you mentioned, there might be a possibility that adding clocking blocks is only a partial solution to the failures you encountered.

tomroberts-lowrisc:
Hi @udinator - I reran at my end. All tests passing except debug_branch_jump and debug_single_step, which I think are due to unrelated issues.

PR #978, opened 2020-06-20
[ibex/dv] Add clocking blocks to Ibex interfaces

udinator:
This PR simply adds clocking blocks to all major Ibex interface agents and updates all interface accesses to use these clocking blocks to enforce clock synchronization.

@tomroberts-lowrisc this should address #954. I tested the `riscv_dret_test` and other similar tests that you noticed some arbitrary errors in, and they all seem to pass successfully with these changes

A few notes about some parts of this PR:

- `ibex_mem_intf` has two driver clocking blocks added, one for host side and one for device side. This is because our Ibex testbench currently provides both host and device agents for the Ibex I/D-side interfaces (of course we only use the reactive device agents in the main testbench). If you feel the host agents should be removed, I can do this in a later PR.
- `csr_if` has only one clocking block - this is because the signals bound to this interface are only ever sampled, never driven (these are signals internal to the CSR 'register file').
- `dut_if` has only one clocking block as well - this is because `fetch_enable` and `debug_req` are only ever driven, never sampled, and the remaining signals are only sampled, never driven (these are signals internal to the `ibex_controller`, so we can make do with a single clocking block here.
- some utility tasks have been added to some interfaces to wait for a specified number of clock cycles - no other significant changes are present in this PR apart from this.

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf.sv
@@ -2,10 +2,13 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-interface ibex_mem_intf#(parameter int ADDR_WIDTH = 32,
-                         parameter int DATA_WIDTH = 32);
+interface ibex_mem_intf#(
+  parameter int ADDR_WIDTH = 32,
+  parameter int DATA_WIDTH = 32
+) (
+  input clk
+);
 
-  logic                    clock;
   logic                    reset;
   logic                    request;
   logic                    grant;
@@ -16,5 +19,52 @@ interface ibex_mem_intf#(parameter int ADDR_WIDTH = 32,
   logic [DATA_WIDTH-1:0]   wdata;
   logic [DATA_WIDTH-1:0]   rdata;
   logic                    error;
+
+  clocking host_driver_cb @(posedge clk);
+    input   reset;
+    output  request;
+    input   grant;
+    output  addr;
+    output  we;
+    output  be;
+    input   rvalid;
+    output  wdata;
+    input   rdata;
+    input   error;
+  endclocking
+
+  clocking device_driver_cb @(posedge clk);
+    input   reset;
+    input   request;
+    output  grant;
+    input   addr;
+    input   we;
+    input   be;
+    output  rvalid;
+    input   wdata;
+    output  rdata;
+    output  error;
+  endclocking
+
+  clocking monitor_cb @(posedge clk);
+    input reset;
+    input request;
+    input grant;
+    input addr;
+    input we;
+    input be;
+    input rvalid;
+    input wdata;
+    input rdata;
+    input error;
+  endclocking
+
+  task automatic wait_clks(input int num);
+    repeat (num) @(posedge clk);
+  endtask
+
+  task automatic wait_neg_clks(input int num);
+    repeat (num) @(negedge clk);
+  endtask
 
 endinterface : ibex_mem_intf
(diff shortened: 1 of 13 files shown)


dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf.sv
```diff
@@ -3,9 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 interface ibex_mem_intf#(parameter int ADDR_WIDTH = 32,
-                         parameter int DATA_WIDTH = 32);
+                         parameter int DATA_WIDTH = 32)(
+                       input clk
+                      );
```
rswarbrick, line 8: I realise that it's not caused by your change, but this spacing looks a little odd. Since you're touching the lines, maybe fix it?

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_master_driver.sv
```diff
@@ -45,42 +45,42 @@ class ibex_mem_intf_master_driver extends uvm_driver #(ibex_mem_intf_seq_item);
...
-    vif.addr    <= 'hz;
-    vif.wdata   <= 'hz;
-    vif.be      <= 'bz;
-    vif.we      <= 'bz;
+    vif.host_driver_cb.request <= 1'b1;
+    vif.host_driver_cb.addr    <= trans.addr;
+    vif.host_driver_cb.be      <= trans.be;
+    vif.host_driver_cb.we      <= trans.read_write;
+    vif.host_driver_cb.wdata   <= trans.data;
+    wait(vif.host_driver_cb.grant === 1'b1);
```
rswarbrick, line 66: Missing space (`wait` is a keyword like `if`, not a function call).

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -28,12 +28,12 @@ class ibex_mem_intf_monitor extends uvm_monitor;
   endfunction: build_phase
 
   virtual task run_phase(uvm_phase phase);
-    wait(vif.reset === 1'b0);
+    wait(vif.monitor_cb.reset === 1'b0);
```
rswarbrick, line 31: I realise that the code already had this looking odd, but could you insert spaces after all the `wait` keywords that this patch touches? We may as well move towards matching the "house style".

dv/uvm/core_ibex/common/irq_agent/irq_master_driver.sv
```diff
@@ -22,7 +22,7 @@ class irq_master_driver extends uvm_driver #(irq_seq_item);
     forever begin
       fork : drive_irq
         get_and_drive();
-        wait(vif.reset === 1'b1);
+        wait(vif.driver_cb.reset === 1'b1);
```
rswarbrick, line 25: Again with the unusual `wait` spacing.

dv/uvm/core_ibex/env/core_ibex_csr_if.sv
```diff
@@ -9,4 +9,14 @@ interface core_ibex_csr_if(input logic clk);
   logic [31:0]              csr_wdata;
   logic [31:0]              csr_rdata;
   ibex_pkg::csr_op_e        csr_op;
+
+  // Only one clocking block needed, we will only sample these signals.
```
rswarbrick, line 13: I'm not sure this comment is really required

dv/uvm/core_ibex/env/core_ibex_dut_probe_if.sv
```diff
@@ -16,6 +16,21 @@ interface core_ibex_dut_probe_if(input logic clk);
   logic                 debug_req;
   ibex_pkg::priv_lvl_e  priv_mode;
 
+  // Only a single clocking block is needed here.
```
rswarbrick, line 19: This comment probably doesn't convey any information (the fact that there's only one block will presumably imply that the author only thought one was needed!). Drop it?

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_master_driver.sv
udinator: thanks for pointing that out, I'll go through and fix the spacings

udinator:
@rswarbrick thanks for the comments! A lot of the code in this testbench is relatively "old" (written around mid-2019), so it's definitely good to go through and update some old formatting styles.
I've also updated the commit message as per your request.

PR #980, opened 2020-06-22
[top_pkg] Fix style lint warnings

msfschaffner:
This fixes several style lint warnings in Ibex. Note that the `top_pkg` from Ibex seems to override the `top_pkg` from OpenTitan (hence we get all these duplicate warnings about missing parameter types in https://reports.opentitan.org/hw/ip/rv_core_ibex/lint/veriblelint/latest/results.html). This is another instance of https://github.com/lowRISC/opentitan/issues/1231 and we should resolve this.

### DIFF
--- a/rtl/ibex_alu.sv
+++ b/rtl/ibex_alu.sv
@@ -591,9 +591,9 @@ module ibex_alu #(
     `define _N(stg) (16 >> stg)
 
     // bext / bdep control bit generation
-    for (genvar stg=0; stg<5; stg++) begin
+    for (genvar stg=0; stg<5; stg++) begin : gen_stage
       // number of segs: 2** stg
-      for (genvar seg=0; seg<2**stg; seg++) begin
+      for (genvar seg=0; seg<2**stg; seg++) begin : gen_segment
 
         assign lrotc_stage[stg][2*`_N(stg)*(seg+1)-1 : 2*`_N(stg)*seg] =
             {{`_N(stg){1'b0}},{`_N(stg){1'b1}}} <<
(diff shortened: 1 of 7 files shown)


msfschaffner:
@udinator could you pipe this through the Ibex DV environment to check whether this does not brake anything?

imphil:
`top_pkg` isn't a duplicated primitive, it contains toplevel-specific global constants. The "icache DV toplevel" is different from the "top_earlgrey OpenTitan toplevel". And as Eunchan recently mentioned in https://github.com/lowRISC/opentitan/issues/2544, we'll have multiple top_pkg files in OT as soon as we have a second toplevel. So lowRISC/opentitan#1231 doesn't really apply here.

The Ibex `top_pkg.sv` is also not vendored into OT, so OT tooling should see only a single top_pkg.sv.

The changes itself LGTM, thanks for taking care of them!

rtl/ibex_alu.sv
```diff
@@ -708,17 +708,17 @@ module ibex_alu #(
...
+        '{32'h2222_2222, 32'h0c0c_0c0c, 32'h00f0_00f0, 32'h0000_ff00};
 
-    localparam logic [31:0] FLIP_MASK_L [0:3] =
-        '{32'h2200_1100, 32'h0044_0000, 32'h4411_0000, 32'h1100_0000};
-    localparam logic [31:0] FLIP_MASK_R [0:3] =
-        '{32'h0088_0044, 32'h0000_2200, 32'h0000_8822, 32'h0000_0088};
+    localparam logic [31:0] FLIP_MASK_L [4] =
+        '{32'h1100_0000, 32'h4411_0000, 32'h0044_0000, 32'h2200_1100};
+    localparam logic [31:0] FLIP_MASK_R [4] =
+        '{32'h0000_0088, 32'h0000_8822, 32'h0000_2200, 32'h0088_0044};
```
vogelpi, line 719: Just for the record, we figured that these changes are incorrect. The order of the elements in the unpacked dimension does not need to be changed, see https://github.com/lowRISC/ibex/pull/999#discussion_r449439788

PR #982, opened 2020-06-23
Prevent writing CSR_SECURESEED to get the seed of dummy instruction

wxjstz:
Although CSR_SECURESEED is unreadable, an attacker can write a new seed, which
brings convenience to the attack. This patch is used to XOR the historical seed,
so that even if the attacker writes a new value to CSR_SECURESEED , he cannot
know the value of the seed.

Signed-off-by: Xiang Wang \<merle@hardenedlinux.org\>

### DIFF
--- a/rtl/ibex_dummy_instr.sv
+++ b/rtl/ibex_dummy_instr.sv
@@ -54,9 +54,20 @@ module ibex_dummy_instr (
   logic [6:0]               dummy_set;
   logic [2:0]               dummy_opcode;
   logic [31:0]              dummy_instr;
+  logic [31:0]              dummy_instr_seed_q, dummy_instr_seed_d;
 
   // Shift the LFSR every time we insert an instruction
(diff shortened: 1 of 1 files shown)


rtl/ibex_dummy_instr.sv
```diff
@@ -88,8 +91,11 @@ module ibex_dummy_instr (
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
       dummy_cnt_q <= '0;
+	  dummy_instr_seed_q <= '0;
```
tomroberts-lowrisc, line 94: The indenting is off here - also can you split it out into a separate always_ff? We try to keep one `if` condition per always_ff block to assist tools inferring clock gates

wxjstz: @tomroberts-lowrisc done

PR #984, opened 2020-06-23
Fix Pointer Authentication lint errors

mmxsrup:
We have some lint errors below in Pointer Authentication. I fixed it.
- Bits of signal are not used
- Case values incompletely covered
- Signal unoptimizable: Feedback to clock or circular logic
- Run clang-format on pa_test.c

### DIFF
--- a/examples/sw/simple_system/pa_test/pa_test.c
+++ b/examples/sw/simple_system/pa_test/pa_test.c
@@ -5,56 +5,58 @@
 #include "simple_system_common.h"
 
 #define PMP_NAPOT 0x18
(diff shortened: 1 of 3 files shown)


vogelpi:
CI currently fails because of lint errors in the C++ code (`examples/sw/simple_system/pa_test/pa_test.c`). Could you please run `clang-format` on that file?

mmxsrup:
I ran `clang-format` on `pa_test.c`. Please check it.

vogelpi:
Excellent, thanks!

PR #987, opened 2020-06-23
[dv/ibex] Add two new interrupt/debug tests

udinator:
As a result of lowRISC/opentitan#2405 and lowRISC/ibex#928, this PR adds some new
directed interrupt and debug tests to check that the core behaves
properly during execution of each supported instruction when some
external irq/debug stimulus comes in.

To do this, we use the two new functions `decode_instr(...)` and
`decode_compressed_instr(...)` in `core_ibex_test_list.sv` to "decode"
every instruction that the `core_ibex_instr_monitor_if` sees in the ID
stage of the pipeline. Once the testbench decodes an instruction that
we have not seen before, it can then drive interrupt or debug stimulus
into the core.

Once any given instruction has been detected by the testbench (and
stimulus driven), it will no longer drive stimulus if this instruction
is seen in the decode pipeline (e.g. if we have previously detected a
`c.addi` instruction in the ID stage and have driven irq/debug stimulus,
we will no longer drive stimulus if we see another `c.addi` instruction,
no matter the operands). This is to avoid driving irq/debug stimulus
after every single instruction as this will add a huge amount
of simulation latency.

A few notes:

- From a reviewer standpoint, the actual decoding logic is a bit esoteric,
  but simply just adheres to the encoding scheme defined by the official
  RISC-V specification
- We drive irq/debug stimulus into the core every time we see a
  `wfi` instruction, as otherwise we will timeout as the core waits
  infinitely for some stimulus from the outside world.
- We ignore some system-level instructions (ebreak/mret/dret) and illegal
  instructions for now, as driving stimulus during these instructions will
  result in a nested trap, which requires special handling.
- The interrupt agent was modified slightly to drive stimulus by
  default on the falling edge of the clock, so this way we can "catch"
  instructions that are in the ID pipeline for only a single cycle.
- The interrupt agent was additionally modified to add a control knob 
  to turn off 'fast' interrupts, as there is no guarantee that a 'fast' interrupt
  into Ibex will be taken, as `mie` is completely randomized.
- The duration for which the testbench raises `debug_req_i` for the core
  is also increased to avoid edge cases where we lower the debug line
  too early (e.g. while long multicycle instructions like `div` are
  executing in the ID stage).

### DIFF
--- a/dv/uvm/core_ibex/common/irq_agent/irq_if.sv
+++ b/dv/uvm/core_ibex/common/irq_agent/irq_if.sv
@@ -11,6 +11,7 @@ interface irq_if(input clk);
   logic        irq_nm;       // non-maskeable interrupt
 
   clocking driver_cb @(posedge clk);
+    default output negedge;
     input   reset;
     output  irq_software;
     output  irq_timer;
@@ -30,6 +31,10 @@ interface irq_if(input clk);
 
   task automatic wait_clks(input int num);
     repeat (num) @(posedge clk);
+  endtask
+
+  task automatic wait_neg_clks(input int num);
+    repeat (num) @(negedge clk);
   endtask
 
 endinterface
--- a/dv/uvm/core_ibex/env/core_ibex_instr_monitor_if.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_instr_monitor_if.sv
@@ -1,0 +1,18 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Interface to probe the instruction moving through the pipeline
+//
+// TODO: does not support dummy instruction insertion right now,
+//       might have to revisit and update.
+interface core_ibex_instr_monitor_if #(parameter DATA_WIDTH = 32);
+
+  // ID stage
+  logic                   valid_id;
+  logic                   err_id;
+  logic                   is_compressed_id;
+  logic [15:0]            instr_compressed_id;
+  logic [DATA_WIDTH-1:0]  instr_id;
+
+endinterface
--- a/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
+++ b/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
@@ -20,6 +20,9 @@ module core_ibex_tb_top;
   // DUT probe interface
   core_ibex_dut_probe_if dut_if(.clk(clk));
 
+  // Instruction monitor interface
+  core_ibex_instr_monitor_if instr_monitor_if();
+
   // RVFI interface
   core_ibex_rvfi_if rvfi_if(.clk(clk));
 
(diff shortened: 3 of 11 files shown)


dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -96,46 +99,56 @@ module core_ibex_tb_top;
...
-  assign rvfi_if.rs2_rdata      = dut.rvfi_rs2_rdata;
-  assign rvfi_if.rd_addr        = dut.rvfi_rd_addr;
-  assign rvfi_if.rd_wdata       = dut.rvfi_rd_wdata;
-  assign rvfi_if.pc_rdata       = dut.rvfi_pc_rdata;
-  assign rvfi_if_pc_wdata       = dut.rvfi_pc_wdata;
-  assign rvfi_if.mem_addr       = dut.rvfi_mem_addr;
-  assign rvfi_if.mem_rmask      = dut.rvfi_mem_rmask;
-  assign rvfi_if.mem_rdata      = dut.rvfi_mem_rdata;
-  assign rvfi_if.mem_wdata      = dut.rvfi_mem_wdata;
+  assign rvfi_if.valid                        = dut.rvfi_valid;
```
rswarbrick, line 102: Are these re-indentations required? It seems to be a whitespace change to code that you're not touching otherwise?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -390,6 +394,7 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
           // Wait for core initialization before starting the stimulus check loop - first write
           // to signature address is guaranteed to be core initialization info
           wait_for_core_setup();
+          clk_vif.wait_clks(50);
```
rswarbrick, line 397: Does this need a comment? Why do we wait another 50 cycles here?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -465,6 +470,222 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
                        "Incorrect dcsr.prv value!")
   endfunction
 
+  // These are helper functions used for directed interrupt tests that will "decode"
+  // an instruction currently in the ID stage of the pipeline.
+  // After "decoding" an instruction that is legal and valid, we add it to a global queue
+  // if it has not been seen before.
```
rswarbrick, line 476: Shouldn't this comment be at the call site for `decode_instr`? I feel like the comment directly above the function should be something more like "Check against a global queue whether we've seen this sort of instruction before. If not, add the instruction to the queue and return 1".

udinator:
> This looks sensible enough. It's a pity that we need to do so much grubbing around in the internals of the decoder stage, but I think I understand the point.

Agreed. It is definitely a bit messy having to dig so deep into the hierarchy, but it wouldn't be possible to test a lot of these features otherwise.

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
udinator: They're not required strictly speaking, but the formatting looks a lot cleaner to me when all of the interface assignments are lined up (especially since there are so many of them).

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: Good catch, this comment was meant more for me during the actual development of this function!
I'll rearrange these as you suggested.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: We want to wait for some extra cycles here because it's not guaranteed that the very next instruction will be in the <main> section of the assembly code. In fact, since PMP is enabled in Ibex the <main> section is in a different area of the program we have to explicitly jump to the correct address after executing the initialization handshake. 
As a result, this delay is to ensure that we start stimulus generation only when the core has started executing <main>. 

I can add a comment here to explain this.

rswarbrick: Ok, makes sense. I guess it would be neater to do something like wait until we see the first instruction from main get fetched, but maybe we don't have that info to hand (the address of main?).

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -465,6 +470,227 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
...
+            if (opcode == seen_instr[i].opcode &&
+                funct3 == seen_instr[i].funct3 &&
+                system_imm == seen_instr[i].system_imm) begin
+              is_seen = 1'b1;
+              break;
+            end
+          end
+        end
+      end
+    endcase
```
cindychip, line 564: Do you think it is necessary to add a default case just in case?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: yep, something like this might be a neater solution.
I've had making this enhancement on the back-burner for a while now, but it never really made much sense to implement in the past since it would have required an `instr_monitor_if` just for this. However now it is a bit more feasible since this interface will be seeing actual use (both in this PR and some upcoming ones that enhance other tests).

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: That's a good point actually. I had been banking on the qualification checks (line 677 of this file) preventing a default case from being necessary, but it's probably good to include one as a safety net here.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -465,6 +470,227 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
...
+  // global seen_compressed_instr[$] queue.
+  // If we have seen it before, it returns 0, otherwise the instruction is added to the
+  // and it returns 1.
+  virtual function bit decode_compressed_instr(bit [15:0] instr);
+    bit        is_seen;
+
+    foreach (seen_compressed_instr[i]) begin
+      if (instr[1:0] == seen_compressed_instr[i][1:0]) begin
+        case (instr[1:0])
+          2'b00: begin
```
sriyerg, line 590: Wondering if something can be done about all these hard-coded numbers here...Consider using enums / parameters / macros wherever suitable.

dv/uvm/core_ibex/common/irq_agent/irq_if.sv
```diff
@@ -11,6 +11,7 @@ interface irq_if(input clk);
   logic        irq_nm;       // non-maskeable interrupt
 
   clocking driver_cb @(posedge clk);
+    default output negedge;
```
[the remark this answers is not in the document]
udinator: Race conditions actually aren't really an issue here.\

What happens pretty frequently is that instructions will stay in the ID stage for just a single cycle, giving us a one cycle window for the sequencer to send an irq item over to the driver. Without driving at the negedge, this irq item will be driven into the core after this one cycle window when the next instruction has entered the ID stage, which is incorrect.

dv/uvm/core_ibex/env/core_ibex_instr_monitor_if.sv
```diff
@@ -0,0 +1,18 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Interface to probe the instruction moving through the pipeline
+//
+// TODO: does not support dummy instruction insertion right now,
+//       might have to revisit and update.
+interface core_ibex_instr_monitor_if#(parameter DATA_WIDTH = 32);
```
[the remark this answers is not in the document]
udinator: good catch, thanks!

dv/uvm/core_ibex/tests/core_ibex_seq_lib.sv
```diff
@@ -70,13 +70,19 @@ class irq_raise_seq extends core_base_seq#(irq_seq_item);
...
+
   bit no_nmi;
+  bit no_fast;
 
   virtual task send_req();
     irq_seq_item irq;
     irq = irq_seq_item::type_id::create($sformatf("irq_raise_single[%0d]", iteration_cnt));
     start_item(irq);
-    `DV_CHECK_RANDOMIZE_WITH_FATAL(irq, num_of_interrupt > 1; irq_nm == ~no_nmi;)
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(irq, num_of_interrupt > 1;
```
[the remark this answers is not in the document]
udinator: sure, that's a good idea.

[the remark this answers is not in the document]
udinator: created an irq_base_seq with all the boilerplate sequence item creation code, and each extended irq sequence simply just overrides the `randomize_item()` function.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -465,6 +470,227 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
...
+              break;
+            end
+          end
+        end
+      end
+    endcase
+
+    // If we've seen the same type of instruction already, return 0.
+    if (is_seen) begin
+      return 1'b0;
```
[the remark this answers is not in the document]
udinator: sure, thanks for the suggestion

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: That's something I'd considered at first too...decoding the compressed instruction set is definitely much more "hardcodey" than any of the rv32i instructions (hence I could abstract everything in `decode_instr(...)` to using `funct3`, `funct7`, `opcode`, `system_imm` variables only).
Using some params/enums here would essentially force me to map each compressed instruction supported by rv32imc to an enum/param, which would turn this whole function into a long switch statement comparing every compressed instruction enum/param to the currently seen instruction.
I felt that wasn't the most elegant solution, the current way I've implemented it allows me to condense groups of checks together (e.g. I can mark an instruction as seen if bits [15:13] are within a wide range of values, as opposed to one individual switch statement per value).

PR #997, opened 2020-07-02
[doc] Add bitmanip spec to introduction page

### LINKED ISSUE #966: Docs: Mention B extension in overview page
We don't mention the B extension on https://ibex-core.readthedocs.io/en/latest/introduction.html yet, we should do so.

@ganoam

imphil:
We list all specifications we implement, even optional ones. Add
Bitmanip there as well.

Fixes #966


imphil:
Debug and many parts of the other specs are also optional, so it seems consistent to simply list all specs in that part.

PR #1003, opened 2020-07-03
Remove outdated documentation

imphil:
By now, ibex depends not only on the a clock gating cell, but also on
assert macros and a LFSR; there will be more to come. This removed piece
of documentation was from the early days, when only a clock gating cell
was to be provided (and we didn't ship one).

Today, users need to either use FuseSoC, or run fusesoc on the simple
system and "harvest" the resulting files if they want to copy-paste Ibex
into their build system. That's not ideal, but not something we can very
easily fix -- so let's remove the outdated documentation first to at
least reduce the confusion.

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -12,17 +12,3 @@ Ibex comes with two different register file implementations.
 Depending on the target technology, either the implementation in ``ibex_register_file_ff.sv`` or the one in ``ibex_register_file_latch.sv`` should be selected.
 For more information about the two register file implementations and their trade-offs, check out :ref:`register-file`.
 
-Clock Gating Cell
------------------
-
-Ibex requires clock gating cells.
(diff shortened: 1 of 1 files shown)


imphil:
This is in preparation for adding a dependency on prim_clock_gating to the ibex_core core file.

imphil:
The easiest option for all those primitives (not only the generates ones) is to run fusesoc once, and then take the files out of the build directory. It's really hard to come up with a simple, yet universal solution for people who don't want to use fusesoc, since there are so many ways this could be done (none of which are particularly beautiful).

vogelpi:
Yeah, I totally agree with you. I thought we could document somewhere that fusesoc will generate those primitives automatically. At the moment, this is not obvious. We don't mention fusesoc one single time in the documentation.

PR #1005, opened 2020-07-03
Enable ibex_simple_system lint in CI

imphil:
This PR contains a rather odd collection of papercut commits, none of them are really spectacular.

The most significant change, besides the end result of enabling lint for simple system, is the addition of the prim_clock_gating dependency. I'll most likely put this one into a separate PR, but I want to see CI for the whole thing first.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -149,3 +149,20 @@ jobs:
         - small
         - experimental-maxperf-pmp
         - experimental-maxperf-pmp-bmfull
+
+  # Run lint on simple system
+  - bash: |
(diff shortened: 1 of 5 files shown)


imphil:
Nice, that seems to do the trick.

I've filed https://github.com/lowRISC/ibex/issues/1006 to get rid of the "simplified" copy of prim_clock_gating.v in the syn folder. Since this flow isn't using fusesoc directly, it continues to work as before.

Since the whole diff is rather small, I'll just keep it in one PR here if that's OK.

I'll rebase this one as soon as https://github.com/lowRISC/opentitan/pull/2697 goes in and I've updated the vendored files here in Ibex.

imphil:
Here we go. Rebased on top of master.

Issue #956, opened 2020-06-15
Replace SRAM_INIT_FILE with MemInitFile


imphil:
Replace SRAM_INIT_FILE with MemInitFile in our examples in Ibex. Seems like I missed a couple uses when updating our primitives.

PR #1008, opened 2020-07-03
Get the simple_system example working with VCS

rswarbrick:
This has two components:

1. Plumbing MemInitFile through properly
1. Rejigging the simple_system core file to allow non-verilator builds

I've checked, and the "hello world" example generates the same trace with VCS (modulo some Xs).


rswarbrick:
This partially solves #956, but there's a bit more work to do yet (for the RISC-V compliance suite).

PR #1009, opened 2020-07-03
Fix / remove last uses of SRAM_INIT_FILE

rswarbrick:


rswarbrick:
I've just run an FPGA synthesis. I don't have an arty board, but I have checked that the parameters make their way down to the sram primitive correctly.

PR #1013, opened 2020-07-03
Clean up Verilator sections in core files

imphil:
- The "PINCONNECTEMPTY" waiver is part of our normal waiver file, no need
  to add it to the tool invocation.
- Recent versions of Verilator choose good defaults for MAKE_OPTS,
  passing it explicitly overrides the settings.
- All Verilator code is now lint clean, we can remove `-Wno-fatal`.
- FST traces are not much slower then VCD traces any more in recent
  Verilator versions, remove the respective comment.
- Align comment about the compile/sim time for tracing with other files
  and OpenTitan.


imphil:
Yes, as usual I missed (at least) one lint error :-)

Issue #687, opened 2020-03-11
[tracking] Use Verible's verilog_format for code formatting


imphil:
Verible's verilog_format is a automatic code formatter. We want to be able to run it on our code base, ultimately in "enforcing" mode, requiring all submissions to be formatted with this tool. This is the same approach we take with clang-format and C/C++ files. This issue is to track the progress towards that.

Ibex TODO:
- [ ] Add Verible format support to edalize
- [ ] Add a fusesoc target to our core files
- [ ] Add Verible format in non-enforcing mode to CI
- [ ] Document how to use it
- [ ] Clean up existing code once the auto-formatted style matches our expectations.
- [ ] Switch to enforcing mode.

Blocking Verible bugs:
- [ ] https://github.com/google/verible/issues/228

imphil:
The first set of code changes to get our code in line with what verible format produces is at https://github.com/lowRISC/ibex/pull/1014. The goal is to have that PR cover all the "easy" cases, getting us step-by-step closer to a full equivalence.

asb:
Thanks for your work in pursuing this, filing upstream issues, and tracking progress.

IMHO #351 should be higher priority - it seems to be a case of producing incorrectly formatted code, which is a much bigger barrier to enforcing verible formatting use than producing output that doesn't match our personal tastes. But once incorrect formatting bugs are addressed I agree with the thrust of your argument in #1014 - don't sweat the small stuff, and in return formatting becomes a solved problem.

We'll want to consult with downstream consumers (OpenTitan) prior to landing a large cosmetic change like #1014. I don't see why it should be problematic at this stage, but it couldn't hurt to explain the plan.

PR #1014, opened 2020-07-03
Format source code with Verible, part 1

imphil:
This commit is the first, incomplete run of Verible format on the source
code in the `rtl` directory. The goal is to get the "easy cases" out of
the way, the ones we agree on that the produced formatting is good
enough. This should leave us with the cases where we either don't agree
with the formatting produced by Verible format (either because it
violates the lowRISC style guide, or because it hurts readability), or
the cases where Verible format breaks down. Ultimately, we should be
able to reduce the diff between what verible-verilog-format produces,
and our code bases, iteratively to zero.

The commit doesn't contain all changes made by Verible format:
- All changes to module headers (parameters and port lists) have been
  excluded due to known formatting issues there.
- Many changes to case statements have been excluded, as there are a
  couple formatting issues and inconsistencies.
- Some assign statement changes have been excluded for too badly hurting
  readbility.
- Some other obvious misformattings have been corrected or excluded.

Most limitations of Verible format are tracked in
https://github.com/lowRISC/ibex/issues/687. Not all problems I've seen
have been filed upstream yet,

The diff contains a rather exhaustive list of changes which I
pre-filtered to remove the most obvious "ugly ones" (see above). That
still leaves a couple ones which are worth discussing:

- Verible doesn't use tabular style for blocks of signal definitions.
  Doing that would most likely be possible in an algorithmic way if we
  agree that readability is hurt too much by not doing it.
- Verible doesn't use tabular style for other assignment code. Doing
  that in a consistent, automated fashion is probably harder. Also, the
  readability trade-offs are different.

I invite you all to look at the diff and classify the changes:
- Changes you're OK with.
- Changes which might not be your preferred style, but are not too bad
  either.
- Changes which you think severely hurt readability, or are against the
  style guide.

I'll go through those comments and remove all contentious items from
the diff; we can then discuss them individually, also involving the
style guide team and the verible developers.

In general, the trade off is as follows: Verible format takes away the
arguing and manual formatting completely out of the code writing and
review process. We win a significant amount of time there. There will
always be cases where a manual formatting might be marginally better,
but expressing that into an automated formatting rule is (close to)
impossible. In those cases, we win by "swallowing the pill" of not fully
optimal styling, for the benefit of never having to think about styling
again (TM).

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -106,8 +106,16 @@ module ibex_controller #(
 
   // FSM state encoding
   typedef enum logic [3:0] {
-    RESET, BOOT_SET, WAIT_SLEEP, SLEEP, FIRST_FETCH, DECODE, FLUSH,
-    IRQ_TAKEN, DBG_TAKEN_IF, DBG_TAKEN_ID
+    RESET,
+    BOOT_SET,
+    WAIT_SLEEP,
+    SLEEP,
+    FIRST_FETCH,
+    DECODE,
+    FLUSH,
+    IRQ_TAKEN,
+    DBG_TAKEN_IF,
+    DBG_TAKEN_ID
   } ctrl_fsm_e;
 
   ctrl_fsm_e ctrl_fsm_cs, ctrl_fsm_ns;
@@ -142,7 +150,7 @@ module ibex_controller #(
   logic handle_irq;
 
   logic [3:0] mfip_id;
-  logic       unused_irq_timer;
+  logic unused_irq_timer;
 
   logic ecall_insn;
   logic mret_insn;
@@ -170,16 +178,16 @@ module ibex_controller #(
   // Exceptions //
   ////////////////
 
-  assign load_err_d  = load_err_i;
+  assign load_err_d = load_err_i;
   assign store_err_d = store_err_i;
 
   // Decoder doesn't take instr_valid into account, factor it in here.
-  assign ecall_insn      = ecall_insn_i      & instr_valid_i;
-  assign mret_insn       = mret_insn_i       & instr_valid_i;
-  assign dret_insn       = dret_insn_i       & instr_valid_i;
-  assign wfi_insn        = wfi_insn_i        & instr_valid_i;
-  assign ebrk_insn       = ebrk_insn_i       & instr_valid_i;
-  assign csr_pipe_flush  = csr_pipe_flush_i  & instr_valid_i;
+  assign ecall_insn = ecall_insn_i & instr_valid_i;
+  assign mret_insn = mret_insn_i & instr_valid_i;
+  assign dret_insn = dret_insn_i & instr_valid_i;
+  assign wfi_insn = wfi_insn_i & instr_valid_i;
+  assign ebrk_insn = ebrk_insn_i & instr_valid_i;
+  assign csr_pipe_flush = csr_pipe_flush_i & instr_valid_i;
   assign instr_fetch_err = instr_fetch_err_i & instr_valid_i;
 
   // "Executing DRET outside of Debug Mode causes an illegal instruction exception."
@@ -203,8 +211,8 @@ module ibex_controller #(
   // the FLUSH state so the cycle following exc_req_q won't remain set for an
   // exception request that has just been handled.
   // All terms in this expression are qualified by instr_valid_i
-  assign exc_req_d = (ecall_insn | ebrk_insn | illegal_insn_d | instr_fetch_err) &
-                     (ctrl_fsm_cs != FLUSH);
+  assign exc_req_d =
+      (ecall_insn | ebrk_insn | illegal_insn_d | instr_fetch_err) & (ctrl_fsm_cs != FLUSH);
 
   // LSU exception requests
   assign exc_req_lsu = store_err_i | load_err_i;
@@ -222,28 +230,27 @@ module ibex_controller #(
   // generic special request signal, applies to all instructions
   // All terms in this expression are qualified by instr_valid_i except exc_req_lsu which can come
   // from the Writeback stage with no instr_valid_i from the ID stage
-  assign special_req_all = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
-      exc_req_d | exc_req_lsu;
+  assign
+      special_req_all = mret_insn | dret_insn | wfi_insn | csr_pipe_flush | exc_req_d | exc_req_lsu;
 
   // special request that can specifically occur during branch instructions
   // All terms in this expression are qualified by instr_valid_i
   assign special_req_branch = instr_fetch_err & (ctrl_fsm_cs != FLUSH);
 
-  `ASSERT(SpecialReqBranchGivesSpecialReqAll,
-    special_req_branch |-> special_req_all)
+  `ASSERT(SpecialReqBranchGivesSpecialReqAll, special_req_branch |-> special_req_all)
 
   `ASSERT(SpecialReqAllGivesSpecialReqBranchIfBranchInst,
-    special_req_all && (branch_set_i || jump_set_i) |-> special_req_branch)
+          special_req_all && (branch_set_i || jump_set_i) |-> special_req_branch)
 
   // Exception/fault prioritisation is taken from Table 3.7 of Priviledged Spec v1.11
   if (WritebackStage) begin : g_wb_exceptions
     always_comb begin
       instr_fetch_err_prio = 0;
-      illegal_insn_prio    = 0;
-      ecall_insn_prio      = 0;
-      ebrk_insn_prio       = 0;
-      store_err_prio       = 0;
-      load_err_prio        = 0;
+      illegal_insn_prio = 0;
+      ecall_insn_prio = 0;
+      ebrk_insn_prio = 0;
+      store_err_prio = 0;
+      load_err_prio = 0;
 
       // Note that with the writeback stage store/load errors occur on the instruction in writeback,
       // all other exception/faults occur on the instruction in ID/EX. The faults from writeback
@@ -251,7 +258,7 @@ module ibex_controller #(
       if (store_err_q) begin
         store_err_prio = 1'b1;
       end else if (load_err_q) begin
-        load_err_prio  = 1'b1;
+        load_err_prio = 1'b1;
       end else if (instr_fetch_err) begin
         instr_fetch_err_prio = 1'b1;
       end else if (illegal_insn_q) begin
@@ -268,11 +275,11 @@ module ibex_controller #(
   end else begin : g_no_wb_exceptions
     always_comb begin
       instr_fetch_err_prio = 0;
-      illegal_insn_prio    = 0;
-      ecall_insn_prio      = 0;
-      ebrk_insn_prio       = 0;
-      store_err_prio       = 0;
-      load_err_prio        = 0;
+      illegal_insn_prio = 0;
+      ecall_insn_prio = 0;
+      ebrk_insn_prio = 0;
+      store_err_prio = 0;
+      load_err_prio = 0;
 
       if (instr_fetch_err) begin
         instr_fetch_err_prio = 1'b1;
@@ -285,19 +292,15 @@ module ibex_controller #(
       end else if (store_err_q) begin
         store_err_prio = 1'b1;
       end else if (load_err_q) begin
-        load_err_prio  = 1'b1;
+        load_err_prio = 1'b1;
       end
     end
     assign wb_exception_o = 1'b0;
   end
 
   `ASSERT_IF(IbexExceptionPrioOnehot,
-             $onehot({instr_fetch_err_prio,
-                      illegal_insn_prio,
-                      ecall_insn_prio,
-                      ebrk_insn_prio,
-                      store_err_prio,
-                      load_err_prio}),
+             $onehot({instr_fetch_err_prio, illegal_insn_prio, ecall_insn_prio, ebrk_insn_prio,
+                      store_err_prio, load_err_prio}),
              (ctrl_fsm_cs == FLUSH) & exc_req_q);
 
   ////////////////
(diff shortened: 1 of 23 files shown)


imphil:
The diff between what verible thinks is right, and what this diff produces, is visible in CI in the "Format all source code with Verible format (experimental)" task, e.g. https://dev.azure.com/lowrisc/ibex/_build/results?buildId=16365&view=logs&j=e92b124b-3ef9-5461-84e3-a5c4020c184a&t=8b62e53f-919c-5319-5e40-bcd998ef015d

rtl/ibex_controller.sv
```diff
@@ -170,16 +178,16 @@ module ibex_controller #(
...
+  assign load_err_d = load_err_i;
   assign store_err_d = store_err_i;
 
   // Decoder doesn't take instr_valid into account, factor it in here.
-  assign ecall_insn      = ecall_insn_i      & instr_valid_i;
-  assign mret_insn       = mret_insn_i       & instr_valid_i;
-  assign dret_insn       = dret_insn_i       & instr_valid_i;
-  assign wfi_insn        = wfi_insn_i        & instr_valid_i;
-  assign ebrk_insn       = ebrk_insn_i       & instr_valid_i;
-  assign csr_pipe_flush  = csr_pipe_flush_i  & instr_valid_i;
```
GregAC, line 182: Personally I rather like using whitespace to align expressions like these, I think it makes it easier to read and in this particular case makes it's very obvious it's just a bunch of incoming signals ANDed with instr_valid_i. Verible has remove alignment cases like this and other across the RTL. I don't think we should do this

GregAC:
To respond to your points around alignment in the initial comment:

> Verible doesn't use tabular style for blocks of signal definitions. Doing that would most likely be possible in an algorithmic way if we agree that readability is hurt too much by not doing it.

I think we should explore this option

> Verible doesn't use tabular style for other assignment code. Doing that in a consistent, automated fashion is probably harder. Also, the readability trade-offs are different.

Can we maybe disable alterations to horizontal whitespace in various cases? E.g. we could allow any alteration to horizontal whitespace that occurs before any printable characters (to allow indentation changes) but then only allow whitespace changes in specific situations after that, e.g. alter:

```verilog
if( condition )
```

to

```verilog
if (condition)
```
So in this case allow horizontal whitespace changes immediately after an `if` and allow horizontal whitespace changes after `(` and before `)`. The condition itself here is tricky you may want to write something like this:

```verilog
if (something && something_else && (foo ||
                                    bar ||
                                    baz))
```

Where you want to keep the alignment but you might want to correct cases like this

```verilog
if (something && something_else &&    something_further)
```

Maybe the rule would be allow whitespace changes in a `if` condition if it's on a single line otherwise leave them alone?

I don't know anything about the innards of Verible so hard to say if this kind of fine-grained control is possible (e.g. if may just parse to a concrete syntax tree without keeping any white-space then pretty print it)

My gut feeling is alignment of expressions is valuable to readability in RTL. You get complex compound expressions you just don't get in software (or would seek to avoid) plus big blocks of assignments where again the alignment is useful for readability.

Alignment of assignments within a block should be doable enough in Verible if it's pretty printing things (say a block is any section of assignments without any other intervening constructs). Good alignment of compound expressions doesn't seem decidable in the general case as a sensible way to align may depend upon the semantic meaning of the expression but perhaps you can do a decent job with some specific rules. Certainly for some specific cases it's quite mechanical e.g. ternaries

```verilog
assign result = cond_1 ? data_1 :
                cond_2 ? data_2 :
                cond_3 ? data_3 :
                         default_data;
```

imphil:
> Can we maybe disable alterations to horizontal whitespace in various cases?

Verible formatter, like clang-format, black, yapf, gofmt and many others, are "opinionated formatters" or "reformatters". They effectively parse the code, and dump it out again in a form they consider appropriate. That's their strength, but also their weakness. It takes away all the arguing by giving you only one acceptable way of formatting: the output of the formatter.

Nothing is ever black or white, so there are options we can explore. However, the first and foremost goal is to teach the formatter the style we'd like to see from "well formatted code". That sounds hard, but there's enough evidence out there in many programming languages that this is possible. The goal of this exercise is to collect examples where we think the formatting is off, analyze them to find a formatting we'd find acceptable, and then work with the Verible developers to get that implemented.

So whenever you find a piece of code you find particularly odd, file a bug either here or in the Verible issue tracker, where we can continue the discussion.

rtl/ibex_controller.sv
imphil: I'll remove those tomorrow.

rtl/ibex_decoder.sv
```diff
@@ -153,20 +153,20 @@ module ibex_decoder #(
...
   ////////////////////
   // Register check //
   ////////////////////
   if (RV32E) begin : gen_rv32e_reg_check_active
-    assign illegal_reg_rv32e = ((rf_raddr_a_o[4] & (alu_op_a_mux_sel_o == OP_A_REG_A)) |
-                                (rf_raddr_b_o[4] & (alu_op_b_mux_sel_o == OP_B_REG_B)) |
-                                (rf_waddr_o[4]   & rf_we));
+    assign illegal_reg_rv32e =
+        ((rf_raddr_a_o[4] & (alu_op_a_mux_sel_o == OP_A_REG_A)) | (
+         rf_raddr_b_o[4] & (alu_op_b_mux_sel_o == OP_B_REG_B)) | (rf_waddr_o[4] & rf_we));
```
GregAC, line 169: I think this is a good example of where we have a complex compound expressions where the alignment helps with the readability. What I'm less certain of is whether it's possible to create a mechnical rule set that will cause such expressions to align into something readable or if it's situation dependent and not something that can be done by an auto-formatter.

We could look at using the formatting on/off comments for such expressions?

GregAC: The other argument is you should just always avoid such expressions by introducing extra signals for the sub-expressions though I'm not convinced this is always a good idea as coming up with good names for the sub-expressions is hard and overall makes it harder to grasp what's going on as you need to look across multiple assignments to understand things.

For example here you might decompose into

```systemverilog
logic rf_a_illegal_reg_rv32e;
logic rf_b_illegal_reg_rv32e;
logic rf_wr_illegal_reg_rv32e;

assign rf_a_illegal_reg_rv32e   = (rf_raddr_a_o[4] & (alu_op_a_mux_sel_o == OP_A_REG_A);
assign rf_b_illegal_reg_rv32e   = (rf_raddr_b_o[4] & (alu_op_b_mux_sel_o == OP_B_REG_B);
assign rf_wr_illlegal_reg_rv32e = (rf_waddr_o[4] & rf_we);

assign illegal_reg_rv32e = rf_a_illegal_reg_rv32e | rf_b_illegal_reg_rv32e | rf_wr_illlegal_reg_rv32e;
```

Which looks pretty reasonable, the one issue is the `rf_a_illegal_reg_rv32e` name is a little confusing as it's not simple `rf_raddr_a` has an illegal address, but it has an illegal address and we are selecting register A as an operand.

rtl/ibex_icache.sv
```diff
@@ -57,158 +57,157 @@ module ibex_icache #(
...
   // This address is also updated if there is a branch that is not granted, since the target
   // address (addr_i) is only valid for one cycle while branch_i is high.
 
   // The captured branch target address is not forced to be aligned since the offset in the cache
   // line must also be recorded for later use by the fill buffers.
-  assign prefetch_addr_d     =
-      lookup_grant_ic0 ? (lookup_addr_aligned + {{ADDR_W-LINE_W-1{1'b0}},1'b1,{LINE_W{1'b0}}}) :
-                         addr_i;
+  assign prefetch_addr_d = lookup_grant_ic0
+      ? (lookup_addr_aligned + {{ADDR_W - LINE_W - 1{1'b0}}, 1'b1, {LINE_W{1'b0}}}) : addr_i;
```
GregAC, line 208: There's various things it's doing with ternaries, can't yet make my mind up about most of them but this one jumps out as being wrong, I feel the `?` should be contained on the same line as the actual condition `lookup_grant_ico` in this case. Indeed I think the original formatting here is the best, I guess it's struggling with juggling line lengths and removing line breaks?

rtl/ibex_icache.sv
```diff
@@ -57,158 +57,157 @@ module ibex_icache #(
...
+  logic reset_inval_q;
+  logic inval_prog_d, inval_prog_q;
+  logic [INDEX_W-1:0] inval_index_d, inval_index_q;
 
   //////////////////////////
   // Instruction prefetch //
   //////////////////////////
 
-  assign lookup_addr_aligned = {lookup_addr_ic0[ADDR_W-1:LINE_W],{LINE_W{1'b0}}};
+  assign lookup_addr_aligned = {lookup_addr_ic0[ADDR_W - 1:LINE_W], {LINE_W{1'b0}}};
```
GregAC, line 199: I'd argue we want to keep things compact in bit select expressions like these (i.e. no space around the `-`), though the style guide allows both options. I'd note that for dimensions in declarations it doesn't add the spacing.

rtl/ibex_icache.sv
```diff
@@ -221,42 +220,38 @@ module ibex_icache #(
   ////////////////////////
 
   // Cache lookup
-  assign lookup_throttle  = (fb_fill_level > FB_THRESHOLD[$clog2(NUM_FB)-1:0]);
+  assign lookup_throttle = (fb_fill_level > FB_THRESHOLD[$clog2(NUM_FB) - 1:0]);
 
-  assign lookup_req_ic0   = req_i & ~&fill_busy_q & (branch_i | ~lookup_throttle) & ~ecc_write_req;
-  assign lookup_addr_ic0  = branch_spec_i ? addr_i :
-                                            prefetch_addr_q;
+  assign lookup_req_ic0 = req_i & ~&fill_busy_q & (branch_i | ~lookup_throttle) & ~ecc_write_req;
```
GregAC, line 225: Couldn't immediately see any style guide guidance on this but I think we should use `()` around unary AND/OR/XOR reduction operators when it's in a compound expression. I.e. I think this line should be written:

```systemverilog
assign lookup_req_ic0 = req_i & ~(&fill_busy_q) & (branch_i | ~lookup_throttle) & ~ecc_write_req;
```

rtl/ibex_icache.sv
```diff
@@ -221,42 +220,38 @@ module ibex_icache #(
...
-  assign tag_index_ic0 = inval_prog_q   ? inval_index_q :
-                         ecc_write_req  ? ecc_write_index :
-                         fill_grant_ic0 ? fill_index_ic0 :
-                                          lookup_index_ic0;
-  assign tag_banks_ic0 = ecc_write_req  ? ecc_write_ways :
-                         fill_grant_ic0 ? fill_ram_req_way :
-                                          {NumWays{1'b1}};
+  assign tag_req_ic0 = lookup_req_ic0 | fill_req_ic0 | inval_prog_q | ecc_write_req;
+  assign tag_index_ic0 = inval_prog_q ? inval_index_q :
+      ecc_write_req ? ecc_write_index : fill_grant_ic0 ? fill_index_ic0 : lookup_index_ic0;
```
GregAC, line 248: Another ternary example where I think the output is clearly less readable, the aligned one condition paired with one select option seems to be the clear winner here.

rtl/ibex_load_store_unit.sv
```diff
@@ -169,12 +171,12 @@ module ibex_load_store_unit
...
-      2'b00:   data_wdata =  lsu_wdata_i[31:0];
-      2'b01:   data_wdata = {lsu_wdata_i[23:0], lsu_wdata_i[31:24]};
-      2'b10:   data_wdata = {lsu_wdata_i[15:0], lsu_wdata_i[31:16]};
-      2'b11:   data_wdata = {lsu_wdata_i[ 7:0], lsu_wdata_i[31: 8]};
-      default: data_wdata =  lsu_wdata_i[31:0];
-    endcase // case (data_offset)
+      2'b00: data_wdata = lsu_wdata_i[31:0];
+      2'b01: data_wdata = {lsu_wdata_i[23:0], lsu_wdata_i[31:24]};
+      2'b10: data_wdata = {lsu_wdata_i[15:0], lsu_wdata_i[31:16]};
+      2'b11: data_wdata = {lsu_wdata_i[7:0], lsu_wdata_i[31:8]};
```
GregAC, line 177: While it's a minor change here that doesn't effect readability a huge amount, in general preserving alignment in these concatenation operations can be very useful in seeing how the various fields line up. I note in ibex_tracer_pkg there is an example of where it's useful to leave the alignment where verible has left things as is.

rtl/ibex_icache.sv
rswarbrick: Hmm, maybe. But don't forget that `~&` is a unary operator in SystemVerilog (the grammar is actually ambiguous, but it doesn't matter because `~(&x)` has the same meaning as "`(~&) x`", if that were valid syntax). Are you saying that we should forbid using it as an operand to a binary `&`?

GregAC: Ah I hadn't realised that `~&` was it's own operator.

Maybe in this case it's fine but certainly where you don't have an `~` the `()` are useful

```systemverilog
assign something = signal_a & &signal_b & signal_c;
```

vs

```systemverilog
assign something = signal_a & (&signal_b) & signal_c;
```

tomroberts-lowrisc: I think things like `~&` are ok as unary operators without their own parentheses. Slightly different but related, some linters (Cadence maybe?) used to warn about things like `foo & &bar` as being ambiguous, preferring `foo & (&bar)`, which is reasonable

GregAC:
Three more verible issues opened, two on ternaries, one on whitespace around operations in bit selects:

https://github.com/google/verible/issues/359
https://github.com/google/verible/issues/360
https://github.com/google/verible/issues/361

imphil:
This PR served its purpose, with Verible lint having picked up many of our suggestions. We'll need to do a next round some time to see how far we've come.

PR #1015, opened 2020-07-05
[doc] Fix spelling of CoreMark

vogelpi:


vogelpi:
Lol. While at it, should I also create a PR upstream?

Thanks for reviewing + approving guys.

PR #1025, opened 2020-07-09
increase time in Icache simulation

dawidzim:
Hey!
Can we increase time of simulation in Icache tests?

### DIFF
--- a/dv/uvm/data/riviera/riviera.hjson
+++ b/dv/uvm/data/riviera/riviera.hjson
@@ -11,7 +11,8 @@
 
   build_opts: ["-sv",
                "-timescale 1ns/1ps",
-               "-uvmver 1.2",
+               "+incdir+\"{ALDEC_PATH}/vlib/uvm-1.2/src\"",
+               "\"{ALDEC_PATH}/vlib/uvm-1.2/src/uvm_pkg.sv\"",
                "-f {sv_flist}"]
 
   run_opts:   ["-sv_seed={seed}",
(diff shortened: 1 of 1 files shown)


rswarbrick:
Please can you give a specific run that fails without the increased timeout? I'd intended this number to be way larger than needed (using a heartbeat for more fine-grained checks).

danielmlynek:
I've took closer look on this issue. You are setting time out in code to 1s by somthing like this in yours design:
```
test_timeout_ns = 1000000000 
uvm_top.set_timeout((test_timeout_ns * 1ns));
```
and you are compiling your design with  -timsecale 1ns/1ps. 
It may not work as you expected.
Value passed to UVM set_timeout is calculated as 1000000000  basing on 1ns/1ps timescale.
But if you are using precompiled UVM it may be compiled with other timescale depending on compilation option used when it was compiled or tools default timescale value (uvm does not set timescale int the code).
In this case for us precompiled UVM  timescale is 1ps/1ps - so UVM gets 1000000000 in set timeout but interprets it as ps. As a result timeout is 1000 times smaller that you expect. That is why we are getting timeouts.

It is hard to find perfect solution. One of them is to recompile the UVM with -timescale 1ns/ps (or whatever you will use for your design).
Another solution is to simple compile the UVM together with design in single invocation of compilation command.
It can be also fixed by change it the code:
`uvm_top.set_timeout((test_timeout_ns * 1ns)/<precompiled_uvm_timeunit>);`
in our case 
`uvm_top.set_timeout((test_timeout_ns * 1ns)/1ps);`
but im not sure if it will for all tools as in each UVM timescale may be different.

dawidzim:
I restore default timeout and I add inline UMV compilation in Riviera hjson file

rswarbrick:
@danielmlynek: Thank you very much for the explanation. That makes lots of sense.

I think that @dawidzim's solution is probably the cleanest, since I don't think we really have any control over the timescale used by a previously-compiled UVM. Could one of you squash these two commits and put an explanation in the commit message? Daniel's explanation from above would be perfect.

PR #1028, opened 2020-07-09
[dv/ibex] Switch to request/response terminology

udinator:
This PR modifies the Ibex DV environment to match OpenTitan terminology
standards of using host/device. These changes are purely aesthetic.


sjgitty:
In general this looks good, but agreeing that the wording is a bit awkward here and there. I like @rswarbrick 's suggest names on the memory side.

udinator:
@rswarbrick @sjgitty after giving this a second look, I'm inclined to agree as well - using request/response probably works better for this environment.

sjgitty:
Recommend you change the title of the PR before merging, but otherwise looks good to go.

PR #1031, opened 2020-07-10
Coremark updates

GregAC:
Felt the README should reflect the numbers actually achieved (as `SpecBranch` reduces performance for the PMP enabled config). Also felt we should show the best results achievable so added ISA configurability to the coremark build as different configs get better results with different ISA strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ These are configurations on which lowRISC is focusing for performance evaluation
 | Config | "small" | "maxperf" | "maxperf-pmp-bmfull" |
 | ------ | ------- | --------- | ---------------- |
 | Features | RV32IMC, 3 cycle mult | RV32IMC, 1 cycle mult, Branch target ALU, Writeback stage | RV32IMCB, 1 cycle mult, Branch target ALU, Writeback stage, 16 PMP regions |
-| Performance (CoreMark/MHz) | 2.44 | 3.09 | 3.09 |
+| Performance (CoreMark/MHz) | 2.47 | 3.13 | 3.05 |
 | Area - Yosys (kGE) | 33.15 | 39.03 | 63.32 |
 | Area - Commercial (estimated kGE) | ~27 | ~31 | ~50 |
 | Verification status | Green | Amber | Amber |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,15 +23,17 @@ These are configurations on which lowRISC is focusing for performance evaluation
...
 | Area - Commercial (estimated kGE) | ~27 | ~31 | ~50 |
 | Verification status | Green | Amber | Amber |
 
 Notes:
 
 * Performance numbers are based on CoreMark running on the Ibex Simple System [platform](examples/simple_system/README.md).
-  Note that CoreMark was compiled without support for the B extension.
+  Note that different ISAs (use of B and C extensions) give the best results for different configurations.
+  See the [Benchmarks README](examples/sw/benchmarks/README.md) for more information.
+  The "maxperf-pmp-bmfull" configuration sets a `SpecBranch` parameter in `ibex_core.sv`, this helps timing but has a small negative performance impact.
```
rswarbrick, line 36: Tiny nit, but it should be "... `ibex_core.sv`; this helps..."

examples/sw/benchmarks/README.md
```diff
@@ -62,6 +62,32 @@ seedcrc          : 0xe9f5
...
+
+```
+make -C ./examples/sw/bencharmsk/coremark clean
+make -C ./examples/sw/benchmarks/coremark RV_ISA=rv32imc
+```
+
+This will build CoreMark using the 'C' extension (compressed instructions).
+
+If CoreMark is already built a `clean` must be done before building
+for a new ISA string to ensure a full rebuild.
```
rswarbrick, line 78: Since I've started with the nitpicks... This sentence is quite hard to read. Maybe simpler: "When changing `RV_ISA`, you must clean out any old build with `make clean` and rebuild."

Also, you changed "CoreMark" to "Coremark" at the top of the file. Which capitalisation is correct?

examples/sw/benchmarks/README.md
```diff
@@ -62,6 +62,32 @@ seedcrc          : 0xe9f5
...
+
+The following ISA strings give the best performance for the Ibex configurations
+listed in the README:
```
rswarbrick, line 81: But poor performance for other configurations? :-)

Maybe something like "The following table shows which ISA string gives the best performance with each Ibex configuration listed in the README"? I'm not sure though - that still sounds slightly ambiguous (it implies that there's an ISA string to rule them all!). Any other ideas?

GregAC: I don't think there's much difference in meaning between the two sentences. Ultimately I'm just trying to simply communicate how someone can reproduce our numbers, I don't think we need to worry too much about the subtleties.

GregAC:
Ah turns out you can still reply to outdated comments but the box disappears if you've got pending review comments and come back to the top-level PR...

PR #1032, opened 2020-07-10
Branch predict wip

GregAC:
Still work to be done here, notably getting it work with the icache. I also need to see what happens in synthesis and ponder if there's better ways for the speculative branch stuff to interact with branch prediction (currently with them both on CoreMark suffers, need to look into this).

However this is the basic implementation and it'd be good to get this in whilst I get further work sorted, looks good in DV though will do a longer run this weekend.

With this you can achieve a CoreMark/MHz of 3.28, that's with bitmanip using rv32imb (no C) for the build.

I've left it turned off in the maxperf configs for now until I've got the other things sorted so I haven't altered the headline CoreMark scores in the README.


rtl/ibex_if_stage.sv
```diff
@@ -343,18 +375,134 @@ module ibex_if_stage #(
...
+        PC_ERET,
+        PC_DRET,
+        PC_BP},
+      pc_set_i)
+
+`ifdef INC_ASSERT
+    /**
+     * Checks for branch prediction interface to fetch_fifo/icache
+     *
+     * The interface has two signals:
```
GregAC, line 427: @tomroberts-lowrisc @rswarbrick please note this hopefully describes the extra interface stuff that will be required for the icache. I will sort out the icache RTL to add support though with @rswarbrick to look at the testbench side of things.

If any of this sounds problematic do let me know.

rtl/ibex_if_stage.sv
```diff
@@ -343,18 +375,134 @@ module ibex_if_stage #(
...
+    end
+
+    // Must only see mispredict after we've performed a predicted branch but before we've accepted
+    // any instruction (with fetch_ready & fetch_valid) that follows that predicted branch.
+    `ASSERT(MispredictOnlyImmediatelyAfterPredictedBranch,
+      nt_branch_mispredict_i |-> predicted_branch_live_q);
+    // Check that on mispredict we get the correct PC for the non-taken side of the branch when
+    // prefetch buffer/icache makes that PC available.
+    `ASSERT(CorrectPCOnMispredict,
+      predicted_branch_live_q & mispredicted_d & fetch_valid |-> fetch_addr == predicted_branch_nt_pc_q);
```
GregAC, line 488: I could dump some/all of the tracking logic by using more sophisticated SVAs but felt it's better keeping it simple as it means it may be usable with our yosys based formal flow and may be useable with verilator when basic SVA support is running (and used by our builds).

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,93 @@
...
+      instr_cj : branch_imm = imm_cj_type;
+      instr_cb : branch_imm = imm_cb_type;
+      default : ;
+    endcase
+  end
+
+  `ASSERT_IF(BranchInsTypeOnehot, $onehot0({instr_j, instr_b, instr_cj, instr_cb}), fetch_valid);
+
+  // Determine branch prediction, taken if offset is negative
+  assign instr_b_taken = (instr_b && imm_b_type[31]) || (instr_cb && imm_cb_type[31]);
```
tomroberts-lowrisc, line 87: I think we're generally favoring bitwise for this sort of assignment

rtl/ibex_prefetch_buffer.sv
```diff
@@ -9,14 +9,18 @@
  * Prefetch Buffer that caches instructions. This cuts overly long critical
  * paths to the instruction cache.
  */
-module ibex_prefetch_buffer (
+module ibex_prefetch_buffer #(
+  parameter bit          BranchPredictor   = 1'b0
```
vogelpi, line 13: Nit: indentation.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -172,13 +187,37 @@ module ibex_prefetch_buffer (
     end
   end
 
+  if (BranchPredictor) begin : g_branch_predictor
```
vogelpi, line 190: Maybe a comment would be useful saying that in case of a predicted branch, we need to store the address in order to go back if the prediction turns out to be wrong.

rtl/ibex_if_stage.sv
```diff
@@ -189,40 +215,46 @@ module ibex_if_stage #(
...
     );
     // ICache tieoffs
     logic unused_icen, unused_icinv;
     assign unused_icen  = icache_enable_i;
     assign unused_icinv = icache_inval_i;
   end
 
-  assign branch_req  = pc_set_i;
+
+  assign branch_req  = pc_set_i | (predict_branch_taken & id_in_ready_i);
```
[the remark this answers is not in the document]
GregAC: Yes it's due to the flush. If we immediately start fetching the predicted branch we can loose the branch instruction itself if the ID stage doesn't accept it the cycle we do the branch predict.

Setting things up so we can begin fetching the predicted branch as soon as we've done that prediction does indeed result in better performance but also adds complexity. I figured that if you have lots of places where the IF stage is getting stalled by instructions ahead the branch prediction probably isn't giving you much in performance anyway so this isn't a big deal. We can always revisit this decision but I wanted to keep it simple for now.

GregAC:
Thanks for the reviews @tomroberts-lowrisc and @vogelpi. I realised I never did do that longer DV run so I'll do that this weekend!

rtl/ibex_if_stage.sv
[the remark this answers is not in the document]
GregAC: I should maybe add a comment to explain this

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,93 @@
...
+  logic instr_j;
+  logic instr_b;
+  logic instr_cj;
+  logic instr_cb;
+
+  logic instr_b_taken;
+
+  // Extract and sign-extend to 32-bit the various immediates that may be used to calculate the
+  // target
+  assign imm_j_type = { {12{fetch_rdata[31]}}, fetch_rdata[19:12], fetch_rdata[20],
```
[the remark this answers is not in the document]
vogelpi: Thanks for changing this.

rtl/ibex_if_stage.sv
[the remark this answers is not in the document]
vogelpi: Thanks @GregAC for explaining and documenting this design decision. It sounds completely reasonable to me :+1:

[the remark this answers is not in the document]
GregAC: Frustratingly it's not that easy after all!

The CI has shown there is a combinational loop in the RISC-V compliance system (but not in simple system). This is because this RTL adds a feedthrough path from data_gnt -> instr_req, data_gnt -> id_in_ready_i -> branch_req -> instr_req (there are some extra steps but that explains what's happening clearly). In the RISCV-V compliance system we don't have a dual ported memory just the bus arbitrates between I and D side. This means that data_gnt depends upon instr_req (instruction requests get priority) so we end up with a loop.

This doesn't appear in simple system as the I and D side use different ports on the memory so there's no coupling between the req and gnt signals.

I did think about having some `restricted_id_in_ready_i` signal that didn't include `data_gnt` so on a branch prediction we'd just send out the request then if it turns out we don't actually have `id_in_ready_i` don't clear the FIFO or any other state and set the appropriate bit to ignore the response when it comes back but this leads to deadlock in some cases (e.g. if you give I side static priority over D side).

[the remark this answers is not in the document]
GregAC: I have two fixes for this feedthrough, one is fiddly and a bit weird but shouldn't add much area, the other is simple but adds a bunch more flops:

1. Introduce the `restricted_in_ready_i` described above however the cycle following the first attempt at a `branch_req` for a predicted branch if we didn't have `id_in_ready_i` with `restricted_in_ready_i` stop `branch_req` until `id_in_ready_i` appears then do `branch_req` the following cycle. What this means is where we have a predicted branch and ID/EX is stalled awaiting a grant on the data interface we'll send out a request on the instruction interface for the predicted branch, the response will then be ignored if ID/EX didn't unstall and we'll wait for the data grant to happen then send the same instruction request out the cycle after the data grant is seen.

2. Add a buffer for the instruction, so 32 extra flops for the instruction bits and 32 extra flops for the instruction address plus a few extra flops for valid bits, error bits etc. Here we if we do a predicted branch where we don't have `id_in_ready_i` the instruction in IF (the one that was a predicted branch) gets placed into the buffer so we can branch and we move from the buffer to ID/EX when `id_in_ready_i` is asserted

I'm tending towards 2. from the simple is better perspective, plus it does offer some performance advantages (you fetch your predicted branch in the shadow of a stalling instruction, though I suspect that this has little advantage practically). The extra area shouldn't be a big concern for anyone looking for a high performance Ibex and branch prediction on what is otherwise a super small config would be an odd configuration point. Thoughts @vogelpi ? I'll wait for @tomroberts-lowrisc to return from holiday to get his input before I merge anything.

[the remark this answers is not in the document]
GregAC: I may be able to be clever on the extra set of instruction address flops also and avoid adding them (given there's already various sets of flops holding addresses).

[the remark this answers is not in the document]
vogelpi: I think I would go for Option 2 as well. It sounds a lot simpler and I agree with you that in such a configuration, some area overhead is probably acceptable.

I am wondering whether this additional instr register for Option 2 could also be used to buffer the next instruction when flushing the fetch FIFO. In case of a mis-predicted taken branch this could perhaps help to avoid the 2-cycle penalty due to refetching a misaligned uncompressed instruction. But maybe this is super complex. It's just a thought.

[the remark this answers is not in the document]
tomroberts-lowrisc: Agreed sounds best to go for option 2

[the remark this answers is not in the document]
GregAC: Good point on the dual use of the extra buffer @vogelpi will also save doing specialised work on the icache to handle that case.

GregAC:
Latest version introduces the skid buffer to break up the timing path. DV looks good though there's a potential issue around predicted branches causing instruction fetches to areas the DV environment considers invalid causing an error.

However this doesn't turn branch prediction on in any of the configs so I think this sufficient quality to go in. We will enable it in some configs later.

rtl/ibex_if_stage.sv
```diff
@@ -191,43 +218,53 @@ module ibex_if_stage #(
...
   end
 
-  assign branch_req  = pc_set_i;
-  assign fetch_ready = id_in_ready_i & ~stall_dummy_instr;
+  // For predicted branches only set branch_req when the ID/EX stage is ready to accept the branch
+  // instruction. Otherwise the branch instruction ends up getting flush out of the IF stage by the
+  // branch_req and is lost. Whilst it is possible to begin fetching the predicted branch without
+  // flushing the branch instruction from IF this adds design complexity and for situations where
+  // ID/EX stage stalls are common more timely fetching of branches is likely to have limited
+  // performance impact.
```
vogelpi, line 263: I think this comment is outdated as we no longer factor in the ID/EX stage ready.

rtl/ibex_if_stage.sv
```diff
@@ -268,11 +305,11 @@ module ibex_if_stage #(
     );
 
     // Mux between actual instructions and dummy instructions
-    assign fetch_valid_out         = insert_dummy_instr | fetch_valid;
+    //assign fetch_valid_out         = insert_dummy_instr | if_instr_valid;
```
vogelpi, line 308: Nit: remove?

rtl/ibex_if_stage.sv
```diff
@@ -298,21 +335,21 @@ module ibex_if_stage #(
     assign unused_dummy_mask       = dummy_instr_mask_i;
     assign unused_dummy_seed_en    = dummy_instr_seed_en_i;
     assign unused_dummy_seed       = dummy_instr_seed_i;
-    assign fetch_valid_out         = fetch_valid;
+    //assign fetch_valid_out         = fetch_valid;
```
vogelpi, line 338: Nit: remove?

rtl/ibex_if_stage.sv
```diff
@@ -371,18 +408,199 @@ module ibex_if_stage #(
...
+
+    // ID stages needs to know if branch was predicted taken so it can signal mispredicts
+    always_ff @(posedge clk_i) begin
+      if (if_id_pipe_reg_we) begin
+        instr_bp_taken_q <= instr_bp_taken_d;
+      end
+    end
+
+    // When branch prediction is enabled a skid buffer between the IF and ID/EX stage is introduced.
+    // If an instruction in IF is predicted to be a branch and ID/EX is not ready the instruction is
```
vogelpi, line 429: Nit: branch -> taken branch

vogelpi: I would emphasize/clarify that the instruction **currently executed in ID/EX** is moved to the skid buffer (and not the newly fetched) such that the prefetch buffer can be flushed. Looking at the code it's obvious that this is what is happening, but I feel that stating this explicitly wouldn't hurt.

rtl/ibex_if_stage.sv
```diff
@@ -371,18 +408,199 @@ module ibex_if_stage #(
...
+    // moved to the skid buffer which becomes the output of the IF stage until the ID/EX stage
+    // accepts the instruction. The skid buffer is required as otherwise the ID/EX ready signal is
+    // coupled to the instr_req_o output which produces a feedthrough path from data_gnt_i ->
+    // instr_req_o (which needs to be avoided as for some interconnects this will result in
+    // a combinational loop).
+
+    assign instr_skid_en = predicted_branch & ~id_in_ready_i & ~instr_skid_valid_q;
+
+    assign instr_skid_valid_d = (instr_skid_valid_q & ~id_in_ready_i & ~stall_dummy_instr) | instr_skid_en;
```
vogelpi, line 438: This statement exceeds 100 chars and should be wrapped.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -209,7 +252,7 @@ module ibex_prefetch_buffer (
       // If a branch is received at any point while a request is outstanding, it must be tracked
       // to ensure we discard the data once received
       assign branch_discard_n[i]    = (valid_req & gnt_or_pmp_err & discard_req_d) |
-                                      (branch_i & rdata_outstanding_q[i]) | branch_discard_q[i];
+                                      (branch_or_mispredict & rdata_outstanding_q[i]) | branch_discard_q[i];
```
vogelpi, line 255: Nit: line limit exceeded.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -223,7 +266,7 @@ module ibex_prefetch_buffer (
                                       rdata_outstanding_q[i];
       assign branch_discard_n[i]    = (valid_req & gnt_or_pmp_err & discard_req_d &
                                        rdata_outstanding_q[i-1]) |
-                                      (branch_i & rdata_outstanding_q[i]) | branch_discard_q[i];
+                                      (branch_or_mispredict & rdata_outstanding_q[i]) | branch_discard_q[i];
```
vogelpi, line 269: Nit: this statement should be wrapped.

doc/instruction_fetch.rst
```diff
@@ -26,6 +26,16 @@ The interfaces of the icache module are the same as the prefetch buffer with two
...
 Secondly a signal to the flush the cache which is set every time a ``fence.i`` instruction is executed.
 
+Branch Prediction
+-----------------
+
+Ibex can be configured to use static branch prediction by setting the ``BranchPrediction`` parameter to 1.
+This improves performance by predicting any branch with a negative offset to be taken and any branch with a positive offset to be not taken.
+When successful the prediction removes a stall cycle from a taken branch.
+There is a mis-predict penalty when a not-taken branch is predicted taken.
+This will be a minimum of 1 cycle where the not-taken side of the branch is aligned (either a 16-bit compressed instruction or a 32-bit instruction aligned to a 32-bit boundary) and a minimum of 2 cycles where it isn't.
```
rswarbrick, line 36: There's an inconsistency here about whether to hyphenate "not taken". I realise that sometimes we want to label branches as taken or not, but the resulting English feels a bit lumpy and I don't think we need it here. How about this?
```
This improves performance by predicting that any branch with a negative offset is taken and that any branch with a positive offset is not.
When successful, the prediction removes a stall cycle from a taken branch.
However, there is a misprediction penalty if a branch is wrongly predicted to be taken.
This penalty is at least one cycle, or at least two cycles if the fall-through instruction is uncompressed and not aligned.
```

ibex_configs.yaml
```diff
@@ -68,7 +72,24 @@ experimental-maxperf-pmp-bmfull:
   BranchTargetALU          : 1
   WritebackStage           : 1
   MultiplierImplementation : "single-cycle"
+  BranchPredictor          : 0
   PMPEnable                : 1
   PMPGranularity           : 0
   PMPNumRegions            : 16
 
+# experimental-maxperf with branch predictor switched on, this exists to allow
```
rswarbrick, line 80: The comma is wrong here: use a colon, semicolon or full stop.

lint/verilator_waiver.vlt
```diff
@@ -12,15 +12,24 @@
...
+// We have some boolean top-level parameters in e.g. ibex_core_tracing.sv.
+// When building with fusesoc, these get set with defines like
+// -GRV32M=1 (rather than -GRV32M=1'b1), leading to warnings like:
+//
+//   Operator VAR '<varname>' expects 1 bits on the Initial value, but
+//   Initial value's CONST '32'h1' generates 32 bits.
+//
+// This signoff rule ignores errors like this. Note that it only
+// matches when you set a 1-bit value to a literal 1, so it won't hide
+// silly mistakes like setting it to 2.
```
rswarbrick, line 24: I love the fact that my text has propagated through the design, but this is now the third copy :-D Not as part of this PR, but I wonder whether there's a way we could consolidate things to avoid the repetition?

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,100 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Branch Predictor
+ *
+ * This implements static branch prediction. It takes an instruction and it's PC and determines if
```
rswarbrick, line 8: Nit: "its PC" (possessive doesn't need the apostrophe)

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,100 @@
...
+ * it's a branch or a jump and calculates its target. For jumps it will always predict taken. For
+ * branches it will predict taken if the PC offset is negative.
+ *
+ * This handles both compressed and uncompressed instructions. Compressed instructions must be in
+ * the lower 16-bits of instr.
+ *
+ * The predictor is entirely combinitorial but takes clk/rst_n signals for use by assertions.
```
rswarbrick, line 15: Nit: combinatorial

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,100 @@
...
+    unique case (1'b1)
+      instr_j  : branch_imm = imm_j_type;
+      instr_b  : branch_imm = imm_b_type;
+      instr_cj : branch_imm = imm_cj_type;
+      instr_cb : branch_imm = imm_cb_type;
+      default : ;
+    endcase
+  end
+
+  `ASSERT_IF(BranchInsTypeOnehot, $onehot0({instr_j, instr_b, instr_cj, instr_cb}), fetch_valid_i);
```
rswarbrick, line 91: Nit: Maybe "...OneHot" for consistent CamelCase? (I refuse to believe that onehot is a word!)

rtl/ibex_branch_predict.sv
```diff
@@ -0,0 +1,100 @@
...
+
+  // Provide short internal name for fetch_rdata_i due to reduce line wrapping
+  assign instr = fetch_rdata_i;
+
+  // Extract and sign-extend to 32-bit the various immediates that may be used to calculate the
+  // target
+
+  // Uncompressed immediates
+  assign imm_j_type = { {12{instr[31]}}, instr[19:12], instr[20], instr[30:21], 1'b0 };
+  assign imm_b_type = { {19{instr[31]}}, instr[31], instr[7], instr[30:25], instr[11:8], 1'b0 };
```
rswarbrick, line 59: Could these be renamed like "b_type_imm"? I had to think a bit when I was reading the definition of `instr_b_taken`, wondering "why does this depend on the top bit of a type?". Renaming to "foo_imm" makes it sound more like it's a "foo immediate", I think.

rtl/ibex_controller.sv
```diff
@@ -26,14 +27,16 @@ module ibex_controller #(
...
-    input  logic [31:0]           instr_i,                 // instr from IF-ID reg, for mtval
-    input  logic [15:0]           instr_compressed_i,      // instr from IF-ID reg, for mtval
-    input  logic                  instr_is_compressed_i,   // instr from IF-ID reg is compressed
-    input  logic                  instr_fetch_err_i,       // instr from IF-ID reg has error
-    input  logic                  instr_fetch_err_plus2_i, // instr from IF-ID reg error is x32
-    input  logic [31:0]           pc_id_i,                 // instr from IF-ID reg address
+    // instr from IF-ID pipeline stage
+    input  logic                  instr_valid_i,           // instr is valid
+    input  logic [31:0]           instr_i,                 // instr data (uncompressed if compressed
+                                                           // otherwise raw data) for mtval
```
rswarbrick, line 33: Could this comment just be "uncompressed instr data for mtval"? Uncompressing an uncompressed instruction is the identity, after all. Also, is the "for mtval" bit needed? Later lines seem to drop it.

rtl/ibex_controller.sv
```diff
@@ -59,6 +64,7 @@ module ibex_controller #(
     // jump/branch signals
     input  logic                  branch_set_i,            // branch taken set signal
     input  logic                  branch_set_spec_i,       // speculative branch signal
+    input  logic                  branch_not_set_i,        // branch was not taken
```
rswarbrick, line 67: This is a bit confusing, I think. If I've understood it right, these signals mean "definitely branch", "probably branch" and "definitely don't branch", respectively. Is this right? If so, could the top and bottom comments gain a "definitely" or similar?

rtl/ibex_controller.sv
```diff
@@ -371,27 +377,28 @@ module ibex_controller #(
     pc_mux_o              = PC_BOOT;
     pc_set_o              = 1'b0;
     pc_set_spec_o         = 1'b0;
+    nt_branch_mispredict_o = 1'b0;
 
-    exc_pc_mux_o          = EXC_PC_IRQ;
-    exc_cause_o           = EXC_CAUSE_INSN_ADDR_MISA; // = 6'h00
+    exc_pc_mux_o           = EXC_PC_IRQ;
```
rswarbrick, line 382: The respacing of the equals signs here is inconsistent. I guess it's supposed to line up with `nt_branch_mispredict_o`? But then I think you should either re-indent everything or nothing.

rtl/ibex_controller.sv
```diff
@@ -495,16 +502,29 @@ module ibex_controller #(
           end
         end
 
-        if ((branch_set_i || jump_set_i) && !special_req_branch) begin
-          pc_set_o       = 1'b1;
+        if (!special_req_branch) begin
+          if ((branch_set_i || jump_set_i)) begin
```
rswarbrick, line 506: Stray parens?

rtl/ibex_id_stage.sv
```diff
@@ -690,11 +698,12 @@ module ibex_id_stage #(
...
   end
 
-  // Holding branch_set/jump_set high for more than one cycle may not cause a functional issue but
-  // could generate needless prefetch buffer flushes and instruction fetches. ID/EX is designed such
-  // that this shouldn't ever happen.
-  `ASSERT(NeverDoubleBranch, branch_set |=> ~branch_set)
-  `ASSERT(NeverDoubleJump, jump_set |=> ~jump_set)
+  // Unless the first branch/jump was predicted holding branch_set/jump_set high for more than one
+  // cycle may not cause a functional issue but could generate needless prefetch buffer flushes and
+  // instruction fetches. ID/EX is designed such that this shouldn't ever happen for non-predicted
```
rswarbrick, line 703: I took a couple of goes to read this sentence. Maybe it would be easier if you split it up? "If the first branch/jump was not predicted then ... functional issue. However, it could generate ..." ?

While you're here, I think the "such that" in the last sentence is a bit ugly. How about "The ID/EX design ensures this never happens for non-predicted branches." ?

rtl/ibex_if_stage.sv
```diff
@@ -141,14 +158,24 @@ module ibex_if_stage #(
...
       PC_BOOT: fetch_addr_n = { boot_addr_i[31:8], 8'h80 };
       PC_JUMP: fetch_addr_n = branch_target_ex_i;
       PC_EXC:  fetch_addr_n = exc_pc;                       // set PC to exception handler
       PC_ERET: fetch_addr_n = csr_mepc_i;                   // restore PC when returning from EXC
       PC_DRET: fetch_addr_n = csr_depc_i;
+      // Without branch predictor will never get pc_mux_internal == PC_BP, still handle no branch
+      // predictor case here to ensure redundant mux logic isn't synthesised.
+      PC_BP:   begin
+        fetch_addr_n = BranchPredictor ? predict_branch_pc : { boot_addr_i[31:8], 8'h80 };
+      end
```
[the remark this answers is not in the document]
rswarbrick: Also, the comment should probably be "Without branch predictor, we will never ... PC_BP. We still handle ..." (otherwise the first clause has no subject and we've joined two independent clauses with a comma).

rtl/ibex_if_stage.sv
rswarbrick: That sounds sensible to me. I think we'll eventually need some timing diagrams in the documentation.

rtl/ibex_prefetch_buffer.sv
```diff
@@ -9,14 +9,18 @@
  * Prefetch Buffer that caches instructions. This cuts overly long critical
  * paths to the instruction cache.
  */
-module ibex_prefetch_buffer (
+module ibex_prefetch_buffer #(
+  parameter bit BranchPredictor = 1'b0
+)(
```
rswarbrick, line 14: Nit: Should there be a space between the back-to-back parens here?

rtl/ibex_branch_predict.sv
GregAC: I did this to keep thing consistent with the decoder that uses the same naming (e.g. `imm_i_type`)

rtl/ibex_if_stage.sv
GregAC: The instruction moved to the skid is the one in IF not the one in ID/EX. When predicting a branch it's done on the instruction in IF so if ID/EX can't immediately accept it it moves to the skid which IF keeps outputting until ID/EX is ready (so from the ID/EX perspective the IF outputs remain stable).

GregAC:
> It would be nice if we could use it also to buffer the next instructions from the prefetch buffer in case of mis-predictions

I agree, something I certainly want to investigate. It's not quite perfect as if you won't be able to use it in all cases, namely where you have a predicted branch but ID/EX isn't ready but I anticipate a lack of ID/EX stalls in many cases so it should work well :)

rtl/ibex_if_stage.sv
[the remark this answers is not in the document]
vogelpi: Thanks, this has been done.

rtl/ibex_if_stage.sv
vogelpi: Thanks for clarifying that, I completely messed up what I wanted to say. But it's all clear now.

dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
```diff
@@ -24,6 +24,7 @@ module ibex_riscv_compliance (
   parameter ibex_pkg::regfile_e RegFile = ibex_pkg::RegFileFF;
   parameter bit BranchTargetALU         = 1'b0;
   parameter bit WritebackStage          = 1'b0;
+  parameter bit BranchPredictor          = 1'b0;
```
vogelpi, line 27: Nit: alignment

vogelpi (approved):
Thanks @GregAC , LGTM!

GregAC:
Issue opened for using the skid buffer for performance improvements: https://github.com/lowRISC/ibex/issues/1097

PR #1034, opened 2020-07-15
[rtl] remove lsu_req_in_id signal

### LINKED ISSUE #1029: Single stepping over `lw` doesn't cause a jump back to Debug Mode
Hi all,
I'm seeing something interesting occur when I run `riscv_debug_single_step_test` on Ibex in the `experimental-maxperf-pmp-bmfull` configuration.
Screenshot from 2020-07-09 16-22-06
In the waveform, Ibex has just finished single stepping over the instruction at `pc_id=0x8000_b25e`, and has just executed the `dret` indicating the end of the debug rom at `pc_id=0x8000_b7ca` (`debug_single_step_i` in the controller is still high, so I am expecting Ibex to execute the next instruction it sees and re-enter debug mode).
Ibex correctly returns to execute the instruction at `pc_id=0x8000_b262`, which is a `lw` instruction, and the controller raises `enter_debug_mode` accordingly as the core stalls to allow the memory transaction to finish. However after the load transaction has cleared (the cycle before marker `M1`), `enter_debug_mode` drops to 0 as `instr_valid_i` suddenly drops to 0. 
This causes the core to incorrectly execute the next instruction (`pc_id=0x8000_b266`) and step into debug mode only after this instruction has passed through the ID stage (at marker `M2`) - Ibex should have stepped into debug mode after `pc_id=0x8000_b262`.

GregAC:
This signal aimed to ensure loads/stores completed succesfully when an
interrupt or debug request appeared at the same time they were being
executed when the writeback stage is present. However other stall logic
suffices for this purpose (debug/interrupt will wait for instruction to
unstall in ID/EX which only happens once request has been sent out, then
first instruction of debug/interrupt handler will stall until load/store
response has been seen based on the generic stall logic for lsu requests
in the writeback stage).

With this signal in place debug single stepping was broken around loads
and stores.

Fixes #1029

This half undoes the fix from this PR: https://github.com/lowRISC/ibex/pull/928 which was in response to this OT issue: https://github.com/lowRISC/opentitan/issues/2405

I'm happy to accept this is not the correct fix but I did have a play with interrupts appearing the same cycle as a load in simple system with various delays on gnt/rvalid and it all seems to work fine. The DV interrupt tests all looked fine. The test in OT that found the original bug was broken on main yesterday (chip_uart_tx_rx) but looking at the latest nightly it seems to be fixed, so I'll have another play with that. Also realised there's some other timing scenarios I can check out with simple system.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -90,7 +90,6 @@ module ibex_controller #(
     input  logic                  csr_mstatus_tw_i,
 
     // stall & flush signals
-    input  logic                  lsu_req_in_id_i,
     input  logic                  stall_id_i,
     input  logic                  stall_wb_i,
     output logic                  flush_id_o,
@@ -510,11 +509,11 @@ module ibex_controller #(
         // If entering debug mode or handling an IRQ the core needs to wait
         // until the current instruction has finished executing. Stall IF
         // during that time.
-        if ((enter_debug_mode || handle_irq) && (stall || lsu_req_in_id_i)) begin
+        if ((enter_debug_mode || handle_irq) && stall) begin
           halt_if = 1'b1;
         end
 
-        if (!stall && !lsu_req_in_id_i && !special_req_all) begin
+        if (!stall && !special_req_all) begin
           if (enter_debug_mode) begin
             // enter debug mode
             ctrl_fsm_ns = DBG_TAKEN_IF;
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -206,7 +206,6 @@ module ibex_id_stage #(
   logic        controller_run;
   logic        stall_ld_hz;
   logic        stall_mem;
-  logic        lsu_req_in_id;
   logic        stall_multdiv;
   logic        stall_branch;
   logic        stall_jump;
@@ -597,8 +596,6 @@ module ibex_id_stage #(
       .debug_ebreaku_i                ( debug_ebreaku_i         ),
       .trigger_match_i                ( trigger_match_i         ),
 
-      // stall signals
-      .lsu_req_in_id_i                ( lsu_req_in_id           ),
       .stall_id_i                     ( stall_id                ),
       .stall_wb_i                     ( stall_wb                ),
       .flush_id_o                     ( flush_id                ),
@@ -883,10 +880,6 @@ module ibex_id_stage #(
     `ASSERT(IbexStallMemNoRequest,
       instr_valid_i & lsu_req_dec & ~instr_done |-> ~lsu_req_done_i)
 
-    // Indicate to the controller that an lsu req is in ID stage - we cannot handle interrupts or
-    // debug requests until the load/store completes
-    assign lsu_req_in_id = instr_valid_i & lsu_req_dec;
-
     assign rf_rd_a_wb_match = (rf_waddr_wb_i == rf_raddr_a_o) & |rf_raddr_a_o;
     assign rf_rd_b_wb_match = (rf_waddr_wb_i == rf_raddr_b_o) & |rf_raddr_b_o;
 
@@ -931,7 +924,6 @@ module ibex_id_stage #(
 
     // No load hazards without Writeback Stage
     assign stall_ld_hz   = 1'b0;
-    assign lsu_req_in_id = 1'b0;
 
     // Without writeback stage any valid instruction that hasn't seen an error will execute
     assign instr_executing = instr_valid_i & ~instr_fetch_err_i & controller_run;


rtl/ibex_controller.sv
```diff
@@ -510,11 +509,11 @@ module ibex_controller #(
         // If entering debug mode or handling an IRQ the core needs to wait
         // until the current instruction has finished executing. Stall IF
         // during that time.
-        if ((enter_debug_mode || handle_irq) && (stall || lsu_req_in_id_i)) begin
+        if ((enter_debug_mode || handle_irq) && stall) begin
           halt_if = 1'b1;
         end
 
-        if (!stall && !lsu_req_in_id_i && !special_req_all) begin
+        if (!stall && !special_req_all) begin
```
tomroberts-lowrisc, line 516: I guess the question is - if we enter debug mode or an interrupt handler with a load still outstanding, what happens if that load triggers an exception?

The issue in that previous case was similar to this - we want to enter debug/interrupt mode, but we've started a load/store so what do we do? The previous solution was to wait until it finishes, and don't execute any instructions, then enter debug mode / interrupt handler. This solution is to start entering debug mode / exception handler but stall the first instruction until the load completes. I think this solution should work, as long as the load/store exceptions do the right thing.

In the previous case, spurious additions loads we being issued to random addresses on the cycle that the load was in WB and the core state machine was in IRQ_TAKEN. I suspect that specific issue was fixed by the other changes to halt the IF stage and clear the ID stage.

GregAC: With the `lsq_req_in_id_i` signal we still entered the debug/interrupt handler before the load/store had finished as it could sit pending in the WB stage whilst the first instruction of the debug/interrupt handler was fetched.

Though looking at the current RTL (including this change) it appears we do end up waiting until the load/store has cleared the WB stage before actually letting the first instruction of the handler into ID/EX. Though I haven't yet unpicked exactly why I think we've got a stall we don't technically need.

I think to correctly deal with an exception on a load/store that is executed at the same time an interrupt/debug request appears avoiding the controller going into DBG_TAKEN_IF/IRQ_TAKEN until the WB stage is cleared may be the simplest way to go but I don't think we had that behaviour before this change or with this change.

Dealing with an interrupt and an exception on same instruction looks tricky, from the RISC-V privileged spec:

> Multiple simultaneous interrupts and traps at the same privilege level are handled in the following
decreasing priority order: external interrupts, software interrupts, timer interrupts, then finally any
synchronous traps.

So if we get an IRQ when a load/store is executing that sees an exception we need to jump to the IRQ handler. However on return from that IRQ handler we need to then take the exception and what do we set MEPC to for the IRQ handler? If it's the instruction after and we then take an exception for the instruction before that doesn't seem to make much sense. Though setting MEPC to the instruction that will cause the exception doesn't seem much better as it's effectively re-executing an already executed instruction so seems to break the definition of MEPC. Plus if we've got some sense of a 'pending exception' whilst an interrupt handler is running how do we deal with an exception that occurs during that interrupt handler that also sees another interrupt handler, we end up with an unlimited stack of pending exceptions.

We could instead say that if we get an IRQ whilst executing an instruction that takes an exception we go to the exception handler instead (which will initially clear the interrupt enable so the interrupt won't happen) so don't need the concept of a pending exception. This seems far simpler to me but doesn't seem to match the spec.

GregAC:
Sadly the OT 'chip_uart_tx_rx' still seems flakey in that it's still failing on the latest for me (though nightly regression has it pass). So haven't yet successfully used it to verify this change.

rtl/ibex_controller.sv
tomroberts-lowrisc: Hmmm true - I was assuming `stall` would be true once the load/store was in WB, but I guess it won't be since there's no `instr_valid_i`.

That is quite odd in the spec - it also says:
> Synchronous  exceptions  are  given  the  lowest  priority  to  minimize  worst-case  interrupt  latency

which has some strange implications for loads. If we take the interrupt, execute some instructions, then deal with the load exception, then it's no-longer a synchronous exception!

Maybe one way of looking at it is that once we have issued the request for the load, any synchronous exception related to that load is "before" any interrupt that might become pending, so it's ok to deal with that first?

udinator:
> Sadly the OT 'chip_uart_tx_rx' still seems flakey in that it's still failing on the latest for me (though nightly regression has it pass). So haven't yet successfully used it to verify this change.

@GregAC - I do not think the current failures in `chip_uart_tx_rx` are due to the same issue solved by #928 (currently we think it might be a software issue), can you try with the `riscv_debug_instr_test` and `riscv_interrupt_instr_test` ?

GregAC:
`riscv_debug_instr_test` and `riscv_interrupt_instr_test` both look fine (that's with 5 iterations of each, along with 5 iterations of all of the other tests that have 'debug' or 'interrupt' in their name). I was just concerned that the initial issue Tom had fixed had only shown up in the OT top-level and not the Ibex DV. So as I was half undoing his fix wanted to see if I could go back to the test that had shown the bug in the first place.

GregAC:
Due to the can of worms I've opened around the treatment of bus errors I think I'll hold off in dealing with the interrupts vs exceptions issue for now.

This does fix the issues in the riscv_debug_single_step test and I've done some testing to show it doesn't break other things. A bus error that is seen the same cycle as an interrupt may do something interesting but I think that's the case in the current Ibex RTL also.

So can we let this through and I'll open an issue to remind us about the exception vs interrupt thing?

vogelpi:
> Due to the can of worms I've opened around the treatment of bus errors I think I'll hold off in dealing with the interrupts vs exceptions issue for now.
> 
> This does fix the issues in the riscv_debug_single_step test and I've done some testing to show it doesn't break other things. A bus error that is seen the same cycle as an interrupt may do something interesting but I think that's the case in the current Ibex RTL also.
> 
> So can we let this through and I'll open an issue to remind us about the exception vs interrupt thing?

This sounds good to me.

However, I had a look at the RTL (with this change) and think I found an issue. To me it seems that now if the WB stage is enabled and we do a single step on a load/store, the following happens:
1. Load/store in ID, we output the request (assume it's granted in the same cycle). `stall_mem` and also `stall` is not set (because `lsu_req_done_i` is high). The controller sees `enter_debug_mode` and goes to `DBG_TAKEN_IF`.
2. While in `DBG_TAKEN_IF`, we get an exception in WB. But we don't see it. `load_err_i/store_err_i` is flopped but we just set everything up to enter debug mode (cause etc.). The controller goes back to `DECODE`.
3. Back in `DECODE` (now we are in debug mode), we "look" again at `special_req_all`. However, this is not set as it only factors in `load_err_i/store_err_i` which is no longer high. We also don't look at `wb_exception_o` (would be still be high as it factors in `load_err_q/store_err_q` as well). So we basically miss the exception as it's happening exactly in the one cycle we spend in `DBG_TAKEN_IF`.

I am not sure what the proper way to handle this is (irrespective of the outcome of the open discussion around the spec). Maybe we should check the `special_req_all` in `DBG_TAKEN_IF` and if it's set, got to `FLUSH` and then back to `DBG_TAKEN_IF` after `FLUSH` (this path from `FLUSH` to `DBG_TAKEN_IF`) is already there. WDYT?

GregAC:
I think when there's a possible exception we should just hold off going out of `DECODE` for an interrupt/debug request until the possible exception is resolved (potentially other than NMI which would basically kill the instruction with the pending request and any resulting response from the memory would get dropped).

I believe the RTL may already be doing this due to some overly conservative stalling from the writeback stage.

GregAC:
Have rebased to trigger a new CI run

PR #1035, opened 2020-07-15
[rtl] Add security hardened PC

tomroberts-lowrisc:
- Checks that PC increments as expected
- Raises an alert if not

### DIFF
--- a/doc/security.rst
+++ b/doc/security.rst
@@ -6,6 +6,13 @@ Security Features
 Ibex implements a set of extra features (when the SecureIbex parameter is set) to support security-critical applications.
 All features are runtime configurable via bits in the **cpuctrl** custom CSR.
 
+Outputs
+-------
+
+Ibex has two alert outputs for signalling security issues.
+The major alert (**alert_major_o**) indicates a critical security issue from which the core cannot recover.
+The minor alert (**alert_minor_o**) indicates potential security issues which can be monitored over time by a system.
+
 Data Independent Timing
 -----------------------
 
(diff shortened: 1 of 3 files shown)


doc/security.rst
```diff
@@ -6,6 +6,13 @@ Security Features
 Ibex implements a set of extra features (when the SecureIbex parameter is set) to support security-critical applications.
 All features are runtime configurable via bits in the **cpuctrl** custom CSR.
 
+Outputs
+-------
+
+Ibex has two alert outputs for signalling security issues.
+The majort alert (**alert_major_o**) indicates a critical security issue from which the core cannot recover.
```
rswarbrick, line 13: "majort" ? :-)

rtl/ibex_if_stage.sv
```diff
@@ -343,6 +345,31 @@ module ibex_if_stage #(
     end
   end
 
+  // Check for expected increments of the PC when security hardening enabled
+  if (SecureIbex) begin : g_secure_pc
+    logic [31:0] prev_instr_addr_incr;
+    logic        prev_instr_seq_q, prev_instr_seq_d;
+
+    // Do not check for sequential increase after a branch, or after reset
```
vogelpi, line 353: I would extend the comment and say that this also holds for jumps, interrupts, debug requests, exceptions etc. Together with branches these are all signaled to IF via `pc_set_i` which is then internally assigned to `branch_req`.

doc/security.rst
tomroberts-lowrisc: Will fix - thanks

rtl/ibex_if_stage.sv
tomroberts-lowrisc: Good plan will do - thanks

Issue #1018, opened 2020-07-07
Illegal Instruction Exception not Raised - Shift Instructions


shetalani:
**RISC-V Specification:**

- _"Chapter 2.4 of Unprivileged ISA"_

**Issue Description:**

An illegal instruction updates the register file and the associated csrs and executes normally without raising an illegal instruction exception.

**Example:**

The instruction 32'h02005913 is decoded at time point t##0 with no illegal instruction being flagged, as illegal_insn_did is de-asserted.

![Issue_0](https://user-images.githubusercontent.com/54367010/86763358-913b1200-c047-11ea-9bfd-afd69a9bbd53.PNG)

---

Product: OneSpin 360 DV-Verify
App: RVV
Tool's version: 2020.1.3

imphil:
Thanks @shetalani for your bug report.

Just to confirm we're speaking about the same issue. The instruction you're seeing would be decoded as:

```
02005913          srli          s2,zero,32
```

And the issue being that the immediate should only be 5 bit wide, and 32 isn't a valid immediate ("the shift amount is encoded in the lower 5 bits of the I-immediate field." as the spec says)

shetalani:
Yes, on the other hand, the spec states that the upper 7 bits should be zero for such an instruction to be decoded as SRLI.

vogelpi:
Hi @shetalani ,
I have created a PR to solve the issue [here](#1036).

vogelpi:
The actual problem was that the decoder did not check bits 26 and 25 for slli, srli and srai instructions.

imphil:
@shetalani can you re-run the tool to check if the issue is now fully resolved? The latest `master` version of Ibex should have the required fix.

shetalani:
The issue is fixed, cheers!

PR #1036, opened 2020-07-16
[rtl] Make sure decoder also checks bits 26 and 25 for slli, srli, srai

vogelpi:
Previously, these bits were not checked when decoding `slli`, `srli` and `srai`, causing some illegal instruction encodings not to trigger an illegal instructions exception.

This resolves lowRISC/Ibex#1018.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -348,7 +348,7 @@ module ibex_decoder #(
 
           3'b001: begin
             unique case (instr[31:27])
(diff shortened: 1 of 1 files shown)


vogelpi:
Regression is currently running, please do not yet merge.

vogelpi:
The regression test suite passes except for `riscv_illegal_instr_test` and `riscv_debug_branch_jump_test`. Both these tests also fail on master, I thus assume they are not related.

vogelpi:
Thanks @GregAC!

Issue #983, opened 2020-06-23
Debug single step test failure


tomroberts-lowrisc:
The debug single step test (small config, seed 1) is currently failing with the following message:
`DPC value [0x8000333e] is not the next instruction after ret_pc [0x8000b23a]`
As far as I can tell, the test is checking that DPC increments by 2 or 4 for every step taken. In this case, the instruction stepped over is a jump, and so the next PC does not equal the previous value - 2/4. The test is not expecting any jump/branch instructions to be generated, but there is one jump at the end (jump to test done) which it ends up stepping to, which causes the faliure.

udinator:
Hi @tomroberts-lowrisc , thanks for opening this issue. You are correct, the single step test is failing because of a new jump instruction right before the main program that got added in as a result of the modifications to program structure to support PMP.
I'm working on expanding this test to support branch and jump instructions, as that will help increase overall coverage.

udinator:
update: currently blocked on #987 - my updates to fix `debug_single_step_test` relies on the infrastructure introduced in this PR.

tomroberts-lowrisc:
I think this is now fixed - no longer seeing failures here so closing the issue

PR #1037, opened 2020-07-16
[dv/ibex] Enhance riscv_debug_single_step test

udinator:
As pointed out by @tomroberts-lowrisc in #983, the current
implementation of riscv_debug_single_step_test cannot handle
single-stepping over instructions that change the PC.
This PR aims to introduce this functionality, utilizing the
new instr_monitor_if.

Now, if the core single-steps onto a branch/jump instruction, the
testbench will log the new target PC and compare it against the actual
target address stored in `dpc`.
"Normal" instructions are checked as usual by incrementing the
instruction's PC by either 2 or 4 (depending whether it is compressed)
and comparing that against `dpc`.

This test passes successfully when combined with the RTL patch in #1034.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -483,7 +483,7 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
     bit [12:0]                                    system_imm;
     instr_t                                       instr_fields;
 
-    opcode      = ibex_pkg::opcode_e'(instr[6:0]);
+    opcode      = ibex_pkg::opcode_e`(instr[6:0]);
     funct3      = instr[14:12];
     funct7      = instr[31:25];
     system_imm  = instr[31:20];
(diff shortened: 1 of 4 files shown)


dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -483,7 +483,7 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
     bit [12:0]                                    system_imm;
     instr_t                                       instr_fields;
 
-    opcode      = ibex_pkg::opcode_e'(instr[6:0]);
```
udinator, line 486: Whoops, just realized that I inadvertently overrode the enum casting, I'll make sure to re-add this casting.

udinator:
Funny that you mention actually, because the Imperas team that developed OVPsim is working to integrate their ISS into the UVM environment for lockstep comparison during simulation time (like you mentioned), which would eliminate the need for all of the handshaking and overly complex checks.
Thanks for the review!

udinator:
this force push re-adds the enum casting to `opcode` and `step_instr[6:0]`, as per my comment in the PR.

Issue #1038, opened 2020-07-17
Sporadic failures in `riscv_illegal_instr_test`


GregAC:
@vogelpi and I have both seen some failures in the `riscv_illegal_instr_test`. I did look into some and I believe they're happening because when running with Spike as a simulator the 'F' extension and potential others are turned on. So illegal instruction for Ibex get executed on Spike and a trace mis-match results.

We should look into the failures in more detail and if my diagnosis is correct ensure Spike runs with the correct ISA.

vogelpi:
Thanks Greg for opening this issue. I have tracked down the bug. You can use the following command to reproduce the issue:
```
make TEST=riscv_illegal_instr_test ITERATIONS=1 SEED=29633 SIMULATOR=vcs ISS=spike
```
This should give the following regression log:
```
Test binary: out/seed-29633/instr_gen/asm_tests/riscv_illegal_instr_test.0.o
ibex uvm log : out/seed-29633/rtl_sim/riscv_illegal_instr_test.0/sim.log
riscv_illegal_instr_test : [PASSED]

ibex : out/seed-29633/rtl_sim/riscv_illegal_instr_test.0/trace_core_00000000.csv
spike : out/seed-29633/instr_gen/spike_sim/riscv_illegal_instr_test.0.csv
Mismatch[1]:
ibex[4305] : pc[8000491c] csrrc	x7,mscratch,x10: t2:0000001e 
spike[4305] : pc[ffffffff80001380] addi    a3, a3, -124: a3:8001efe8 
Mismatch[2]:
ibex[4306] : pc[80004920] xori	x10,x19,747: a0:000002e9 
spike[4306] : pc[ffffffff800013f0] csrr    a2, mepc: a2:80004918 
Mismatch[3]:
ibex[4307] : pc[80004924] and	x5,x5,x11: t0:00000000 
spike[4307] : pc[ffffffff800013f4] csrr    a2, mcause: a2:00000002 
Mismatch[4]:
ibex[4308] : pc[80004928] xori	x16,x5,274: a6:00000112 
spike[4308] : pc[ffffffff800013f8] c.li    t6, 3: t6:00000003 
Mismatch[5]:
ibex[4309] : pc[8000492c] csrrwi	x27,mscratch,24: s11:00000000 
spike[4309] : pc[ffffffff80001402] c.li    t6, 8: t6:00000008 
24 instructions left in trace spike
[FAILED]: 2942 matched, 16896 mismatch

[...]

PR #1040, opened 2020-07-20
[dv] Add custom CSRs to yaml description file

vogelpi:
This is related to lowRISC/ibex#1038. It does not yet do the trick and @udinator mentioned that some more work in riscv-dv is required for this. Thus, I am leaving this as a draft.


udinator:
Thanks for opening this PR! At a first glance, the new yaml entries lgtm, but as you said let's leave this as a draft until I get riscv-dv working with custom CSRs, and then I'll "formally" approve this when the CSR and illegal_instr tests are up and running.

vogelpi:
Thanks @udinator , I am merging this one then.

Issue #985, opened 2020-06-23
[DV] Remove all copies of OpenTitan files except when vendored in


imphil:
We are vendoring in Ibex into OT, and we need to avoid files being duplicated, or core files in Ibex overriding ones in OT.

General rules:
- If code from OT is needed, it *must be* vendored in.
- If code is not vendored in, it *may not* use the same `name` in the core file, and ideally also not the same file names.

One example we know of right now is the mem_model at https://github.com/lowRISC/ibex/tree/master/dv/uvm/core_ibex/common/mem_model; there might be others.

This issue is to double-check all core files, and to ensure that the rules outlined hold.

rswarbrick:
I did a super-simple search for stuff duplicated between the two with these commands:

```sh
find opentitan -type f | grep -v '\.git' | grep -v 'vendor/' | xargs -n1 basename | sort -u >ot.txt
find ibex -type f | grep -v '\.git' | grep -v 'vendor/' | xargs -n1 basename | sort -u >ibex.txt
comm -12 ot.txt ibex.txt >both.txt
```

Obviously, this gives false positives (stuff like the azure pipeline), but it's a reasonable place to start. Grab the paths to these files, discarding Makefiles and readmes, in Ibex with:

```sh
xargs -n1 <both.txt find ibex -name | sort | grep -v vendor/ | grep -v Makefile | grep -v README.md >ibex-paths.txt
```

The resulting list ([ibex-paths.txt](https://github.com/lowRISC/ibex/files/4869366/ibex-paths.txt)) has 39 entries. I think the following items need looking at (in no particular order):

- `check_tool_requirements.core`, `util/check_tool_requirements.py`: We could vendor this in, I think.
- `dv/uvm/core_ibex/common/mem_model/*`: We could vendor this in.
- `dv/uvm/core_ibex/common/utils/clk_if.sv`, `dv/uvm/core_ibex/common/utils/dv_macros.svh`, `dv/uvm/core_ibex/common/utils/dv_utils_pkg.sv`: Manually copied to `dv/uvm/core_ibex` as a temporary hack. There's another copy properly vendored in. Long-term, the `core_ibex` tests should probably be ported to use fusesoc properly. Shorter term, we could probably just point them at the vendored code instead of the manually copied code.
[...]

PR #1042, opened 2020-07-21
Define an Ibex-specific top_pkg core

rswarbrick:
The idea is that this can supply top_pkg.sv, a top-level thing in
OpenTitan, for DV code we vendor from there. It's probably better to
do this than to directly vendor in OpenTitan's top_pkg, because the
latter has information about e.g. flash memory layout, which doesn't
really have any meaning for Ibex.

Why do this? Partly because it's a dependency of getting rid of the copied code in `dv/uvm/core_ibex/common/utils` (see issue #985 for details): it turns out that the correctly-vendored versions of this code depend on a top_pkg core file that we incorrectly copied into `dv/uvm/icache` (yes, really). This patch straightens everything out.


rswarbrick:
Gah! I accidentally merged this patch as part of #1043.

@imphil: Does this look like a reasonable way to do things? If not, I'll happily rejig to do it differently (and push up a PR with the fix).

imphil:
I'm not convinced that this fix goes in the right direction, but also didn't have time to wrap my head around a proper fix. 

The problem with top_pkg.sv in Ibex is that I'm not sure why we actually need these constants at all; so there's most likely a bit of refactoring of the DV code to be done not to rely on bus constants for a bus which isn't used. 

For the OpenTitan use case, there are issues around more than one toplevel and these global constants, which are tracked in https://github.com/lowRISC/opentitan/issues/2544.

All of it doesn't have a straightforward solution, so I'm fine with this PR for now, as it seems to solve a problem we're having. We'll fall over this one again as soon as vendored in DV code adds or uses other parameters in top_pkg; hopefully by that time I've finished thinking :smile:

PR #1043, opened 2020-07-21
Remove duplicated files from dv/uvm/core_ibex/common/utils

rswarbrick:
This PR currently has 2 patches, but the first is a copy of PR #1042 on which it depends. We should merge that first, and a rebase will make this a single commit. Below is the contents of that second commit message:

> We now have a clock/reset interface and the dv_utils stuff vendored
from OpenTitan so can delete the duplicates and point the file list at
the vendored files.
>
> The only difficulty is that the clock interfaces are slightly
different, so there are a couple of minor changes to the core_ibex
test lib, renaming "clk_if" to "clk_rst_if" and changing how we apply
resets.
>
> Note that the testbench (core_ibex_tb_top.sv) starts the clock and
resets the DUT at the start of time. This is different from how other
OpenTitan VIP does things (where the reset happens in the sequence),
but this is the smallest change I could make from how it worked
before (where the reset happened in the clock interface itself).

This work is part of the tidy-up that is tracked in issue #985 (item 1 on the TODO list there)


rswarbrick:
Yep, they work (they didn't at first, because I wasn't starting the clock: it took me a good hour to work out what on earth I'd missed!)

Issue #1044, opened 2020-07-22
PMP TOR writes and behavior


sorear:
1. There appears to be a functional difference between [spike](https://github.com/riscv/riscv-isa-sim/blob/67b7edd0270d5d074136eb45474477d15aeee725/riscv/processor.cc#L809) and [ibex](https://github.com/lowRISC/ibex/blob/d7c13aa9045838e7b9751e87b6b4630e7893db39/rtl/ibex_cs_registers.sv#L828) in the interaction between L and A=TOR; riscv/riscv-isa-manual#545 endorsed the spike interpretation.

2. Should https://github.com/lowRISC/ibex/blob/master/rtl/ibex_pmp.sv#L92-L93 be generating both ends of the TOR range from the same pmpaddr?

tomroberts-lowrisc:
1. Do we differ from spike here? Locking of the address registers is implemented [here](https://github.com/lowRISC/ibex/blob/d7c13aa9045838e7b9751e87b6b4630e7893db39/rtl/ibex_cs_registers.sv#L827). I think we verified the Ibex implementation against spike - @udinator will be able to confirm

2. The start address of the TOR range comes from `region_start_address` which will be the previous PMP region address - see [here](https://github.com/lowRISC/ibex/blob/d7c13aa9045838e7b9751e87b6b4630e7893db39/rtl/ibex_pmp.sv#L53)

GregAC:
On 1. I think we'll disallow writes to `pmp_addr[i]` if entry `i+1` has A=TOR regardless of the lock bit on entry `i+1` where the spec update (which matches current Spike behavior) indicates we should allow those unless entry `i+1` has both A=TOR and the lock bit set?

tomroberts-lowrisc:
Yep - just re-read it myself - will fix to match the new spec for 1.

rswarbrick:
Does this just show that we haven't tried these tests against a recent Spike, or is it a proper test hole?

GregAC:
We should also check what the OVPsim behavior is here. Given our behavior was spec compliant until https://github.com/riscv/riscv-isa-manual/pull/545 went in 7 hours ago OVPsim may also have similar behavior and not match Spike.

udinator:
@tomroberts-lowrisc - yes, we verified the Ibex implementation against both Spike and OVPsim, not taking into account this recent spec change.
[...]

PR #1054, opened 2020-07-29
[ibex/rtl] Fix pmpaddr write enable signal

udinator:
Updates the write enable for `pmpaddr[i]` CSRs to deny writes if
`pmpcfg[i+1].lock == 1` and `pmpcfg[i+1].mode == TOR`, as per the
spec.
Relates to #1044.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -825,7 +825,7 @@ module ibex_cs_registers #(
       // --------------------------
       if (i < PMPNumRegions - 1) begin : g_lower
         assign pmp_addr_we[i] = csr_we_int & ~pmp_cfg[i].lock &
-                                (pmp_cfg[i+1].mode != PMP_MODE_TOR) &
+                                (~pmp_cfg[i+1].lock | (pmp_cfg[i+1].mode != PMP_MODE_TOR)) &
                                 (csr_addr == (CSR_OFF_PMP_ADDR + i[11:0]));
       end else begin : g_upper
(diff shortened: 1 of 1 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -825,7 +825,7 @@ module ibex_cs_registers #(
       // --------------------------
       if (i < PMPNumRegions - 1) begin : g_lower
         assign pmp_addr_we[i] = csr_we_int & ~pmp_cfg[i].lock &
-                                (pmp_cfg[i+1].mode != PMP_MODE_TOR) &
+                                (!pmp_cfg[i+1].lock || pmp_cfg[i+1].mode != PMP_MODE_TOR) &
```
tomroberts-lowrisc, line 828: These should be bitwise operations here - plus I generally prefer to add the extra parentheses around the equality (as in the previous line)

rswarbrick: To be fair to Udi, `a != b` is a logical test (we don't use `~|(a^b)` !) and this patch replaces `(a != b)` with `(!x || a != b)` which is at least consistent. Are you suggesting `(~x | (a != b))`?

tomroberts-lowrisc: yep I'm suggesting `(~x | (a != b))` - it's at least a bit more consistent with the two previous lines then

udinator: ah yep you're absolutely right it should be a bitwise operator - I'll modify that as well as the formatting change you also suggested.
Thanks for the comment!

PR #1056, opened 2020-07-30
Pulpissimo interrupt fix

heavySea:
This is a modification of the additional "CLINTx" Interrupt Interface for the Pulpissimo µC.
The Interface is intended to work with the platform-level Interrupt Controller. This controller is managed via an APB-Interface and brings its own Interrupt Enable Mask. Therefore the miex register is redundant. Because the mask register is also never written by the PULP SDK the core does not react to an interrupt.

This modification goes hand in hand with some other modifications in the [pulp-rt](https://github.com/pulp-platform/pulp-rt/pull/36) and [pulp-configs](https://github.com/pulp-platform/pulp-configs/pull/12) Repositories.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -726,7 +726,7 @@ module ibex_cs_registers #(
       mstack_epc_q   <= '0;
       mstack_cause_q <= '0;
 
-      miex_q         <= '0;
+      miex_q         <= {32{1'b1}};
       mtvecx_q       <= 32'b01;
 
     end else begin


vogelpi:
Hi @heavySea ,
thanks for your PR. Question: Where are you located geographically?
Best,
Pirmin

heavySea:
Hi @vogelpi ,
I'm in Germany. Why?
Kind Regards,
Max

vogelpi:
I just saw that you did a couple of PRs to various PULP repos recently and thought you might be even in-house ;-)

heavySea:
No (;
But I work on modifying pulpissimo for my master thesis. I like to give back the changes I think everybody can find useful. :)

vogelpi:
That's very nice of you, thanks :+1:

vogelpi:
I had a brief look at the code and commit messages. If I understand it correctly, this PR is mainly about removing the `miex` register. And the reason for doing this that PULP SDK never sets this register to all ones (because it uses a similar register in the external PLIC). Wouldn't it be cleaner to let the PULP runtime just write the whole register to all-one once in the beginning? Or even change the reset value of the register to all ones?

Reasoning: When adding this register, we agreed in the PULP team that `mipx`, `miex` etc. should simply be an extension to the regular spec-compliant `mip`, `mie` but with all 32 interrupts being available to the platform. As such it should behave exactly the same way.

heavySea:
Yes this would practically have the same effect and would be closer to the spec. From my point of view the register was obsolete. But I can change it this way.

heavySea:
Changed it and tested it.

rtl/ibex_pkg.sv
```diff
@@ -230,7 +230,44 @@ typedef enum logic [5:0] {
...
+  EXC_CAUSE_IRQ_X_23           = {1'b1, 5'd23},
+  EXC_CAUSE_IRQ_X_24           = {1'b1, 5'd24},
+  EXC_CAUSE_IRQ_X_25           = {1'b1, 5'd25},
+  EXC_CAUSE_IRQ_X_26           = {1'b1, 5'd26},
+  EXC_CAUSE_IRQ_X_27           = {1'b1, 5'd27},
+  EXC_CAUSE_IRQ_X_28           = {1'b1, 5'd28},
+  EXC_CAUSE_IRQ_X_29           = {1'b1, 5'd29},
+  EXC_CAUSE_IRQ_X_30           = {1'b1, 5'd30}
+  //EXC_CAUSE_IRQ_X_31           = {1'b1, 5'd031} // decoded as EXC_CAUSE_IRQ_NM
+
```
vogelpi, line 270: This does not seem to be needed as it is not referenced anyhwere. Can you please undo these changes?

rtl/ibex_cs_registers.sv
```diff
@@ -726,7 +726,7 @@ module ibex_cs_registers #(
       mstack_epc_q   <= '0;
       mstack_cause_q <= '0;
 
-      miex_q         <= '0;
+      miex_q         <= '1;
```
vogelpi, line 729: Please change this to `{32{1'b1}}`. In the past I have seen some tools turning `'1` into `0...0001`.

rtl/ibex_pkg.sv
heavySea: Removed.

rtl/ibex_cs_registers.sv
heavySea: Done

heavySea:
Thanks aswell. The {32{1'b1}} fixed a problem on my fpga synthesis :+1:

PR #1061, opened 2020-08-03
Set `FPGA_XILINX` define whenever Vivado is used

vogelpi:
This PR makes sure that whenever Vivado is used, the Verilog define `FPGA_XILINX` is set. This define can be used to enable Xilinx specific pragmas in the RTL code. It is currently used to implement the performance counters with DSP slices (incl. integrated output registers for counter widths <= 32) which helps to save logic resources and FFs (e.g. -7% and -15% when using 10 32-bit counters).

Previously, we had to set this parameter in every single top-level .core file using Ibex.

### DIFF
--- a/examples/fpga/artya7/top_artya7.core
+++ b/examples/fpga/artya7/top_artya7.core
@@ -31,12 +31,6 @@ parameters:
     default: "../../../../../examples/sw/led/led.vmem"
     paramtype: vlogparam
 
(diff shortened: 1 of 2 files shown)


vogelpi:
Thanks @rswarbrick . Yes indeed, the only place we are using `FPGA_XILINX` now is inside `ibex_core.core` and `rtl/ibex_counter.sv`.

PR #1063, opened 2020-08-05
[ibex/dv] Update Ibex PMP tests

udinator:
This PR adds a new riscv-dv option `enable_write_pmp_csr=1` to each of the
tests in the existing PMP test suite.

This will enable the core to execute a directed test sequence to write
random values to each `pmpaddr` and `pmpcfg` CSR in order to explicitly test write
accessibility and spec compliance.

The original values stored in these CSRs are restored afterwards.

Note: This patch has exposed what looks like a bug in OVPsim,
for the time being please run all PMP tests with Spike.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -640,6 +640,7 @@
   gen_opts: >
     +instr_cnt=6000
     +pmp_max_offset=00024000
+    +enable_write_pmp_csr=1
   rtl_test: core_ibex_base_test
 
 - test: riscv_pmp_disable_all_regions_test
@@ -668,6 +669,7 @@
     +pmp_region_13=X:0,W:0,R:0
     +pmp_region_14=X:0,W:0,R:0
     +pmp_region_15=X:0,W:0,R:0
+    +enable_write_pmp_csr=1
   rtl_test: core_ibex_base_test
 
 - test: riscv_pmp_out_of_bounds_test
@@ -680,6 +682,7 @@
   gen_opts: >
     +instr_cnt=6000
     +pmp_max_offset=00024000
+    +enable_write_pmp_csr=1
     +directed_instr_0=riscv_load_store_rand_addr_instr_stream,50
   rtl_test: core_ibex_base_test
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
Would it make sense to either update the documentation to warn against OVPSim or (more drastically) comment out support for that ISS from the Makefile while things are still broken? I'm worried we'll get bug reports of "your PMP tests are broken with OVPSim".

udinator:
There has been a long-standing issue with Spike involving timer interrupts which is why I switched to OVPsim as the primary ISS for Ibex testing - so this is actually creating a bigger issue right now of not being able to run the full Ibex test suite with just one ISS.
I need to finish pinpointing exactly what is going wrong with OVPsim to file a bug and have this issue resolved; I will have a report filed later today.
If it takes longer than anticipated, I'll be allowing individual tests to specify an ISS, that way full regressions will not be broken.

rswarbrick:
> There has been a long-standing issue with Spike involving timer interrupts which is why I switched to OVPsim as the primary ISS for Ibex testing

Do we have an issue tracking the fact that Spike doesn't work for timer interrupts? Since riscv-isa-sim is rather more open-source than OVPSim, it might even be easier to get fixed than a problem in OVPSim.

udinator:
Currently no we do not - the switch to OVPsim was also made due to the relatively higher ease of configurability of the simulator, but now is probably a good time to do so! 
I'll reproduce the Spike issue and file a bug later today also.

rswarbrick:
> I'll reproduce the Spike issue and file a bug later today also.

Brill, thanks! Assuming you're in California, I hope you get some sleep first :-)

Imperas:
@udinator 

> Note: This patch has exposed what looks like a bug in OVPsim,

Hi, we are unaware of any issues in this area, could you provide some detailed information regarding the error you are seeing. Additionally if you can tell us how to reproduce that would be very helpful
Thx
Lee

udinator:
@Imperas yes definitely! I will fe an issue as necessary on Imperas GitHub page a little later today.

Imperas:
Just to follow up on this, I am pretty sure this is due to a misconfiguration of riscvOVPsim described here 
https://github.com/riscv/riscv-ovpsim/issues/15

@udinator if you can confirm our thoughts, that would be great
Thx
Lee

udinator:
@Imperas as I mentioned in riscv/riscv-ovpsim#15, changing the physical address size to 34 bits has fixed the issue, thank you for the help!

PR #1064, opened 2020-08-06
[ibex] Update CSR description of cpuctrl/secureseed

udinator:
This patch does a few things:
- Fixes some yaml syntax errors by deleting the appropriate `-` characters to align with format of all other entries
- Temporarily changes type of all `cpuctrl` fields to read-only, as we do not enable the SecureIbex parameter in Ibex regressions yet to turn on all security features (and add a TODO item to do so)
- Changes type of `secureseed` to read-only, as according to the [Ibex CSR spec](https://ibex-core.readthedocs.io/en/latest/cs_registers.html#security-feature-seed-register-secureseed), "Seed values are not actually stored in a register and so reads to this register will always return zero"

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
@@ -297,6 +297,8 @@
       lsb: 0
 
 # CPUCTRL
+#
+# TODO : change fields to WARL type once SecureIbex parameter is enabled
 - csr: cpuctrl
   description: >
     CPU control register (custom)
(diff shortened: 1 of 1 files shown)


udinator:
Hi @vogelpi - yes, that definitely does make sense.
The issue right now is that the CSR description model doesn't have a way to capture the type of r/w behavior that these custom CSRs have,so all the CSR tests are failing, shall we just comment out `cpuctrl/secureseed` entries for the time being then?

vogelpi:
Ah I see, we can't model the WARL behavior.

I think the solution depends on what the behavior of the IS generator and simulator. If the CSRs are changed to read-only:
1. Will the IS generator just generate read accesses? If this is the case, I think the best is to just change them to read-only as you initially suggested. 
2. Will the IS generator also generate write accesses? If this is the case, the question is how the IS simulator handles these. If it does expect Ibex to generate an illegal instruction exception, then changing to read-only does not help and we should comment out those entries.

But if we comment them out, does it mean the IS simulator expects to see an illegal instruction exception on every access? If yes, we still don't win anything. In this case we should change the RTL to cause an illegal instruction exception upon accessing these CSRs if the security features are disabled.

udinator:
I think a little context might be needed for the CSR test as it's different from the rest of the randomly generated assembly tests.
We use the standalone Python script [gen_csr_test.py](https://github.com/google/riscv-dv/blob/master/scripts/gen_csr_test.py) to create the CSR test from the YAML description - the YAML description is not used by the riscv-dv generator otherwise.
The CSR test generated by this script will essentially just generate a bunch of reads/writes to each CSR in the YAML description and will self-check that the correct data has been read and written to the CSR (taking advantage of the fact that CSR instructions are a combination of atomic read and atomic write). As a result, we do not use an IS simulator (Spike/OVPsim) with this test, so no need to worry about any unexpected illegal instruction exceptions.

In the scenario here, setting the `cpuctrl/secureseed` fields to read-only will mean the self-checking mechanism will expect the values stored in these registers to not deviate from reset value of 0x0, sort of like a write-ignored-read-legal field type - this means that while both reads and writes will be generated to these CSRs, the test will only ever expect that 0 be read back from the CSRs.
This is a little inaccurate, as `cpuctrl` can be written with anything in M-mode, but will only return 0 when read (if SecureIbex is disabled), so it's definitely a bit of a hack, but is the closest way to model the current behavior of these CSRs.

tomroberts-lowrisc:
That sounds sensible for now. We can also change the behavior of these registers if you think there's a better arrangement of WARL/trapping that's more consistent.

vogelpi:
Thanks @udinator for the clarification! I wasn't aware of the distinction between those CSR tests and what RISCV-DV does. But it makes all much more sense to me. I agree with your suggestion.

PR #1065, opened 2020-08-07
[syn] Use latch-based register file in yosys

tomroberts-lowrisc:
- Add a technology map for latches (only works with nandgate45 library
  at the moment)
- Add a real latch-based clock gating cell
- Update timing path reporting to differentiate between register and
  latch paths
- Update summary results in README to reflect the latch-based numbers,
  plus add numbers for a micro-riscy-style (RV32EC) config


tomroberts-lowrisc:
Note that the timing path groups changed as part of this PR to group all I/Os into in2reg, reg2out and in2out groups as I thought that was a bit simpler. Can change it back if anyone strongly prefers the original style.

tomroberts-lowrisc:
Good plan - I'll increase the number to 1k

PR #1066, opened 2020-08-10
[ibex/dv] Update OVPsim to use 34-bit address range

udinator:
This patch updates the OVPsim config file to use a 34-bit address space, as per discussion in https://github.com/riscv/riscv-ovpsim/issues/15.
Needed to resolve bugs in PMP tests that are popping up due to OVPsim currently only being configured for a 32-bit address space (PMP uses a 34-bit space).

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
+++ b/dv/uvm/core_ibex/riscv_dv_extension/riscvOVPsim.ic
@@ -21,3 +21,4 @@
 --override riscvOVPsim/cpu/defaultsemihost=F
 --override riscvOVPsim/cpu/wfi_is_nop=T
 --override riscvOVPsim/cpu/tval_ii_code=T
+--addressbits 34


udinator:
> Looks sensible to me.
> 
> If I've understood the discussion at the issue, the extra 2 bits will mean that lots of randomly chosen ranges won't do very much (because the lower bound for the PMP range lies within the address space with probability 1/2^2 = 25%). Do we do anything to weight the chosen ranges to make them more likely to have an effect?

Thanks for the review!

That's a good question.
What we do in the Ibex DV environment is always set 'pmpaddr0' to the address of the '<main>' section of assembly and then provide an upper bound to prevent situations like what you describe from occurring.

Technically speaking, in programs generated by riscv-dv that randomly impose PMP access restrictions, the bits [33:32] aren't used due to the 32-bit range of the program itself but Ibex can still write to them without restrictions - so we want the same behavior in the ISS so CSR accessibility tests don't fail.

rswarbrick:
Ah, I see. So we do occasionally blat rubbish into the upper bits, but only do that when we don't care about their effect (except that they are restored correctly on the next CSR read). Sounds good to me.

udinator:
Yep, exactly!

PR #1075, opened 2020-08-17
[ibex/dv] Add dependency on bus_params_pkg

udinator:
This patch updates the Ibex environment's dependency on the old top_pkg
to now depend on bus_params_pkg.


udinator:
Thanks for the quick review @imphil !

@rswarbrick I'm merging this PR now to prevent the Ibex environment from breaking.

rswarbrick:
Oops, thanks Udi. I'd forgotten that the core_ibex stuff used this too.

Issue #89, opened 2019-06-24
[tracking] Implement  bit manipulation (B) RISC-V ISA extension


imphil:
The bit manipulation (bitmanip) extension is currently being specified. Bit manipulation is a common task in many scenarios where Ibex could be used.

The bitmanip extension adds many new instructions to Ibex, not all of them have a straightforward implementation. Please have a look at the implementation plan and the implementation tracking sheet for an overview 

* [Draft standard](https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-draft.pdf)
* [Implementation plan](https://docs.google.com/document/d/1HS6tLeCac1MJPdrhnzD1k7PBBLxwb66W2R26moWZmok/edit)
* [Implementation tracking sheet](https://docs.google.com/spreadsheets/d/1XzAnCXw4GVD8dAEqZTnDEFYrvZ7v0_8zb27CVaWIfvQ/edit)

**All implementations of individual instructions are tracked in separate issues.**

esshiu:
Hi, Philipp,

I'm a new Github user. I'd like to start a project. Can I try to take on this RTL task?

-Eric

imphil:
Hi Eric,

thanks for your interest in working on this project! This is a reasonably large project; before we can move on to implementation, we need to do a bit of planning and break it down into smaller actionable chunks. Would you be interested in driving that? (If not that's fine as well, just let us know.)

A potentially good way of moving forward would be a shared Google Doc (share the link in this issue) with an implementation plan, describing 

- which (subsets of) instructions would be good to implement in an Ibex-class microcontroller
- what latency you'd expect for these instructions (single or multiple cycles?)
- a rough first idea of an implementation architecture, e.g. in a block diagram (https://ibex-core.readthedocs.io/en/latest/pipeline_details.html could be a starting point)
- a rough outline of milestones (which parts do you plan to tackle first?)
- your plans for testing during development, and help you need there from the people working on DV

None of that needs to be polished, feel free to share an incomplete draft and ask for help. You can also join our chat channel at https://lowrisc.zulipchat.com/#narrow/stream/198227-ibex (most people are in European time zones).

taoliug:
Add one note on the DV side. Right now riscv-dv doesn't support B-extension. As far as I know, the spec is not ratified yet, so it's subject to changes. Like Philipp mentioned, we need to define the scope we want to support. In the meantime, make sure the toolchain(instruction generator, compiler, ISS) is ready.

[...]

PR #1076, opened 2020-08-19
Parameter cleanup

vogelpi:
This PR contains two commits:

1. A commit for cleaning up the documentation and usage of the `RV32B` parameter for the bitmanip extension. In some places we were still expecting a bit parameter (instead of enum), and in some places it was still marked as "experimental".

1. A commit to replace the previous bit parameter `RV32M` and the string parameter `MultiplierImplementation` by a single enum parameter `RV32M` that can have the values: `RV32MNone`, `RV32MSlow`, `RV32MFast` and `RV32SingleCycle`.

I am currently running a full regression with RISCV-DV.


rtl/ibex_pkg.sv
```diff
@@ -8,9 +8,16 @@
...
-/////////////////////////
+/////////////////////
+// Parameter Enums //
+/////////////////////
+
+typedef enum integer {
+  RV32MNone,
+  RV32MSlow,
+  RV32MFast,
+  RV32MSingleCycle
```
imphil, line 19: Can you explicitly assign values to these enums, as we need to use them by that sometimes?

syn/tcl/lr_synth_flow_var_setup.tcl
```diff
@@ -15,8 +15,8 @@ set_flow_bool_var flatten 1 "flatten"
 set_flow_bool_var timing_run 0 "timing run"
 set_flow_bool_var ibex_branch_target_alu 0 "Enable branch target ALU in Ibex"
 set_flow_bool_var ibex_writeback_stage 0 "Enable writeback stage in Ibex"
-set_flow_var ibex_bitmanip 0 "Bitmanip extenion setting for Ibex (0,1,2 - enums not supported)"
-set_flow_var ibex_multiplier "fast" "Multiplier implementation for Ibex (slow/fast/single-cycle)"
+set_flow_var ibex_bitmanip 0 "Bitmanip extenion setting for Ibex (0,1,2 for None, Balanced, Full - enums not supported)"
+set_flow_var ibex_multiplier 2 "Multiplier extension setting for Ibex (0,1,2,3 for None/Slow/Fast/SingleCycle - enums not supported)"
```
imphil, line 19:
```suggestion
set_flow_var ibex_multiplier 2 "Multiplier extension setting for Ibex (see ibex_pkg::rv32m_e for permitted values. Enum names are not supported in Yosys.)"
```

doc/instruction_decode_execute.rst
```diff
@@ -67,10 +67,13 @@ Other blocks use the ALU for the following tasks:
 Bit Manipulation Extension
   Support for the `RISC-V Bit Manipulation Extension (draft version 0.92 from November 8, 2019) <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.92.pdf>`_ is optional. [#B_draft]_
   It can be enabled via the enumerated parameter ``RV32B`` defined in :file:`rtl/ibex_pkg.sv`.
+  By default, this parameter is set to "RV32BNone" to disable the bit manipulation extension.
```
imphil, line 70:
```suggestion
  By default, this parameter is set to "ibex_pkg::RV32BNone" to disable the bit manipulation extension.
```

(same below)

doc/integration.rst
```diff
@@ -93,10 +93,9 @@ Parameters
...
 | ``RV32M``                    | bit               | 1          | M(ultiply) extension enable                                     |
 +------------------------------+-------------------+------------+-----------------------------------------------------------------+
-| ``RV32B``                    | ibex_pkg::rv32b_e | RV32BNone  | *EXPERIMENTAL* - B(itmanipulation) extension select:            |
+| ``RV32B``                    | ibex_pkg::rv32b_e | RV32BNone  | B(itmanipulation) extension select:                             |
 |                              |                   |            | "RV32BNone": No B-extension                                     |
-|                              |                   |            | "RV32BBalanced": Sub-extensions Zbb, Zbs, Zbf and               |
-|                              |                   |            | Zbt                                                             |
+|                              |                   |            | "RV32BBalanced": Sub-extensions Zbb, Zbs, Zbf and Zbt           |
 |                              |                   |            | "RV32Full": All sub-extensions                                  |
```
imphil, line 102: Please prefix the enum names with `ibex_pkg` as well.

dv/riscv_compliance/ibex_riscv_compliance.core
```diff
@@ -40,7 +40,7 @@ parameters:
     datatype: str
     default: ibex_pkg::RV32BNone
     paramtype: vlogdefine
-    description: "Bitmanip implementation parameter enum. See ibex_pkg.sv (EXPERIMENTAL)"
+    description: "Bitmanip implementation parameter enum. See ibex_pkg.sv"
```
imphil, line 43:
```suggestion
    description: "Bitmanip implementation parameter enum. See the ibex_pkg::rv32_e enum in ibex_pkg.sv for permitted values."
```

syn/tcl/lr_synth_flow_var_setup.tcl
```diff
@@ -15,7 +15,7 @@ set_flow_bool_var flatten 1 "flatten"
 set_flow_bool_var timing_run 0 "timing run"
 set_flow_bool_var ibex_branch_target_alu 0 "Enable branch target ALU in Ibex"
 set_flow_bool_var ibex_writeback_stage 0 "Enable writeback stage in Ibex"
-set_flow_var ibex_bitmanip 0 "Bitmanip extenion setting for Ibex (0,1,2 - enums not supported)"
+set_flow_var ibex_bitmanip 0 "Bitmanip extenion setting for Ibex (0,1,2 for None, Balanced, Full - enums not supported)"
```
imphil, line 18:
```suggestion
set_flow_var ibex_bitmanip 0 "Bitmanip extension setting for Ibex (see ibex_pkg::rv32b_e for permitted values)"
```

doc/instruction_decode_execute.rst
```diff
@@ -116,18 +116,18 @@ The Multiplier/Divider (MULT/DIV) is a state machine driven block to perform mul
 The fast and slow versions differ in multiplier only. All versions implement the same form of long division algorithm. The ALU block is used by the long division algorithm in all versions.
 
 Multiplier
-  The multiplier can be implemented in three variants controlled via the parameter ``MultiplierImplementation``.
+  The multiplier can be implemented in three variants controlled via the enumerated parameter ``RV32M`` defined in :file:`rtl/ibex_pkg.sv`.
```
imphil, line 119:
```suggestion
  The multiplier can be implemented in three variants controlled via the parameter ``RV32M`` of the enum type `ibex_pkg::rv32m_e`. See  :file:`rtl/ibex_pkg.sv` for permitted values.
```

doc/instruction_decode_execute.rst
```diff
@@ -116,18 +116,18 @@ The Multiplier/Divider (MULT/DIV) is a state machine driven block to perform mul
...
 
   Single-Cycle Multiplier
-    This implementation is chosen by setting the ``MultiplierImplementation`` parameter to "single-cycle". The single-cycle multiplier makes use of three parallel multiplier units, designed to be mapped to hardware multiplier primitives on FPGAs. It is therefore the **first choice for FPGA synthesis**.
+    This implementation is chosen by setting the ``RV32M`` parameter to "RV32MSingleCycle". The single-cycle multiplier makes use of three parallel multiplier units, designed to be mapped to hardware multiplier primitives on FPGAs. It is therefore the **first choice for FPGA synthesis**.
```
imphil, line 122: Please use the fully-qualified enum name.

While you're at it: can you make it one sentence per line?

imphil (approved):
LGTM with some smaller comments.

Just to confirm I'm still remembering things right: we need to take the detour with the define setting the parameter because fusesoc always adds quotes around strings when passing them to the tools, which we shouldn't do for names of enum constants?

rtl/ibex_pkg.sv
vogelpi: Good idea! Done.

syn/tcl/lr_synth_flow_var_setup.tcl
vogelpi: Done.

doc/instruction_decode_execute.rst
vogelpi: Done, also everywhere else in the file.

doc/integration.rst
vogelpi: Done.

dv/riscv_compliance/ibex_riscv_compliance.core
vogelpi: Done.

syn/tcl/lr_synth_flow_var_setup.tcl
vogelpi: Done.

doc/instruction_decode_execute.rst
vogelpi: Done.

doc/instruction_decode_execute.rst
vogelpi: Done. Good catch about one sentence per line.

vogelpi:
Thanks @imphil for your review! 

> Just to confirm I'm still remembering things right: we need to take the detour with the define setting the parameter because fusesoc always adds quotes around strings when passing them to the tools, which we shouldn't do for names of enum constants?

I thought it was just a problem with certain tools like VCS but I don't exactly remember why we did the work around with the define.

vogelpi:
Update: I found the original discussion that lead us to add the defines: https://github.com/lowRISC/ibex/issues/89#issuecomment-641416561

Seems they were originally required for passing strings to VCS. I shall have a look if we can remove them for the enums.

vogelpi:
Alright, what's actually happening: For both the SystemVerilog enum parameters as well as the previous SystemVerilog string parameter `MultiplierImplementation` we are declaring the datatype in FuseSoC as string or `str`. We then use FuseSoC command line arguments to override the default values in the top level of the build. Overwriting these FuseSoC string parameters from the command line is what seems not to be supported by VCS. That's why the detour via `define` is needed.

This also means, the define is only needed in top levels that we build in VCS such as the Ibex simple system and the UVM testbench. Other files like `ibex_core.sv` and `ibex_core_tracing.sv` are not affected as they are never used as top level. I will remove the defines from these files and document this issue.

vogelpi:
FYI, I keep getting one failure in the regression suite (`riscv_debug_branch_jump_test`) but that one is also failing on `master`.

vogelpi:
I am merging this now.

imphil:
Thanks Pirmin, really nice clean-up (and annoying to do, as the diff shows).

PR #1077, opened 2020-08-19
Use a RV32IMCB toolchain in CI

imphil:
This update also switches from GCC 9.2 to GCC 10 experimental, see
https://github.com/lowRISC/lowrisc-toolchains/releases/tag/20200626-1
for more information about the toolchain builds.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,8 @@
 
 variables:
   VERILATOR_VERSION: 4.032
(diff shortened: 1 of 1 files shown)


imphil:
We will need this toolchain version when running the UVM DV exercising the B extension in CI; this PR is a preparation step for that.

vogelpi (approved):
LGTM!

PR #1082, opened 2020-08-20
Fix PRJ_DIR in DV Makefile

imphil:
PRJ_DIR is presumably meant to point to the root of the git repository;
instead, it was pointing one directory higher up in the hierarchy. This
worked as long as the repository was checked out into a directory called
"ibex" -- but not otherwise, as in CI.


imphil:
@udinator I might have misinterpreted something here, does that patch break anything on your side?

PR #1084, opened 2020-08-20
Add UVM regression runs in private CI

imphil:
Add necessary configuration to be able to run RISC-V DV-based UVM regressions in CI. Due to licensing concerns these runs won't be publicly available. We therefore use the same setup as we do in OpenTitan, and add these runs as a second pipeline. The pipeline configuration and detailed results will only be visible to the core team, success/failure status will be available to everybody in the GitHub pull request view.

This PR is therefore mostly refactoring, the job configuration itself is in a separate repository.

### DIFF
--- a/ci/vars_to_logging_cmd.py
+++ b/ci/vars_to_logging_cmd.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Read an Azure Pipelines-compatible variables file, and convert it into
+# logging commands that Azure Pipelines understands, effectively setting the
+# variables at runtime.
+#
+# This script can be used as a workaround if variables cannot be included in the
+# Pipeline definition directly.
+#
+# See https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands
+# for more information on logging commands.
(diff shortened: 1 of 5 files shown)


ci/vars_to_logging_cmd.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# See https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands
+# for more information on logging commands.
+
+import sys
+import yaml
+
+def vars_to_logging_cmd(vars_file):
+    data = {}
+    print(vars_file)
```
rswarbrick, line 21: I understand that these won't break anything but, if we're keeping the handy debugging prints, could they gain a bit more info? Maybe this one could be replaced by the "Setting variables from..." line below (moving it above the load and assert which might fail). And do we really need the `print(data)` on line 24? Maybe it would be best to change the assertion to something like this?
```python
if not (isinstance(data, dict) and 'variables' in data):
    print("YAML file wasn't a dictionary with a 'variables' key. Got: {}"
          .format(data))
```

rswarbrick (approved):
Looks good to me, modulo the small nit about the Python. Thanks for doing this!

ci/vars_to_logging_cmd.py
imphil: The print(data) was indeed a debugging leftover, I removed it. Also replaced the assert with your suggestion. Thanks!

PR #1096, opened 2020-08-27
[rtl] Remove outdated assertion

tomroberts-lowrisc:
- The core does not rely on instr_err_i being always valid, and most bus
  protocols don't support that


vogelpi (approved):
Thanks @tomroberts-lowrisc for fixing this, LGTM!

PR #1098, opened 2020-08-27
[syn] Add script to print kGE equivalent

tomroberts-lowrisc:

### DIFF
--- a/syn/python/get_kge.py
+++ b/syn/python/get_kge.py
@@ -1,0 +1,80 @@
+#!/usr/bin/python3
+
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Parse a yosys area report and give a kGE equivalient
+
+import argparse
+
+
+def read_lib(lib_file_path, ref_cell):
+    with open(lib_file_path, 'r') as f:
+        lib_file = f.readlines()
+
+    cell_dict = {}
+    weighted_dict = {}
+    cell_name = None
+    for line_idx, line in enumerate(lib_file):
+        if line.startswith('  cell ('):
+            if cell_name is not None:
+                raise RuntimeError('{}:{} Found cell while searching for area'
+                                   .format(lib_file_path, line_idx + 1))
+            cell_name = line.split()[1].strip('()')
+        elif line.startswith('\tarea'):
+            if cell_name is None:
+                raise RuntimeError('{}:{} Found area while searching for cell'
+                                   .format(lib_file_path, line_idx + 1))
+            try:
+                cell_area = line.split()[2].strip(';')
+                cell_dict[cell_name] = float(cell_area)
+                cell_name = None
+            except (IndexError, ValueError):
+                raise RuntimeError('{}:{} Area declaration misformatted'
+                                   .format(lib_file_path, line_idx + 1))
+
+    if ref_cell not in cell_dict:
+        raise RuntimeError('Specified reference cell: {} was not found in '
+                           'library: {}' .format(ref_cell, lib_file_path))
+
+    for cell in cell_dict:
+        weighted_dict[cell] = cell_dict[cell] / cell_dict[ref_cell]
+    return weighted_dict
+
+
+def get_kge(report_path, weighted_dict):
+    with open(report_path, 'r') as f:
+        report = f.readlines()
+    ge = 0.0
+    for line_idx, line in enumerate(report):
+        data = line.split()
+        if not data:
+            continue
+        weight = weighted_dict.get(data[0])
+        if weight is not None:
+            try:
+                ge += float(data[1]) * weight
+            except (IndexError, ValueError):
+                raise RuntimeError('{}:{} Cell {} matched but was misformatted'
+                                   .format(report_path, line_idx + 1, data[0]))
+    print("Area in kGE = ", round(ge/1000, 2))
+
+
+def main():
+    arg_parser = argparse.ArgumentParser(
+        description="""Calculate kGE from a Yosys report and LIB file""")
+
+    arg_parser.add_argument('lib_file_path', help='Path to the LIB file')
+    arg_parser.add_argument('report_path', help='Path to the report')
+    arg_parser.add_argument('--cell', help='Reference cell (default:NAND2_X1)',
+                            default='NAND2_X1')
+
+    args = arg_parser.parse_args()
+
+    weighted_dict = read_lib(args.lib_file_path, args.cell)
+    get_kge(args.report_path, weighted_dict)
+
+
+if __name__ == "__main__":
+    main()
--- a/syn/syn_yosys.sh
+++ b/syn/syn_yosys.sh
@@ -52,3 +52,5 @@ yosys -c ./tcl/yosys_run_synth.tcl | tee ./$LR_SYNTH_OUT_DIR/log/syn.log
 sta ./tcl/sta_run_reports.tcl | tee ./$LR_SYNTH_OUT_DIR/log/sta.log
 
 ./translate_timing_rpts.sh
+
+python/get_kge.py $LR_SYNTH_CELL_LIBRARY_PATH $LR_SYNTH_OUT_DIR/reports/area.rpt


syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def read_lib(lib_file_path):
+    with open(lib_file_path, 'r') as f:
+        lib_file = f.readlines()
+
+    cell_dict = {}
+    cell_name = ''
+    for line in lib_file:
+        if re.search(r'^  cell \(', line):
```
rswarbrick, line 22: As a general note, `re.search(r'^foo')` is the same thing as `re.match(r'foo')`. Probably cleaner to use that explicitly? And, in this case, you have a literal string. So you could just write
```python
if line.startswith('    cell ('):
    ...
```
which I think is easier to understand.

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+            if cell_name != '':
+                print("Size missing")
+                quit()
```
rswarbrick, line 25: This would probably be cleaner using exceptions (`raise RuntimeError('size missing')`) or with a `die_horribly` function, which takes a message to write to `sys.stderr` and then exits with code 1. For bonus points, a line number might be helpful. The easiest way to do that is to change line 21 to `for line_idx, line in enumerate(lib_file):` (and remember that it starts at zero, so error messages need to talk about line `line_idx + 1`).

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+            cell_name = line.split()[1].strip('()')
+        elif re.search(r'^\s+area', line):
```
rswarbrick, line 27: Probably use `re.match` here.

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+            if cell_name == '':
+                print("Size found without cell")
+                quit()
+            cell_area = line.split()[2].strip(';')
+            cell_dict[cell_name] = float(cell_area)
```
rswarbrick, line 32: Maybe add some error handling here? I don't think you know that `line.split()` will generate at least 3 words, and you don't know that `cell_area` is a valid number.

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Parse a yosys area report and give a kGE equivalient
+
+import re
+import argparse
+
+weighted_dict = {}
```
rswarbrick, line 12: Eugh, globals! Probably cleaner to return `weighted_dict` from `read_lib` and then pass it as an argument to `get_kge`.

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def get_kge(report_path):
+    with open(report_path, 'r') as f:
+        report = f.readlines()
+    ge = 0.0
+    for line in report:
+        data = line.split()
+        if len(data) == 0:
+            continue
```
rswarbrick, line 45: A more "Pythony" way to write that is `if not data:` (because the empty string is "falsy" in Python)

syn/python/get_kge.py
```diff
@@ -0,0 +1,65 @@
...
+        if data[0] in weighted_dict:
+            ge += float(data[1]) * weighted_dict[data[0]]
```
rswarbrick, line 47: This is fine, but you could avoid the double lookup with:
```
weight = weighted_dict.get(data[0])
if weight is not None:
    ge += float(data[1]) * weight
```

And, as earlier, do you want to do any error checking to make sure there are exactly 2 words in the line and the second one is a valid float?

syn/python/get_kge.py
tomroberts-lowrisc: Will do

syn/python/get_kge.py
tomroberts-lowrisc: Yep will do

syn/python/get_kge.py
tomroberts-lowrisc: Nice - thanks for the suggestion, I'll have a go at that

syn/python/get_kge.py
tomroberts-lowrisc: Since the other one is matching a literal string, I'll just switch this over too. Weirdly these lines all start with a tab, whereas the cell lines all start with two spaces?

syn/python/get_kge.py
tomroberts-lowrisc: I'll have a go at this - let me know if I've done it wrong

syn/python/get_kge.py
tomroberts-lowrisc: Will do

syn/python/get_kge.py
tomroberts-lowrisc: Will do thanks

tomroberts-lowrisc (commented):
Thanks for the review - I've had a go at adding the error handling stuff, but let me know if there's a cleaner way of doing it.

syn/python/get_kge.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def read_lib(lib_file_path):
+    with open(lib_file_path, 'r') as f:
+        lib_file = f.readlines()
+
+    cell_dict = {}
+    weighted_dict = {}
+    cell_name = ''
+    for line_idx, line in enumerate(lib_file):
```
rswarbrick, line 19: As a note, you could also iterate over the file here instead of slurping it at the start. The code would be something like
```python3
with open(lib_file_path, 'r') as lib_file:
    for line_idx, line in enumerate(lib_file):
        ...
```
Sadly, you gain another level of indentation, so I'm not convinced it's better. Also, it's not like we're streaming through GB of data, so it doesn't really matter to read everything first.

syn/python/get_kge.py
```diff
@@ -0,0 +1,73 @@
...
+import argparse
+
+
+def read_lib(lib_file_path):
+    with open(lib_file_path, 'r') as f:
+        lib_file = f.readlines()
+
+    cell_dict = {}
+    weighted_dict = {}
+    cell_name = ''
```
rswarbrick, line 18: I'd consider using `None` instead of the empty string here. Or, I suppose, you could assert that `cell_name` isn't empty after line 24.

syn/python/get_kge.py
```diff
@@ -0,0 +1,73 @@
...
+    for line_idx, line in enumerate(lib_file):
+        if line.startswith('  cell ('):
+            if cell_name != '':
```
rswarbrick, line 21: A more "pythonic" way to say this is just `if cell_name:` (because the empty string is falsy in Python). If you go with changing the "no cell name" case to use None, this becomes `if cell_name is not None`.

Similarly, line 26 becomes `if not cell_name` or `if cell_name is None`, depending on which version you go with.

syn/python/get_kge.py
```diff
@@ -0,0 +1,73 @@
...
+                raise RuntimeError('Found cell while searching for area at \
+                    line ' + str(line_idx + 1))
```
rswarbrick, line 23: This is fine, but it's probably easier to write with the format function:
```python3
raise RuntimeError('Found cell while searching for area at line {}.'.format(1 + line_idx))
```

Conveniently, Python works like C and pastes adjacent string literals, so you can do things like this to break lines:
```python3
raise RuntimeError('This is my rather {} string, which '
                   'I am going to break to avoid long '
                   '{}.'
                   .format('long', 'lines'))
```

Also, I'd consider generating errors like a compiler "foo.txt:23: Oh no, this looks malformed" (you have the path, so it's not too painful to construct the message).

rswarbrick (commented):
Hi Tom! A couple of other "Python tips". If you don't care and would like to just get the damn thing in, no problem :-)

syn/python/get_kge.py
tomroberts-lowrisc: Will do

syn/python/get_kge.py
tomroberts-lowrisc: Left as-is for now

syn/python/get_kge.py
tomroberts-lowrisc: Will do

syn/python/get_kge.py
tomroberts-lowrisc: Thanks - was wondering about line breaks. Will add the compiler-style messages too.

tomroberts-lowrisc (commented):
Thanks - more updates will follow

syn/python/get_kge.py
```diff
@@ -0,0 +1,73 @@
...
+                                   .format(lib_file_path, line_idx + 1))
+            try:
+                cell_area = line.split()[2].strip(';')
+                cell_dict[cell_name] = float(cell_area)
+                cell_name = None
+            except (IndexError, ValueError):
+                raise RuntimeError('{}:{} Area declaration misformatted'
+                                   .format(lib_file_path, line_idx + 1))
+    for cell in cell_dict:
+        weighted_dict[cell] = cell_dict[cell] / cell_dict['NAND2_X1']
```
GregAC, line 37: Have the 1 GE cell specified as a config parameter rather than hardcoded? The name will probably differ between libraries.

GregAC (approved):
I note Yosys outptus a total 'chip area' so instead of working out the GE area for each cell then iterating over every cell in the area report summing up the GE areas for each cell instance you can just grab the final area and divide by the 1 GE area.

Though this way is fine, especially as it's now done. Having easy access to a per-cell GE and doing the sum in the script may prove useful for future work that wants to do a more detailed analysis anyway.

tomroberts-lowrisc (commented):
Agreed the cell by cell parsing is a bit pointless - it makes more sense with RAMs so might be useful in the future

syn/python/get_kge.py
```diff
@@ -0,0 +1,75 @@
...
+                                   .format(lib_file_path, line_idx + 1))
+            try:
+                cell_area = line.split()[2].strip(';')
+                cell_dict[cell_name] = float(cell_area)
+                cell_name = None
+            except (IndexError, ValueError):
+                raise RuntimeError('{}:{} Area declaration misformatted'
+                                   .format(lib_file_path, line_idx + 1))
+    for cell in cell_dict:
+        weighted_dict[cell] = cell_dict[cell] / cell_dict[ref_cell]
```
GregAC, line 37: Probably worth a check that `ref_cell` is in `cell_dict` after `cell_dict` with a useful error message if it isn't.

tomroberts-lowrisc: Good point - will add

PR #1099, opened 2020-08-31
Port ICache formal core file to final Edalize design

rswarbrick:
The PR that I pushed up to add sby support to
Edalize (https://github.com/olofk/edalize/pull/173) ended up with a
different input format from the one it started with.

That's now merged into master, so this patch updates the Ibex code to
match the final syntax.

### DIFF
--- a/formal/icache/run.sby.j2
+++ b/formal/icache/run.sby.j2
@@ -17,11 +17,14 @@ cv: depth 32
 smtbmc boolector
 
 [script]
-read -sv @INPUT@
+{{"-sv"|gen_reads}}
 
 # Our formal properties are currently just about control logic, which
 # isn't affected by the exact behaviour of the memories in the design.
 # Blackbox them.
 blackbox $abstract\prim_generic_ram_1p
 
-prep -top ibex_icache
+prep -top {{top_level}}
+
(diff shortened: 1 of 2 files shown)


formal/icache/run.sby.j2
```diff
@@ -17,11 +17,14 @@ cv: depth 32
...
 [script]
-read -sv @INPUT@
+{{"-sv"|gen_reads}}
 
 # Our formal properties are currently just about control logic, which
 # isn't affected by the exact behaviour of the memories in the design.
 # Blackbox them.
 blackbox $abstract\prim_generic_ram_1p
 
 prep -top ibex_icache
```
imphil, line 27: I'd probably use the "toplevel" template variable here as well to ensure changes to the toplevel in the fusesoc .core file get reflected here as well. But fine either way, it's rather hypothetical at this point.

imphil (approved):

formal/icache/run.sby.j2
rswarbrick: I wasn't sure whether the future-proofing was worth the indirection. I'm still on the fence a bit, but let's show off all the features for now :-)

Issue #1106, opened 2020-09-10
The step operation is error when the SW bp is set in a J instruction.


apollojason:
Hi Guys： 
      I use the Ibex core running a simple test C code.  
       I set a SW bp before a “jump” instruction,  when the cpu fetched the SW bp instruction, the cpu is halted. 
       But after that, i use the “step" cmd in telnet,  the PC in CPU is wrong, it's not the correct PC which the JUMP instruction expected , the dpc has been changed. 
      When I use the HW bp， the dpc has‘t been changed.
       So, i think this is a bug
       please see the following figure.
      

![image](https://user-images.githubusercontent.com/71075865/92725183-51254200-f39e-11ea-81fc-698ca50d9b6b.png)

![image](https://user-images.githubusercontent.com/71075865/92725197-571b2300-f39e-11ea-8c1b-1e2a62d5aed7.png)

![image](https://user-images.githubusercontent.com/71075865/92725252-70bc6a80-f39e-11ea-8581-3ed01bcebc3b.png)

apollojason:
@GregAC Could u help me checking this issue？ Thanks

GregAC:
I think this is an issue with Ibex's `EBREAK` handling, it's setting the `depc` for the instruction following the `EBREAK` not the `EBREAK` itself.

From what I can tell from the waveform you've supplied, Ibex is first fetching the instruction immediately following the `jal 0x30000090`, which is the instruction at `0x300006c2` (fetches are aligned so it fetches at `0x300006c0`). The instruction at that address is a 4-byte `EBREAK` (`0x00100073`) which is split across 2 instruction fetches as it's unaligned.

Ibex executes that `EBREAK` giving you `depc = 0x300006c6` which is the address of the instruction after the `EBREAK` where it should produce `depc = 0x300006c2`

From a brief look at the RTL this should work correctly and uses the same mechanism as normal exceptions (and indeed it is tested in the verification environment) so there may be something more complex going on that causes it to break. If you could provide a full .vcd of the failure that would be useful in debugging it.

tomroberts-lowrisc:
I would guess that `enter_debug_mode` is set when the core goes to `FLUSH` as a result of the `EBREAK`. That causes the core to go to `DBG_TAKEN_IF` rather than `DBG_TAKEN_ID` hence it captures the wrong PC.

What debug module are you using? and did you definitely clear the HW breakpoint before setting the SW breakpoint?

apollojason:
[...]

PR #1107, opened 2020-09-14
[rtl] Fix debug step over EBREAK issue

tomroberts-lowrisc:
- If there is a request to enter debug mode (either due to a halt
  request or single stepping mode) on the same cycle as an EBREAK
  instruction, continue with the EBREAK and capture the ID PC rather
  than the IF PC.
- relates to #1106


GregAC (approved):
LGTM Thanks

PR #1108, opened 2020-09-16
Add formal checks for data independent timing for multdiv

towoe:
Move the sv2v script into util so it can be reused.
Add basic checks for multdiv. Count the number of cycles after starting an operation. Check that after a fixed amount of cycles the output is valid and only is this cycle.

I am open to suggestions how to manage the combination of different test in order to reduce repetition of code.

Also I am not sure how the setting of RV32M could be improved to reflect the original implementation as a `str` and `vlogdefine`. Maybe the call to sv2v puts some restrictions to it.


formal/data_ind_timing/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+slow_mull slow_mulh slow_div slow_rem:
+	$(call mk-fusesoc-cmd,$@)
+fast_mull fast_mulh fast_div fast_rem:
+	$(call mk-fusesoc-cmd,$@)
+single_mull single_mulh single_div single_rem:
+	$(call mk-fusesoc-cmd,$@)
+
+lint:
+	mypy --strict sv2v_in_place.py
```
rswarbrick, line 38: I don't think you want this bit :-) Also, could you move that target from the icache directory in the previous commit?

formal/data_ind_timing/run.sby.j2
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+[tasks]
+slow sl
+fast fs
+single si
+
+[options]
+mode bmc
+depth 50
```
rswarbrick, line 12: Wow, that's a long way! Do we actually need a depth of 50? I assume that a single operation only takes a few cycles, no?

rswarbrick (commented):
This looks neat! And I'm thrilled to see my hacky sv2v script being useful :-)

formal/data_ind_timing/run.sby.j2
towoe: Only multiplication operations are fast. A division needs 37 cycles plus some init cycles, that would be then around 40, so to round it up I landed on 50.
It might be possible to separate this, but I wanted to keep it simple. And it does not compute for so long.

formal/data_ind_timing/Makefile
towoe: Yeah, that does not make much sense :) Thanks for spotting that.

Moving to where? The script is not part of a make system anymore. Would it be okay to just remove this target altogether as it seems to be a specific developer only target?

formal/data_ind_timing/Makefile
```diff
@@ -0,0 +1,38 @@
...
+slow: slow_mull slow_mulh slow_div slow_rem
+fast: fast_mull fast_mulh fast_div fast_rem
+single: single_mull single_mulh single_div single_rem
+
+slow_mull slow_mulh slow_div slow_rem:
+	$(call mk-fusesoc-cmd,$@)
+fast_mull fast_mulh fast_div fast_rem:
+	$(call mk-fusesoc-cmd,$@)
+single_mull single_mulh single_div single_rem:
+	$(call mk-fusesoc-cmd,$@)
```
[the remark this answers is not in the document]
towoe: Yes, should be possible to do that.
I repeated the code only because I did not want to have a long line with all the targets. But I did not have the idea to generate the targets as you suggested. Will give it a go, thanks.

formal/data_ind_timing/run.sby.j2
rswarbrick: Ahh! That makes sense. :+1:

formal/data_ind_timing/Makefile
rswarbrick: Could you just stick a Makefile with this lint target in the same directory as the script? I agree that it's somewhat developer-only, but I dream of a day when we enforce these lint checks in CI :-)

towoe: Okay, will find a nice place for this :)
True, CI should do something like this. We can sneak in a github action super-linter ;)

towoe: Added a target in the root Makefile. With a check for all python files in the util folder. Two of three files report some errors.

tomroberts-lowrisc (approved):
This looks good to me pending @rswarbrick 's comments - thanks @towoe !

I have a vague notion that we should be able to frame an assertion that says "this operation always takes the same number of cycles" without actually specifying what that number is, but that's something we can mess with later!

rswarbrick (approved):
Looks great!

towoe:
Thanks for the review @rswarbrick and @tomroberts-lowrisc.

> I have a vague notion that we should be able to frame an assertion that says "this operation always takes the same number of cycles" without actually specifying what that number is, but that's something we can mess with later!

That is an intriguing thought.

PR #1109, opened 2020-09-16
Instantiate CSR module

tomroberts-lowrisc:
This PR is an example to discuss how we might move the CSRs to separate modules. The objective is mostly to allow parameterized addition of shadow copies, but there should be some other side benefits (clock gating, tidy-up etc.).

* I have only converted two CSRs here as an example to gather feedback.
* There is more scope for tidy-up and rationalization of the write muxing etc. once all CSRs are converted.
* There is also the option of making the CSR module type-parametrizable, rather than casting at the module boundaries. It's not currently supported by sv2v but I'll file that as an issue separately.

### DIFF
--- a/dv/uvm/core_ibex/ibex_dv.f
+++ b/dv/uvm/core_ibex/ibex_dv.f
@@ -37,6 +37,7 @@ ${PRJ_DIR}/rtl/ibex_alu.sv
 ${PRJ_DIR}/rtl/ibex_branch_predict.sv
 ${PRJ_DIR}/rtl/ibex_compressed_decoder.sv
 ${PRJ_DIR}/rtl/ibex_controller.sv
+${PRJ_DIR}/rtl/ibex_csr.sv
 ${PRJ_DIR}/rtl/ibex_cs_registers.sv
 ${PRJ_DIR}/rtl/ibex_counter.sv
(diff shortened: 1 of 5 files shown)


GregAC (approved):
Looks like a great approach to me.

This may also help our FPGA utilization significantly (CSRs were always a major resource user in FPGA), we'll have a bunch of flops with csr_wdata_int fed in rather than a bunch of flops muxing between the current mux value and csr_wdata_int depending upon what the decoder does assuming the fabric implements flop enables.

From a power perspective I wonder if it's worth adding an instantiated clock gate inside ibex_cs_registers so we can totally turn off the clock tree to the CSRs when we're not doing a CSR access (along with a few inevitable exceptions that can run off the main clock)?

GregAC:
Oh and 

> There is also the option of making the CSR module type-parametrizable, rather than casting at the module boundaries. It's not currently supported by sv2v but I'll file that as an issue separately.

+1 to this. Can Verilator deal with this currently?

tomroberts-lowrisc:
> From a power perspective I wonder if it's worth adding an instantiated clock gate inside ibex_cs_registers

Good plan - I'll add that

> +1 to this. Can Verilator deal with this currently?

Yep Verilator is happy with it. Not convinced it's better in this application since it adds its own messiness, but worth having support for it for other things.

PR #1110, opened 2020-09-16
In util, restrict mypy linting to sv2v_in_place.py

rswarbrick:
The other scripts in the directory don't have typing annotations.


towoe (approved):
As I introduced this I am also okay with reverting it back to just the sv2v python file :)

PR #1112, opened 2020-09-22
[simple_system] Update CSR Access for LLVM and GCC

lenary:
At some point in the future, the version of binutils we're using, and
the released version of LLVM, will coincide to use the correct name. In
the meantime, this change just uses the number which is supported by
both compilers.

This is, as I understand it, the only stumbling block to compiling the
in-repo C benchmark code with both GCC and LLVM. Being able to do so
would make it much easier for us to do comparisons on the benchmarks for
both compilers.

---

I'm not sure if the changes to `ibex_tracer.sv` are correct, it seems `0x321` and `0x322` are not defined in the privileged spec v1.11, and nor are they defined in https://ibex-core.readthedocs.io/en/latest/cs_registers.html

@luismarques are there other stumbling blocks to running ibex simple system with LLVM? I think this is the only one Yuichi told us about during the pointer authentication work.

### DIFF
--- a/examples/sw/simple_system/common/simple_system_common.c
+++ b/examples/sw/simple_system/common/simple_system_common.c
@@ -105,9 +105,13 @@ void pcount_reset() {
 void pcount_enable(int enable) {
   // Note cycle is disabled with everything else
   unsigned int inhibit_val = enable ? 0x0 : 0xFFFFFFFF;
-  // mucounteren == mcountinhibit but binutils doesn't seem to known the
-  // mcountinhibit name
-  asm volatile("csrw  mucounteren, %0\n" : : "r"(inhibit_val));
+  // CSR 0x320 was called `mucounteren` in the privileged spec v1.9.1, it was
+  // then dropped in v1.10, and then re-added in v1.11 with the name
+  // `mcountinhibit`. Unfortunately, the version of binutils we use only allows
+  // the old name, and LLVM only supports the new name (though this is changed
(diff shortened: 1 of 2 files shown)


imphil:
> I'm not sure if the changes to ibex_tracer.sv are correct, it seems 0x321 and 0x322 are not defined in the privileged spec v1.11, and nor are they defined in ibex-core.readthedocs.io/en/latest/cs_registers.html

The list of CSRs was created from the binutils & friends source code at the time I wrote the tracer to give identical output. It's very likely that since then things have moved on, so I'm fine with this update.

imphil (approved):

imphil:
Thanks Sam, you're a hardware engineer now!

lenary:
:ohno:

Also wait I get a "rebase and merge" button?! :ohnooooo:

imphil:
Go for it!

rswarbrick:
For extra entertainment, see https://github.com/riscv/riscv-binutils-gdb/issues/202.

luismarques:
> @luismarques are there other stumbling blocks to running ibex simple system with LLVM? I think this is the only one Yuichi told us about during the pointer authentication work.

I think was the only one. When I used Ibex for benchmarking I didn't use the entire Ibex SS sources, I only copied the relevant parts to the benchmarks (so that the benchmarks wouldn't depend on the Ibex software distribution, making it easier to upstream those ports), so maybe there could be something else, but probably not.

PR #1115, opened 2020-09-23
[simple-system] Add missing linker sections

### LINKED ISSUE #1092: Possible Problem with Coremark on Simple System: Access to Uninitialized Memory
Hi all, 

I may have run into a problem when simulating coremark on the simple system example.

I have been trying to run the simulation with synopsys VCS (since I am using the same framework to benchmark another core, whose tracer is not supported by verilator), and have found that the simulation fails throwing illegal instruction exceptions after some time. I have compared the traces and found that the differences start an access to apparently uninitialized memory. Since verilator defaults unknown values to 0, it seems that there is no problem and the program continues to the end. However when simulating with VCS, the program starts deviating at some point and starts to offend a range of assert statements due to unknown signals.
I can make the simulation run with VCS by padding the entire memory with 0. The results are then the same as with verilator. However, I am not sure if we should trust those results.

You can look into the trace logs here (large files):
https://drive.google.com/drive/folders/16SJRj9_kDt-uGIyC2Aie4j6XNU0LWnH1?usp=sharing

The first occurance of an uninitialized memory access is at time 2402. 

The Process I used for simulating with verilator: ( as specified in the readme).
```
#compile benchmark
make -C examples/sw/benchmark/coremark

#build the simulation
fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system

#run the simulation
./build/lowrisc_ibex_ibex_simple_system_0/sim_verilator/Vibex_simple_system --meminit=ram,examples/sw/benchmark/coremark/coremark.elf
```

For simulation with VCS:
```
#compile benchmark
make -C examples/sw/benchmark/coremark

#create vmem
riscv32-unknown-elf-objcopy -O binary examples/sw/coremark/coremark.elf examples/sw/coremark/coremark.bin
srec_cat examples/sw/coremark/coremark.bin --binary --offset 0x0000 --byte-swap 4 --fill 0xff -within examples/sw/coremark/coremark.bin -binary -range-pad 4 --output examples/sw/coremark/coremark.vmem --vmem 32

#build the simulation
fusesoc --cores-root=. run --target=sim --tool=vcs --setup --build lowrisc:ibex:ibex_simple_system --SRAMInitFile=examples/sw/coremark/coremark.vmem

#run the simulation
./build/lowrisc_ibex_ibex_simple_system_0/sim-vcs/lowrisc_ibex_ibex_simple_system_0
```
To pad the memory with 0 I add `--pad-to=0x1fffff` to the `objcopy` command.

tomroberts-lowrisc:
Adding the "small" sections to the example linker script. This ensures
that all sections of memory needing to be zeroed are captured.

Fixes #1092

### DIFF
--- a/examples/sw/simple_system/common/link.ld
+++ b/examples/sw/simple_system/common/link.ld
@@ -47,12 +47,18 @@ SECTIONS
 
     .rodata : {
         . = ALIGN(4);
+        /* Small RO data before large RO data */
+        *(.srodata)
+        *(.srodata.*)
         *(.rodata);
         *(.rodata.*)
     } > ram
 
     .data : {
         . = ALIGN(4);
+        /* Small data before large data */
+        *(.sdata)
+        *(.sdata.*)
         *(.data);
         *(.data.*)
     } > ram
@@ -61,6 +67,9 @@ SECTIONS
     {
         . = ALIGN(4);
         _bss_start = .;
+        /* Small BSS before large BSS */
+        *(.sbss)
+        *(.sbss.*)
         *(.bss)
         *(.bss.*)
         *(COMMON)


rswarbrick (approved):
Looks good to me. Should we also be adding `--orphan-handling=error` to our Makefiles to spot this sort of thing quicker in future? (See [LD docs](https://sourceware.org/binutils/docs/ld/Options.html#Options))

GregAC:
> Should we also be adding --orphan-handling=error to our Makefiles to spot this sort of thing quicker in future? (See LD docs)

Sounds like a good idea to me, was unaware of that option!

tomroberts-lowrisc:
Made a separate issue to add the option - there were a lot of other sections missing!

Issue #1070, opened 2020-08-12
Add more HW breakpoints


tomroberts-lowrisc:
Support for additional HW breakpoints would make software debug easier / more flexible

PR #1125, opened 2020-10-06
[rtl] Add support for additional HW breakpoints

towoe:
I was recently debugging some software and having more hardware breakpoints was really useful. Especially for setting a breakpoint and then returning up without the need to delete / disable.
It is also really helpful for simulation based software debugging as the "real" code is executed directly without all the debug code interception, so it helps to speed things up.

I think the requirement for this PR can be discussed in #1070 but since I was in need of the capability I wanted to share the current implementation for now and maybe it can be useful for others as a local temporary change.


imphil:
@tomroberts-lowrisc you're probably the best person to have a look at this PR?

rtl/ibex_cs_registers.sv
```diff
@@ -986,66 +987,89 @@ module ibex_cs_registers #(
...
+    always_comb begin
+      tmatch_control_d            = tmatch_control_q;
+      tmatch_control_d[tselect_q] = tmatch_control_we ? csr_wdata_int[2] :
+                                                        tmatch_control_q[tselect_q];
+    end
     // tmatch_value has its own clock gate
-    assign tmatch_value_d   = csr_wdata_int[31:0];
+    for (genvar i = 0; i < DbgHwBreakNum; i++) begin
+      always_comb begin
+        tmatch_value_d[i]         = tmatch_value_q[i];
```
tomroberts-lowrisc, line 1016: Better to create an enable for each of these based on the value of tselect_q rather than muxing the old value

rtl/ibex_cs_registers.sv
```diff
@@ -986,66 +987,89 @@ module ibex_cs_registers #(
...
+    assign tselect_d = tselect_we && (csr_wdata_int < DbgHwBreakNum) ?
+                        csr_wdata_int[$clog2(DbgHwBreakNum)-1:0] : tselect_q;
+    always_comb begin
+      tmatch_control_d            = tmatch_control_q;
+      tmatch_control_d[tselect_q] = tmatch_control_we ? csr_wdata_int[2] :
+                                                        tmatch_control_q[tselect_q];
+    end
     // tmatch_value has its own clock gate
-    assign tmatch_value_d   = csr_wdata_int[31:0];
+    for (genvar i = 0; i < DbgHwBreakNum; i++) begin
```
tomroberts-lowrisc, line 1014: Generate loops should all be named

tomroberts-lowrisc (commented):
Thanks @towoe 
A few nits but looks good to me so far - I have some changes about to go in that change the CSR structure. They will affect this a bit, but not drastically!

towoe:
> Thanks @towoe
> A few nits but looks good to me so far - I have some changes about to go in that change the CSR structure. They will affect this a bit, but not drastically!

Thanks for the review @tomroberts-lowrisc.
Will wait for your changes and then update the PR. Will also add some documentation and change the default number to 1.

rtl/ibex_cs_registers.sv
towoe: After your changes to the CSR structure I forward the value to each register and use the write enable signals.

rtl/ibex_cs_registers.sv
towoe: Done

doc/03_reference/cs_registers.rst
```diff
@@ -309,7 +309,11 @@ Reset Value: ``0x0000_0000``
 
 Accessible in Debug Mode or M-Mode when trigger support is enabled (using the DbgTriggerEn parameter).
 
-Ibex implements a single trigger, therefore this register will always read as zero.
+Number of the currently selected trigger.
+The number of triggers is configured by the DbgHwNumLen parameter.
+
+Writing a value larger than the number of supported triggers will write a truncated number to the register.
```
tomroberts-lowrisc, line 315: Might just be my interpretation, but truncated kind-of implies something different to what happens here. Looks like we actually just retain the old value, but could probably also set it to the supported maximum.

rtl/ibex_cs_registers.sv
```diff
@@ -1199,75 +1200,107 @@ module ibex_cs_registers #(
...
-    logic [31:0] tmatch_value_d, tmatch_value_q;
+    logic [DbgHwNumLen-1:0]   tselect_d, tselect_q;
+    logic                     tmatch_control_d;
+    logic [DbgHwBreakNum-1:0] tmatch_control_q;
+    logic [31:0]              tmatch_value_d;
+    logic [31:0]              tmatch_value_q[DbgHwBreakNum];
     // Write enables
-    logic tmatch_control_we;
-    logic tmatch_value_we;
+    logic tselect_we;
```
tomroberts-lowrisc, line 1211: Nit: Could align these as above, but not important

rtl/ibex_cs_registers.sv
```diff
@@ -1199,75 +1200,107 @@ module ibex_cs_registers #(
...
+    logic [DbgHwBreakNum-1:0] tmatch_control_we;
+    logic [DbgHwBreakNum-1:0] tmatch_value_we;
 
     // Write select
-    assign tmatch_control_we = csr_we_int & debug_mode_i & (csr_addr_i == CSR_TDATA1);
-    assign tmatch_value_we   = csr_we_int & debug_mode_i & (csr_addr_i == CSR_TDATA2);
+    assign tselect_we = csr_we_int & debug_mode_i & (csr_addr_i == CSR_TSELECT);
+    for (genvar i = 0; i < DbgHwBreakNum; i++) begin : g_dbg_tmatch_we
+      assign tmatch_control_we[i] = (i == tselect_q) & csr_we_int & debug_mode_i &
```
tomroberts-lowrisc, line 1218: Will need a width specifier on the genvar to prevent some lint warnings (e.g. `(i[DbgHwNumLen-1:0] == tselect_q)`)
The genvar is 32bits

rtl/ibex_cs_registers.sv
```diff
@@ -1199,75 +1200,107 @@ module ibex_cs_registers #(
...
+                                    (csr_addr_i == CSR_TDATA1);
+      assign tmatch_value_we[i]   = (i == tselect_q) & csr_we_int & debug_mode_i &
```
tomroberts-lowrisc, line 1220: Same as above

rtl/ibex_cs_registers.sv
```diff
@@ -1199,75 +1200,107 @@ module ibex_cs_registers #(
...
+        .rst_ni     (rst_ni),
+        .wr_data_i  (tmatch_value_d),
+        .wr_en_i    (tmatch_value_we[i]),
+        .rd_data_o  (tmatch_value_q[i]),
+        .rd_error_o ()
+      );
+    end
+
     // Assign read data
     // TSELECT - only one supported
```
tomroberts-lowrisc, line 1274: Maybe update this comment

rtl/ibex_cs_registers.sv
```diff
@@ -1199,75 +1200,107 @@ module ibex_cs_registers #(
...
+                                   1'b0,                         // store   : not supported
+                                   1'b0};                        // load    : not supported
     // TDATA1 - address match value only
-    assign tmatch_value_rdata = tmatch_value_q;
+    assign tmatch_value_rdata = tmatch_value_q[tselect_q];
 
     // Breakpoint matching
     // We match against the next address, as the breakpoint must be taken before execution
-    assign trigger_match_o = tmatch_control_q & (pc_if_i[31:0] == tmatch_value_q[31:0]);
+    logic [DbgHwBreakNum-1:0] trigger_match;
```
tomroberts-lowrisc, line 1299: Nit: I'd probably put this signal declaration above with the others

tomroberts-lowrisc (commented):
Thanks @towoe 
LGTM once few nits and lint things are fixed

doc/03_reference/cs_registers.rst
towoe: True, I will have a look at the specification and just check what should actually happen.

rtl/ibex_cs_registers.sv
towoe: No you are right, that should be aligned!

rtl/ibex_cs_registers.sv
towoe: Thanks! Done

rtl/ibex_cs_registers.sv
towoe: Moved to the aligned block

doc/03_reference/cs_registers.rst
tomroberts-lowrisc: I think what we do now is probably ok, but it might be more useful to software to return the max rather than the current value. Then you could write, say 7 and read back 4 and know right away that there are 4 supported.

towoe: I think the way gdb works is that the first time a breakpoint is set, it selects values until the maximum is determined.
So it might be more useful to software to return the max, but probably only gdb is the software which actually will interface with it? Only guessing, but maybe it is easier in software to determine the max by iterating than to guess a value and then guess until the max is reached?

Anyway, the spec allows some flexibility to `tselect` but looking at the description for `tdata` it would make more sense to align it and write the max value.

For reference:
current (0.13.2) debug spec
5.2.1 "Writes of values greater than or equal to the number of supported triggers may result in a different value  in  this  register  than  what  was  written.  To  verify  that  what  they  wrote  is  a  valid  index,debuggers can read back the value and check that tselect holds what they wrote."

5.2 "Most trigger functionality is optional.  All tdata registers follow write-any-read-legal semantics. If a debugger writes an unsupported configuration, the register will read back a value that is supported(which  may  simply  be  a  disabled  trigger)."

tomroberts-lowrisc (approved):
LGTM pending one comment on the documentation
Thanks @towoe

imphil:
@towoe can you open an issue in the OT repository to ask if SW people would like more than one HW breakpoint enabled for OT?

PR #1128, opened 2020-10-13
[rtl] Add plusarg to disable trace log

towoe:
As the tracer is built into a simulation it will currently always produce the trace log.
Instead of having to remove it manually I think it is a good idea to be able to disable it for a specific simulation run.
It is strictly not needed, but when I use a simulation together with a gdb debug session I sometimes would prefer not to have this huge file filled up.
Any feedback on the necessity of this?


rtl/ibex_tracer.sv
```diff
@@ -92,6 +94,18 @@ module ibex_tracer (
   localparam logic [4:0] MEM = (1 << 4);
   logic [4:0] data_accessed;
 
+  logic trace_log_disabled;
+  logic trace_log_disable_checked;
+  always_ff @(posedge clk_i) begin
+    if (!trace_log_disable_checked) begin
+      if (|$test$plusargs("ibex_tracer_disable")) begin
+        $display("Writing execution trace file disabled.");
```
rswarbrick, line 102: This message might be slightly easier to read as something more like "Instruction trace disabled", maybe? (Otherwise it sounds at first like we're writing something that is disabled)

Also, couldn't the check go in an `initial` block to avoid needing the `trace_log_disable_checked` signal?

rtl/ibex_tracer.sv
```diff
@@ -92,6 +94,18 @@ module ibex_tracer (
   localparam logic [4:0] MEM = (1 << 4);
   logic [4:0] data_accessed;
 
+  logic trace_log_disabled;
+  logic trace_log_disable_checked;
+  always_ff @(posedge clk_i) begin
+    if (!trace_log_disable_checked) begin
+      if (|$test$plusargs("ibex_tracer_disable")) begin
```
rswarbrick, line 101: It looks like Verible prefers `$value$plusargs` (see CI run).

rtl/ibex_tracer.sv
towoe: Your right, will reword the message.

I tried `initial` first, but it did not work. My guess was that the plusargs parsing is not available when testing in initial. But not sure on that.

rtl/ibex_tracer.sv
towoe: Yeah, I saw that and started to investigate. I found the Verible lint definition, but not the style it refers so.
And I am not sure which way we should go because I have not experience with plusgars.
I see that it could be misleading if values are dropped if `test` is used, but the extra step to drop a value for this particular use case seems also not ideal.

rtl/ibex_tracer.sv
rswarbrick: > My guess was that the plusargs parsing is not available when testing in initial. But not sure on that.

I think that should work: most plusargs parsing is done in `initial` blocks in OpenTitan.

rtl/ibex_tracer.sv
rswarbrick: Why not add something like `+ibex_tracer_enable=0` or `+ibex_tracer_enable=1`, defaulting to `1`? That also allows us to get rid of the negative name (which often gets confusing, when one starts "disabling the disable" and similar).

rtl/ibex_tracer.sv
towoe: You were right, obviously :)
Not sure what was wrong the last time.

rtl/ibex_tracer.sv
towoe: That's a good idea. Now `ibex_tracer_enable=0` should disable the tracer.

rswarbrick (approved):
Looks good to me!

towoe:
> Looks good to me!

Thanks for the review. I guess the wording in the documentation could be improved later.

PR #1140, opened 2020-10-17
[dv/ibex] Enable icache in Ibex environment

udinator:
This PR enables the icache in DV simulations of Ibex.

There's sort of a mixed bag of changes in this PR, explained below:
1) A new config entry is added in `ibex_configs.yaml` with the icache and ecc enabled - I did this to still allow flexibility for anyone who wants to simulate the core without the cache, but I have no objections to enabling the icache in the existing `experimental-maxperf-pmp-bmfull` config.
2) A small initialization routine is added in `ibex_asm_program_gen.sv` to write `1` to `cpuctrl` in order to actually enable the icache during runtime.
3) The `Makefile` now uses Spike as the default ISS, and uses the new config with cache enabled by default.
4) The yaml description of `cpuctrl.icache_enable` in `csr_description.yaml` is changed to be a `WARL` field since writes are now visible with the `ICache` parameter enabled in the default core configuration.
5) Explicit parameterization was added to `prim_ram_1p.sv` when instantiating the `prim_generic_ram_1p` as there are currently parameter width mismatches that cause compilation to fail with the cache enabled.

All of these changes were bundled into a single commit to make them all atomic so that previous commits don't encounter any strange compilation/simulation failures due to the cache being only partially enabled in the testbench.


dv/uvm/core_ibex/common/prim/prim_ram_1p.sv
```diff
@@ -30,7 +30,11 @@ module prim_ram_1p #(
   parameter prim_pkg::impl_e Impl = `PRIM_DEFAULT_IMPL;
 
   if (Impl == prim_pkg::ImplGeneric) begin : gen_generic
-    prim_generic_ram_1p u_impl_generic (
+    prim_generic_ram_1p #(
+      .Width          ( Width           ),
+      .Depth          ( Depth           ),
+      .DataBitsPerMask( DataBitsPerMask )
+    ) u_impl_generic (
```
imphil, line 37: Not sure how we ended up with this slightly broken file (most likely my fault). Actually, this file should match the one generated by fusesoc during normal builds. Can you please rectify this by copying the generated file to this one? It should look like this (keep the comment at the top of the file):

```
module prim_ram_1p

#(

  parameter  int Width           = 32, // bit
  parameter  int Depth           = 128,
  parameter  int DataBitsPerMask = 1, // Number of data bits per bit of write mask
  parameter      MemInitFile     = "", // VMEM file to initialize the memory with

  localparam int Aw              = $clog2(Depth)  // derived parameter

) (
  input  logic             clk_i,

  input  logic             req_i,
  input  logic             write_i,
  input  logic [Aw-1:0]    addr_i,
  input  logic [Width-1:0] wdata_i,
  input  logic [Width-1:0] wmask_i,
  output logic [Width-1:0] rdata_o // Read data. Data is returned one cycle after req_i is high.
);

  if (1) begin : gen_generic
    prim_generic_ram_1p #(
      .Depth(Depth),
      .MemInitFile(MemInitFile),
      .Width(Width),
      .DataBitsPerMask(DataBitsPerMask)
    ) u_impl_generic (
      .*
    );

  end

endmodule
```

rswarbrick:
This looks good to me, but I assume it needs the new custom Spike? In which case, could you add some notes about that to a README somewhere?

Eventually, generated programs should probably also be switching the icache on and off, but that can go in a future commit.

dv/uvm/core_ibex/common/prim/prim_ram_1p.sv
udinator: yup will do, thanks for the pointer!

udinator:
> This looks good to me, but I assume it needs the new custom Spike? In which case, could you add some notes about that to a README somewhere?

Yup, these changes will need the new Spike. Good point, I'll add a few words to the verification.rst documentation.
 
> Eventually, generated programs should probably also be switching the icache on and off, but that can go in a future commit.

The change to `ibex_asm_program_gen` allows the generated programs to turn the icache on actually - if you are referring to making this configurable at runtime though I agree this can wait until a future commit.

rswarbrick:
About enabling/disabling the icache, I mean that this is something a program can do in the middle of its execution, and we definitely need to verify that doing so works. Since we're not doing anything funky with self-modifying code, I think we can essentially just sprinkle the instruction stream with occasional "turn icache on" and "turn icache off" sequences and all should be well.

udinator:
> About enabling/disabling the icache, I mean that this is something a program can do in the middle of its execution, and we definitely need to verify that doing so works. Since we're not doing anything funky with self-modifying code, I think we can essentially just sprinkle the instruction stream with occasional "turn icache on" and "turn icache off" sequences and all should be well.

Yeah this is what I meant - sorry if I wasn't clear!
I definitely agree that randomly enabling/disabling the cache is something that needs to be tested.

udinator:
Latest force push fixes the RAM primitive as per @imphil's comment, and adds a note to the verification documentation about use of lowrisc/riscv-isa-sim.

imphil (approved):

Issue #1132, opened 2020-10-14
Illegal Instruction Retires


shetalani:
####  RISC-V Specification

> 3.1.11: "The minstret CSR counts the number of instructions the hart has retired. The mcycle and minstret registers
have 64-bit precision on all RV32 and RV64 systems.

> The counter registers have an arbitrary value after system reset, and can be written with a given value. Any CSR write takes effect after the writing instruction has otherwise completed."

> The counter-inhibit register mcountinhibit is a 32-bit WARL register that controls which of the hardware performance-monitoring counters increment. The settings in this register only control whether the counters increment; their accessibility is not affected by the setting of this register.

> When the CY, IR, or HPMn bit in the mcountinhibit register is clear, the cycle, instret, or hpmcountern register increments as usual. When the CY, IR, or HPMn bit is set, the corresponding counter does not increment.

####  Issue Description
An illegal instruction increments the minstret counter as if it was a legal one.

#### Steps to Reproduce

As shown below, the minstret counter is incremented for the illegal instruction csrrw x8, 0x010, x3, knowing that IR bit of mcountinhibit is set to '0 and thus allowing for the minstret to be incremented.  

![minstret_ill](https://user-images.githubusercontent.com/54367010/95972699-0e2e1280-0e13-11eb-87f8-9621c1352d6b.PNG)

---
Product: OneSpin 360 DV-Verify
App: Processor Verification App
Tool's version: 2020.2.0

shetalani:
Same happens for ecall and ebreak instructions, as well as load/ store access faults. For your reference, please check [https://github.com/openhwgroup/cv32e40p/issues/533](url).

imphil:
@tomroberts-lowrisc is this issue fixed by https://github.com/lowRISC/ibex/pull/1141, or are there outstanding issues?

tomroberts-lowrisc:
AFAIK
@shetalani are you able to rerun and confirm this is now fixed?

shetalani:
Hi @tomroberts-lowrisc, not really. It's not for the illegal cases of executing dret outside debug mode, or mret in U mode.

tomroberts-lowrisc:
Ok - thanks for confirming @shetalani 
I'll have a look at adding the missing things you mentioned

GregAC:
[...]

PR #1141, opened 2020-10-20
[rtl] Fix performance counter bug

tomroberts-lowrisc:
- Make sure performance counters only count retired, non-trapping
  instructions excluding ebrk/ecall
- Rewire some signalling through the writeback stage to allow
  instruction retire to be signalled from one place
- Relates to #1132

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -131,6 +131,7 @@ module ibex_core #(
                                                // ease fan-out)
   logic [15:0] instr_rdata_c_id;               // Compressed instruction sampled inside IF stage
   logic        instr_is_compressed_id;
+  logic        instr_perf_count_id;
   logic        instr_bp_taken_id;
   logic        instr_fetch_err;                // Bus error on instr fetch
   logic        instr_fetch_err_plus2;          // Instruction error is misaligned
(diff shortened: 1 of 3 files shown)


rtl/ibex_core.sv
```diff
@@ -296,9 +298,10 @@ module ibex_core #(
   // signals relating to instruction movements between pipeline stages
   // used by performance counters and RVFI
   logic        instr_id_done;
-  logic        instr_id_done_compressed;
   logic        instr_done_wb;
 
+  logic        perf_count_wb;
+  logic        perf_count_comp_wb;
```
GregAC, line 304: We could do with some more specific naming here (with matching changes to similarly named signals) to make it more obvious these are the signals that specifically cause the minstret counters to increment.

Perhaps

```
logic perf_instr_ret_count_wb;
logic perf_instr_compressed_ret_count_wb;
```

(I use 'compressed' in full as that's how it's used elsewhere and 'comp' can be confused with 'complete' which is also a word you'd associated with instruction retired so can be confusing).

rtl/ibex_wb_stage.sv
```diff
@@ -91,6 +98,8 @@ module ibex_wb_stage #(
         rf_wdata_wb_q   <= rf_wdata_id_i;
         wb_instr_type_q <= instr_type_wb_i;
         wb_pc_q         <= pc_id_i;
+        wb_comp_q       <= instr_is_compressed_id_i;
```
GregAC, line 101: As above let's use 'compressed' in full, maybe `wb_instr_compressed_q`.

GregAC (approved):
Thanks for the fix. The logic looks good to me but I think we could do with some naming tweaks.

PR #1144, opened 2020-10-23
[doc] Mark icache verification as "done for now"

imphil:
The icache verification is complete until new bugs are found, let's mark
it as such in the docs.


rswarbrick (approved):
Sounds good to me.

PR #1145, opened 2020-10-23
[ibex/dv] return non-zero exit code upon failure

udinator:
Currently the Ibex run scripts return 0 no matter what the test result
is.
To get Ibex sims correctly integrated into CI, the Makefile needs to
return 1 upon seeing a log comparison failure to indicate an error.


imphil (approved):
Easy enough, thanks!

PR #1147, opened 2020-10-24
[ibex/dv] Update CSV log conversion script

udinator:
This PR makes a few changes to how we convert the Ibex trace logs into
CSV format for the riscv-dv functional coverage flow.

1) We now write an entry for every executed instruction into the output
CSV file, as not doing this will skew functional coverage results.

2) We now parse and organize instruction operands a little differently,
to match recent updates to the CSV format (in the upstream riscv-dv
repo).
Operands for instructions that look like this:
`<instr> <rd>, <rs1>(<imm>)`
are now displayed in the CSV file as `<rd>,<rs1>,<imm>` whereas
previously they were displayed as `<rd>,<imm>,<rs1>`.

3) The parsing script now supports bitmanipulation pseudo-instructions.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
@@ -22,7 +22,7 @@ try:
     from riscv_trace_csv import (RiscvInstructionTraceCsv,
                                  RiscvInstructionTraceEntry,
                                  get_imm_hex_val)
-    from lib import RET_FATAL, gpr_to_abi, sint_to_hex
+    from lib import RET_FATAL, gpr_to_abi, sint_to_hex, convert_pseudo_instr
     import logging
 
 finally:
@@ -33,7 +33,9 @@ INSTR_RE = \
     re.compile(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+(?P<pc>[0-9a-f]+)\s+"
                r"(?P<bin>[0-9a-f]+)\s+(?P<instr>\S+\s+\S+)\s*")
 RD_RE = re.compile(r"(x(?P<rd>[1-9]\d*)=0x(?P<rd_val>[0-9a-f]+))")
-ADDR_RE = re.compile(r"(?P<imm>[\-0-9]+?)\((?P<rs1>.*)\)")
+ADDR_RE = re.compile(r"(?P<rd>[a-z0-9]+?),"
+                     r"(?P<imm>[\-0-9]+?)"
+                     r"\((?P<rs1>[a-z0-9]+)\)")
 
 
 def _process_ibex_sim_log_fd(log_fd, csv_fd, full_trace=True):
@@ -60,27 +62,26 @@ def _process_ibex_sim_log_fd(log_fd, csv_fd, full_trace=True):
 
         # Extract instruction information
         m = INSTR_RE.search(line)
-        if m:
-            instr_cnt += 1
-            # Write the extracted instruction to a csvcol buffer file
-            trace_entry = RiscvInstructionTraceEntry()
-            trace_entry.instr_str = m.group("instr")
-            trace_entry.instr = m.group("instr").split()[0]
-            trace_entry.pc = m.group("pc")
-            trace_entry.binary = m.group("bin")
-            if full_trace:
-                # Convert the operands into ABI format for
-                # the functional coverage flow
-                operands = m.group("instr").split()[1]
-                trace_entry.operand = convert_operands_to_abi(operands)
-                process_trace(trace_entry)
+        if m is None:
+            continue
+
+        instr_cnt += 1
+        # Write the extracted instruction to a csvcol buffer file
+        trace_entry = RiscvInstructionTraceEntry()
+        trace_entry.instr_str = m.group("instr")
+        trace_entry.instr = m.group("instr").split()[0]
+        trace_entry.pc = m.group("pc")
+        trace_entry.binary = m.group("bin")
+        if full_trace:
+            expand_trace_entry(trace_entry, m.group("instr").split()[1])
 
         c = RD_RE.search(line)
         if c:
-            trace_entry.gpr.append('{}:{}'
-                                   .format(gpr_to_abi("x%0s" % c.group("rd")),
-                                           c.group("rd_val")))
-            trace_csv.write_trace_entry(trace_entry)
+            abi_name = gpr_to_abi("x{}".format(c.group("rd")))
+            gpr_entry = "{}:{}".format(abi_name, c.group("rd_val"))
+            trace_entry.gpr.append(gpr_entry)
+
+        trace_csv.write_trace_entry(trace_entry)
 
     return instr_cnt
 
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -33,7 +33,8 @@
     re.compile(r"^\s*(?P<time>\d+)\s+(?P<cycle>\d+)\s+(?P<pc>[0-9a-f]+)\s+"
                r"(?P<bin>[0-9a-f]+)\s+(?P<instr>\S+\s+\S+)\s*")
 RD_RE = re.compile(r"(x(?P<rd>[1-9]\d*)=0x(?P<rd_val>[0-9a-f]+))")
-ADDR_RE = re.compile(r"(?P<imm>[\-0-9]+?)\((?P<rs1>.*)\)")
+ADDR_RE = re.compile(
+    r"(?P<rd>[a-z0-9]+?),(?P<imm>[\-0-9]+?)\((?P<rs1>[a-z0-9]+)\)")
```
rswarbrick, line 37: Nit: For this sort of thing, I'd probably use Python's implicit string concatenation to split the string literal, rather than do the ugly "hanging indent" thing. As a bonus, you also get to split the regex somewhere meaningful, which makes it a bit easier to read. I'd actually consider splitting it over 3 lines like this:
```python
ADDR_RE = re.compile(r"(?P<rd>[a-z0-9]+?),"
                     r"(?P<imm>[\-0-9]+?)"
                     r"\((?P<rs1>[a-z0-9]+)\)")
```

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -69,17 +70,17 @@ def _process_ibex_sim_log_fd(log_fd, csv_fd, full_trace=True):
...
             trace_entry.binary = m.group("bin")
             if full_trace:
+                trace_entry.operand = m.group("instr").split()[1]
+                process_trace(trace_entry)
                 # Convert the operands into ABI format for
                 # the functional coverage flow
-                operands = m.group("instr").split()[1]
-                trace_entry.operand = convert_operands_to_abi(operands)
-                process_trace(trace_entry)
+                trace_entry.operand = convert_operands_to_abi(trace_entry.operand)
```
rswarbrick, line 77: This code is pretty hard to follow (setting a field, calling a mutating function on the object, then setting the field again), and I think we could do it more simply. Since we don't use `process_trace` anywhere else, could we move the `convert_operands_to_abi` call inside it? Maybe something like this would be cleaner? (untested)
```python
            ...
            trace_entry.binary = m.group("bin")
            if full_trace:
                expand_trace_entry(trace_entry, m.group("instr").split()[1])

            c = RD_RE.search(line)
            ...

...

def expand_trace_entry(entry, operands):
    '''Expand a trace for a single instruction

    Add operands to the entry, converting from raw register names (x0, x1 etc.)
    to ABI names (a1, t1 etc.).

    '''
    process_imm(entry)
    entry.instr, operands = \
        convert_pseudo_instr(entry.instr, operands, entry.binary)

    # process any instructions of the form:
    # <instr> <reg> <imm>(<reg>)
    n = ADDR_RE.search(operands)
    if n:
        entry.imm = get_imm_hex_val(n.group("imm"))
        operands = ",".join([n.group("rd"), n.group("rs1"), n.group("imm")])

    # Convert the operands into ABI format for the functional coverage flow and
    # store them into the trace entry.
    entry.operand = convert_operands_to_abi(operands)

```

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -69,17 +70,17 @@ def _process_ibex_sim_log_fd(log_fd, csv_fd, full_trace=True):
...
-        c = RD_RE.search(line)
-        if c:
-            trace_entry.gpr.append('{}:{}'
-                                   .format(gpr_to_abi("x%0s" % c.group("rd")),
-                                           c.group("rd_val")))
+            c = RD_RE.search(line)
+            if c:
+                trace_entry.gpr.append('{}:{}'
+                                       .format(gpr_to_abi("x%0s" % c.group("rd")),
+                                               c.group("rd_val")))
```
rswarbrick, line 83: The extra 4 char indent pushes this over 80 chars, which is why I noticed it. Maybe it would be better to use some local variables to avoid the deep nesting?
```python
            c = RD_RE.search(line)
            if c:
                abi_name = gpr_to_abi('x{}'.format(c.group('rd')))
                gpr_entry = '{}:{}'.format(abi_name, c.group("rd_val"))
                trace_entry.gpr.append(gpr_entry)
```
Same number of lines, but they're shorter and it's more obvious what's going on (I think).

Another thing that might make this easier to read (sort of orthogonal to the above) is to change the check after `INSTR_RE.search` to continue on failure. This drops 4 spaces of indentation, which I think is probably better (and "skip this line if I don't understand it" isn't particularly confusing control flow). With the changes I suggested above, the result ends up looking something like this:
```python
    for line in log_fd:
        if re.search("ecall", line):
            break

        # Extract instruction information
        m = INSTR_RE.search(line)
        if m is None:
            continue

        instr_cnt += 1
        # Write the extracted instruction to a csvcol buffer file
        trace_entry = RiscvInstructionTraceEntry()
        trace_entry.instr_str = m.group("instr")
        trace_entry.instr = m.group("instr").split()[0]
        trace_entry.pc = m.group("pc")
        trace_entry.binary = m.group("bin")
        if full_trace:
            expand_trace_entry(trace_entry, m.group("instr").split()[1])

        c = RD_RE.search(line)
        if c:
            abi_name = gpr_to_abi('x{}'.format(c.group('rd')))
            gpr_entry = '{}:{}'.format(abi_name, c.group("rd_val"))
            trace_entry.gpr.append(gpr_entry)

        trace_csv.write_trace_entry(trace_entry)
```

rswarbrick (commented):
This looks good but, as we're touching the code, I'd love to make it cleaner as we go past.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
rswarbrick: Gaargh! That doesn't work because `process_imm` looks at `entry.operand` too. I probably prefer the "doubling down on not mutating stuff" approach, so `process_imm` changes to this:
```python3
...
    operands = process_imm(entry.instr, entry.pc, operands)
...

def process_imm(instr_name, pc, operands):
    """Process any immediate operand to follow RISC-V standard convention"""
    if instr_name in ['beq', 'bne', 'blt', 'bge', 'bltu', 'bgeu', 'c.beqz',
                      'c.bnez', 'beqz', 'bnez', 'bgez', 'bltz', 'blez',
                      'bgtz', 'c.j', 'j', 'c.jal', 'jal']:
        idx = operands.rfind(',')
        if idx == -1:
            imm = operands
            return str(sint_to_hex(int(imm, 16) - int(pc, 16)))

        imm = operands[idx + 1:]
        imm = str(sint_to_hex(int(imm, 16) - int(pc, 16)))
        return operands[0:idx + 1] + imm
```

The idea is you slowly move towards "compute some fields and make an (immutable) object out of them", which generally leads to much simpler code.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: thanks for the suggestion, I'll give that a shot! that definitely does make this regex look a lot cleaner.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: hm you make a good point re: the jumble of setting fields and mutating objects - I'll restructure this along the lines of what you suggested (making sure it works ofc) and will update this PR accordingly.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: whoops, thanks for catching the line length violation!
I'll restructure this part as well to match your suggested style.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
```diff
@@ -131,29 +132,44 @@ def convert_operands_to_abi(operand_str):
...
-            imm = str(sint_to_hex(int(imm, 16) - int(trace.pc, 16)))
-            trace.operand = trace.operand[0:idx + 1] + imm
+            imm = operands
+            return str(sint_to_hex(int(imm, 16) - int(pc, 16)))
+
+        imm = operands[idx + 1:]
+        imm = str(sint_to_hex(int(imm, 16) - int(pc, 16)))
+        return operands[0:idx + 1] + imm
+    else:
+        return operands
```
udinator, line 172: since we now rely on the return value from `process_imm`, we need to make sure we can pass through the operands to prevent not recognizing any instruction that isn't a branch or jump.

rswarbrick: Oops! I didn't notice that, thanks!

Some linters complain about
```python
if some_test:
    ...
    return 123
else:
   ...
```
because the "else" isn't strictly necessary (if you go in to the first branch, you always return). You've currently got an example of this at line 171. That complaint is why I suggested dropping the `else` after the `idx == -1` check. I don't have strong feelings here, but it seems some linter has slapped me across the knuckles enough times to do it automatically :-)

For this particular example, I'd actually suggest switching the outer `if` around. So you get:
```python
if instr_name not in [some, long, list]:
   return operands

idx = operands.rfind(',')
...
```
This avoids the single line floating off away from the check.

Anyway, this is a very nitty style question: I'm happy either way!

rswarbrick (approved):
Looks great! Thanks, Udi.

dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py
udinator: Sure, I have no issues changing this slightly!

PR #1149, opened 2020-10-27
Update tool versions

imphil:
Updates to tools as used in CI, see the individual commits for slightly more details.


rswarbrick (approved):
Looks good to me.

PR #1150, opened 2020-10-27
Enable configuration with Icache in public CI

imphil:
We now have an Ibex configuration
experimental-maxperf-pmp-bmfull-icache, which also enables the
instruction cache. Enable this configuration in public CI, which makes
it run the lint and RISC-V compliance jobs.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -95,6 +95,7 @@ jobs:
         - small
         - experimental-maxperf-pmp
         - experimental-maxperf-pmp-bmfull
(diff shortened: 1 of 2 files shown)


tomroberts-lowrisc (approved):
LGTM - we could perhaps trim this list down a bit later if it's taking too long

imphil:
@tomroberts-lowrisc I added a second commit to tie off two branch predictor-related signals. Can you have a look?

tomroberts-lowrisc:
LGTM thanks!

PR #1151, opened 2020-10-27
Pass through ICache-related parameters in compliance test

imphil:
The RISC-V compliance toplevel needs to pass through the instruction
cache-related parameters, which are set by the ibex configuration
script.


imphil:
This unbreaks CI (I didn't see that there are checks missing when merging https://github.com/lowRISC/ibex/pull/1140).

tomroberts-lowrisc (approved):
LGTM

PR #1152, opened 2020-10-27
Enable Verible lint

imphil:
Turns out, just having a nice error message with the correct tool
invocation isn't enough: you actually need to call the tool with the
arguments you document.

Let's do that and actually run Verilator lint and Verible lint, and not
just Verilator lint twice.


tomroberts-lowrisc (approved):
Whoops!

PR #1153, opened 2020-10-27
[ci] Add a comment on private CI

imphil:
Private CI has a separate list of ibex configurations it runs (there's
no sane way around that). Add a note to the pipeline configuration to
hopefully remember to keep them in sync, if that's desired (they don't
*have to* be in sync).


rswarbrick (approved):
Good idea!

Issue #1154, opened 2020-10-27
[dv] Regression failure for riscv_multiple_interrupt_test with config experimental-branch-predictor


imphil:
* Seed: 3116
* `make -C dv/uvm/core_ibex ITERATIONS=1 TEST=riscv_rand_instr_test,riscv_multiple_interrupt_test,riscv_debug_basic_test ISS=spike IBEX_CONFIG=experimental-branch-predictor`
* CI run: https://dev.azure.com/lowrisc/lowrisc-private/_build/results?buildId=24876&view=logs&j=89b1484e-e59b-56b8-bef8-ec534b74be7c&t=69a04726-f3d2-5f59-d3f4-a33300a2ed9f
* Full build log in CI: https://artproduks1.artifacts.visualstudio.com/Ace3fc167-bb91-4c33-a65a-ae648df7f974/4efdbdf8-a184-4518-9864-57d2a49c2aed/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2xvd3Jpc2MvcHJvamVjdElkLzRlZmRiZGY4LWExODQtNDUxOC05ODY0LTU3ZDJhNDljMmFlZC9idWlsZElkLzI0ODc2L2FydGlmYWN0TmFtZS9vdXQ1/content?format=file&subPath=%2Fout.tar.xz

```
Tue, 27 Oct 2020 15:01:21 ERROR    Timeout[300s]: env SIM_DIR=out/seed-3116/rtl_sim/riscv_multiple_interrupt_test.0   out/seed-3116/rtl_sim/vcs_simv +vcs+lic+wait +signature_addr=8ffffffc  -ucli -do ./vcs.tcl       +ntb_random_seed=3116 +UVM_TESTNAME=core_ibex_debug_intr_basic_test +bin=out/seed-3116/instr_gen/asm_tests/riscv_multiple_interrupt_test_0.bin     +ibex_tracer_file_base=out/seed-3116/rtl_sim/riscv_multiple_interrupt_test.0/trace_core     -l out/seed-3116/rtl_sim/riscv_multiple_interrupt_test.0/sim.log  +require_signature_addr=1 +enable_irq_multiple_seq=1 
```

@udinator @GregAC is this something you can have a look at? The logs should contain all necessary information.

udinator:
Sure, I will take a look into this failure today.

udinator:
[...]

PR #1155, opened 2020-10-28
[dv] Increase simulation timeout

imphil:
Increase the RTL simulation timeout when not in LSF-mode from 5 to 10
minutes.

Could potentially solve #1154.


rswarbrick (approved):
Looks fine, but I wonder whether we should be raising an issue about the fact that the tests take so long. Assuming we manage 1KHz, a test of 300 seconds must be running at least 100k instructions. Surely we don't need that many!

imphil:
Yeah, this is certainly more a stop-gap than a solution. I'm working with Udi in #1154 to give him access to the CI run that actually failed to first of all see if we actually had a timeout because the simulation took too long, of if we picked a seed that actually shows a RTL/DV problem.

PR #1157, opened 2020-10-29
Fix lint issues

lukamac:
Fixed linter issues found by the `svlint` tool.

### DIFF
--- a/.svlint.toml
+++ b/.svlint.toml
@@ -1,0 +1,42 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Rules for svlint, a SystemVerilog linter commonly used in editors.
+# The configuration matches the lowRISC SystemVerilog style guide at
+# https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md.
+# See https://github.com/dalance/svlint/blob/master/RULES.md for a list of rules.
+
+[option]
+exclude_paths = ["build.*", "sw/.*", ".sv.tpl$", "vendor/.*"]
+
+[rules]
+case_default = true
+enum_with_type = true
+for_with_begin = true
+function_same_as_system_function = true
+function_with_automatic = true
+generate_for_with_label = true
+generate_if_with_label = true
+generate_keyword_forbidden = true
+generate_keyword_required = false
+genvar_declaration_in_loop = true
+genvar_declaration_out_loop = false
+if_with_begin = true
+inout_with_tri = false
+input_with_var = false
+interface_port_with_modport = false
+legacy_always = false
+level_sensitive_always = true
+loop_variable_declaration = true
+non_ansi_module = true
+output_with_var = false
+parameter_in_package = false
+priority_keyword = true
+tab_character = true
+unique0_keyword = false
+unique_keyword = false
+wire_reg = true
+generate_keyword = false
+tab_charactor = false
+genvar_declaration = false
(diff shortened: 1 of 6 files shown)


.svlint.toml
```diff
@@ -0,0 +1,42 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Rules for svlint, a SystemVerilog linter commonly used in editors.
+# The configuration matches the lowRISC SystemVerilog style guide at
+# https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md.
+# See https://github.com/dalance/svlint/blob/master/RULES.md for a list of rules.
+
+[option]
+exclude_paths = ["build.*", "sw/.*", ".sv.tpl$", "vendor/*"]
```
imphil, line 11: exclude_paths is a regular expression, where "any character" is `.*`. So the last item should be `"vendor/.*`.

dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
```diff
@@ -34,13 +34,13 @@ module ibex_riscv_compliance (
   assign rst_sys_n = IO_RST_N;
 
   // Bus hosts, ordered in decreasing priority
-  typedef enum {
+  typedef enum int {
```
imphil, line 37: We prefer four-state data types over two-state ones for synthesizable code. So that would be integer (4 state, i.e. 1, 0, Z and X are valid values) instead of int (2 state, only 1 and 0 are valid values). Even better are data types with an explicit width, so that would be a vector of logic values.

In short, make this 
```suggestion
  typedef enum logic [1:0] {
```

See also https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#enumerations

dv/riscv_compliance/rtl/ibex_riscv_compliance.sv
```diff
@@ -34,13 +34,13 @@ module ibex_riscv_compliance (
...
     TestUtilHost,
     CoreD,
     CoreI
   } bus_host_e;
 
-  typedef enum {
+  typedef enum int {
```
imphil, line 43:
```suggestion
  typedef enum logic {
```

dv/riscv_compliance/rtl/riscv_testutil.sv
```diff
@@ -72,6 +72,8 @@ module riscv_testutil (
         ADDR_SET_END_SIGNATURE: begin
           end_signature_addr_d = dev_wdata_i;
         end
+        default:
+          ;
```
imphil, line 76: That's correct, even though a slightly more common way of writing this is

```suggestion
        default: ;
```

Btw, we discussed this yesterday, but for more background, here's a paragraph in our style guide on FSMs and default statements: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#case-statements

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -50,11 +50,11 @@ module ibex_simple_system (
 
   logic clk_sys = 1'b0, rst_sys_n;
 
-  typedef enum {
+  typedef enum int {
```
imphil, line 53: Can you also use logic data types for these enums, as in the other file?

imphil (commented):
Thanks a lot @lukamac for this this contribution! It looks already very good. I have a couple small comments inline below. Can you fix those and update the pull request? (This will involve `git rebase -i` to fold the changes I requested into the existing commits. Let me know if you're lost in "git land" somewhere, and I'll be happy to help you out.)

imphil:
Ah, and please ignore the "Ibex Private CI" check failing. It's unrelated to your changes, and I filed #1159 to see where it actually came from.

lukamac:
@imphil Did the changes :smile:

imphil (approved):
Great, thanks! Thanks a lot for your contribution, I'm going to merge this now. Let me know if you look for inspiration for next tasks :smile:

PR #1164, opened 2020-11-01
Add VCS compile option for unicode

hodjat91:
Similar to simulator.yaml, this changes has to be applied to rtl_simulation as well to avoid weird compilation errors.


rswarbrick (approved):
Thanks for spotting this! Just out of interest, what command doesn't work without this fix? (I've got a recent GCC here, and I'm slightly surprised that everything was working)

PR #1171, opened 2020-11-02
Misc fixes

GregAC:


rtl/ibex_multdiv_fast.sv
```diff
@@ -249,7 +249,7 @@ module ibex_multdiv_fast #(
       end
     end
 
-    assign unused_mult1_res_uns = mult1_res[33:32];
+    assign unused_mult1_res_uns = mult1_res_uns[33:32];
```
GregAC, line 252: This fix is weird in that the existing version doesn't cause a lint problem when building the ibex_core_tracing target but does cause a problem when building simple system and when running the simple system lint target.

GregAC:
We're currently running simple system lint for a single config, I think we may want to add it to the set of per config tests as it would have spotted the icache parameter issue. It's a bit frustrating as core_tracing lint and simple system lint are very very similar, but not quite identical! There's also the odd point that simple system lint found an issue that core_tracing lint didn't and I cannot determine why.

imphil (approved):
LGTM, thanks. Let's just add the simple system with all configurations to CI. the other option is to reduce simple system to only allow a single config, which is not what we want probably. (A clear example why multiple configs and multiple entry points make the testing matrix explode quite rapidly.)

PR #1176, opened 2020-11-05
[dv] Fix riscv_debug_single_step_test

### LINKED ISSUE #1167: riscv_debug_single_step_test fail
Hi ,
when SEED=27754, IBEX_CONFIG=small, ISS=ovpsim, riscv_debug_single_step_test fail because dpc record the branch target address when core executes a branch instruction, but the test bench think the dpc of the core should record the address of the branch instruction address +4, I think the RTL is right, please help double check it.

Thanks,
Lihua

GregAC:
Testbench computation of next PC when single stepping was broken. A
branch may stall before branch_set is asserted and this wasn't being
taken into account. Where the branch target ALU is present branch_set
can be set on the first cycle of the branch so this bug wasn't apparent.
Even with the branch target ALU the branch may stall awaiting the result
of a load when the writeback stage is present.

To make matters more complicated for a jump the PC being jumped to is
only available when jump_set is asserted which happens whilst the jump
instruction is stalling.

This alters the riscv_debug_single_step_test so when waiting for a
single step it looks for either a branch_set or jump_set signal being
asserted. Otherwise it waits until the instruction has unstalled. This
should give correct behaviour in all scenarios (potentially further
thought required where exceptions and interrupts occur).

Fixes #1167

### DIFF
--- a/dv/uvm/core_ibex/env/core_ibex_instr_monitor_if.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_instr_monitor_if.sv
@@ -21,6 +21,8 @@ interface core_ibex_instr_monitor_if #(
   logic [DATA_WIDTH-1:0]  pc_id;
   logic                   branch_taken_id;
   logic [DATA_WIDTH-1:0]  branch_target_id;
+  logic                   stall_id;
+  logic                   jump_set_id;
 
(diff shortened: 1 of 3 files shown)


udinator (approved):
lgtm, this fix makes sense to me - thanks for catching it!

Issue #1159, opened 2020-10-30
[DV] Failing regression at riscv_debug_basic_test.0 for experimental-maxperf-pmp-bmfull


imphil:
* Config: experimental-maxperf-pmp-bmfull
* Test: riscv_debug_basic_test
* Seed: 11904
* Message: No dret detected, or incorrect privilege mode switch in                                     timeout period of 100000 cycles

See https://dev.azure.com/lowrisc/lowrisc-private/_build/results?buildId=25194&view=results for full logs (you can find the artifacts under the "Related" heading on that page where it says "1 published").

The test log says:

```
Test binary: out/seed-11904/instr_gen/asm_tests/riscv_rand_instr_test.0.o
ibex uvm log : out/seed-11904/rtl_sim/riscv_rand_instr_test.0/sim.log
riscv_rand_instr_test : [PASSED]

ibex : out/seed-11904/rtl_sim/riscv_rand_instr_test.0/trace_core_00000000.csv
spike : out/seed-11904/instr_gen/spike_sim/riscv_rand_instr_test.0.csv
[PASSED]: 694 matched

Test binary: out/seed-11904/instr_gen/asm_tests/riscv_debug_basic_test.0.o
ibex uvm log : out/seed-11904/rtl_sim/riscv_debug_basic_test.0/sim.log
riscv_debug_basic_test : [FAILED]

Test binary: out/seed-11904/instr_gen/asm_tests/riscv_multiple_interrupt_test.0.o
ibex uvm log : out/seed-11904/rtl_sim/riscv_multiple_interrupt_test.0/sim.log
riscv_multiple_interrupt_test : [PASSED]

ibex : out/seed-11904/rtl_sim/riscv_multiple_interrupt_test.0/trace_core_00000000.csv
spike : out/seed-11904/instr_gen/spike_sim/riscv_multiple_interrupt_test.0.csv
[PASSED]: 0 matched

2 PASSED, 1 FAILED
```

sim.log says:
```
ucli% run
UVM_INFO @ 0: reporter [RNTST] Running test core_ibex_debug_intr_basic_test...
UVM_INFO /azp/agent/_work/1/s/dv/uvm/core_ibex/tests/core_ibex_base_test.sv(111) @ 202500: uvm_test_top [uvm_test_top] Running test : out/seed-11904/instr_gen/asm_tests/riscv_debug_basic_test_0.bin
[...]

PR #1177, opened 2020-11-06
[dv/uvm] Fix ecall / irq race condition

tomroberts-lowrisc:
If an irq is driven on the same cycle as the end of test ecall
instruction, the irq checks can fail since the ecall will have modified
expected csr values. This fix just abandons the irq sequence if the
ecall has already happened.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -210,6 +210,12 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
     // there are multiple irqs asserted at once.
     irq_valid = get_max_valid_irq_id(irq);
     `uvm_info(`gfn, $sformatf("irq_id: 0x%0x", irq_id), UVM_LOW)
+    // Check if the test already ended and drop the irq request if so. The ecall can cause the irq
+    // checks to fail if we continue driving interrupts.
+    if (dut_vif.dut_cb.ecall === 1'b1) begin
+      irq_valid = 1'b0;
+    end
+
     // If the interrupt is maskable, and the corresponding bit in MIE is not set, skip the next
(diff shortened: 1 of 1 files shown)


tomroberts-lowrisc:
@udinator this fixes the issue, but might not be the best way of doing it. Let me know if you have a better way.
I pushed this now as it is failing in CI periodically (as in #1141).

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -210,6 +210,12 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
     // there are multiple irqs asserted at once.
     irq_valid = get_max_valid_irq_id(irq);
     `uvm_info(`gfn, $sformatf("irq_id: 0x%0x", irq_id), UVM_LOW)
+    // Check if the test already ended and drop the irq request if so. The ecall can cause the irq
+    // checks to fail if we continue driving interrupts.
+    if (dut_vif.dut_cb.ecall === 1'b1) begin
```
udinator, line 215: thanks @tomroberts-lowrisc for this!

could you try changing this conditional to `if (dut_vif.ecall === ...`?
I'm making a similar fix for the debug sequences on my end to address #1159, in which this logic will not work due to the fact that `dut_vif.dut_cb.ecall` won't necessarily be high when an ecall instruction is in the ID stage (which is really what we care about here).

tomroberts-lowrisc: Hi @udinator - confirmed that also fixes the issue, thanks
I'll leave this here until you push that fix then close this

udinator: great! I just opened up #1183 to address both of these issues.

tomroberts-lowrisc:
Superseded by #1183

PR #1179, opened 2020-11-06
[rtl] Fix NAPOT address matching (#1178)

### LINKED ISSUE #1178: PMP NAPOT masks are incorrect when the region address spans the entire granule
The address masking logic in [`ibex_pmp.sv`](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_pmp.sv) always considers the lowest bit of the granule of the PMP address configuration register (i.e. `csr_pmp_addr_i[r][PMPGranularity+2]`) not to be part of the range address in NAPOT mode. This is correct when `PMPGranularity == 0`, since then NA4 mode must be selected if the range address uses all bits, but it is incorrect when `PMPGranularity > 0`, since then it is legal in NAPOT mode to have a range address span the entire granule, with the next-lowest bit (i.e. `csr_pmp_addr_i[r][PMPGranularity+1]`) cleared to indicate that all higher bits are part of the range address.

For instance with `PMPGranularity == 28`, the granule covers bits `csr_pmp_addr_i[r][33:30]`, yet bit `csr_pmp_addr_i[r][29]` must also be considered to check whether a range address spans all 4 bits of the granule. A range address of `csr_pmp_addr_i[r] == 34'b00_0101_1111_1111_1111_1111_1111_1111_1111` could be used to specify a range from 1 GiB to 2 GiB for example and in that case the region address mask should be `region_addr_mask [r] == 4'b1111` (i.e. the entire granule). However, since only bits `csr_pmp_addr_i[r][33:30]` are currently considered when generating the mask, it is incorrectly set to `region_addr_mask [r] == 4'b1100`.

The specification states that bits `G-2:0` (i.e. bits `csr_pmp_addr_i[r][PMPGranularity:0]`) are implicitly set to all ones and hence do not need to be considered, but that implies that bit `G-1` (i.e. `csr_pmp_addr_i[r][PMPGranularity+1]`) is still used to specify the range:

> When G≥1, the NA4 mode is not selectable.  When G≥2 and pmpcfg<sub>i</sub>.A[1] is set, i.e. the mode is NAPOT, then bits pmpaddr<sub>i</sub>[G-2:0] read as all ones.
> -- <cite>[Volume II: RISC-V Privileged Architectures V20190608-Priv-MSU-Ratified, Sect. 3.6.1](https://github.com/riscv/riscv-isa-manual/releases/download/Ratified-IMFDQC-and-Priv-v1.11/riscv-privileged-20190608.pdf)</cite>

michael-platzer:
Generate correct masks for NAPOT range addresses covering the entire granule. When PMPGranularity > 0, then a NAPOT range can span the whole granule, which requires the next-lowest bit (i.e. highest bit that is not part of the granule) to be 0. Fixes #1178

### DIFF
--- a/rtl/ibex_pmp.sv
+++ b/rtl/ibex_pmp.sv
@@ -55,8 +55,8 @@ module ibex_pmp #(
     end
     // Address mask for NA matching
     for (genvar b = PMPGranularity+2; b < 34; b++) begin : g_bitmask
-      if (b == PMPGranularity+2) begin : g_bit0
-        // Always mask bit (G+2) for NAPOT
+      if (b == 2) begin : g_bit0
+        // Always mask bit 2 for NAPOT
         assign region_addr_mask[r][b] = (csr_pmp_cfg_i[r].mode != PMP_MODE_NAPOT);
       end else begin : g_others
         // We will mask this bit if it is within the programmed granule
@@ -65,7 +65,7 @@ module ibex_pmp #(
         //                  | This bit pos is the top of the mask, all lower bits set
         // thus mask = 1111 0000
         assign region_addr_mask[r][b] = (csr_pmp_cfg_i[r].mode != PMP_MODE_NAPOT) |
-                                        ~&csr_pmp_addr_i[r][b-1:PMPGranularity+2];
+                                        ~&csr_pmp_addr_i[r][b-1:PMPGranularity+1];
       end
     end
   end


tomroberts-lowrisc (approved):
Agree with this - thanks for the fix

byllgrim:
Hi @tomroberts-lowrisc and @michael-platzer .
I think maybe this merge introduced another bug.

Please consider line 68:
https://github.com/lowRISC/ibex/blob/master/rtl/ibex_pmp.sv#L68
`assign region_addr_mask[r][b] = [...]` was changed from using `[b-1:PMPGranularity+2]` to `[b-1:PMPGranularity+1]`.
I think this breaks NAPOT matching for G=0.

Here is an example that demonstrates my point.
The cover property below _is_ reachable, it also works from `PMPGranularity>1`, but it does not work for `PMPGranularity==0`.
The example demonstrates how the "ignore patterns" (e.g. yyyy...y011) works.
That is, `pmp_req_addr_i` is purposefully set to not match `csr_pmp_addr_i` in the 5 lsb (except for [1:0] which is not part of `pmpaddress` register!).
It is perfectly fine that they don't match because of the 01111 pattern at the end of `csr_pmp_addr_i`.
```
cov_please_napot: cover property (
  // given:
  (PMPGranularity == 1)
  && (csr_pmp_cfg_i[0].mode == ibex_pkg::PMP_MODE_NAPOT)
  && (priv_mode_i === ibex_pkg::PRIV_LVL_U)
  // then:
  && (csr_pmp_addr_i[0][33:2] == 8'b11101111)
  && (pmp_req_addr_i[33:2]    == 8'b11110101) //5-bit mismatch ignored?
  && (pmp_req_err_o == 0 ) //hope to get a match
  );
  ```
As said, G>=1 is ok, but try it with G=0.
Because, in order to ignore the bits as dictated by the low en of `pmpaddress` register (e.g. `pmpaddress[2:0]=3'b011`), then one would need to have a 1 at `csr_pmp_addr_i[1]` (aka `pmpaddress[-1]`), or else the masking logic will not work.

I think it is a bug. But since there are so many nuances of the PMP spec that I might have misunderstood, I thought it best to ask if you guys agree or not.
If it is not a bug, but some mistake on my part, then please let me know :)
If my describe was hard to interpret then also let me know and I can word myself differently :)

**Comment by rswarbrick** [comment]
Hi there and thanks for the message. We might take a couple of weeks to get back to this properly because most lowRISC employees are having a Christmas break at the moment. See you in the new year!

**Comment by byllgrim** [comment]
Btw, 3 more comments to my post above.

This seemingly only fails when `csr_pmp_addr_i[1] = 0`.
But that should be the case anyway, as seen here:
https://github.com/lowRISC/ibex/blob/master/rtl/ibex_cs_registers.sv#L1071
`assign csr_pmp_addr_o[i] = {pmp_addr_rdata[i], 2'b00};`
And the Vol II spec seem to want that as well.

Above, the merge says "3 checks passed", but if you open it up
https://github.com/lowRISC/ibex/runs/1364421198
It says "10 errors".

At a first glance, it seems like a quick (but not so elegant) fix is as follows.
I have _not_ tested this.
```
      end else begin : g_others
        // We will mask this bit if it is within the programmed granule
        // i.e. addr = yyyy 0111
        //                  ^
        //                  | This bit pos is the top of the mask, all lower bits set
        // thus mask = 1111 0000
        if (PMPGranularity == 0) begin
          assign region_addr_mask[r][b] = (csr_pmp_cfg_i[r].mode != PMP_MODE_NAPOT) |
                                          ~&csr_pmp_addr_i[r][b-1:2];
        end else begin
          assign region_addr_mask[r][b] = (csr_pmp_cfg_i[r].mode != PMP_MODE_NAPOT) |
                                          ~&csr_pmp_addr_i[r][b-1:PMPGranularity+1];
        end
      end
```

michael-platzer:
Hi @byllgrim ,

thanks for pointing this out and apologies for overlooking the consequences this change would have for a PMPGranularity of 0!

Changing the indexing of `csr_pmp_addr_i` on [line 68](https://github.com/lowRISC/ibex/blob/0199bbae665b9b142144e6688279e2ecef7d83a0/rtl/ibex_pmp.sv#L68) from `[b-1:PMPGranularity+2]` to `[b-1:PMPGranularity+1]` was correct nonetheless, since the standard requires that "When G≥2 and pmpcfgi. A[1] is set, i.e. the mode is NAPOT, then bits pmpaddri[G-2:0] read as all ones.", so bit G-1 (i.e. PMPGranularity+1) must still be taken into account.

However, as you wrote, this change breaks the behavior for a PMPGranularity of 0, which now includes bit 1 of `csr_pmp_addr_i`. For NAPOT this bit should be implicitely 1, but it is set to 0 in [ibex_cs_registers.sv](https://github.com/lowRISC/ibex/blob/0199bbae665b9b142144e6688279e2ecef7d83a0/rtl/ibex_cs_registers.sv#L1071). So the situation where PMPGranularity is 0 requires dedicated handling.

The fix that you propose seems fine to me.

About the error messages in the merge checks: I do not know why these appear, but I do not think that they have been introduced by this merge, since they already appear in earlier merges (for instances in [1171](https://github.com/lowRISC/ibex/pull/1171/checks) and [1177](https://github.com/lowRISC/ibex/pull/1177/checks)).

GregAC:
Hi @byllgrim thanks for raising this and providing a fix. I have opened a PR #1234 to implement the fix. In general please do submit a PR when you have a fix like this. In this instance I needed the fix for some other work I was doing (Implementing extended PMP behaviour #1232) so did the PR myself.

With regards to the checks passed message along with errors this is an issue with the compliance suite expecting an exception on unaligned accesses (where Ibex deals with these in hardware which is spec compliant), so the errors are ignored for the checks, see here for more details: https://github.com/lowRISC/ibex/issues/100

Issue #1174, opened 2020-11-03
[DV] SV disable fork label - Xcelium simulator differences


PaulOKCrevinn:
While running and debugging Ibex regression we came across some differences in how Cadence Xcelium simulates disable of fork labels. 
We've locally fixed this issue in 6 different testbench files and we can submit the change requests ourselves,
I wanted to raise an issue on this as it is something that needs to be monitored in the future. 
I am not sure if the testbench code is linted, but is it possible to lint for the disable label construct and flag it in future? 

Below is a summary of the issue we've found, we will follow up with the changes to the Ibex git repo this week. 

**Issue: Disabling a fork process by label name.** 
Cadence support site recommends that when a fork is to be disabled it is to be disabled using the "disable fork" statement and not the "disable fork_process_label" construct, 
the reason being that the "disable fork_process_label" construct for forks is not defined in the IEEE Std 1800-2017 LRM. See section 9.6.2 of the LRM (2017). 
As far as we have seen Cadence Xcelium will ignore a "disable fork_process_label" line and continue, thereby leaving the process running and altering the function of the testcase. 
This cause multiple test failures in a Ibex Xcelium regression and lead to particular problems with the riscv_reset_test. 

So for example in the following Ibex code in core_ibex_test_lib.sv: The "**disable imem_fork**" should be replaced with "**disable fork**" in order to be compatible with Cadence Xcelium. 
We have done very limited testing on these changes on EDA playground for Cadence Xcelium 19.09, Mentor Questa 2020.1, and Synopsys VCS 2019.06 and the "disable fork" code seem executes correctly for all these simulators. 

          fork : imem_fork
            begin
              check_next_core_status(HANDLING_EXCEPTION, "Core did not jump to exception handler");
              check_priv_mode(PRIV_LVL_M);
              latched_imem_err = 1'b1;
              `uvm_info(`gfn, $sformatf("latched_imem_err: 0x%0x", latched_imem_err), UVM_LOW)
            end
            begin
              clk_vif.wait_clks(750);
            end
          join_any
          **disable fork;    //  (Changed from:   disable imem_fork;)** 

Cadence support ticket referencing the above issue (NOTE: you will need a Cadence support account to view these): 

https://support.cadence.com/apex/ArticleAttachmentPortal?id=a1Od0000000syTQEAY&pageName=ArticleContent

Other Cadence support issues that deal with fork-join disable behaviour: 

[...]

PR #1180, opened 2020-11-06
[dv] Fixes for disabling forks, for Cadence Xcelium

CathalMCrevinn:
This makes changes to support Cadence Xcelium 20.09.001.

Cadence recommends that disabling forks should be done using a *"disable fork;"* statement, not the *"disable fork_process_label;"* construct.
The *"disable fork_process_label"* construct for forks is not defined in the IEEE Std 1800-2017 LRM. See section 9.6.2 of the LRM (2017).
For a more detailed explanation, please see the description from my colleague in Issue #1174.

In many cases, Xcelium will ignore a disable statement if it is not in the *"disable fork;"* form.
This causes problems in many tests, most notably in the riscv_reset_test.
Processes not being disabled when reset is asserted cause not just test failures, but the entire regression to crash.

Most changes here are simple substitutes of *"disable fork;"* where *"disable label;"* was.
Some changes in core_ibex_test_lib.sv require the layout of the fork to be adjusted.

Signed-off-by: CathalMCrevinn <cathal_minnock@crevinn.com>

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
@@ -36,7 +36,7 @@ class ibex_mem_intf_monitor extends uvm_monitor;
         wait (vif.monitor_cb.reset === 1'b1);
       join_any
       // Will only reach this point when mid-test reset is asserted
-      disable check_mem_intf;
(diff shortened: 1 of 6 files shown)


udinator (approved):

imphil:
Thanks a lot for your changes to better support Xcelium (and standards compliance!) @CathalMCrevinn and @PaulOKCrevinn!

PR #1182, opened 2020-11-06
[rtl] Reduce size of PMP addr CSRs to minimum

### LINKED ISSUE #1181: PMP CSR enhancement / bug
Looking PMP granularity as part of #1178 I noticed we always keep the PMP address CSRs as 32bits despite the granularity. At best this is inefficient as we're missing out on area savings, at worst it's a bug since we use the raw address in the matching logic (which doesn't match the modified version the user sees on a CSR read).

michael-platzer:
- Reduce the size of the PMP address configuration registers to the permissible minimum depending on granularity.
- Ensure consistency between PMP address CSR reads and the addresses forwarded to the PMP module.
- Follow the specification more strictly, by returning bits pmpaddr[G-2:0] as all ones when pmpcfg.A[1] is set (i.e. mode is NAPOT or NA4, though the latter is invalid for G > 0) and bits pmpaddr[G-1:0] as all zeros when pmpcfg.A[0] is clear (i.e. mode is OFF or TOR).
- Fixes #1181

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -120,6 +120,7 @@ module ibex_cs_registers #(
   import ibex_pkg::*;
 
   localparam int unsigned RV32MEnabled = (RV32M == RV32MNone) ? 0 : 1;
+  localparam int unsigned PMPAddrWidth = (PMPGranularity > 0) ? 33 - PMPGranularity : 32;
 
   // misa
   localparam logic [31:0] MISA_VALUE =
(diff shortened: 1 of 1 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -952,13 +952,13 @@ module ibex_cs_registers #(
...
-    pmp_cfg_t                    pmp_cfg         [PMPNumRegions];
-    pmp_cfg_t                    pmp_cfg_wdata   [PMPNumRegions];
-    logic [31:0]                 pmp_addr        [PMPNumRegions];
-    logic [PMPNumRegions-1:0]    pmp_cfg_we;
-    logic [PMPNumRegions-1:0]    pmp_cfg_err;
-    logic [PMPNumRegions-1:0]    pmp_addr_we;
-    logic [PMPNumRegions-1:0]    pmp_addr_err;
+    pmp_cfg_t                                        pmp_cfg         [PMPNumRegions];
+    pmp_cfg_t                                        pmp_cfg_wdata   [PMPNumRegions];
+    logic [31:(PMPGranularity>0)?PMPGranularity-1:0] pmp_addr        [PMPNumRegions];
```
tomroberts-lowrisc, line 957: Could add a localparam for this lower bound declaration, since it's a bit messy and is used in other places in the code within this generate block. Not a big deal though.

tomroberts-lowrisc (approved):
LGTM with one optional nit

rtl/ibex_cs_registers.sv
michael-platzer: I added a localparam declaration for the PMP address width, I thought that this might be even less messy than a localparam for the lower bound. But if you prefer, I can change it to a declaration for the lower bound instead.

tomroberts-lowrisc: Perfect - Thanks for fixing this!

PR #1183, opened 2020-11-09
[ibex/dv] Modify EOT ecall check

udinator:
This PR slightly modifies how we wait for the end of an Ibex sim.
Currently we wait on dut_vif.dut_cb.ecall, which will be delayed by a
cycle as it is in a clocking block.
However we should end the test immediately when an ecall is seen, so we
should rather wait on dut_vif.ecall instead to be more accurate and
prevent potential race conditions caused by irq/debug stimulus being
sent on the same cycle as an ecall instruction is executed by the core.

Addresses #1177 and #1159.


tomroberts-lowrisc (approved):
LGTM thanks @udinator

PR #1199, opened 2020-11-25
Ibex fcov v3

GregAC:
I've now adopted @sriyerg's initial suggestion for a single interface that gets bound into the design and probes into it with hierarchical references.

This doesn't represent a full set of functional coverage for Ibex but rather illustrates how I intend to go about writing the coverpoints.

In particular before detailed review I want to focus on the new macros I've created there's a couple of questions:

1. Do we want these macros?
2. Where should they live? For now they're in a vendored in directory. We'll either need to add them with a patch or add them to OpenTitan when we merge the final version of this PR

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -1,0 +1,141 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+interface core_ibex_fcov_if import ibex_pkg::*; (
+  input clk_i,
+  input rst_ni,
+
+  input priv_lvl_e priv_mode_id,
+  input priv_lvl_e priv_mode_if,
+  input priv_lvl_e priv_mode_lsu
+);
+  `include "dv_macros.svh"
+  import uvm_pkg::*;
+
+  typedef enum {
+    InsnCategoryALU,
+    InsnCategoryMult,
+    InsnCategoryDiv,
+    InsnCategoryBranch,
+    InsnCategoryJump,
+    InsnCategoryLoad,
+    InsnCategoryStore,
+    InsnCategoryOther,
+    InsnCategoryNone
+  } insn_category_e;
+
+  typedef enum {
+    IdStallTypeNone,
+    IdStallTypeInsn,
+    IdStallTypeLdHz,
+    IdStallTypeMem
+  } id_stall_type_e;
+
+  insn_category_e id_instr_category;
+
+  // Set `id_instr_category` to the appropriate category for the uncompressed instruction in the
+  // ID/EX stage.  Compressed instructions are not handled (`id_stage_i.instr_rdata_i` is always
+  // uncompressed).  When the `id_stage.instr_valid_i` isn't set `InsnCategoryNone` is the given
+  // instruction category.
+  // TODO: Illegal instructions
+  always_comb begin
+    case (id_stage_i.instr_rdata_i[6:0])
+      ibex_pkg::OPCODE_LUI:    id_instr_category = InsnCategoryALU;
+      ibex_pkg::OPCODE_AUIPC:  id_instr_category = InsnCategoryALU;
+      ibex_pkg::OPCODE_JAL:    id_instr_category = InsnCategoryJump;
+      ibex_pkg::OPCODE_JALR:   id_instr_category = InsnCategoryJump;
+      ibex_pkg::OPCODE_BRANCH: id_instr_category = InsnCategoryBranch;
+      ibex_pkg::OPCODE_LOAD:   id_instr_category = InsnCategoryLoad;
+      ibex_pkg::OPCODE_STORE:  id_instr_category = InsnCategoryStore;
+      ibex_pkg::OPCODE_OP_IMM: id_instr_category = InsnCategoryALU;
+      ibex_pkg::OPCODE_OP: begin
+        if(id_stage_i.instr_rdata_i[31:25] == 7'b0000000) begin
+          id_instr_category = InsnCategoryALU; // RV32I ALU reg-reg ops
+        end else if (id_stage_i.instr_rdata_i[31:25] == 7'b0000001) begin
+          if (id_stage_i.instr_rdata_i[14]) begin
+            id_instr_category = InsnCategoryMult; //MUL*
+          end else begin
+            id_instr_category = InsnCategoryDiv; // DIV*
+          end
+        end
+      end
+      default: id_instr_category = InsnCategoryOther;
+    endcase
+
+    if (!id_stage_i.instr_valid_i) begin
+      id_instr_category = InsnCategoryNone;
+    end
+  end
+
+  id_stall_type_e id_stall_type;
+
+  // Set `id_stall_type` to the appropriate type based on signals in the ID/EX stage
+  always_comb begin
+    id_stall_type = IdStallTypeNone;
+
+    if (id_stage_i.instr_valid_i) begin
+      if (id_stage_i.stall_mem) begin
+        id_stall_type = IdStallTypeMem;
+      end
+
+      if (id_stage_i.stall_ld_hz) begin
+        id_stall_type = IdStallTypeLdHz;
+      end
+
+      if (id_stage_i.stall_multdiv || id_stage_i.stall_branch ||
+          id_stage_i.stall_jump) begin
+        id_stall_type = IdStallTypeInsn;
+      end
+    end
+  end
+
+  `DV_FCOV_SVA(instruction_unstalled, id_stall_type != IdStallTypeNone ##1
+                                      id_stall_type == IdStallTypeNone)
+
+  covergroup uarch_cg @(posedge clk_i);
+    type_option.strobe = 1;
+
+    `DV_FCOV_EXPR_SEEN(cp_insn_unstalled, instruction_unstalled.triggered)
+
+    cp_insn_category_id: coverpoint id_instr_category;
+
+    cp_stall_type_id: coverpoint id_stall_type;
+
+    cp_wb_reg_hz: coverpoint id_stage_i.fcov_rf_rd_wb_hz;
+    cp_wb_load_hz: coverpoint id_stage_i.fcov_rf_rd_wb_hz &&
+                              wb_stage_i.outstanding_load_wb_o;
+
+    cp_ls_error_exception: coverpoint load_store_unit_i.fcov_ls_error_exception;
+    cp_ls_pmp_exception: coverpoint load_store_unit_i.fcov_ls_pmp_exception;
+
+    cp_branch_taken: coverpoint id_stage_i.fcov_branch_taken;
+    cp_branch_not_taken: coverpoint id_stage_i.fcov_branch_not_taken;
+
+    cp_priv_mode_id: coverpoint priv_mode_id;
+    cp_priv_mode_if: coverpoint priv_mode_if;
+    cp_prov_mode_lsu: coverpoint priv_mode_lsu;
+
+    `DV_FCOV_EXPR_SEEN(cp_interrupt_taken, id_stage_i.controller_i.fcov_interrupt_taken)
+    `DV_FCOV_EXPR_SEEN(cp_debug_entry_if, id_stage_i.controller_i.fcov_debug_entry_if)
+    `DV_FCOV_EXPR_SEEN(cp_debug_entry_id, id_stage_i.controller_i.fcov_debug_entry_id)
+    `DV_FCOV_EXPR_SEEN(cp_pipe_flush, id_stage_i.controller_i.fcov_pipe_flush)
+
+    wb_reg_hz_instr_cross: cross cp_insn_category_id, cp_wb_reg_hz;
+    stall_cross: cross cp_insn_category_id, cp_stall_type_id;
+    pipe_cross: cross cp_insn_category_id, if_stage_i.if_instr_valid,
+                      wb_stage_i.fcov_wb_valid;
+
+    interrupt_taken_instr_cross: cross cp_insn_unstalled, cp_interrupt_taken, cp_insn_category_id;
+    debug_entry_if_instr_cross: cross cp_debug_entry_if, cp_insn_category_id;
+    pipe_flush_instr_cross: cross cp_pipe_flush, cp_insn_category_id;
+  endgroup
+
+  bit en_uarch_cov;
+
+  initial begin
+   void'($value$plusargs("enable_uarch_cov=%d", en_uarch_cov));
+  end
+
+  `DV_INSTANTIATE_CG(uarch_cg, en_uarch_cov)
+endinterface
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -8,6 +8,7 @@
  */
 
 `include "prim_assert.sv"
+`include "dv_fcov.svh"
 
 module ibex_controller #(
     parameter bit WritebackStage  = 0,
@@ -828,6 +829,14 @@ module ibex_controller #(
       illegal_insn_q <= illegal_insn_d;
     end
   end
+
+  //////////
+  // FCOV //
+  //////////
+  `FCOV_SIGNAL(logic, interrupt_taken, (ctrl_fsm_cs != IRQ_TAKEN) & (ctrl_fsm_ns == IRQ_TAKEN));
+  `FCOV_SIGNAL(logic, debug_entry_if, (ctrl_fsm_cs != DBG_TAKEN_IF) & (ctrl_fsm_ns == DBG_TAKEN_IF));
+  `FCOV_SIGNAL(logic, debug_entry_id, (ctrl_fsm_cs != DBG_TAKEN_ID) & (ctrl_fsm_ns == DBG_TAKEN_ID));
+  `FCOV_SIGNAL(logic, pipe_flush, (ctrl_fsm_cs != FLUSH) & (ctrl_fsm_ns == FLUSH));
 
   ////////////////
   // Assertions //
--- a/vendor/lowrisc_ip/dv_utils/dv_fcov.svh
+++ b/vendor/lowrisc_ip/dv_utils/dv_fcov.svh
@@ -1,0 +1,39 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Coverage support is not available in Verilator but it's useful to include extra fcov signals for
+// linting purposes. They need to be marked as unused to avoid warnings.
+`ifdef VERILATOR
+  `define FCOV_MARK_UNUSED(__var_type, __var_name) \
+    __var_type unused_fcov_``__var_name; \
+    assign unused_fcov_``__var_name = fcov_``__var_name;
+`else
+  `define FCOV_MARK_UNUSED(__var_type, __var_name)
+`endif
+
+`ifndef FCOV_SIGNAL
+`define FCOV_SIGNAL(__var_type, __var_name, __var_definition) \
+  `ifdef DV_FCOV \
+    __var_type fcov_``__var_name; \
+\
+    assign fcov_``__var_name = __var_definition; \
+\
+    `FCOV_MARK_UNUSED(__var_type, __var_name) \
+  `endif
+`endif
+
+`ifndef FCOV_SIGNAL_GEN_IF
+`define FCOV_SIGNAL_GEN_IF(__var_type, __var_name, __var_definition, __generate_test, __default_val = '0) \
+  `ifdef DV_FCOV \
+    __var_type fcov_``__var_name; \
+\
+    if (__generate_test) begin : g_fcov_``__var_name \
+      assign fcov_``__var_name = __var_definition; \
+    end else begin : g_no_fcov_``__var_name \
+      assign fcov_``__var_name = __default_val; \
+    end \
+\
+    `FCOV_MARK_UNUSED(__var_type, __var_name) \
+  `endif
+`endif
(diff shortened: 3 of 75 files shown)


GregAC:
For ease of reviewing I'd recommend looking at this commit alone: https://github.com/lowRISC/ibex/pull/1199/commits/adbee89322d7715e7b2c7e3f119640530459d9f1 the other two are just mechanistic things around vendoring in the latest OT IP.

vendor/lowrisc_ip/dv_utils/dv_fcov.svh
```diff
@@ -0,0 +1,43 @@
+// Copyright lowRISC contributors.
```
GregAC, line 1: These are the macros I'm less sure about. They exist so that logic around interesting things we want to cover that aren't directly a single signal in the design can be written in the design file but explicitly flagged as being for coverage only.

The `DV_FCOV_LINT_ONLY` exists so we can build verilator systems with these included and get errors/warnings if the signals are messed up. Otherwise we'll only see failures in VCS builds. CI includes VCS builds but it makes life easier if the fcov signals are checked in verilator builds too.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,143 @@
...
+  input clk_i,
+  input rst_ni,
+
+  input priv_lvl_e priv_mode_id,
+  input priv_lvl_e priv_mode_if,
+  input priv_lvl_e priv_mode_lsu
+);
+`ifdef UVM
+  import uvm_pkg::*;
+`endif
```
rswarbrick, line 15: What is this used for?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,143 @@
...
+            end else begin
+              return InsnCategoryDiv; // DIV*
+            end
+          end
+        end
+        default: return InsnCategoryOther;
+      endcase
+    end
+
+    return InsnCategoryNone;
```
rswarbrick, line 68: I'd consider switching this around to start with
```systemverilog
  if (!insn_valid) begin
    return InsnCategoryNone;
  end
```
so that you can lose a level of indentation for the case statement.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,143 @@
...
+  endfunction
+
+  // Look at signals for ID stage (using uarch_fcov_vif) to determine what kind of stall, if any, is
+  // occurring in the ID stage and return it.
+  function id_stall_type_e determine_id_stall_type();
```
rswarbrick, line 73: This looks fine, but is there any reason to define this as a function? Couldn't you just define a signal? (It might well be easier to debug weird failures that way too).

This goes for `insn_category_from_bits` too (since it's only ever called with the signals from the ID stage).

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,143 @@
...
+      end
+
+      return IdStallTypeNone;
+    end
+
+   return IdStallTypeNone;
+  endfunction
+
+  `DV_FCOV_SVA(instruction_unstalled,
+    determine_id_stall_type() != IdStallTypeNone ##1 determine_id_stall_type() == IdStallTypeNone)
```
rswarbrick, line 95: Nit: I think it might be easier to read if you put the line break after the `##1` (so you get a "line per cycle"). That also avoids needing the ugly out-denting.

vendor/lowrisc_ip/dv_utils/dv_macros.svh
```diff
@@ -517,4 +517,23 @@
...
+  cover property (@(posedge __clk) disable iff (__rst == 0) (__sva)) begin \
+    -> __ev_name; \
+  end
+`endif
+
+// Creates a coverpoint for a 1-bit signal where only the asserted state is of interest for coverage
+// (e.g. where the signal indicates an event has occured).
+`ifndef DV_FCOV_SEEN
+`define DV_FCOV_SEEN(__cp_name, __sig_name) __cp_name: coverpoint __sig_name { bins seen = {1}; }
+`endif
```
rswarbrick, line 537: I think this probably needs a name that's more different from `DV_FCOV_SVA`, above (since they do quite different things). Maybe `DV_FCOV_SEEN_POINT`?? (I'm not massively convinced about that).

Also, maybe this should be generalised to take an arbitrary (boolean) expression. At the moment, the call sites look like
```systemverilog
    `DV_FCOV_SEEN(cp_insn_unstalled, instruction_unstalled.triggered)
```
Maybe
```systemverilog
    `DV_FCOV_SEEN(cp_insn_unstalled, instruction_unstalled.triggered != 0)
```
would make it clearer what's going on (and then you could also ask to see something being zero etc.).

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,143 @@
...
+
+    interrupt_taken_instr_cross: cross cp_insn_unstalled, cp_interrupt_taken, cp_insn_category_id;
+    debug_entry_if_instr_cross: cross cp_debug_entry_if, cp_insn_category_id;
+    pipe_flush_instr_cross: cross cp_pipe_flush, cp_insn_category_id;
+  endgroup
+
+  bit en_uarch_cov;
+
+  initial begin
+    en_uarch_cov = $test$plusargs("enable_uarch_cov");
```
rswarbrick, line 139: I think we've got some linters in OpenTitan that don't like `$test$plusargs` (sadly). This probably needs to go through `$value$plusargs` and a `+enable_uarch_cov=1`-style plusarg.

vendor/lowrisc_ip/dv_utils/dv_fcov.svh
rswarbrick: Instead of a special `DV_FCOV_LINT_ONLY` define, couldn't we just predicate it on the tool and special-case Verilator that way?

rtl/ibex_controller.sv
```diff
@@ -829,6 +830,14 @@ module ibex_controller #(
     end
   end
 
+  //////////
+  // FCOV //
+  //////////
+  `FCOV_SIGNAL(logic, interrupt_taken, (ctrl_fsm_cs != IRQ_TAKEN) & (ctrl_fsm_ns == IRQ_TAKEN));
+  `FCOV_SIGNAL(logic, debug_entry_if, (ctrl_fsm_cs != DBG_TAKEN_IF) & (ctrl_fsm_ns == DBG_TAKEN_IF));
+  `FCOV_SIGNAL(logic, debug_entry_id, (ctrl_fsm_cs != DBG_TAKEN_ID) & (ctrl_fsm_ns == DBG_TAKEN_ID));
+  `FCOV_SIGNAL(logic, pipe_flush, (ctrl_fsm_cs != FLUSH) & (ctrl_fsm_ns == FLUSH));
```
rswarbrick, line 839: I think I'd consider moving the `ifdef DV_FCOV` bit out of the definition of `FCOV_SIGNAL` and around this block. It's only a couple of lines per module, and I think it probably makes it clearer what's going on.

Once you've done that, I'm not sure whether the FCOV_SIGNAL macro is particularly useful. Maybe it's clearer just to define the signals and call FCOV_UNUSED manually? There will be a few more lines, but the code no longer defines signals (that a reader cares about) inside macros. What do you think?

rswarbrick (commented):
This looks good! I definitely like the general design, and the way we can define some signals in the design (close to the code in question) and then grab them from a bound-in interface.

I've got a couple of questions about the contents of the interface, and how the macros should work, but this looks like a good way to go.

vendor/lowrisc_ip/dv_utils/dv_fcov.svh
GregAC: Yes indeed, like how we do the assertion macros.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: As in a signal in this interface? Happy to do it that way. I didn't want to introduce it as a signal inside the design as I'm trying to separate signals that indicate microarchitectural things and will alter with the design (like the `pipe_flush` signal in the controller) from signals that purely related to how the coverage wants to view the world (such as putting things into categories).

vendor/lowrisc_ip/dv_utils/dv_macros.svh
GregAC: Yeah I couldn't come up with a great name, my other thought was `DV_FCOV_ONESHOT` but that's not great either.

I think it might just work with boolean expressions (you can use them directly in coverpoint expressions after all). I will need to experiment and find out.

I prefer leaving it without the `!=0` as that will be the primary (and maybe only) usage, it should be fairly clear what's going on.

rtl/ibex_controller.sv
GregAC: I wanted to highlight these were fcov only signals so there was no chance of them accidentally getting used as design signals. Though what you suggest should do the trick also. It will introduce a bit more boiler plate though

```
logic fcov_interrupt_taken;
logic fcov_debug_entry_if;
logic fcov_debug_entry_id;
logic fcov_pipe_flush;

assign fcov_interrupt_taken = (ctrl_fsm_cs != IRQ_TAKEN) & (ctrl_fsm_ns == IRQ_TAKEN);
assign debug_entry_if = (ctrl_fsm_cs != DBG_TAKEN_IF) & (ctrl_fsm_ns == DBG_TAKEN_IF);
assign debug_entry_id = (ctrl_fsm_cs != DBG_TAKEN_ID) & (ctrl_fsm_ns == DBG_TAKEN_ID);
assign pipe_flush = (ctrl_fsm_cs != FLUSH) & (ctrl_fsm_ns == FLUSH);

`FCOV_UNUSED(fcov_interrupt_taken);
`FCOV_UNUSED(fcov_debug_entry_if);
`FCOV_UNUSED(fcov_debug_entry_id);
`FCOV_UNUSED(fcov_pipe_flush);
```

So I'd slightly prefer my original macro.

There's also the `FCOV_SIGNAL_GEN_IF` macro. Which uses a generate if to check if some condition is met and if so assigns the signal to the expression given, but otherwise to a default. Again you can do without it but I prefer using it to avoid the boilerplate.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: It's required to use `DV_INSTANTIATE_CG` (which performs a call to the uvm logging framework).

GregAC:
Thanks for the review @rswarbrick

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
rswarbrick: Eugh, that's horrible - we have a macro that needs particular namespaces instantiated?! Let's get that fixed on the OT side.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
rswarbrick: Yep, that's what I meant.

vendor/lowrisc_ip/dv_utils/dv_macros.svh
rswarbrick: Oh, yeah, that's fine. I'd just got confused about the naming of the macro arguments: probably just rename `__sig_name` to `__expr` or similar.

GregAC: Will do, had failed to realize I'd accidentally written a macro that worked with arbitrary expressions!

rtl/ibex_controller.sv
rswarbrick: I guess this is a preference thing. Personally, I don't like macros that declare variables without looking like they're declaring variables. (This isn't just a SystemVerilog thing - I feel the same in C!) The problem is that it makes it more difficult to scan through code to see what's going on.

Note that you don't actually need the `assign`: you can fold the definitions into the signal declarations. And you could also collapse `FCOV_UNUSED` and pass it the concatenation of the nets. So
```
`FCOV_UNUSED(fcov0, {foo, bar, baz})
```
The only pain point problem is that you'll need an extra argument with a name to uniquify the unused signal.

End result:
```systemverilog
logic fcov_interrupt_taken = (ctrl_fsm_cs != IRQ_TAKEN) & (ctrl_fsm_ns == IRQ_TAKEN);
logic fcov_debug_entry_if = (ctrl_fsm_cs != DBG_TAKEN_IF) & (ctrl_fsm_ns == DBG_TAKEN_IF);
logic fcov_debug_entry_id = (ctrl_fsm_cs != DBG_TAKEN_ID) & (ctrl_fsm_ns == DBG_TAKEN_ID);
logic fcov_pipe_flush = (ctrl_fsm_cs != FLUSH) & (ctrl_fsm_ns == FLUSH);

`FCOV_UNUSED(fcov,
             {fcov_interrupt_taken, fcov_debug_entry_if, fcov_debug_entry_id, fcov_pipe_flush});
```
which I'd say compares pretty well.

GregAC: The `logic foo = value;` syntax sets an initial value (rather like assigning it in an `initial` block though not precisely, all those initialisations occur before any `initial` block is scheduled) rather than creating a continuous assignment so won't do what we want in this case.

You can do this with nets, i.e. this would do what we want:

```systemverilog
wire logic fcov_var = expr;
```

Though we don't use nets anywhere else and I don't think this alone justifies their inclusion (for one thing I don't know if Verilator even supports them).

Good point on the aggregate unused, one thing to deal with is the unknown width of the aggregate (at least unknown within the macro). I think we can use a unary or to deal with this, e.g.:

```systemverilog
`define FCOV_UNUSED(__name, __sigs) assign unused``__name == |__sigs;
```

> I guess this is a preference thing.

Pretty much, I'm not too opposed to doing it with explicit extra `logic` declaration lines though. Let's give @imphil the casting vote.

GregAC: Hmm style guide does allow use of `wire` 'allowed when necessary' and looks like it's supported in Verilator from a quick look at the manual.

I think the main objection to using that style would be it's a bit error prone, if someone attempts to copy it when setting up some new fcov signals and leaves off the `wire` it won't work and they may not notice.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
sriyerg: UVM macros call UVM methods underneath, so `uvm_pkg` needs to be imported wherever they are invoked. Not sure there is a good way to fix this.

rswarbrick: Oh, I see. Hmm, is `import uvm_pkg::*` idempotent? If so, we could fold it in to the macro?

rtl/ibex_controller.sv
rswarbrick: Yeah, I'm not sure what's better there. About `FCOV_UNUSED`, I think you probably want a unary XOR (Michael Schaffner pointed this out to me in the past). As I understand it, the problem with a unary OR is that some tools get upset if it they manage to prove that one of the bits is always true and then throw away the lot.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
sriyerg: See relevant link on this here: https://github.com/lowRISC/opentitan/pull/3822#discussion_r505939186
In that PR, I had coded the `` `DV_INFO`` .. etc. macros to directly invoke the UVM report functions scoped with `uvm_pkg::` rather than invoking the `` `uvm_info`` .. etc. macros underneath, precisely to solve this problem. However, I changed my mind and decided to no deviate from UVM provided macros. I have now reverted this back in lowrisc/opentitan#4419.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
udinator: agree with @rswarbrick, prefer to define a plusarg and use `$value$plusargs` here

udinator (commented):
One nit about the use of `$test$plusargs`, but otherwise I think the general design of this looks good!

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -0,0 +1,141 @@
...
+interface core_ibex_fcov_if import ibex_pkg::*; (
+  input clk_i,
+  input rst_ni,
+
+  input priv_lvl_e priv_mode_id,
+  input priv_lvl_e priv_mode_if,
+  input priv_lvl_e priv_mode_lsu
+);
+  `include "dv_macros.svh"
+  import uvm_pkg::*;
```
GregAC, line 14: This is still here as  `DV_INSTANTIATE_CG` uses `UVM_MEDIUM` directly. I will open an PR to fix this on the OT side.

ibex_core_tracing.core
```diff
@@ -101,6 +101,10 @@ parameters:
     paramtype: vlogparam
     description: "Number of PMP regions"
 
+  DV_FCOV_LINT_ONLY:
+    datatype: bool
+    paramtype: vlogdefine
+
```
GregAC, line 107: Forgot to remove this, will do so when it turns into a 'proper' PR

GregAC (commented):
As discussed in the HW meeting for the macros that live in the vendor/ directory once agreed I'll add them in properly with a patch. In parallel I'll open an PR in OT and aim to get them upstreamed so we can ditch the patch.

ibex_core_tracing.core
GregAC: Had another thing to fix so latest update removes this

rtl/ibex_controller.sv
GregAC: Ok let's draw this to a close:

@rswarbrick if we drop the `FCOV_SIGNAL` as you suggest would you suggest we use `wire logic fcov_var = expr;` style expressions or have separate `logic fcov_var;` and `assign fcov_var = expr;` expressions? Would you be happy with keeping a form of `FCOV_SIGNAL_GEN_IF` that drops the variable declaration and just does the if/else boiler plate or should we just drop that too?

@udinator @imphil @tomroberts-lowrisc @vogelpi any opinions here?

tomroberts-lowrisc: I'd vote against `wire logic = expr` - just seems too error prone and messy. No strong views on having the `FCOV_SIGNAL` macro or not!

rswarbrick: If we drop `FCOV_SIGNAL`, I'd probably suggest the wire declaration because it's more compact. (Of course, other code can define signals however it likes, so we're not exactly tying ourselves into a decision here).

On the `FCOV_SIGNAL_GEN_IF` question, I don't have very strong feelings. Presumably, we're picking between something like:
```systemverilog
logic my_signal0, my_signal1;
`FCOV_SIGNAL_GEN_IF(my_signal0, some_expression0, enable_param, default_val0)
`FCOV_SIGNAL_GEN_IF(my_signal1, some_expression1, enable_param, default_val1)
```
and
```systemverilog
logic my_signal0 = default_val0;
logic my_signal1 = default_val1;
generate if (enable_param) begin
  assign my_signal0 = some_expression0;
  assign my_signal1 = some_expression1;
end
```
I think the second is probably more readable and isn't that much longer. I "cheated", making two signals dependent on the same parameter, but I suspect that will be the usual use-case, and think it might be nicer to group things explicitly.

GregAC: I'd agree with Tom that I don't like the use of `wire` whilst allowed by the style guide, given we don't commonly use it (or use it all as far as I'm aware) it feels too error prone. Though perhaps Verilator has a lint warning if you do `logic foo = blah;`? I'll do a quick test and see.

As for `FCOV_SIGNAL_GEN_IF` unfortunately you can't fold the default value into the declaration. As it's just a shorthand for setting up a continuous assignment so if done as you suggest when `enable_param` is true you'll have two continuous assignments for the same signal. Meaning you have to do:

```
logic my_signal0;
logic my_signal1;

generate if (enable_param) begin
  assign my_signal0 = some_expression0;
  assign my_signal1 = some_expression1;
end else begin
  assign my_signal0 = default_val0;
  assign my_signal1 = default_val1;
end
```

If we're not using `wire` I would argue `FCOV_SIGNAL` as presented is the better choice to cut down on the repetition.

rswarbrick: > As for FCOV_SIGNAL_GEN_IF unfortunately you can't fold the default value into the declaration. As it's just a shorthand for setting up a continuous assignment so if done as you suggest when enable_param is true you'll have two continuous assignments for the same signal.

Maybe I've misunderstood something, but I thought an initial value is exactly what `logic foo = value;` set. (Note that it's `logic`, not `wire`). Isn't that what you reminded me of earlier in this thread?

GregAC: Ah this is true, it might work then, just mixing a proceedural assignment with a continuous assignment (if `enable_param` is set you get a proceedural assignment at time 0 with the default val along with the continuous assignment) may not be great, if indeed it's possible!

rswarbrick: Oh, I think you're right. It's actually explicitly disallowed in section 10.3.2 of 1800-2017. Boo.

GregAC: Verilator doesn't warn on `logic foo = value;`. Without lint to find such assignments without `wire` I think we're better off avoiding that style altogether.

udinator: A bit late to the party, but I think I'm inclined to agree with @tomroberts-lowrisc - the `wire logic` declaration just feels a bit too messy, and I'm ok if we leave the `FCOV_SIGNAL` as-is.

vendor/lowrisc_ip/dv_utils/dv_macros.svh
GregAC: I've renamed this to `DV_FCOV_EXPR_SEEN` I think that keeps it sufficiently different from `DV_FCOV_SVA` whilst alter capturing the intent of the macro

GregAC:
Closing in favour of https://github.com/lowRISC/ibex/pull/1239

PR #1202, opened 2020-11-27
bugfix tracer irq_x_i

micprog:
Hi @vogelpi,
When using ibex_core_tracing in PULP, the irq_x_i lines were in an X state due to this bug.


vogelpi (approved):
Thanks for the fix @micprog , LGTM!

PR #1203, opened 2020-11-27
Fix Jump Latency for Configuration with BTALU

ganoam:
The jump latency is not affected by the presence of a a branch target
ALU. The pipeline must stall for an instruction fetch from the new
address and flushing the prefetch buffer.

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -776,10 +776,8 @@ module ibex_id_stage #(
               perf_branch_o = 1'b1;
             end
             jump_in_dec: begin
-              // uncond branch operation
-              // BTALU means jumps only need one cycle
-              id_fsm_d      = BranchTargetALU ? FIRST_CYCLE : MULTI_CYCLE;
-              stall_jump    = ~BranchTargetALU;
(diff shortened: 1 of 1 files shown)


ganoam:
I'm not sure if this is the correct thing to do. I will open an issue instead.

ganoam:
I think this is correct. The branch target ALU accelerates conditional branches, by computing the branch target and the branch condition in parallel. For unconditional branches, i.e. jumps it makes no difference.

vogelpi (changes requested):
Hi @ganoam,
thanks for reporting the issue and for doing the PR.

Your fix seems to solve the issue but I fear it's not the right approach. Namely, because it causes a stall cycle in the ID stage for any jump. But this stall is not needed, the ID/EX stage is busy for a single cycle only. There is no other work to be done in ID that would motivate the stalling.

I think the problem you are describing requires a different fix. Probably the `lw` should not enter the ID stage in the first place. By the way, I can't see the `lw` nor the spurious data request in the screenshots you've provided.

I suggest to close this PR, file an issue and upload a wave dump/simple application there to allow us reproducing the problem.

ganoam:
Alright, I'll do that. Thanks!

PR #1205, opened 2020-11-30
Add branch prediction feature to icache

tomroberts-lowrisc:
Two commits in this PR -
The first is some minor performance updates to the icache when running in cache disabled mode, the second adds the branch prediction signals and deals with them in the same way as the prefetch buffer.

### DIFF
--- a/rtl/ibex_prefetch_buffer.sv
+++ b/rtl/ibex_prefetch_buffer.sv
@@ -286,7 +286,7 @@ module ibex_prefetch_buffer #(
   // Push a new entry to the FIFO once complete (and not cancelled by a branch)
   assign fifo_valid = rvalid_or_pmp_err & ~branch_discard_q[0];
 
-  assign fifo_addr = branch_mispredict_i ? branch_mispredict_addr : addr_i;
+  assign fifo_addr = branch_i ? addr_i : branch_mispredict_addr;
 
   ///////////////
   // Registers //
--- a/doc/03_reference/icache.rst
+++ b/doc/03_reference/icache.rst
@@ -52,14 +52,6 @@ The following table describes the available configuration parameters.
 +-------------------------+-----------+-----------------------------------------------+
 | ``NumWays``             | ``2``     | The number of ways.                           |
 +-------------------------+-----------+-----------------------------------------------+
-| ``SpecRequest``         | ``1'b0``  | When set, the system will attempt to          |
-|                         |           | speculatively request data from memory in     |
-|                         |           | parallel with the cache lookup. This can give |
-|                         |           | improved performance for workloads which      |
-|                         |           | cache poorly (at the expense of power).       |
-|                         |           | When not set, only branches will make         |
-|                         |           | speculative requests.                         |
-+-------------------------+-----------+-----------------------------------------------+
 | ``BranchCache``         | ``1'b0``  | When set, the cache will only allocate the    |
 |                         |           | targets of branches + two subsequent cache    |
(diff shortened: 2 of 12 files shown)


rtl/ibex_prefetch_buffer.sv
```diff
@@ -286,7 +286,7 @@ module ibex_prefetch_buffer #(
   // Push a new entry to the FIFO once complete (and not cancelled by a branch)
   assign fifo_valid = rvalid_or_pmp_err & ~branch_discard_q[0];
 
-  assign fifo_addr = branch_mispredict_i ? branch_mispredict_addr : addr_i;
+  assign fifo_addr = branch_i ? addr_i : branch_mispredict_addr;
```
rswarbrick, line 289: I think this hunk is probably because of our discussions about what happens when `branch_i & branch_mispredict_i`. Does it need a note in the commit message too?

GregAC: Could you explain what motivated this change? Following from what @rswarbrick said if we have `branch_i & branch_mispredict_i` the mis-predict should take priority. In that scenario we have a mis-predicted branch in ID/EX and a predicted taken branch in IF (which asserts `branch_i`). The predicted taken branch should never have been fetched due to the mis-predicted branch in ID/EX so the mispredict takes priority.

tomroberts-lowrisc: I'll add a comment to the commit message, but my thinking was the following:

The `branch_mispredict_i` already suppresses the valid output from the prefetch buffer, which prevents the branch predictor from being able to drive any predicted `branch_i` that cycle.

Likewise, the `branch_i` cannot be related to the mispredict (the mispredict exists because we thought there would be a taken branch this cycle, but the branch turned out to be not-taken, hence no `branch_i`).

That means the `branch_i` can only be some kind of exception / IRQ / debug type thing. In which case, we need to capture that address and ignore the mispredict, since it is being squashed by the exception. Having said that, I'm pretty sure these things can't happen on the same cycle with the current exception logic, but there is no assertion guaranteeing that, and things have changed around on exception timing in the past.

GregAC: Ah makes sense, been a while since I've thought about any of this! From a look at controller I think `branch_i & branch_mispredict_i` is impossible as we're stopping branch prediction in IF when we have a mispredict in ID/EX (for now at least, perhaps timing wise there's a better way to deal with this issue) and if you have an IRQ/Debug request at the same time that will take priority and you'll see a `branch_i` driven from ID/EX and no mispredict.

So as things currently stand I think this works either way round, however a comment (probably in ibex_if_stage.sv as it's a property of the interface to prefetch buffer/icache) to explain things somewhere plus an assertion may be worthwhile.

tomroberts-lowrisc: Extra comments, assertion and commit message comment now added

rswarbrick (approved):

GregAC (approved):
LGTM, has there been a DV run with icache and branch prediction enabled yet? Tests was looking pretty clean with branch predict on last time I tried it.

tomroberts-lowrisc:
Yep tests look clean with icache too (focused on the loop test which exercises the BP)

tomroberts-lowrisc:
@rswarbrick I have cherry-picked your reset-related testbench fixes to go in before the rtl changes as part of this PR (fixes CI)

rswarbrick:
Oops, yeah I should have said they were needed. Sorry!

PR #1208, opened 2020-12-01
Update CI to use Ubuntu 18.04

imphil:
Follow OpenTitan's lead and move to Ubuntu 18.04 for the CI runner. Ubuntu 16.04 comes with Python 3.5, which is EOL and causes more and more issues.

Specificially, this update is triggered by rst2pdf not being installable with pip's new dependency resolver (see also https://github.com/rst2pdf/rst2pdf/issues/489#issuecomment-736652808 for a detailed analysis).

### DIFF
--- a/ci/install-build-deps.sh
+++ b/ci/install-build-deps.sh
@@ -57,21 +57,12 @@ case "$ID-$VERSION_ID" in
 
       # Python dependencies
       #
-      # Installing six is a workaround for pip dependency resolution: six is
-      # already installed as system package with a version below the required
-      # one. Explicitly installing six through pip gets us a supported version.
-      #
       # Updating pip and setuptools is required to have these tools properly
       # parse Python-version metadata, which some packages uses to specify that
       # an older version of a package must be used for a certain Python version.
       # If that information is not read, pip installs the latest version, which
       # then fails to run.
-      $SUDO_CMD pip3 install -U pip six
-
-      # There's been a bit of a kerfuffle about setuptools version 50, which
-      # breaks importing distutils on Debian/Ubuntu systems. Make sure we don't
-      # pick it up until the dust has settled and things work again.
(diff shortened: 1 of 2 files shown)


tomroberts-lowrisc (approved):
Nice!
Thanks @imphil

ci/install-build-deps.sh
```diff
@@ -57,21 +57,12 @@ case "$ID-$VERSION_ID" in
...
       # an older version of a package must be used for a certain Python version.
       # If that information is not read, pip installs the latest version, which
       # then fails to run.
-      $SUDO_CMD pip3 install -U pip six
-
-      # There's been a bit of a kerfuffle about setuptools version 50, which
-      # breaks importing distutils on Debian/Ubuntu systems. Make sure we don't
-      # pick it up until the dust has settled and things work again.
-      $SUDO_CMD pip3 install -U 'setuptools < 50.0.0'
+      $SUDO_CMD pip3 install -U pip setuptools
```
rswarbrick, line 65: Does moving to Ubuntu 18.04 fix the problem with setuptools? And do we actually need versions newer than 18.04? If not, I'd be inclined to get rid of this line entirely.

rswarbrick (approved):
Looks good, modulo a pip question

ci/install-build-deps.sh
imphil: It's quite common for older pip/setuptools to be unable to parse metadata within packages that was generated with newer versions; hence upgrading to the latest version is something (in experience) is always necessary at some point.

The setuptools < 50.0.0 workaround isn't necessary any more as the dependency resolver shouldn't pick up this version any more with the `-U` flag passed to it.

So we could remove this line and wait until it bites us again, or leave it in.

rswarbrick: Cool, let's leave it in. Thanks for the explanation.

PR #1210, opened 2020-12-02
Add issue templates to GitHub project

imphil:
Add two issue templates to the GitHub project: one for questions, and
one for reporting bugs. These templates are not mandatory, users can
freely change them, or get a blank issue template instead.

To avoid having too much description/example text in the final issue the
instructions are written in HTML comments (which is not beautiful, but
should do the trick and seen commonly in projects).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,57 @@
+---
+name: bug
+title: Report a bug in Ibex
+about: Have you found a bug in Ibex or associated tooling?
+labels: Type:Bug
+---
+
+<!--
+Thank you for your interest in Ibex. We are continuously improving Ibex, and your bug report will help us with that. To help us understand and reproduce your problem as quickly as possible we have prepared a little template below.
+
+No text in blocks starting with < !-- will be shown in the final bug report.
+
+You can also copy/paste images into the bug report, and you can attach files to it by simply dragging them into the issue window.
+-->
+
+## Observed Behavior
+<!--
+Please describe the problem you saw. Include waveforms (in VCD or Verilator/GtkWave FST format only please) and log files if possible.
+-->
+
+## Expected Behavior
+<!-- Please describe the behavior you expected to see instead. -->
+
+## Steps to reproduce the issue
+<!--
+Please help us to reproduce the issue you're seeing by including a detailed description of what you did to get to the problem.
+
+For example:
+* What command did you run? (Also include the output of the command.)
+* What parameters/defines did you set?
+* Did you make modifications to the source code?
+-->
+
+## My Environment
+
+<!--
+Let us know how you're using Ibex. This helps us to better understand your problem and find a solution more quickly.
+-->
+
+**EDA tool and version:**
+<!--
+If you are using an EDA tool to work with Ibex let us know which tool and which version you're using.
+For example, "Xilinx Vivado 2020.1" or "Synopsys VCS 2020.03-SP1"
+-->
+
+**Operating system:**
+<!--
+Please let us know the operating system/Linux distribution you're using.
+For example "Ubuntu Linux 18.04" or "CentOS 7.3".
+-->
+
+**Version of the Ibex source code:**
+<!--
+Which version of the Ibex source code are you using?
+You can call `git rev-parse HEAD` in your Ibex source directory to get this information.
+Please also let us know if you have made changes to the code.
+-->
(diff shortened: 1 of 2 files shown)


imphil:
This is one way of doing it; we can also go less structured, which is what we've done in cocotb (see here: https://github.com/cocotb/cocotb/issues/new). It's probably something we need to try out to see what works best to give us the necessary information faster.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,48 @@
+---
+name: bug
+title: Report a bug in Ibex
+about: Have you found a bug in Ibex or associated tooling?
+labels: Type:Bug
+---
+
+<!--
+Thank you for your interest in Ibex. Are are continuously improving Ibex, and your bug report will help us with that. To help us understand and reproduce your problem as quickly as possible we have prepared a little template below.
```
rswarbrick, line 9: Nit: "We are", not "Are are" :-)

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,48 @@
...
+
+All text in blocks starting with < !-- won't be shown in the final bug report.
```
rswarbrick, line 11: Nit: Remove "All" (Or change it to "No text in blocks ... will be shown")

rswarbrick (approved):

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,48 @@
...
+
+You can also copy/paste images into the bug report, and you can attach files to it by simply dragging them into the issue window.
+-->
+
+## Observed Behavior
+<!-- Please describe the problem you saw. Include waveforms and log files if possible. -->
```
GregAC, line 17: waveforms - ideally an .fst or gzipped .vcd (you can attach to the github issue with drag/drop)

GregAC: though waveform images are also useful and easier to deal with initially, the issue is they often leave off pertinent signals. Clearly we need to spend lots of time building some kind of .vcd/.fst renderer that integrates nicely with github issues.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,48 @@
...
+-->
+
+**Operating system:**
+<!-- Please let us know the operating system/Linux distribution you're using.
+For example "Ubuntu Linux 18.04" or "CentOS 7.3".
+-->
+
+**Version of the Ibex source code:**
+<!-- Which version of the Ibex source code are you using?
+You can call `git rev-parse HEAD` in your Ibex source directory to get this information. -->
```
GregAC, line 48: Have you modified the code in any way?

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,34 @@
...
+For example, "Xilinx Vivado 2020.1" or "Synopsys VCS 2020.03-SP1"
+-->
+
+**Operating system:**
+<!-- Please let us know the operating system/Linux distribution you're using.
+For example "Ubuntu Linux 18.04" or "CentOS 7.3".
+-->
+
+**Version of the Ibex source code:**
+<!-- Which version of the Ibex source code are you using?
```
GregAC, line 33: Again question whether it's been modified

GregAC (approved):
Thanks for doing this

tomroberts-lowrisc (approved):
LGTM - thanks!
Lets see how they work out

vogelpi (approved):
Thanks @imphil for doing this, LGTM!

I think the template for questions could be extended, but I am not sure. Often, the environment (tools, OS, Ibex version) does not really matter.  Looking at previous questions, the differences are huge. We also get questions that could be answered by looking up the RISC-V spec or the Ibex manual. So, I am wondering if we should add a point asking what part of the spec or manual is actually relevant for this question (in the hope that people would look it up answer the question themselves).

But, this we can also add later. I don't want to delay this PR.

imphil:
@vogelpi I added a bit of text to the question template to ask about specification/doc links. I'm worried that the text will get too long and nobody reads it if we add too much detail to it, so I tried to keep it short (but I'm sure there's further room for improvement). We're not in a hurry to get this PR merged, so let me know if you have further ideas.

vogelpi:
Sorry @imphil I forgot about this PR. I agree with you, it shouldn't be too long, otherwise no one will read it. The text you've added to the question template looks good, it points people in the right direction. Let's merge this.

PR #1218, opened 2020-12-29
add RISC-V debug module to FPGA example

majbthrd:
I thought the FPGA Ibex offering might be enhanced by including a ready-to-use (but with the flexibility to opt-out, if desired, at synthesis time) example RISC-V debug implementation.

My goal was as much design re-use as possible.  SiFive already has a defacto standard for JTAG pinouts on the Arty boards, so I followed that.  The RISC-V debug module is a direct copy of the OpenTitan variant of the pulp-platform/riscv-dbg module.

The top level of the debug was customized to suit the FPGA example (the OpenTitan variant has a nicer integration of JTAG signals and IDCODE implementation, but no TileLink bus is used on the Ibex project), and top_artya7.sv was overhauled to instantiate the top-level debug module and provide bus arbitration between all three bus hosts.


GregAC:
Hi @majbthrd thank you for this contribution. We are interested in having a good debug module example in a basic Ibex system. We're just having some internal discussion as to where it should live. We'd like the Ibex repository to remain focused on Ibex itself so as examples and systems get more complex we'll want a different place for them to be.

What you've done looks like a great starting point for our more complete example system. It will take us a while longer to work out what we're doing here but rest assured this has not been discarded or forgotten.

majbthrd:
Thanks for the message.  I agree that makes sense to decouple the core Ibex RTL from other code.

FYI: previously, the [SEGGER J-Link](https://www.segger.com/products/debug-probes/j-link/) debug pod was not compatible with the pulp-platform/riscv-dbg module.  It would be unrealistic to expect that SEGGER should devise their own Ibex implementation from scratch for testing, but by providing them with pre-compiled bitstreams for the Arty FPGA board (using this PR), they were able to test it locally, the problem was identified, and V6.94c+ (05 Feb 2021) does now work with the target (Ibex + riscv-dbg).

```
- [J-Link EDU Mini V1]: RISC-V: IBEX core: Reading / writing memory did not work properly because J-Link did not handle 32-bit only system bus implementation properly. Fixed.
- [J-Link EDU/BASE/PLUS V10]: RISC-V: IBEX core: Reading / writing memory did not work properly because J-Link did not handle 32-bit only system bus implementation properly. Fixed.
- [J-Link EDU/BASE/PLUS V11]: RISC-V: IBEX core: Reading / writing memory did not work properly because J-Link did not handle 32-bit only system bus implementation properly. Fixed.
```

PR #1225, opened 2021-01-08
Revert "Clear MAKEFLAGS when running dvsim.py"

rswarbrick:
This reverts commit 31a18ad: the problem that it was working around
was fixed in OpenTitan with commit 249a544, vendored into Ibex as
b1daf9e.


imphil (approved):
I knew we'd need follow ups to the vendoring pr :) thanks!

rswarbrick:
Well, this isn't "needed": it's just removing a bodge that I remembered while I was in the shower this morning! :-)

PR #1226, opened 2021-01-08
CI: The distro-provided pyyaml package is enough

imphil:
In the past, we did explicitly install pyyaml through PIP to get a
version newer than the one provided in Ubuntu 18.04. Since then we
changed the calling code to not rely on newer pyyaml features and
are thus able to rely on an older version of pyyaml.

The distro-provided version of pyyaml is built with C bindings, which
are significantly faster than the pip-installed version, which uses a
pure Python-implementation of the parser/dumper.


rswarbrick (approved):
Looks great!

Issue #1228, opened 2021-01-09
[tracking] Activate pseudo zext.[b|h] decoding


towoe:
The tracer currently does not decode the pseudo instruction.
Use it once it is supported by more tools. Actually not quite which, but it was recently mentioned in https://github.com/llvm/llvm-project/commit/249d7de1190f50178181d2477aa661cd252e294c

vogelpi:
I just noted that more recent development versions of the tools we use (e.g. the Spike version for @GregAC 's cosim framework) as well as GCC v.10.2.0 (see https://github.com/riscv-collab/riscv-gcc/tree/riscv-gcc-10.2.0-rvb) and Binutils (https://github.com/riscv-collab/riscv-binutils-gdb) needed for Bitmanip support `zext.h` (`zext.b` is not part of Bitmanip v.1.0.0). So we could change our tracer to output `zext.h`, WDYT? @luismarques do you maybe know what the status in LLVM is on this?

luismarques:
> @luismarques do you maybe know what the status in LLVM is on this?

Both the current toolchain and the under evaluation one (based on LLVM 13) support those two instructions (with the appropriate command-line options). I did some quick testing and the encodings didn't seem to have changed. I can double-check that against the specs, if needed.

vogelpi:
Thanks for your feedback @luismarques ! This sounds great, I suggest that once we provide both an LLVM and GCC toolchain that support these pseudo-instructions, we make the tracer to output them. I think the encoding is still the same, no need to check.

Issue #1227, opened 2021-01-09
Remove Verilator warnings


towoe:
Following the reports in #1219 I had a look at Verilator and with https://github.com/verilator/verilator/commit/47eeef485d1781f104211d96ffb7498f55ae1b2b warnings are now created for unused parameters.

It also will create a warning for latches.

Remove those occurrences.

towoe:
Fixed with #1229

Issue #1219, opened 2021-01-07
build simple system error


guqian110:
## My Environment

Hi all,

I have compiled verilator from the source and installed fusesoc. when I was building the simple system under the root of ibex directory,

```
make build-simple-system
```

it gives warning information as following,

```
fusesoc --cores-root=. run --target=sim --setup --build \
	lowrisc:ibex:ibex_simple_system \
	--RV32E=0 --RV32M=ibex_pkg::RV32MFast --RV32B=ibex_pkg::RV32BNone --RegFile=ibex_pkg::RegFileFF --BranchTargetALU=0 --WritebackStage=0 --ICache=0 --ICacheECC=0 --BranchPredictor=0 --PMPEnable=0 --PMPGranularity=0 --PMPNumRegions=4
WARNING: Unknown item formal in section Target
WARNING: Unknown item formal in section Target
WARNING: Unknown item formal in section Target
INFO: Preparing lowrisc:dv_verilator:ibex_pcounts:0
INFO: Preparing lowrisc:dv_verilator:memutil_dpi:0
INFO: Preparing lowrisc:dv_verilator:simutil_verilator:0
INFO: Preparing lowrisc:ibex:ibex_pkg:0.1
INFO: Preparing lowrisc:prim:assert:0.1
INFO: Preparing lowrisc:prim:primgen:0.1
INFO: Preparing lowrisc:prim:secded:0.1
INFO: Preparing lowrisc:tool:check_tool_requirements:0.1
INFO: Preparing lowrisc:dv_verilator:memutil_verilator:0
INFO: Preparing lowrisc:ibex:ibex_tracer:0.1
INFO: Preparing lowrisc:lint:common:0.1
INFO: Preparing lowrisc:prim:prim_pkg:0.1
INFO: Preparing lowrisc:prim:clock_gating:0
INFO: Preparing lowrisc:prim:lfsr:0.1
INFO: Preparing lowrisc:prim:ram_1p:0
INFO: Preparing lowrisc:prim:ram_2p:0
INFO: Preparing lowrisc:ibex:ibex_icache:0.1
INFO: Preparing lowrisc:ibex:sim_shared:0
INFO: Preparing lowrisc:ibex:ibex_core:0.1
INFO: Preparing lowrisc:ibex:ibex_core_tracing:0.1
INFO: Preparing lowrisc:ibex:ibex_simple_system:0
INFO: Generating lowrisc:prim:prim_pkg-impl:0.1
[...]

PR #1229, opened 2021-01-09
Verilator warnings

towoe:
Not sure about the rules we have for turning of the lint warnings. Are they allowed inline? For this temporary situation it seems like a good case not to the the `vlt` file.

Feedback to using the defines in the tracer is also welcome :)

Relates to #1227 #1228 #1219

### DIFF
--- a/rtl/ibex_tracer_pkg.sv
+++ b/rtl/ibex_tracer_pkg.sv
@@ -23,7 +23,6 @@ parameter logic [31:0] INSN_BLT     = { 17'h?,             3'b100, 5'h?, {OPCODE
 parameter logic [31:0] INSN_BGE     = { 17'h?,             3'b101, 5'h?, {OPCODE_BRANCH} };
 parameter logic [31:0] INSN_BLTU    = { 17'h?,             3'b110, 5'h?, {OPCODE_BRANCH} };
 parameter logic [31:0] INSN_BGEU    = { 17'h?,             3'b111, 5'h?, {OPCODE_BRANCH} };
-parameter logic [31:0] INSN_BALL    = { 17'h?,             3'b010, 5'h?, {OPCODE_BRANCH} };
 
 // OPIMM
 parameter logic [31:0] INSN_ADDI    = { 17'h?,             3'b000, 5'h?, {OPCODE_OP_IMM} };
--- a/rtl/ibex_tracer_pkg.sv
+++ b/rtl/ibex_tracer_pkg.sv
@@ -80,10 +80,13 @@ parameter logic [31:0] INSN_CTZ  = { 12'b011000000001, 5'h?,  3'b001, 5'h?, {OPC
 parameter logic [31:0] INSN_PCNT = { 12'b011000000010, 5'h?,  3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTB = { 12'b011000000100, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTH = { 12'b011000000101, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
+// The ZEXT.B and ZEXT.H pseudo-instructions are currently not emitted by the tracer due to a lack
+// of support in the LLVM and GCC toolchains. Enabling this functionality when the time is right is
+// tracked in https://github.com/lowRISC/ibex/issues/1228
 // sext -- pseudoinstruction: andi rd, rs 255
-parameter logic [31:0] INSN_ZEXTB = { 4'b0000, 8'b11111111, 5'h?, 3'b111, 5'h?, {OPCODE_OP_IMM} };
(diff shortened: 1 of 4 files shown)


rtl/ibex_tracer_pkg.sv
```diff
@@ -80,10 +80,14 @@ parameter logic [31:0] INSN_CTZ  = { 12'b011000000001, 5'h?,  3'b001, 5'h?, {OPC
 parameter logic [31:0] INSN_PCNT = { 12'b011000000010, 5'h?,  3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTB = { 12'b011000000100, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTH = { 12'b011000000101, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
+// Currently not used as 9bd3350 removes it following missing definitions of instruction set
+// simulators. Tracked in https://github.com/lowRISC/ibex/issues/1228
+// verilator lint_off UNUSED
```
rswarbrick, line 85: I think I'd probably prefer to actually comment out the code. That way we can't forget to remove the `lint_off / lint_on` pair when we start using it again.

rswarbrick (commented):
Thanks for doing this!

rtl/ibex_tracer_pkg.sv
towoe: That's a good idea. Hopefully the line length does not matter too much ;)

rtl/ibex_tracer_pkg.sv
```diff
@@ -23,7 +23,6 @@ parameter logic [31:0] INSN_BLT     = { 17'h?,             3'b100, 5'h?, {OPCODE
 parameter logic [31:0] INSN_BGE     = { 17'h?,             3'b101, 5'h?, {OPCODE_BRANCH} };
 parameter logic [31:0] INSN_BLTU    = { 17'h?,             3'b110, 5'h?, {OPCODE_BRANCH} };
 parameter logic [31:0] INSN_BGEU    = { 17'h?,             3'b111, 5'h?, {OPCODE_BRANCH} };
-parameter logic [31:0] INSN_BALL    = { 17'h?,             3'b010, 5'h?, {OPCODE_BRANCH} };
```
imphil, line 26: Too bad we loose this nicely named constant :smile:

rtl/ibex_tracer_pkg.sv
```diff
@@ -81,10 +80,12 @@ parameter logic [31:0] INSN_CTZ  = { 12'b011000000001, 5'h?,  3'b001, 5'h?, {OPC
 parameter logic [31:0] INSN_PCNT = { 12'b011000000010, 5'h?,  3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTB = { 12'b011000000100, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
 parameter logic [31:0] INSN_SEXTH = { 12'b011000000101, 5'h?, 3'b001, 5'h?, {OPCODE_OP_IMM} };
+// Currently not used as 9bd3350 removes it following missing definitions of instruction set
+// simulators. Tracked in https://github.com/lowRISC/ibex/issues/1228
```
imphil, line 84: Let's slightly reword that:

"The ZEXT.B and ZEXT.H pseudo-instructions are currently not emitted by the tracer due to a lack of support in the LLVM and GCC toolchains. Enabling this functionality when the time is right is tracked in https://github.com/lowRISC/ibex/issues/1228."

shared/rtl/bus.sv
```diff
@@ -68,6 +69,7 @@ module bus #(
 
   // Device select
   always_comb begin
+    device_sel_req = 'b0;
```
imphil, line 72:
```suggestion
    device_sel_req = '0;
```

(For consistency with the code below, and more importantly, because I think it emits a lint warning somewhere otherwise.)

shared/rtl/bus.sv
```diff
@@ -59,6 +59,7 @@ module bus #(
 
   // Master select prio arbiter
   always_comb begin
+    host_sel_req = 'b0;
```
imphil, line 62:
```suggestion
    host_sel_req = '0;
```

imphil (approved):
LGTM with a couple of small comments. Thanks @towoe!

shared/rtl/bus.sv
towoe: Will update it, but just out of curiosity, can you remember where this lint warning is coming from? We have some other places where we use `'b0`, might as well clean that up at some point. And the style guide suggest to use `'0`, just to reinforce your suggestion.

rtl/ibex_tracer_pkg.sv
towoe: Thanks for the nice formulation Philipp!

rtl/ibex_tracer_pkg.sv
towoe: Well, instead of removing it we could rename and use it. `INSN_IBEX` or `INSN_KA...` ;)

towoe:
Thanks @rswarbrick and @imphil for the review.

PR #1239, opened 2021-01-15
Add Ibex functional coverage

GregAC:
This is very similar to my previous PR: https://github.com/lowRISC/ibex/pull/1199 though with some clean-ups and proper vendoring.

I've kept the `FCOV_SIGNAL` and `FCOV_SIGNAL_GEN_IF` macros as Tom and Udi expressed a slight preference for keeping them. They now live in a new `dv/fcov` directory. Perhaps we'll want to use them in OpenTitan but seemed simpler to have them as Ibex repository files rather than living in a patch only and applied to the vendored OT stuff.

The `DV_FCOV_SVA` and `DV_FCOV_EXPR_SEEN` macros are patched into dv_utils

### DIFF
--- a/dv/fcov/dv_fcov.core
+++ b/dv/fcov/dv_fcov.core
@@ -1,0 +1,17 @@
+CAPI=2:
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:dv:dv_fcov"
+description: "DV FCOV utilities"
+
+filesets:
+  files_fcov:
+    files:
+      - dv_fcov_macros.svh: {is_include_file: true}
+    file_type: systemVerilogSource
+
+targets:
+  default:
+    filesets:
+      - files_fcov
--- a/dv/uvm/core_ibex/ibex_dv.f
+++ b/dv/uvm/core_ibex/ibex_dv.f
@@ -65,6 +65,7 @@ ${PRJ_DIR}/vendor/google_riscv-dv/src/riscv_signature_pkg.sv
 +incdir+${PRJ_DIR}/dv/uvm/core_ibex/tests
 +incdir+${PRJ_DIR}/dv/uvm/core_ibex/common/ibex_mem_intf_agent
 +incdir+${PRJ_DIR}/dv/uvm/core_ibex/common/irq_agent
++incdir+${PRJ_DIR}/dv/fcov
 +incdir+${PRJ_DIR}/vendor/lowrisc_ip/dv/sv/mem_model
 +incdir+${PRJ_DIR}/vendor/lowrisc_ip/dv/sv/dv_utils
 +incdir+${PRJ_DIR}/vendor/lowrisc_ip/dv/sv/str_utils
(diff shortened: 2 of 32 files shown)


rswarbrick (approved):
Looks sensible enough. I still really dislike the macros that introduce un-greppable signal names, but I think I've lost that argument.

dv/uvm/core_ibex/ibex_dv.f
```diff
@@ -7,6 +7,7 @@
 +define+BOOT_ADDR=2147483648 // 32'h8000_0000
 +define+TRACE_EXECUTION
 +define+RVFI
++define+DV_FCOV
```
sriyerg, line 10: What is the need for adding this AND the plusarg? In general runtime controls should be sufficient.

dv/fcov/dv_fcov.core
```diff
@@ -0,0 +1,17 @@
...
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:dv:dv_fcov"
+description: "DV FCOV utilities"
+
+filesets:
+  files_fcov:
+    files:
+      - dv_fcov.svh: {is_include_file: true}
```
sriyerg, line 11: Generally suffix files that only contain macros: example: `dv_fcov_macros.svh`.

dv/uvm/core_ibex/ibex_dv.f
GregAC: It's for the FCOV macros that are used in design files. If it's not defined then the those macros don't produce anything. This is useful in things like synthesis runs or formal verification or just plain simulation where you don't want the extra fcov_ signals cluttering things up.

sriyerg: Why not protect them with `` `ifdef SYNTHESIS`` instead?

GregAC: As I said synthesis is one reason you'd want to exclude them, but there may be others. It seemed reasonable to use a new define to control whether or not they are included. It's similar to the `INC_ASSERT` define.

sriyerg: `INC_ASSERT` does not need to be passed externally. It is defined based on the toolflow. 
https://github.com/lowRISC/opentitan/blob/577bcb8364eb792edad5d096868476f7054a8249/hw/ip/prim/rtl/prim_assert.sv#L72

Perhaps we could follow the same approach here and define `INC_FCOV` inline also based on the toolflow? I prefer to keep the list of macros that need to be defined to enable / disable features to a minimum.

GregAC: I've updated things to use an `EXCLUDE_FCOV` define instead. So the FCOV stuff is in by default and we've got some tool specific `ifdef`s in the header to exclude it for synthesis.

I did it this way round as the situations you might not want it could be quite specific so preferred an option for people to remove it as part of setting up options for their flow rather than having to make modifications to dv_fcov_macros.svh.

dv/fcov/dv_fcov.core
GregAC: I've renamed it to `dv_fcov_macros.svh`

PR #1240, opened 2021-01-18
Add misa b extension

towoe:
I am not 100% sure if this should be done as the B extension bit is still "Tentatively reserved for Bit-Manipulation extension".

As I made changes to the testbench I checked it with Verilator and added some minor changes to it as well into this PR.

### DIFF
--- a/dv/cs_registers/tb_cs_registers.core
+++ b/dv/cs_registers/tb_cs_registers.core
@@ -124,3 +124,4 @@ targets:
           - '-CFLAGS "-std=c++14 -Wall -DTOPLEVEL_NAME=tb_cs_registers -DVM_TRACE_FMT_FST -g"'
           - '-LDFLAGS "-pthread -lutil -lelf"'
           - "-Wall"
+          - '-Wno-fatal' # Do not fail on (style) issues, only warn about them.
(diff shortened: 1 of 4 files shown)


dv/cs_registers/tb_cs_registers.core
```diff
@@ -124,3 +124,5 @@ targets:
           - '-CFLAGS "-std=c++14 -Wall -DTOPLEVEL_NAME=tb_cs_registers -DVM_TRACE_FMT_FST -g"'
           - '-LDFLAGS "-pthread -lutil -lelf"'
           - "-Wall"
+          # Only a subset is test, print messages but do not fail with unsed warnings.
+          - '-Wno-fatal'
```
imphil, line 128: How about only disabling the unused warnings (`-Wno-UNUSED`)?

imphil (approved):
LGTM. Even though the bit is only "tentatively reserved" I find it very unlikely that this bit would ever change to something other than the bitmanip extension -- which itself will evolve, of course, but that's independent of the MISA flag.

dv/cs_registers/tb_cs_registers.core
towoe: Because then the warning is not printed. I thought it would be better to see the warning and build it.
The "optimal" would be to disable the warning from being an error and still get the message.

imphil: Ah I see. In that case let's slightly reword the comment to "Do not fail on (style) issues, only warn about them."

towoe: Thanks, now reading the comment again I see that it is not only vague but also has an error.

PR #1243, opened 2021-01-19
[rtl] Fix instruction encodings in tracer package

vogelpi:
This PR contains fixes overlapping instruction encodings for the bitmanip extension in the tracer package.

@micprog could you please check that this PR resolves the remaining warning around FSRI?


vogelpi:
@ganoam since you did the entire bitmanip implementation, it would be good to get your opinion. Does this look right to you?

micprog:
@vogelpi Yes, the warning regarding FSRI is now fixed with these changes, thanks for the update!

ganoam:
Hi @vogelpi, 
Yes, this makes sense. In fact, I think the immediate field for all for all of these shift- and single bit instructions is specified to be one bit too wide. Even for RV64, bit 26 should never be used.

vogelpi:
Okay, thanks both for your feedback.

rswarbrick (approved):
Looks good to me!

vogelpi:
Thanks @rswarbrick !

PR #1247, opened 2021-01-22
[rtl] Fix lint "fix" with DbgTriggerEn

rswarbrick:
The previous change was wrong: it was trying to define a signal with
`DbgHwNumLen` bits that contained `DbgHwBreakNum - 1`. Unfortunately, `'1`
is *not* the same as a zero-extended version of `1'b1`.


GregAC (approved):
Well spotted! I probably should have noticed this one in my previous review.

rswarbrick:
Well, it was actually spotted by a failing CI test in an OpenTitan PR where I'd made the same stupid mistake :-)

Issue #1233, opened 2021-01-11
Add RISC-V Compliance v0.2 check


towoe:
RISC-V defines a [Compliance Framework](https://riscv.org/technical/specifications/) which is available in v0.2 as [RISCOF](https://gitlab.com/incoresemi/riscof).
It is still marked as beta but is now at a point where it can be utilised.
It allows for a more flexible way to check for compliance.

I hacked together a first test which runs some checks successfully. This is based on the current dv/riscv-compliance work as it will also compare the generated signatures.
It sill needs some work for an initial version.
This issue should keep track of the open points and questions.

- [ ] Implement Ibex model plugin according to [riscof-plugins](https://gitlab.com/incoresemi/riscof-plugins)
  - [ ] Make it variable so different configurations of Ibex can be checked
- [ ] Can / should [ibex_config.py](https://github.com/lowRISC/ibex/blob/master/util/ibex_config.py) be extended to generate a YAML specification which is needed by [riscv-config](https://github.com/riscv/riscv-config) which serves as an input checker for RISCOF
- [ ] Add it to CI?
- [ ] Choose the reference model to which Ibex is compared
- [ ] Document the usage
- [ ] Remove [riscv-compliance](https://github.com/riscv/riscv-compliance)? I am not yet sure how RISCOF is integrated into riscv-compliance as mentioned in https://github.com/riscv/riscv-compliance#issuecomment-747524231. Maybe it is only a way to run the test created in the new framework?

Documentation for reference [RISCOF](https://riscof.readthedocs.io/), [riscv-config](https://riscv-config.readthedocs.io), [riscv-isac](https://riscv-isac.readthedocs.io/)

imphil:
>  Add it to CI?

Absolutely.

PR #1251, opened 2021-01-25
Add support for RISCOF based checks

towoe:
This PR addresses #1233 and provides some basic functionality to run riscv-compliance checks based on RISCOF.
RISCOF is still in development and the switch to it in riscv-compliance is only partially completed.

Add a Python model of Ibex to integrate with RISCOF.
Add a pseudo model to compare the output of Ibex to reference signatures.
Run the check in CI, it is intended not to fail even though some tests will not pass. The same specification is used for different configurations of Ibex, this needs to be extended in the future. (Provide more `ibex_*_isa.yaml` files.)


towoe:
Anyone know what would be the correct solution to `ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.`? Uninstall it, use `pip --ignore-installed`?

rswarbrick:
For the pyYAML problem, I think you probably just want to use the distro version. See PR #1226.

towoe:
Sorry, I should have explained it in more detail. pyYAML is pulled in as a dependency from `riscof` which I specified. How do I say that the distro version should be used in this case?

rswarbrick:
Oh, I see. I guess `--ignore-installed` makes sense (but possibly with a comment explaining what's going on?)

towoe:
```
-I, --ignore-installed
              Ignore  the installed packages, overwriting them. This can break your system if the existing package is of a different version or was installed with
              a different package manager!
```

reading the documentation more carefully, maybe the safer bet would be remove the local installation again?
But I don't really understand why pyYAML is even needed/updated if it is already available?

towoe:
Okay, I just saw in the log that pyYAML 3.12 is available, that's from 2016. I did not directly find which package requires what, but I would guess somewhere a newer version is required. Soo, remove system installation? Maybe @imphil has an opinion here?

imphil:
Yes, we explicitly installed the distribution's version of pyyaml as the PyPi-distributed version didn't include the libyaml (C) library bindings, but only the pure-Python version, which is significantly slower. Debian 18.04 gives us pyyaml 3.x.

The requirement for a newer pyyaml version comes from riscv-config (https://github.com/riscv/riscv-config/blob/master/setup.py#L38). It's hard to tell where the exact version requirement came from looking at the relevant commit (https://github.com/riscv/riscv-config/commit/4617b312ecd5182ce20e085f27a25cd252ed50f8).

But at the same time I just saw that pyyaml is now installing wheels for Linux which include the libyaml bindings, so we actually don't need to use the distribution package any more.

You can therefore remove the python3-yaml distribution package from the list of packages, and pip should pick up a compatible version from PyPi.

towoe:
Thanks a lot Philipp for your investigation!

towoe:
Should pip install update the packages again? `pip install -U` or was there another reason why it was removed?

imphil:
The update flag shouldn't be needed, we're happy with any version satisfying the requirements.

vogelpi:
Hey @towoe , @imphil , @rswarbrick ,
more than 7 months have passed since the last activity in this PR. Do you know what the status of this PR is? Are there open points to be addressed, things that need to be discussed?

towoe:
Hi @vogelpi,
as far as I can remember the new version of the compliance framework was not ready when I created the PR.
This PR shows how the new compliance checks could be realized and should make it easy to rebase or "re-implement" it.
I wasn't sure if this PR is wanted, but if there is a need I could at some point revisit this topic.

PR #1255, opened 2021-01-25
[syn] Add missing include path

GregAC:
With the introduction of dv_fcov_macros.svh we need to add it as an
included path for sv2v.


rswarbrick (approved):
Ahah! Thanks :-)

PR #1256, opened 2021-01-26
[rtl] Remove paths between dmem and imem signals

GregAC:
Prior to this change Ibex had multiple feedthrough paths from the data
memory interface to the instruction memory interface. This existed
because Ibex would hold off doing a instruction fetch for a jump or
branch if there was a outstanding memory request. It would wait for the
response to be available so either the jump or branch would occur or an
exception was taken.

With this change the branch or jump will speculatively begin the
instruction fetch whilst there is an outstanding memory request. Should
an exception result from the memory request the fetch will be discarded
and the exception taken as normal.

An alternative fix would not factor the data error response
(data_err_i) directly into the controller logic for branches and jumps.
With this option new stall cycles would be introduced anywhere a branch
or jump immediately follows a memory instruction which would have an
adverse impact on performance.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -142,8 +142,9 @@ module ibex_controller #(
   logic illegal_dret;
   logic illegal_umode;
   logic exc_req_lsu;
-  logic special_req_all;
-  logic special_req_branch;
+  logic special_req;
+  logic special_req_pc_change;
+  logic special_req_flush_only;
   logic enter_debug_mode_d;
   logic enter_debug_mode_q;
   logic ebreak_into_debug;
@@ -219,29 +220,18 @@ module ibex_controller #(
 
 
   // special requests: special instructions, pipeline flushes, exceptions...
+  // All terms in these expressions are qualified by instr_valid_i except exc_req_lsu which can come
+  // from the Writeback stage with no instr_valid_i from the ID stage
 
-  // To avoid creating a path from data_err_i -> instr_req_o and to help timing the below
-  // special_req_all has a version that only applies to branches. For a branch the controller needs
-  // to set pc_set_o but only if there is no special request. If the generic special_req_all signal
-  // is used then a variety of signals that will never cause a special request during a branch
-  // instruction end up factored into pc_set_o. The special_req_branch only considers the special
-  // request reasons that are relevant to a branch.
+  // These special requests only cause a pipeline flush and in particular don't cause a PC change
+  // that is outside the normal execution flow
+  assign special_req_flush_only = wfi_insn | csr_pipe_flush;
+
+  // These special requests cause a change in PC
+  assign special_req_pc_change = mret_insn | dret_insn | exc_req_d | exc_req_lsu;
 
   // generic special request signal, applies to all instructions
-  // All terms in this expression are qualified by instr_valid_i except exc_req_lsu which can come
-  // from the Writeback stage with no instr_valid_i from the ID stage
-  assign special_req_all = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
-      exc_req_d | exc_req_lsu;
-
-  // special request that can specifically occur during branch instructions
-  // All terms in this expression are qualified by instr_valid_i
-  assign special_req_branch = instr_fetch_err & (ctrl_fsm_cs != FLUSH);
-
-  `ASSERT(SpecialReqBranchGivesSpecialReqAll,
-    special_req_branch |-> special_req_all)
-
-  `ASSERT(SpecialReqAllGivesSpecialReqBranchIfBranchInst,
-    special_req_all && (branch_set_i || jump_set_i) |-> special_req_branch)
+  assign special_req = special_req_pc_change | special_req_flush_only;
 
   // Exception/fault prioritisation is taken from Table 3.7 of Priviledged Spec v1.11
   if (WritebackStage) begin : g_wb_exceptions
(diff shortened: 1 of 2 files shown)


rtl/ibex_controller.sv
```diff
@@ -219,29 +218,12 @@ module ibex_controller #(
...
-  // to set pc_set_o but only if there is no special request. If the generic special_req_all signal
-  // is used then a variety of signals that will never cause a special request during a branch
-  // instruction end up factored into pc_set_o. The special_req_branch only considers the special
-  // request reasons that are relevant to a branch.
-
   // generic special request signal, applies to all instructions
   // All terms in this expression are qualified by instr_valid_i except exc_req_lsu which can come
   // from the Writeback stage with no instr_valid_i from the ID stage
-  assign special_req_all = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
+  assign special_req = mret_insn | dret_insn | wfi_insn | csr_pipe_flush |
```
GregAC, line 224: The `special_req_all`/`special_req_branch` separation was the first attempt at cutting the dmem -> imem feedthrough. It works in the two stage design but a different approach is required with the writeback stage.

Now we're doing speculative fetches for branch or jump when an exception might override them we just don't factor `special_req` in the `pc_set_o` logic in the controller at all. The existing `special_req_branch` only considered instruction fetch errors anyway and those will prevent the branch_set/jump_set signals from being set at all in `ibex_id_stage` (`instr_executing_spec` will not be set).

rtl/ibex_controller.sv
```diff
@@ -852,4 +832,38 @@ module ibex_controller #(
...
 
+  `ifdef INC_ASSERT
+    logic instr_seen_special_req;
+    logic instr_doing_special_req;
+    logic instr_seen_csr_pipe_flush;
+
+    // csr_pipe_flush is the only special_req that doesn't require a pc_set_o at the end to ensure
+    // correct behaviour so exclude it from instr_doing_special_req here to prevent checking for
+    // this.
+    assign instr_doing_special_req = special_req & ~csr_pipe_flush;
```
GregAC, line 843: Current behaviour on `csr_pipe_flush` looks a little odd, I'm planning to investigate this some more.

GregAC:
I've run a full DV regression on this and all looks good.

I need to do some further work with the synthesis flow to get a full picture of all remaining feedthrough paths (OpenSTA should be reporting all in2out paths but seems to be missing some for reasons I haven't yet understood, it looks like it might just be omitting paths that are included in longer ones). I think this fixes the most major ones.

GregAC:
Telling a lie, I ran a full dv regression on a earlier version that retained `special_req_branch` and that all looked fine. I ran a restricted regression targetting imem errors against this and that reported no failure. I will run a full regression against this tonight.

rtl/ibex_controller.sv
```diff
@@ -852,4 +832,38 @@ module ibex_controller #(
...
+          (instr_seen_special_req | instr_doing_special_req) & ~instr_valid_clear_o;
+        instr_seen_csr_pipe_flush <= (instr_seen_csr_pipe_flush | csr_pipe_flush) & ~instr_valid_clear_o;
+      end
+    end
+
+    // If there's been a special request during execution when the instruction is leaving ID Ibex
+    // must do some kind of jump to an exception handler, this cannot be a normal jump (as used by
+    // branch and jump instructions).
+    `ASSERT(IbexSetExceptionPCOnSpecialReq,
+      instr_doing_special_req & instr_valid_clear_o |-> pc_set_o & (pc_mux_o != PC_JUMP));
```
imphil, line 860: No semicolon after ASSERT (here and below) :-)

rtl/ibex_id_stage.sv
```diff
@@ -874,14 +899,32 @@ module ibex_id_stage #(
...
+                                  ~instr_fetch_err_i &
+                                  controller_run     &
+                                  ~stall_ld_hz;
+
     assign instr_executing = instr_valid_i              &
                              ~instr_kill                &
                              ~stall_ld_hz               &
                              ~outstanding_memory_access;
 
+    `ASSERT(IbexExecutingSpecIfExecuting, instr_executing |-> instr_executing_spec);
```
tomroberts-lowrisc, line 920: Stray semicolon

rtl/ibex_id_stage.sv
```diff
@@ -874,14 +899,32 @@ module ibex_id_stage #(
...
+
     `ASSERT(IbexStallIfValidInstrNotExecuting,
       instr_valid_i & ~instr_kill & ~instr_executing |-> stall_id)
 
+    `ASSERT(IbexCannotRetireWithPendingExceptions,
+      instr_done |-> ~(wb_exception | outstanding_memory_access));
```
tomroberts-lowrisc, line 926: Stray semicolon

rtl/ibex_controller.sv
```diff
@@ -853,4 +841,72 @@ module ibex_controller #(
...
+    end
+
+    // Once an exception request has been accepted it must be handled before controller goes back to
+    // DECODE
+    `ASSERT(IbexNoDoubleExceptionReq, exception_req_accepted |-> ctrl_fsm_cs != DECODE)
+
+    // Only signal ready, allowing a new instruction into ID, if there is no exception request
+    // pending or it is done this cycle.
+    `ASSERT(IbexDontSkipExceptionReq,
+      id_in_ready_o |-> !exception_req_pending || exception_req_done);
```
tomroberts-lowrisc, line 897: Stray semicolon

tomroberts-lowrisc (approved):
This looks like a good way to go, thanks - just a few semicolons to fix!

GregAC:
@tomroberts-lowrisc just made a minor update. Found some dv errors and needed to alter one of the assertions, `IbexNoDoubleSpecialReqPCSet`

tomroberts-lowrisc:
LGTM

PR #1259, opened 2021-01-27
[dv] Wire up alerts to dut probe interface

rswarbrick:
This silences VCS warnings about the DUT having missing port
connections. It doesn't add any actual testing for these signals.

The patch also re-orders the signals to match the order in
`ibex_core_tracing.sv`, to make it easier to spot what's going on by
eye.


udinator (approved):
LGTM

PR #1262, opened 2021-01-27
[dv] Be explicit about the target priv_mode in wait_ret test

rswarbrick:
The previous code contained

    wait (dut_vif.dut_cb.priv_mode === select_mode())

and VCS warns that this `wait` block will only trigger on changes to
explicit arguments. That is, if the `in_nested_trap` field changes, so
the return value of the `select_mode()` method would change to match
priv_mode, the `wait` statement won't finish.

This patch explicitly stores a snapshot of the value of `select_mode()`
just before the wait line. I think this is the intended behaviour, and
will no longer trigger warnings from VCS.

@udinator: Have I understood this code correctly? This change isn't right if we expect `in_nested_trap` to change "behind our backs".

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -328,6 +328,7 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
     run.raise_objection(this);
     fork
       begin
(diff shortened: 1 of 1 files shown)


udinator (approved):
LGTM - this change shouldn't mess with expected behavior, as `in_nested_trap` should never get set without us knowing about it.

PR #1264, opened 2021-01-29
[dv] Improve sim.py error reporting

GregAC:
This fixes up a few frustrations I've had with the way the regression reporting works. We could clean things up further and improve the python but we're a bit constrained by the way `compare_trace_csv` works from RISCV DV. It outputs to a named logfile and directly reports [PASSED] or [FAILED] into the log so we can't alter how that works nor easily redirect its output somewhere unless we use a temp file for it to write to and we read back from it. I think we're better off changing `compare_trace_csv` in RISCV DV to be more flexible for reporting or reimplementing it on our side.

* Handle missing log files with error messages rather than terminating
  on unhandled exceptions
* Output potential failure causes from sim log file into regression log
* Alter per test output to make it clearer what line corresponds to what
  test
* Only output [PASSED] or [FAILED] a single time per test
* Don't output [PASSED] where sim log is good but ISS comparison is not

### DIFF
--- a/dv/uvm/core_ibex/sim.py
+++ b/dv/uvm/core_ibex/sim.py
@@ -368,6 +368,9 @@ def compare_test_run(test, idx, iss, output_dir, report):
     logging.info("Comparing %s/DUT sim result : %s" % (iss, elf))
 
     with open(report, 'a') as report_fd:
+        test_name_idx = '{}.{}'.format(test_name, idx)
+        test_underline = '-' * len(test_name_idx)
+        report_fd.write('\n{}\n{}\n'.format(test_name_idx, test_underline))
         report_fd.write('Test binary: {}\n'.format(elf))
 
     rtl_dir = os.path.join(output_dir, 'rtl_sim',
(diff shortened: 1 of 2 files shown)


dv/uvm/core_ibex/sim.py
```diff
@@ -380,33 +383,48 @@ def compare_test_run(test, idx, iss, output_dir, report):
...
+                report_fd.write(line)
+            report_fd.write('[FAILED]: sim error seen\n')
+
+            return False
 
-    if no_post_compare:
-        return True
+        if no_post_compare:
+            report_fd.write('[PASSED]\n')
+            return True
```
rswarbrick, line 409: Nit: I don't think this needs to be in the `with` block.

rswarbrick (approved):
Much better, thanks!

dv/uvm/core_ibex/sim.py
GregAC: It needs access to `report_fd` that the `with` opens. Unless you very quick off the mark and saw my initial version that needlessly nested a second `with` here?

rswarbrick: Ah, no. I'm just being stupid. Looks good to me!

udinator (approved):
lgtm!

PR #1265, opened 2021-02-01
Don't automatically generate disassembly in the example Makefile

### LINKED ISSUE #1263: Doubt in .dis file generated under examples/sw/
Hi,
I have a doubt in the .dis file generated in the example/sw/simple_system.
In the .dis file generated there above every pc count I am seeing one more instruction, how is it generated and what is the purpose of it?

IN the below snapshot if you see  the bld instructions. on top of some instructions its not getting generated (1006e0 & 1006e4)

  **addi x10, x0, 0**
  1006d8:	4501                	li	a0,0
  
  **addi x11, x0, 0**
  1006da:	4581                	li	a1,0
  
  **jal x1, main**
  1006dc:	00a030ef          	jal	ra,1036e6 <main>
  
  **li x5, SIM_CTRL_BASE + SIM_CTRL_CTRL**
  1006e0:	000202b7          	lui	t0,0x20
  1006e4:	02a1                	addi	t0,t0,8
  
  **li x6, 1**
  1006e6:	4305                	li	t1,1

  **sw x6, 0(x5)**
  1006e8:	0062a023          	sw	t1,0(t0) # 20000 <_stack_len+0x18000>

image

rswarbrick:
Anyone who needs to disassemble their generated ELF can probably just
call objdump directly and the precise set of flags have already
confused at least one potential contributor[1].

We're keeping the canned objdump command for "engineers that know
where to look" because some have said they find it useful.

[1] https://github.com/lowRISC/ibex/issues/1263

Fixes #1263 (marking as a fix because the question becomes moot!)

### DIFF
--- a/examples/sw/simple_system/common/common.mk
+++ b/examples/sw/simple_system/common/common.mk
@@ -33,15 +33,20 @@ OBJS := ${C_SRCS:.c=.o} ${ASM_SRCS:.S=.o} ${CRT:.S=.o}
 DEPS = $(OBJS:%.o=%.d)
 
 ifdef PROGRAM
-OUTFILES := $(PROGRAM).elf $(PROGRAM).vmem $(PROGRAM).bin $(PROGRAM).dis
+OUTFILES := $(PROGRAM).elf $(PROGRAM).vmem $(PROGRAM).bin
 else
 OUTFILES := $(OBJS)
 endif
 
 all: $(OUTFILES)
 
+ifdef PROGRAM
 $(PROGRAM).elf: $(OBJS) $(LINKER_SCRIPT)
 	$(CC) $(CFLAGS) -T $(LINKER_SCRIPT) $(OBJS) -o $@ $(LIBS)
+
+.PHONY: disassemble
+disassemble: $(PROGRAM).dis
+endif
 
 %.dis: %.elf
 	$(OBJDUMP) -fhSD $^ > $@


GregAC:
I do find these a useful convenience so have a slight preference to keep them in.

rswarbrick:
In which case, can you suggest an improvement to the documentation that explains our choice of objdump flags?

GregAC:
Will do, doesn't look like objdump has a way to only include .c source, but does have a way to prefix source lines with a given string, I'll experiment with that and see if it gives something more readable where you get the confusing mix of .s source and actual disassembly.

rswarbrick:
I still think that it's not worth spending time on it. Someone who needs to read disassembly can *surely* drive objdump by now...

towoe:
At some time in the past I found it helpful to have the disassembly generated automatically.
I think it helps people who are not familiar with a project or the tooling. But I guess a good documentation would be equally good.

If we keep objdump maybe we should move from `-SD` to just `-d`?

rswarbrick:
Hi @GregAC! Could you take a quick look at this? I think the updated patch does what we agreed in the meeting.

GregAC:
I was thinking more along the lines of a flag that could be used with make to have the disassembly done along with the build e.g. `make DISASSEMBLE_SW=1` or we could have a specific make target `make disassemble` that would do the disassembly.

rswarbrick:
I personally think that we should expect a CPU designer to be able to run binutils. But, yep, I can add a phone disassemble target.

GregAC (approved):
Thanks for sorting this out

PR #1266, opened 2021-02-02
Dv fixes

GregAC:
Fixing up a variety of small issues I came across when looking into DV errors on my feedthrough PR


rswarbrick:
I don't know enough about our debug tests to have an opinion about the first patch, but the rest looks good. However, please could you squash the last patch with the vendoring one to avoid a broken commit?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -655,12 +655,19 @@ class core_ibex_interrupt_instr_test extends core_ibex_directed_test;
     vseq.irq_raise_single_seq_h.max_delay = 0;
     vseq.irq_raise_single_seq_h.max_interval = 0;
     forever begin
-      // hold until we see a valid instruction in the ID stage of the pipeline.
-      wait (instr_vif.instr_cb.valid_id && !(instr_vif.instr_cb.err_id || dut_vif.illegal_instr));
+      // hold until we see a valid instruction in the ID stage of the pipeline or the core goes to
+      // sleep
+      wait ((instr_vif.instr_cb.valid_id && !(instr_vif.instr_cb.err_id || dut_vif.illegal_instr))
+        || dut_vif.core_sleep);
```
GregAC, line 661: I'm not hugely satisfied with this fix but seemed like the most pragmatic way to make things work without wider changes.

I consider doing the sleep check at the very beginning of the function, outside of the forever loop but there's the issue WFI takes a few cycles to enter. If we've just executed a WFI instruction we won't yet see `core_sleep` asserted but may not see a valid instruction either so we'll have the same missed wake-up issue just with a smaller chance it will occur.

GregAC:
> However, please could you squash the last patch with the vendoring one to avoid a broken commit?

Can do, I didn't initially as I thought we didn't like squashing things into vendoring commits but now recall it's been acceptable to do so to avoid broken commits in the past.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: I think I'm fine with this change, like you said I think it's the easiest way to avoid having to overhaul some of these functions.

udinator (approved):
lgtm overall, thanks!

PR #1269, opened 2021-02-09
Update README to match design

imphil:
Since this part of the README was written the design moved on. Let's
update it. This update follows the text at https://ibex-core.readthedocs.io/en/latest/index.html.

I also removed the block diagram, since it is misleadingly showing only
the 2 stage design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,21 @@
 
 # Ibex RISC-V Core
 
-Ibex is a small and efficient, 32-bit, in-order RISC-V core with a 2-stage pipeline that implements
-the RV32IMC instruction set architecture.
+Ibex is a production-quality open source 32-bit RISC-V CPU core written in
+SystemVerilog. The CPU core is heavily parametrizable and well suited for
+embedded control applications. Ibex is being extensively verified and has
+seen multiple tape-outs. Ibex supports the Integer (I) or Embedded (E),
+Integer Multiplication and Division (M), Compressed (C), and B (Bit
+Manipulation) extensions.
 
+The block diagram below shows the *small* parametrization with a 2-stage
+pipeline.
 <p align="center"><img src="doc/03_reference/images/blockdiagram.svg" width="650"></p>
 
-This core was initially developed as part of the [PULP platform](https://www.pulp-platform.org)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,16 +2,17 @@
 
 # Ibex RISC-V Core
 
-Ibex is a small and efficient, 32-bit, in-order RISC-V core with a 2-stage pipeline that implements
-the RV32IMC instruction set architecture.
-
-<p align="center"><img src="doc/03_reference/images/blockdiagram.svg" width="650"></p>
-
-This core was initially developed as part of the [PULP platform](https://www.pulp-platform.org)
+Ibex is a production-quality open source 32 bit RISC-V CPU core written in
```
vogelpi, line 5: 32 bit -> 32-bit?

README.md
```diff
@@ -2,16 +2,17 @@
 
 # Ibex RISC-V Core
 
-Ibex is a small and efficient, 32-bit, in-order RISC-V core with a 2-stage pipeline that implements
-the RV32IMC instruction set architecture.
-
-<p align="center"><img src="doc/03_reference/images/blockdiagram.svg" width="650"></p>
```
vogelpi, line 8: I would not remove the block diagram because it's the only figure we have in the readme and I would argue that in the meantime it has gained quite some recognition value. It may not accurately represent the configuration with the writeback stage/BT ALU. But if we wanted to accurately represent this configuration, how would this new block diagram look like? On register in the line between LSU and RF?

vogelpi: If we want to be really accurate we can also put a caption and state that this figure shows the "small" configuration (according to the verification status still the most important one).

vogelpi (commented):
Thanks @imphil for the PR. The changes look good but I wouldn't remove the block diagram. Without a block diagram the readme/GitHub page becomes less attractive and recognizable. I would rather adapt the block diagram or state in a caption that the figure shows the small config.

imphil:
Uh, I totally forgot about this PR. I added the block diagram back, PTAL.

vogelpi (approved):
LGTM, thanks @imphil !

PR #1270, opened 2021-02-11
[rtl] Debug mode controller changes

GregAC:
* `if` in `DBG_TAKEN_IF` is needless as the conditions it checks will be
  true if controller enters `DBG_TAKEN_IF` state

* flop `enter_debug_mode` so `FLUSH` state looks at what
  `enter_debug_mode` was when it was seen in `DECODE` state rather than
  what it has become. In particular the controller could enter `FLUSH`
  on the basis of performing a WFI then divert down the debug control
  path due to a new debug request being raised. In this instance it is
  preferable for the WFI to complete entering `SLEEP` before the debug
  request wakes the core back up.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -144,7 +144,8 @@ module ibex_controller #(
   logic exc_req_lsu;
   logic special_req_all;
   logic special_req_branch;
-  logic enter_debug_mode;
+  logic enter_debug_mode_d;
+  logic enter_debug_mode_q;
   logic ebreak_into_debug;
   logic handle_irq;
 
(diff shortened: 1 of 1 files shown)


rtl/ibex_controller.sv
```diff
@@ -317,8 +318,8 @@ module ibex_controller #(
   // due to a recently flushed IF (or a delay in an instruction returning from
   // memory) before it has had anything to single step.
   // Also enter debug mode on a trigger match (hardware breakpoint)
-  assign enter_debug_mode = (debug_req_i | (debug_single_step_i & instr_valid_i) |
-                             trigger_match_i) & ~debug_mode_q;
+  assign enter_debug_mode_d = (debug_req_i | (debug_single_step_i & instr_valid_i) |
+                              trigger_match_i) & ~debug_mode_q;
```
tomroberts-lowrisc, line 322: nit: this wants one more space since it's inside the bracket above

tomroberts-lowrisc (approved):
LGTM

PR #1271, opened 2021-02-11
[dv] Reorder checks in sim.py

GregAC:
The UVM log should be checked for failures before attempting to process
the core trace log. A simulation failure could mean the trace log
doesn't exist and is is preferable to report the simulation error from
the log rather than trace not found as a failure cause.

### DIFF
--- a/dv/uvm/core_ibex/sim.py
+++ b/dv/uvm/core_ibex/sim.py
@@ -376,22 +376,10 @@ def compare_test_run(test, idx, iss, output_dir, report):
     rtl_dir = os.path.join(output_dir, 'rtl_sim',
                            '{}.{}'.format(test_name, idx))
 
-    rtl_log = os.path.join(rtl_dir, 'trace_core_00000000.log')
-    rtl_csv = os.path.join(rtl_dir, 'trace_core_00000000.csv')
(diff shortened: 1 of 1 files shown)


udinator (approved):

rswarbrick (approved):
Much better, thanks!

dv/uvm/core_ibex/sim.py
```diff
@@ -404,9 +392,25 @@ def compare_test_run(test, idx, iss, output_dir, report):
...
+    try:
+        # Convert the RTL log file to a trace CSV.
+        process_ibex_sim_log(rtl_log, rtl_csv, 1)
+    except (OSError, RuntimeError) as e:
+        with open(report, 'a') as report_fd:
+            report_fd.write('[FAILED]: Log processing failed: {}\n'.format(e))
+
+        return False
+
+    no_post_compare = test.get('no_post_compare')
```
rswarbrick, line 407: Nit: It doesn't really matter, but I'd be inclined to do this as `test.get('no_post_compare', False)` to make it clear that this is always supposed to be a boolean. And maybe assert `isinstance(no_post_compare, bool)` too.

PR #1276, opened 2021-02-18
Fix deprecated sphinx html_context usage in conf.py

imphil:
We were using the old html_context which has been deprecated
for a while. This PR switches to html_css_files instead.
See sphinx-doc/sphinx#8885 for more information.


GregAC (approved):
Thanks for fixing

imphil:
Other CI is failing for unrelated reasons.

PR #1287, opened 2021-02-26
[dv] Fix bug in sim.py and type in testlist

GregAC:


dv/uvm/core_ibex/sim.py
```diff
@@ -410,7 +410,9 @@ def compare_test_run(test, idx, iss, output_dir, report):
     # no_post_compare skips the final ISS v RTL log check, so if we've reached
     # here we're done when no_post_compare is set.
     if no_post_compare:
-        report_fd.write('[PASSED]\n')
+        with open(report, 'a') as report_fd:
+            report_fd.write('[PASSED]\n')
+
```
rswarbrick, line 415: I've just looked at the surrounding code. Is there any need to keep opening and closing the file? Can I suggest instead pulling the `with open(report, 'a') as report_fd` out to the `compare` function?

rswarbrick (approved):

PR #1290, opened 2021-03-01
[fcov] Move fcov files out of dv tree

tomroberts-lowrisc:
This is required to vendor Ibex into OpenTitan since these files are a
dependency of the design, but everything under dv/ is stripped when
vendoring.


tomroberts-lowrisc:
Couldn't go in shared/ as that is also excluded when vendoring!

GregAC (approved):

PR #1291, opened 2021-03-01
[dv/ibex] update how coverage is merged

udinator:
Currently, the `cov` step in the DV Makefile will only merge coverage
databases emitted directly from Ibex simulations, and will not pick up
any coverage databases generated by the RISCV-DV functional coverage
flow.

This PR updates the `gen_cov()` function in `sim.py` to recursively
search for any generated coverage directories and then merges them all.

Resultant coverage reports include all code coverage, Ibex functional
coverage, and RISCV-DV functional coverage.

The coverage-related targets in the Makefile have also been renamed to
improve clarity.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -92,7 +92,7 @@ all: sim
 
 instr: iss_sim
 
-sim: post_compare cov
+sim: post_compare gen_cov
 
 .PHONY: clean
 clean:
@@ -417,7 +417,8 @@ post_compare: $(OUT-SEED)/regr.log
 ###############################################################################
 # Generate RISCV-DV functional coverage
 # TODO(udi) - add B extension
-fcov:
+.PHONY: riscv_dv_cov
+riscv_dv_fcov:
 	$(verb)python3 ${GEN_DIR}/cov.py \
           --core ibex \
           --dir ${OUT-SEED}/rtl_sim \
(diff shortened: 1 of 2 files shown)


dv/uvm/core_ibex/Makefile
```diff
@@ -426,7 +426,9 @@ fcov:
           --custom_target riscv_dv_extension
 
 # Merge all output coverage directories into the <out>/rtl_sim directory
-cov:
+#
+# Any coverage databases generated from the riscv_dv_fcov target will be merged as well.
+gen_cov:
```
rswarbrick, line 431: Wouldn't it make more sense for `gen_cov` to depend on `riscv_dv_fcov`? Otherwise you have to run `sim` in order to get everything to happen in the right order.

dv/uvm/core_ibex/sim.py
```diff
@@ -478,13 +478,14 @@ def gen_cov(base_dir, simulator, lsf_cmd):
...
-                         (base_dir, entry.name))
-            if os.path.exists(vdb_path):
-                dir_list.append(vdb_path)
+
+    # All generated coverage databases will be named "test.vdb"
+    vdb_dir_name = "test.vdb"
+
+    for path, dirs, files in os.walk(base_dir):
+      if vdb_dir_name in dirs:
+        dir_list.append(os.path.join(path, vdb_dir_name))
```
rswarbrick, line 487: This file uses 4-space indentation elsewhere: can you update this to match? Also, would it make sense to include a `logging.info` call to show the database we're adding? The code that's being replaced (sort of) had that, and it might be handy for working out what's going into the result.

dv/uvm/core_ibex/Makefile
udinator: hm, that's a good point, will update accordingly!

dv/uvm/core_ibex/sim.py
udinator: Sure, I can add that logging call back in.

rswarbrick (approved):
Looks good to me!

PR #1293, opened 2021-03-03
edited ibex_cs_registers.sv

vrozic:
[rtl] Add MCOUNTEREN CSR

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -319,6 +319,12 @@ module ibex_cs_registers #(
         csr_rdata_int[CSR_MFIX_BIT_HIGH:CSR_MFIX_BIT_LOW] = mie_q.irq_fast;
       end
 
+      // mcounteren: machine counter enable
+      CSR_MCOUNTEREN: begin
+        csr_rdata_int = '0;
+        illegal_csr   = ~DbgTriggerEn;
+      end
+
       CSR_MSCRATCH: csr_rdata_int = mscratch_q;
 
       // mtvec: trap-vector base address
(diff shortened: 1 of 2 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -456,6 +456,11 @@ module ibex_cs_registers #(
         csr_rdata_int = '0;
         illegal_csr   = ~DbgTriggerEn;
       end
+      CSR_MCOUNTEREN: begin
+        csr_rdata_int = '0;
+        illegal_csr   = ~DbgTriggerEn;
+      end
```
vogelpi, line 462: This change looks good. I would just move it up in the file a bit. FYI: we try to keep the case statements according to the CSR address. `mcounteren` is at address 0x306. That's just after `mtvec` at 0x305.

vogelpi (commented):
Thanks for your first PR @vrozic , that's great!

One more thing is needed to make this work: You need to define the value for `CSR_MCOUNTEREN` inside `rtl/ibex_pkg.sv` (it's 0x306, you can add the new entry directly below `CSR_MTVEC`). The missing definition is why CI currently fails.

In addition, I recommend to change the commit message to something like
```
[rtl] Add MCOUNTEREN CSR

This commit adds the MCOUNTEREN CSR as required by the RISC-V spec.
The register is defined as WARL. At the moment, Ibex doesn't enable U-mode
access to the performance montiors. Consequently, writes to the register are
ignored and it reads as zero which is okay according to the spec.

This resolves lowRISC/Ibex#1278 .
```
The last line is very useful because it automatically links the PR to the corresponding issue on GitHub and once the PR is merged, the issue is even closed automatically.

vogelpi (approved):
Nice @vrozic , LGTM!

PR #1298, opened 2021-03-05
[dv/ibex] filter out tests on a per-config basis

udinator:
This PR adds functionality to filter out tests during regressions for a
particular config.

e.g. if a full regression is kicked off using the `small` config, we
don't want to attempt to run any PMP and bitmanip tests as the RTL
parameter-set will not support it.

To do this, a new YAML field called `rtl_params` is added to relevant
test entries, to indicate what parameters (if any) are required to be
able to run the particular test, along with the required value of said
parameters.

`sim.py` will then parse this field (if it exists), and using
information from `ibex_configs.yaml` pertaining to the current config,
will remove tests from being run on-the-fly.

This also gives us the convenient side effect of not having to re-run
instruction generation if there is a parameter/config mismatch, we can
just rerun the RTL compilation and simulation stages safely.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -353,6 +353,7 @@ $(OUT-DIR)rtl_sim/.compile.stamp: \
 	$(verb)./sim.py \
 	 --o=$(OUT-DIR) \
 	 --steps=compile \
+	 --ibex_config=$(IBEX_CONFIG) \
 	 ${COMMON_OPTS} \
 	 --simulator="${SIMULATOR}" --simulator_yaml=yaml/rtl_simulation.yaml \
 	 $(cov-arg) $(wave-arg) $(lsf-arg) \
@@ -389,6 +390,7 @@ $(metadata)/rtl_sim.run.stamp: \
 	$(verb)./sim.py \
 	 --o=$(OUT-SEED) \
 	 --steps=sim \
+	 --ibex_config=$(IBEX_CONFIG) \
 	 ${TEST_OPTS} \
 	 --simulator="${SIMULATOR}" --simulator_yaml=yaml/rtl_simulation.yaml \
 	 $(cov-arg) $(wave-arg) $(lsf-arg) \
@@ -407,6 +409,7 @@ $(OUT-SEED)/regr.log: \
 	$(verb)./sim.py \
      --o=$(OUT-SEED) \
      --steps=compare \
+     --ibex_config=$(IBEX_CONFIG) \
      ${TEST_OPTS} \
      --simulator="${SIMULATOR}" \
      --iss="${ISS}"
--- a/dv/uvm/core_ibex/sim.py
+++ b/dv/uvm/core_ibex/sim.py
@@ -31,6 +31,7 @@ _OLD_SYS_PATH = sys.path
 # as it started.
 try:
     sys.path = ([os.path.join(_CORE_IBEX, 'riscv_dv_extension'),
+                 os.path.join(_IBEX_ROOT, 'util'),
                  os.path.join(_RISCV_DV_ROOT, 'scripts')] +
                 sys.path)
 
@@ -44,6 +45,8 @@ try:
     from instr_trace_compare import compare_trace_csv
 
     from ibex_log_to_trace_csv import process_ibex_sim_log, check_ibex_uvm_log
+
+    from ibex_config import parse_config
 
 finally:
     sys.path = _OLD_SYS_PATH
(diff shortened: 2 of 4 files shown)


dv/uvm/core_ibex/Makefile
```diff
@@ -389,6 +390,7 @@ $(metadata)/rtl_sim.run.stamp: \
 	$(verb)./sim.py \
 	 --o=$(OUT-SEED) \
 	 --steps=sim \
+   --ibex_config=$(IBEX_CONFIG) \
```
rswarbrick, line 393: Nit: Indentation seems off here?

dv/uvm/core_ibex/sim.py
```diff
@@ -190,6 +193,74 @@ def get_simulator_cmd(simulator, yaml_path, enables):
             subst_cmd(entry['sim']['cmd'], enables, entry['sim'], env_vars))
 
 
+def filter_tests_by_config(cfg, test_list):
+    '''Filter the provided testlist to ensure all required RTL parameters are
+       provided by the Ibex core configuration.
```
rswarbrick, line 198: Nit: The usual docstring style is to start with a single-line sentence (a bit like a git commit message)

dv/uvm/core_ibex/sim.py
```diff
@@ -190,6 +193,74 @@ def get_simulator_cmd(simulator, yaml_path, enables):
...
+            matching_params = False
+            for p in param_dict.keys():
+                # Throw an error if required RTL parameters in the testlist
+                # have been formatted incorrectly (typos, wrong parameters, etc)
+                if p not in config:
+                    logging.error('parameter % not found in config %' % (p, cfg))
+                    sys.exit(RET_FAIL)
+                else:
+                    config_val = config.get(p)
+                    p_val = param_dict.get(p)
```
rswarbrick, line 232: I'm not a fan of looking things up twice. There's a small performance cost (who cares!) but there's also a slight difficulty when reading: the reader has to look back to convince themselves the result exists. If you know that it does (like you do here), then `config[p]` is probably preferable to `config.get(p)` **but** you could do this more nicely still by iterating over `p, p_val in param_dict.items()` above and then moving the `config_val = config.get(p)` before the check on line 227, then checking whether `config_val is None` or not.

dv/uvm/core_ibex/sim.py
```diff
@@ -190,6 +193,74 @@ def get_simulator_cmd(simulator, yaml_path, enables):
...
+                    # Ibex has some enum parameters, so as a result some tests
+                    # are able to run with several of these parameter values
+                    # (like bitmanipulation tests).
+                    # If this is the case, the testlist will specify all legal
+                    # enum values, check if any of them match the config.
+                    if isinstance(p_val, list):
+                        for val in p_val:
+                            if val == config_val:
+                                matching_params = True
+                                break;
```
rswarbrick, line 243: Couldn't this be `matching_params = config_val in p_val`?

dv/uvm/core_ibex/sim.py
```diff
@@ -190,6 +193,74 @@ def get_simulator_cmd(simulator, yaml_path, enables):
...
+                            function will filter out all tests that require
+                            B-extension and PMP parameters
+    '''
+    filtered_test_list = []
+    config = parse_config(cfg, os.path.join(_IBEX_ROOT, "ibex_configs.yaml"))
+
+    for test in test_list:
+        if "rtl_params" in test:
+            param_dict = test['rtl_params']
+            matching_params = False
```
rswarbrick, line 223: I think I'd consider moving the stuff that sets `matching_params` into a separate function. It'll get rid of some of the indentation and I think the code will become a bit clearer (you're essentially just filtering a list, and that will look much more obvious then, I think).

util/ibex_config.py
```diff
@@ -214,6 +214,46 @@ def get_config_file_location():
     return os.environ.get('IBEX_CONFIG_FILE', _DEFAULT_CONFIG_FILE)
 
 
+def parse_config(config_name, config_filename=get_config_file_location()):
```
rswarbrick, line 217: I think that both call sites set `config_filename`, so I'd suggest removing the default argument.

util/ibex_config.py
```diff
@@ -214,6 +214,46 @@ def get_config_file_location():
...
+    try:
+        config_file = open(config_filename)
+        config_dicts = get_config_dicts(config_file)
+
+        if config_name not in config_dicts:
+            print('ERROR: configuration',
+                  config_name,
+                  'not found in',
+                  config_filename,
+                  file=sys.stderr)
```
rswarbrick, line 241: I know this isn't your code but, if you're touching it anyway, could you turn this into a call to `format` to tidy things up? Something like `print('ERROR configuration {!r} not found in {!r}.'.format(config_name, config_filename), file=sys.stderr)`

rswarbrick (approved):
Some nitty style suggestions, but this looks good!

dv/uvm/core_ibex/Makefile
udinator: ah you're right, should be fixed now.

dv/uvm/core_ibex/sim.py
udinator: fixed, thanks!

dv/uvm/core_ibex/sim.py
udinator: good point! this would remove a lot of the unnecessary indents here, probably not as much need to move the `matching_params` stuff into a separate function now

dv/uvm/core_ibex/sim.py
udinator: sure, that sounds sensible to me, updated!

util/ibex_config.py
udinator: done

util/ibex_config.py
udinator: done

util/ibex_config.py
```diff
@@ -214,6 +214,52 @@ def get_config_file_location():
...
+    Raises a FileNotFoundError if there are errors opening the chosen config file.
+    """
+    try:
+        config_file = open(config_filename)
+        config_dicts = get_config_dicts(config_file)
+
+        if config_name not in config_dicts:
+            print('ERROR: configuration {!r} not found in {!r}.'.format(
+                  config_name, config_filename), file=sys.stderr)
+            #print('ERROR: configuration',
```
GregAC, line 239: Please remove the commented out code

GregAC (approved):
LGTM modulo a couple of tweaks, thanks for sorting it out

PR #1300, opened 2021-03-10
[rtl/icache] Switch ECC granularity to 32bits

tomroberts-lowrisc:
Changes the ECC granularity in the data RAMs from 64bit to 32bit. This
is to align with an upcoming change in bus ECC. Relates to
lowRISC/opentitan#5450

### DIFF
--- a/rtl/ibex_icache.sv
+++ b/rtl/ibex_icache.sv
@@ -64,14 +64,15 @@ module ibex_icache #(
   // Request throttling threshold
   localparam int unsigned FB_THRESHOLD = NUM_FB - 2;
   // Derived parameters
-  localparam int unsigned LINE_SIZE_ECC   = ICacheECC ? (LineSize + 8) : LineSize;
   localparam int unsigned LINE_SIZE_BYTES = LineSize/8;
   localparam int unsigned LINE_W          = $clog2(LINE_SIZE_BYTES);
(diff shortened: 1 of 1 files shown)


rtl/ibex_icache.sv
```diff
@@ -64,9 +64,10 @@ module ibex_icache #(
   // Request throttling threshold
   localparam int unsigned FB_THRESHOLD = NUM_FB - 2;
   // Derived parameters
-  localparam int unsigned LINE_SIZE_ECC   = ICacheECC ? (LineSize + 8) : LineSize;
+  localparam int unsigned LINE_SIZE_ECC   = ICacheECC ? (LineSize + 14) : LineSize;
```
rswarbrick, line 67: I think this formula was a bit dubious, since it kind of assumed that `LineSize` is 64. But we can do better here now. Can we add an assertion that checks `BusWidth` is 32 when ECC is enabled and then move the computation of `LINE_SIZE_ECC to the end of the list, changing the value to `BUS_SIZE_ECC * LINE_BEATS`?

tomroberts-lowrisc:
Any advance on this - would be good to get it vendored back with the other changes this week

rswarbrick (approved):
Sorry, I didn't notice that you'd pushed up a new version. Looks good to me!

tomroberts-lowrisc:
Thanks!

PR #1303, opened 2021-03-12
Add RAM cfg signals

tomroberts-lowrisc:
Update vendored-in lowrisc_ip and add missing ram config signals for icache.


imphil (approved):

imphil:
Needs a bit more grepping, but then it should be alright :-)

rswarbrick:
@tomroberts-lowrisc: Possibly a stupid question, but... Do these signals have any behavioural effect? If so, should we be trying to capture them in the icache testing?

tomroberts-lowrisc:
They are config signals for real physical rams - don't need to touch them for verification

PR #1309, opened 2021-03-18
Irq in debug mode fix

GregAC:


GregAC:
I have been doing regressions to check things but still need to run a full one on this precise set of commits and changes.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -195,33 +195,54 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
     check_next_core_status(INITIALIZED, "Core initialization handshake failure", 5000);
   endtask
 
+  function int determine_irq_from_txn();
```
udinator, line 198: nit: could you have this function take in the `irq_txn` object as a parameter?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -195,33 +195,54 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
...
+    vseq.start_nmi_raise_seq();
+    send_irq_stimulus_inner(ret_val);
+  endtask
+
+  virtual task send_irq_stimulus_inner(output bit ret_val);
+    bit irq_valid;
     irq_collected_port.get(irq_txn);
-    irq = {irq_txn.irq_nm, irq_txn.irq_fast, 4'b0, irq_txn.irq_external, 3'b0,
-           irq_txn.irq_timer, 3'b0, irq_txn.irq_software, 3'b0};
     `uvm_info(`gfn, $sformatf("irq: 0x%0x", irq), UVM_LOW)
```
udinator, line 223: could this `uvm_info` be moved inside of `determine_irq_from_txn`?
since the assignment to `irq` has been moved inside of that function, it's not not as obvious where `irq` is getting assigned so I feel printing out its value should happen directly after the assignment.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -730,29 +760,89 @@ class core_ibex_irq_in_debug_test extends core_ibex_directed_test;
...
+      // for the IRQ being handled during debug and deal with the IRQ remained asserted after DRET
+      // case.
       fork
         begin : wait_irq
-          wait_for_core_status(HANDLING_IRQ);
-          `uvm_fatal(`gfn, "Core is handling interrupt detected in debug mode")
+          // Get IRQ raise transaction from IRQ monitor
+          irq_collected_port.get(irq_txn);
+          detected_irq = 1'b1;
+          irq_valid = determine_irq_from_txn();
```
udinator, line 789: nit: perhaps the order of these two lines should be switched so we have assurance that this fork will only be disabled by `wait_dret`after the proper irq_valid and irq have been calculated

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -730,29 +760,89 @@ class core_ibex_irq_in_debug_test extends core_ibex_directed_test;
...
+            check_irq_handle();
+            send_irq_stimulus_end();
+          end else begin
+            `uvm_info(`gfn, "IRQ is disabled, no interrupt should be taken", UVM_LOW)
+            // IRQ is disabled so just drop IRQ
+            vseq.start_irq_drop_seq();
+            irq_collected_port.get(irq_txn);
+          end
         end
         begin
```
udinator, line 823: nit: could you label this subprocess, to match the other two?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
GregAC: > can this function take in the irq_txn object as an argument?

It could, I did consider it when doing the initial work, the issue is `core_ibex_debug_intr_basic_test` contains a bunch of shared state around IRQs (`irq_id`, `irq_txn`, `irq`) and `determine_irq_from_txn` is updating that shared state based upon `irq_txn` so with the current design calling `determine_irq_from_txn` with anything other than the `irq_txn` from the class would be an error. So I felt it was better to just avoid having it as an argument altogether.

> perhaps this function should be renamed to something like determine_irq_valid(...)

As it updating the `irq` and `irq_id` from `irq_txn` are a core part of it's functionality I preferred a name that indicated that. Happy to change it to something else but `determine_irq_valid` doesn't really hint at the updating it's doing.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
GregAC: Good point, I forget the processes can be paused an any point to allow another to continue rather than just at explicit waits.

GregAC:
Thanks for the review @udinator made most of the suggested changes but not certain about a couple.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
udinator: sgtm!

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -195,33 +195,59 @@ class core_ibex_debug_intr_basic_test extends core_ibex_base_test;
     check_next_core_status(INITIALIZED, "Core initialization handshake failure", 5000);
   endtask
 
+  function bit determine_irq_from_txn();
+    bit irq_valid;
+
+    `uvm_info(`gfn, $sformatf("irq: 0x%0x", irq), UVM_LOW)
```
udinator, line 201: sorry another nit about this `uvm_info` - can you place it after the `irq` is assigned from the `irq_txn` object so we print out the updated value?

udinator (approved):
had one more small nit, but LGTM overall! thanks for making these updates.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
GregAC: Well spotted, certainly not much use where it is, I'll move it.

PR #1310, opened 2021-03-18
[rtl] Add MSECCFGH CSR

### LINKED ISSUE #1297: [epmp] question about handling mseccfg high bits
In the [ePMP specification](https://ibex-core.readthedocs.io/en/latest/_downloads/adbc55b5b9bad5f92968b58e7b12fc57/riscv-epmp.pdf) `mseccfg` is specified as a 64-bit register that is split into two parts (low at 0x390 and high at 0x391) on 32-bit architectures.

The Ibex documentation doesn't mention this so I am curious as to how/if we are currently handling the high bits. Should we document a `mseccfgh` zero read-only register at 0x391 so that portable software can read the full 64-bit value?

GregAC:
This is the top 32 bits of MSECCFG. It currently has no specified bits
so reads as 0 and ignores writes.

Fixes #1297

### DIFF
--- a/doc/03_reference/cs_registers.rst
+++ b/doc/03_reference/cs_registers.rst
@@ -36,6 +36,8 @@ Ibex implements all the Control and Status Registers (CSRs) listed in the follow
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x390  | ``mseccfg``        | WARL   | Machine Security Configuration                |
 +---------+--------------------+--------+-----------------------------------------------+
+|  0x391  | ``mseccfgh``       | WARL   | Upper 32 bits of ``mseccfg``                  |
++---------+--------------------+--------+-----------------------------------------------+
 |  0x3A0  | ``pmpcfg0``        | WARL   | PMP Configuration Register                    |
 +---------+--------------------+--------+-----------------------------------------------+
 |     .             .               .                    .                              |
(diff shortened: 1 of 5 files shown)


vogelpi (approved):
LGTM, thanks @GregAC for fixing this.

Reading the ePMP spec I noted that actually, if PMP is disabled the `mseccfg(h)` CSRs shouldn't be available at all:

"Machine Security Configuration (mseccfg) ​is a new Machine mode CSR, used forconfiguring various security mechanisms present on the hart, and only accessible toMachine mode. It is only available on harts with PMP support."

So actually, Ibex should still trigger an illegal instruction exception if these are accessed while `PMPEnable = 1'b0`. However, I don't know if DV is flexible enough to support both cases. I think to remember that there were quite a few issues around "custom" CSRs.

GregAC:
Ah well spotted. It's probably fine to make it an illegal instruction for `PMPEnable = 1'b0` cases as right now the DV isn't emitting any reads or writes to it (even in illegal instruction tests, as otherwise we'd see a Spike/Ibex mismatch in PMP on configs). We've got already got functionality to include/exclude specific tests for specific configs so can exclude tests that do stimulate it once they exist.

vogelpi:
Perfect, thanks Greg for adding the exception if PMP is disabled. LGTM!

PR #1316, opened 2021-03-24
Create new top level "ibex_top" to wrap core, register file and icache RAMs

tomroberts-lowrisc:
This is the precursor to adding dual-core lockstep, but is a huge set of changes unfortunately!

There are some other changes in here to move some (effectively) fixed icache parameters into the package which reduces the number of parameters needing to be passed around.

The icache TB now uses a wrapper to instantiate the icache rtl and the RAMs which mirrors the setup in ibex_top.

Finally, I had to switch to
`module foo import ibex_pkg::*; #(`
In places to avoid line lengths > 100 in the IO listings. This seems to be the defacto standard in OT now, and I can't see any reason not to do it in ibex too.


tomroberts-lowrisc:
Superseded by #1318

PR #1318, opened 2021-03-25
Add dual core lockstep option

tomroberts-lowrisc:
Note this includes commits from the previous PR (wiring to support lockstep).

Also note that the lockstep checking if effectively disabled pending further work to add a configuration which resets all registers.


GregAC (approved):
LGTM. We could consider making the outputs comparison more detailed by factoring in relevant valid signals (only compare the memory bus signals for a request if the request is valid etc), though the current method should provide a stronger check as all being well (ignoring Xs) things should match perfectly even if the valid is unset.

Issue #1313, opened 2021-03-22
[dv] riscv_debug_ebreakmu_test: Core did not enter debug mode after execution of ebreak


imphil:
* Config: experimental-maxperf-pmp-bmfull-icache
* `regr.log`: 
```
riscv_debug_ebreakmu_test.9
---------------------------
Test binary: out/seed-27831/instr_gen/asm_test/riscv_debug_ebreakmu_test.9.o
sim log: out/seed-27831/rtl_sim/riscv_debug_ebreakmu_test.9/sim.log
UVM_FATAL /azp/agent/_work/1/s/dv/uvm/core_ibex/tests/core_ibex_base_test.sv(204) @ 8886400: reporter [uvm_test_top] Check failed signature_data == core_status (8 [0x8] vs 1 [0x1]) Core did not enter debug mode after execution of ebreak
[FAILED]: sim error seen
```
* Log files and waves: [simerror-riscv_debug_ebreakmu_test.zip](https://github.com/lowRISC/ibex/files/6183819/simerror-riscv_debug_ebreakmu_test.zip)

@udinator @GregAC

GregAC:
Was there just this single failure in the run or did all of them fail? Regression run I just completed using #1309 passed all iterations of this test.

I'll do a bunch of iterations of this test specifically with latest master to see if it's still occurring.

imphil:
The `riscv_debug_ebreakmu_test` test ran 15 times, one out of those 15 failed, all other 14 were successful.

GregAC:
I had a go at recreating this with a local run and found 1 failure in 30 that I thought was the same, however just seen they're subtly different but may have the same root cause. First the issue I saw:

It turned out to be an interesting corner case and looks like Ibex is doing the correct thing and the testbench is getting confused.

The test first stimulates a `debug_req` to enter debug mode so it can set the bit in `dcsr` that turns `ebreak` into a breakpoint (so it enters debug mode rather than just causing an exception much like an `ecall`).

Occasionally that initial `debug_req` will occur the same cycle the first `ebreak` is executed. Ibex correctly sets up the `mstatus`, `mcause`, `mepc` for the exception the `ebreak` should cause then immediately enters debug mode setting up the debug registers based upon the exception entry (so return PC and privilege levels are the ones required for the exception handler entry). The testbench isn't expecting this and in particular expects a normal debug mode entry for the `debug_req` from U-mode in the failing case so complains `dcsr.prv` has the wrong value.

[...]

PR #1319, opened 2021-03-26
[dv] Add known failure detection to riscv_debug_ebreakmu_test

GregAC:
This will produce a 'KNOWN FAILURE' message when we hit the #1313

This still turns into a failure for the regression. I don't think it should become a pass but if we're happy with this we can modify 'sim.py' to have a third known failure category using regexs to match against errors to decide which failures are known failures.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -1055,19 +1055,47 @@ class core_ibex_debug_ebreakmu_test extends core_ibex_directed_test;
   `uvm_component_utils(core_ibex_debug_ebreakmu_test)
   `uvm_component_new
 
+  bit seen_ebreak;
+
+  virtual task send_stimulus();
+    seen_ebreak = 0;
+    fork
+      begin : detect_ebreak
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1055,19 +1055,45 @@ class core_ibex_debug_ebreakmu_test extends core_ibex_directed_test;
...
-                           "Core did not enter debug mode after debug_req stimulus", 2000);
-    check_priv_mode(PRIV_LVL_M);
-    // Read dcsr and verify the appropriate ebreak(m/s/u) bit has been set based on the prv field,
-    // as well as the cause field
-    wait_for_csr_write(CSR_DCSR, 500);
-    check_dcsr_prv(init_operating_mode);
-    check_dcsr_ebreak();
-    check_dcsr_cause(DBG_CAUSE_HALTREQ);
-    wait_ret("dret", 5000);
+    fork
```
udinator, line 1074: nit: could you wrap the `fork...join_any...disable fork;` inside of an isolation fork? something like:
```systemverilog
fork begin
  fork
    <new code added in this task>
  join_any
  disable fork;
end join
```

udinator (approved):

PR #1329, opened 2021-04-08
[rtl] Fix RF read enables for illegal instruction/fetch error

GregAC:
The read enables should only be asserted where an actual RF read will
occur. Where there is an illegal instruction or a fetch error the raw
decoder signals might still be asserted but should be squashed before
they become the true enable signals.

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -237,6 +237,11 @@ module ibex_id_stage #(
   rf_wd_sel_e  rf_wdata_sel;
   logic        rf_we_dec, rf_we_raw;
   logic        rf_ren_a, rf_ren_b;
+  logic        rf_ren_a_dec, rf_ren_b_dec;
+
+  // Read enables should only be asserted for valid and legal instructions
+  assign rf_ren_a = instr_valid_i & ~instr_fetch_err_i & ~illegal_insn_o & rf_ren_a_dec;
+  assign rf_ren_b = instr_valid_i & ~instr_fetch_err_i & ~illegal_insn_o & rf_ren_b_dec;
 
   assign rf_ren_a_o = rf_ren_a;
   assign rf_ren_b_o = rf_ren_b;
@@ -461,8 +466,8 @@ module ibex_id_stage #(
       .rf_raddr_a_o                    ( rf_raddr_a_o         ),
       .rf_raddr_b_o                    ( rf_raddr_b_o         ),
       .rf_waddr_o                      ( rf_waddr_id_o        ),
-      .rf_ren_a_o                      ( rf_ren_a             ),
-      .rf_ren_b_o                      ( rf_ren_b             ),
+      .rf_ren_a_o                      ( rf_ren_a_dec         ),
+      .rf_ren_b_o                      ( rf_ren_b_dec         ),
 
       // ALU
       .alu_operator_o                  ( alu_operator         ),
@@ -849,9 +854,17 @@ module ibex_id_stage #(
 
   `ASSERT(StallIDIfMulticycle, (id_fsm_q == FIRST_CYCLE) & (id_fsm_d == MULTI_CYCLE) |-> stall_id)
 
-  // Stall ID/EX stage for reason that relates to instruction in ID/EX
+
+  // Stall ID/EX stage for reason that relates to instruction in ID/EX, update assertion below if
+  // modifying this.
   assign stall_id = stall_ld_hz | stall_mem | stall_multdiv | stall_jump | stall_branch |
                       stall_alu;
+
+  // Generally illegal instructions have no reason to stall, however they must still stall waiting
+  // for outstanding memory requests so exceptions related to them take priority over the illegal
+  // instruction exception.
+  `ASSERT(IllegalInsnStallMustBeMemStall, illegal_insn_o & stall_id |-> stall_mem &
+    ~(stall_ld_hz | stall_multdiv | stall_jump | stall_branch | stall_alu))
 
   assign instr_done = ~stall_id & ~flush_id & instr_executing;
 
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -256,7 +256,7 @@ module ibex_cs_registers #(
   logic [31:0] csr_wdata_int;
   logic [31:0] csr_rdata_int;
   logic        csr_we_int;
-  logic        csr_wreq;
+  logic        csr_wr;
 
   // Access violation signals
   logic        illegal_csr;
@@ -279,7 +279,7 @@ module ibex_cs_registers #(
 
   // See RISC-V Privileged Specification, version 1.11, Section 2.1
   assign illegal_csr_priv   = (csr_addr[9:8] > {priv_lvl_q});
-  assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wreq;
+  assign illegal_csr_write  = (csr_addr[11:10] == 2'b11) && csr_wr;
   assign illegal_csr_insn_o = csr_access_i & (illegal_csr | illegal_csr_write | illegal_csr_priv);
 
   // mip CSR is purely combinational - must be able to re-enable the clock upon WFI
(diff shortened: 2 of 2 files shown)


rtl/ibex_id_stage.sv
```diff
@@ -237,6 +237,11 @@ module ibex_id_stage #(
   rf_wd_sel_e  rf_wdata_sel;
   logic        rf_we_dec, rf_we_raw;
   logic        rf_ren_a, rf_ren_b;
+  logic        rf_ren_a_dec, rf_ren_b_dec;
+
+  // Read enables should only be asserted for valid and legal instructions
+  assign rf_ren_a = instr_valid_i & ~instr_fetch_err_i & ~illegal_insn_o & rf_ren_a_dec;
+  assign rf_ren_b = instr_valid_i & ~instr_fetch_err_i & ~illegal_insn_o & rf_ren_b_dec;
```
vogelpi, line 244: Nit: you could assign directly `rf_ren_a/b_o` here and remove the `rf_ren_a/b` signals completely.

rtl/ibex_id_stage.sv
```diff
@@ -849,10 +854,18 @@ module ibex_id_stage #(
...
+  // modifying this.
   assign stall_id = stall_ld_hz | stall_mem | stall_multdiv | stall_jump | stall_branch |
                       stall_alu;
 
+  // Generally illegal instructions have no reason to stall, however they must still stall waiting
+  // for outstanding memory requests so exceptions related to them take priority over the illegal
+  // instruction exception.
+  `ASSERT(IllegalInsnStallMustBeMemStall, illegal_insn_o & stall_id |-> stall_mem &
+    ~(stall_ld_hz | stall_multdiv | stall_jump | stall_branch | stall_alu))
+
```
vogelpi, line 868: I am not sure if this is correct. `stall_mem` is high during the address phase of the transaction only. That's when we would get a PMP error. But the instruction is still in ID/EX stage, so there can't be an illegal instruction exception, otherwise the load/store would not have been started in the first place.

In contrast, `stall_ld_hz` will be high when the current instruction in ID/EX needs the outcome of the instruction in WB. However, with this change it will be low if the current instruction in ID/EX is illegal. This means we wouldn't wait for the result of the load (and a possible bus error) and take the illegal instruction exception directly. Is that correct and intended?

vogelpi (commented):
Thanks @GregAC for your PR. I have a question regarding the assertion and intended behavior.

rtl/ibex_id_stage.sv
GregAC: I had a slight distaste for using output signals directly as inputs into other logic within a module. Though Ibex has quite a lot of that so I may be fighting a loosing battle here.

rtl/ibex_id_stage.sv
GregAC: `stall_mem` will be asserted when there's an outstanding memory access in writeback as well as during the address phase of a memory access instruction. It is the stall you have when waiting for a memory access in writeback to complete that we still need to consider for illegal instructions. Perhaps I should split these out into two separate signals as they're different stall reasons anyway.

All other stalls are do to with things the instruction in ID/EX itself may be doing. An illegal instruction should be doing nothing so we shouldn't be seeing any stalls.

> This means we wouldn't wait for the result of the load (and a possible bus error) and take the illegal instruction exception directly. Is that correct and intended?

`stall_mem` is what prevents that from happening.

vogelpi: Ah, now I understand. Thanks for the clarification Greg.

Regarding splitting `stall_mem`: I think this is a good idea and would be required for imprecise load/store exceptions. But it doesn't need to happen right now.

vogelpi (approved):
LGTM!

Note it seems that CI fails because of an UNOPTFLAT warning:
```
%Warning-UNOPTFLAT: ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:214:16: Signal unoptimizable: Feedback to clock or circular logic: 'ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.stall_ld_hz'
  214 |   logic        stall_ld_hz;
      |                ^~~~~~~~~~~
                    ... Use "/* verilator lint_off UNOPTFLAT */" and lint_on around source to disable this message.
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:214:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.stall_ld_hz
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:927:33:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:210:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.instr_executing_spec
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:759:3:      Example path: ALWAYS
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:217:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.stall_branch
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:860:19:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:219:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.stall_id
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:869:21:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:212:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.instr_done
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:981:28:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_core.sv:323:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.instr_id_done
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:640:34:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_core.sv:250:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.csr_op_en
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_cs_registers.sv:739:19:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_cs_registers.sv:259:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.cs_registers_i.csr_wreq
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_cs_registers.sv:283:29:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_core.sv:254:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.illegal_csr_insn_id
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:531:25:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_core.sv:339:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.illegal_insn_id
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:243:19:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:239:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.rf_ren_a
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:975:24:      Example path: ASSIGNW
                    ../src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_id_stage.sv:214:16:      Example path: ibex_top_tracing.u_ibex_top.u_ibex_core.id_stage_i.stall_ld_hz

```

GregAC:
Ah yes, had missed that as I was just running VCS DV which doesn't check for such things.

It's due to how we're generating the illegal CSR signal. It's factoring in the enable (indicating the instruction is clear to execute) which it doesn't need to do. I had actually spotted this a couple of weeks ago when playing around with FPGA timing had made a note to fix it then totally forgotten about it! (I'll make sure I'll do a github issue next time I see something similar and don't fix immediately).

GregAC:
Requesting new review as this now has a second change for the illegal CSR logic.

vogelpi (approved):
Oh yes, this makes sense. And your way of fixing the loop looks like the preferred way for doing this. LGTM, thanks @GregAC !

PR #1331, opened 2021-04-12
Ibex fcov2

GregAC:
A previous PR (#1329) has the RTL fixes, this is just to review the fcov changes. Other PR will get merged first.

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+`include "prim_assert.sv"
+
 interface core_ibex_fcov_if import ibex_pkg::*; (
   input clk_i,
   input rst_ni,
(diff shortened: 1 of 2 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -14,57 +14,111 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
-        end else if (id_stage_i.instr_rdata_i[31:25] == 7'b0000001) begin
+        if ({id_stage_i.instr_rdata_i[26], id_stage_i.instr_rdata_i[13:12]} == {1'b1, 2'b01}) begin
+          id_instr_category = InstrCategoryALU; // reg-reg/reg-imm ops
+        end else if (id_stage_i.instr_rdata_i[31:25] inside {7'b000_0000, 7'b010_0000, 7'b011_0000,
+              7'b011_0100, 7'b001_0100, 7'b001_0000, 7'b000_0101, 7'b000_0100, 7'b010_0100}) begin
+          id_instr_category = InstrCategoryALU; // RV32I and RV32B reg-reg/reg-imm ops
+        end else if (id_stage_i.instr_rdata_i[31:25] == 7'b000_0001) begin
           if (id_stage_i.instr_rdata_i[14]) begin
-            id_instr_category = InsnCategoryMult; //MUL*
+            id_instr_category = InstrCategoryDiv; // DIV*
```
rswarbrick, line 73: I wonder whether there's some cross-checking we could add somewhere, to make sure that if we mark an instruction as `InstrCategoryDiv` then we actually see an internal "I've got a division" signal go high? (And similarly for the other categories, but this is the one that had a bug!)

I'm a bit worried about this because I can imagine a bug in the "decode logic" here causing a big coverage hole.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -85,21 +139,40 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
-    `DV_FCOV_EXPR_SEEN(insn_unstalled, instruction_unstalled.triggered)
+  // Keep track of previous values for some signals. These are used for some of the crosses relating
+  // to exception and debug entry. We want to cross different instruction categories and stalling
+  // behaviour with exception and debug entry but signals indicating entry occur a cycle after the
+  // relevant information is flushed from the pipeline.
+  always @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      // Explicit X out of reset so first cycle out of reset case (where there is no last stall
+      // type) is correctly handled
+      id_stall_type_last     <= 'x;
```
rswarbrick, line 160: I think I'd probably prefer adding a 1-bit "has_last" signal here. Magic "x" handling is really easy to get wrong: I think it's probably better to be explicit.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -112,23 +185,69 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    debug_entry_if_instr_cross: cross cp_debug_entry_if, instr_unstalled_last, id_instr_category_last;
+    pipe_flush_instr_cross: cross cp_pipe_flush, instr_unstalled, cp_instr_category_id;
+
+    exception_stall_instr_cross: cross cp_ls_pmp_exception, cp_ls_error_exception,
+      cp_instr_category_id, cp_stall_type_id, instr_unstalled, cp_irq_pending, cp_debug_req {
+      illegal_bins illegal =
+        // Only Div, Mul, Branch and Jump instructions can see an instruction stall
+        (!binsof(cp_instr_category_id) intersect {InstrCategoryDiv, InstrCategoryMul,
+                                                 InstrCategoryBranch, InstrCategoryJump} &&
+         binsof(cp_stall_type_id) intersect {IdStallTypeInstr})
```
rswarbrick, line 242: I don't know whether this is actually possible, but is there a way to share the logic between this and `stall_cross`? I can sort of see a way to do it with an auxiliary signal, but I'm not sure whether there's something cleverer. Ideally we wouldn't have to duplicate this non-trivial set of intersections.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: I've added a bunch of asserts to check the categories.

There may be an argument for using the decoder signals directly but I prefer this mini decode approach. I prefer to minimize micro architectural probing for coverage purposes so I think the decode done here with the assertions to check there's no disagreement with the decoder is a good approach.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: I don't think there's a good way to do it. I want to use the features here for performing logical operations on sets of bins and they're only available inside a cross coverage group. So some auxiliary signal would need to implement them 'manually'. I don't think there's any way to cross a cross cover group with another (you can just cross coverpoints) so you can't do some 'base' cross coverage that does these illegal bins and then 'inherit' from it.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
rswarbrick: Yep, I agree. Looks good to me.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
rswarbrick: Yeah, I thought that might be the case. Ho hum...

udinator (approved):
lgtm in general, thanks @GregAC !

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -14,60 +16,168 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
 
+  // Check instruction categories calculated from instruction bits match what decoder has produced.
+
+  // The ALU category is tricky as there's no specific ALU enable and instructions that actively use
+  // the result of the ALU but aren't themselves ALU operations (such as load/store and JALR). This
+  // categorizes anything that selects the ALU as the source of register write data and enables
+  // register writes minus some exclusions as an ALU operation.
+  `ASSERT(InstrCategoryALUCorrect, id_instr_category == InstrCategoryALU |->
+      (id_stage_i.rf_wdata_sel == RF_WD_EX) && id_stage_i.rf_we_dec && ~id_stage_i.mult_sel_ex_o &&
+      ~id_stage_i.div_sel_ex_o && ~id_stage_i.lsu_req_dec && ~id_stage_i.jump_in_dec);
```
imphil, line 135: I knew I would find a semicolon after an assert :-) Sorry for being late.

PR #1334, opened 2021-04-13
[dv] Remove MISA from csr_description.yaml

GregAC:
The value of `misa` will change depending on whether M or B are enabled.
The presence and read values of other CSRs may also depend upon the Ibex
configuration. A fix is required to allow riscv_csr_test to deal with
different CSR descriptions for different Ibex configurations. For now
just comment out `misa` from the descriptions file to enable
riscv_csr_test to run on a wider range of configurations.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
@@ -39,27 +39,29 @@
 #      - ...
 
 
+# TODO: Enable multiple configurations for CSR test. MISA read value will depend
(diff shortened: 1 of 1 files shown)


GregAC:
@imphil this will fix 5 failures in each of the DV runs on configs without bitmanip.

imphil (approved):
Good find! @udinator I assume there is no other quick way to get the test do the right thing for different configurations?

udinator:
> Good find! @udinator I assume there is no other quick way to get the test do the right thing for different configurations?

Hmm, off the top of my head the only way to get the correct behavior here would be to enhance the tank with the correct values to use if certain parameters are enabled, similwr to what we did with the CI flows.

Issue #1335, opened 2021-04-13
Usage of prim_clock_gating and latch_map.v during synthesis


rahulraveendran15-coder:
(1) Do I need to use both prim_clock_gating.v and latch_map.v file during synthesis when using RegfileFF as my configuration?
Also In reference with closed case #1299 @GregAC  told that 

`prim_clock_gating.v is a file required to build Ibex, it is used for the top-level clock gate the enables/disables the CPU clock. latch_map.v allow yosys to build the latch based synthesis file by telling it what latch cells to use. This too may require adjustment for library other than nangate.`

If answer for (1) is no, then if I want to use RegfileLatch do I need to change both files if I am using a differenct library for synthesis?changing latch_map.v makes sense. What about prim_clock_gating.v, I can't see any latch maping there?

vogelpi:
Hi @rahulraveendran15-coder,

I am pretty sure that you need both `prim_clock_gating.v` and `latch_map.v` independent of the selected register file. As you said `latch_map.v` is needed to tell Yosys which cell to map generic latches to that are inferred when synthesizing the Verilog code. The `prim_clock_gating.v` is just a latch-based clock gate. Without `latch_map.v` Yosys most likely won't know to which cell of the library it should map the latch inside `prim_clock_gating.v` during the technology mapping stage.

Now, if you want to switch to a different library, you just need to change `latch_map.v`.

rahulraveendran15-coder:
Hi @vogelpi 
So are you telling eventhough if I am using the RegfileFF , do  I need to map the latch cell inside my library to the latch_map.v for prim_clock_gating.v file proper working(since clock gating is needed in all cases.)

Also how it is mapping the latches, I can see the module name for latch_map.v as like below. what will be the module name here? Where this module is instantiated?

module $_DLATCH_P_ (input E, input D, output Q);

rahulraveendran15-coder:
Also Greg told that "latch_map.v used because of refileLatch mode used in synthesis". In that case does it needed for RegfileFF?
#1299

vogelpi:
> So are you telling eventhough if I am using the RegfileFF , do I need to map the latch cell inside my library to the latch_map.v for prim_clock_gating.v file proper working(since clock gating is needed in all cases.)

Correct, even though you are using RegFileFF, you'll need `latch_map.v` for `prim_clock_gating.v`.

[...]

PR #1336, opened 2021-04-14
Fix a couple of synthesis bugs

tomroberts-lowrisc:
1. Missing prim_assert in ibex_top.sv (more of an rtl bug but only found
in running synthesis scripts)
2. Write out the pre-mapped netlist before mapping latches

Relates to #1335


vogelpi (approved):
LGTM, thanks @tomroberts-lowrisc !

PR #1338, opened 2021-04-14
[rtl] Implement mvendorid/marchid/mimpid CSRs

GregAC:


doc/03_reference/cs_registers.rst
```diff
@@ -576,6 +582,33 @@ The User Mode ``time(h)`` registers are not implemented in Ibex.
...
+Use the ``MVENDORID_VALUE`` localparam in :file:`rtl/ibex_cs_registers.sv` to change the fixed value.
+
+Machine Architecture ID (marchid)
+---------------------------------
+
+CSR Address: ``0xF12``
+
+Reset Value: ``0x0000_0016``
+
+Use the ``MARCHID_VALUE`` localparam in :file:`rtl/ibex_cs_registers.sv` to change the fixed value.
```
imphil, line 601: Maybe add a note that these numbers are assigned to us, and people should change them if they make significant changes (can we be more specific?) to the core?

rtl/ibex_cs_registers.sv
```diff
@@ -125,6 +125,21 @@ module ibex_cs_registers #(
   localparam int unsigned RV32MEnabled = (RV32M == RV32MNone) ? 0 : 1;
   localparam int unsigned PMPAddrWidth = (PMPGranularity > 0) ? 33 - PMPGranularity : 32;
 
+  // Vendor ID
+  // 0 indicates a non-commercial implementation
```
imphil, line 129: Ibex is commercial. I think we should just get a JEDEC ID, we'll need that one day for JTAG and similar purposes as well.

@asb

imphil (commented):
Thanks Greg! 

I'd be interested in adding a bit more guidance to users of Ibex what they should and should not do with those IDs. E.g. when to change them (not fully detailed, but at least roughly). And if we want to encourage changing the IDs, maybe putting them in a place that requires less RTL modifications (such as a package) or even a top-level parameter?

GregAC:
> I'd be interested in adding a bit more guidance to users of Ibex what they should and should not do with those IDs

Given what these are for is part of the RISC-V specification I was trying to avoid saying too much about them in the documentation as we don't just want to repeat the specification. I could add something saying look at the specific for further details.

> And if we want to encourage changing the IDs, maybe putting them in a place that requires less RTL modifications (such as a package) or even a top-level parameter?

I can move them to the top-level package. Trying to avoid drowning Ibex in too many top-level parameters and these aren't things you're likely to want to change on an instantiation by instantiation basis.

rtl/ibex_cs_registers.sv
GregAC: We can certainly look into it, could switch this comment to say 'No JEDC ID has been allocated to lowRISC so value is 0'?

rswarbrick: Sounds like a good formulation to me.

imphil: Yeah, let's do that for the moment. And it's "JEDEC"

imphil:
> Given what these are for is part of the RISC-V specification I was trying to avoid saying too much about them in the documentation as we don't just want to repeat the specification.

Yes, but nobody ever reads the spec :smile:  So if we can give people a hint in the "how to use ibex for your project" section of the docs (https://ibex-core.readthedocs.io/en/latest/02_user/getting_started.html) that they should maybe change mimpid if they make small changes to Ibex, and mvendorid/marchid when they make massive changes to it, but leave the identifiers as they are if they take ibex as-is?

imphil (approved):

rtl/ibex_pkg.sv
```diff
@@ -558,4 +561,21 @@ parameter int unsigned CSR_MSECCFG_MML_BIT  = 0;
...
+// Top bit is unset to indicate an open source project. The lower bits are an ID allocated by the
+// RISC-V Foundation. Note this is allocated specifically to Ibex, should significant changes be
+// made a different architecture ID should be supplied.
+localparam logic [31:0] CSR_MARCHID_VALUE = {1'b0, 31'd22};
+
+// Implementation ID
+// 0 indicates this field is not implemeted. Ibex implementors may wish to indicate an RTL/netlist
+// version here using their own unique encoding (e.g. 32 bits of the git hash of the implemented
+// commit).
+localparam logic [31:0] CSR_MIMPID_VALUE = 32'b0;
```
GregAC, line 579: I am now wondering if we should use some of these bits to indicate Ibex configuration, e.g. if writeback stage is enabled. We could have one bit per config option but this could add up to a lot of bits. We could reserve some number of bits for a 'config ID' and we give configurations of interest a config ID (e.g. the configurations listed in the README would each have their own config ID).

We could also/in addition implement an Ibex specific CSR that gives detailed config information.

GregAC:
@imphil I've added a paragraph to the docs in getting started

doc/02_user/getting_started.rst
```diff
@@ -11,3 +11,16 @@ Register File
 Ibex comes with three different register file implementations that can be selected using the enumerated parameter ``RegFile`` defined in :file:`rtl/ibex_pkg.sv`.
 Depending on the target technology, either the flip-flop-based ("ibex_pkg::RegFileFF", default), the latch-based ("ibex_pkg::RegFileLatch") or an FPGA-targeted ("ibex_pkg::RegFileFPGA") implementation should be selected.
 For more information about the three register file implementations and their trade-offs, check out :ref:`register-file`.
+
+Identification CSRs
+-------------------
+
+The RISC-V Privileged Architecture specifies several read-only CSRs that can be to identify the vendor and microarchitecture of a CPU.
```
imphil, line 18:
```suggestion
The RISC-V Privileged Architecture specifies several read-only CSRs that identify the vendor and micro-architecture of a CPU.
```

micro-architecture for consistency with the usage below.

doc/02_user/getting_started.rst
```diff
@@ -11,3 +11,16 @@ Register File
...
+These are ``mvendorid``, ``marchid`` and ``mimpid``.
+The fixed, read-only values for these CSRs are defined in :file:`rtl/ibex_pkg.sv`.
+Implementers should carefully consider appropriate values for these registers.
+Ibex, as an open source implementation, has an assigned architecture ID (``marchid``) of 22 (Allocations are specified in `marchid.md of the riscv-isa-manual repository <https://github.com/riscv/riscv-isa-manual/blob/master/marchid.md>`_).
```
imphil, line 22:
```suggestion
Ibex, as an open source implementation, has an assigned architecture ID (``marchid``) of 22. 
(Allocations are specified in `marchid.md of the riscv-isa-manual repository <https://github.com/riscv/riscv-isa-manual/blob/master/marchid.md>`_.)
```

doc/03_reference/cs_registers.rst
```diff
@@ -576,6 +582,38 @@ The User Mode ``time(h)`` registers are not implemented in Ibex.
...
 
+Machine Vendor ID (mvendorid)
+-----------------------------
+
+CSR Address: ``0xF11``
+
+Reset Value: ``0x0000_0000``
+
+Use the ``CSR_MVENDORID_VALUE`` parameter in :file:`rtl/ibex_pkg.sv` to change the fixed value.
+Details of what the ID represents can be found in the RISC-V Privileged Specification, version 1.11.
```
imphil, line 593:
```suggestion
Details of what the ID represents can be found in the RISC-V Privileged Specification.
```

(We specify the version of the RISC-V spec that we follow in one place to avoid having to update them everywhere if we change it.)

Also below.

imphil (approved):

vogelpi:
Hey @GregAC ,
it has been a while since the last activity in this PR. But it's already approved and if I understand correctly, the changes are absolutely uncontroversial. We could just rebase and merge this PR. Or is there is there still something missing?

GregAC:
Ah well spotted, think I simply forgot to hit the merge button! There's a potential add-on in using mimpid to differentiate different Ibex configs but can deal with that in a follow up if desired.

vogelpi:
Ah, great. Thanks for the quick feedback Greg.

PR #1341, opened 2021-04-16
[lockstep] Introduce optimization barrier around lockstep Ibex

msfschaffner:
Certain synthesis tools like DC are very smart at optimizing away redundant logic.
Hence, we have to insert an optimization barrier at the IOs of the lockstep Ibex.
This is achieved by manually buffering each bit using `prim_buf`.
Our Xilinx and DC synthesis flows make sure that these buffers cannot be optimized
away using `keep` attributes (Vivado) and `size_only` constraints (DC).

### DIFF
--- a/ibex_top.core
+++ b/ibex_top.core
@@ -10,6 +10,7 @@ filesets:
     depend:
       - lowrisc:ibex:ibex_pkg
       - lowrisc:ibex:ibex_core
+      - lowrisc:prim:buf
     files:
       - rtl/ibex_register_file_ff.sv # generic FF-based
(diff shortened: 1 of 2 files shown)


msfschaffner:
Could you integrate this and vendor it back into the OT repo? Thanks!

tomroberts-lowrisc (approved):
LGTM
I guess at the moment, the alert outputs aren't connected to anything so the logic will definitely be optimized away. Once they are connected I'd be surprised if a tool would optimize this far but better safe than sorry!

PR #1344, opened 2021-04-19
Add cast to enum

dawidzim:
Hey!
In this PR I added cast to enum, because we can't assign non-enum to enum LRM 6.19.3

Signed-off-by: Dawid Zimonczyk <dawidz@aldec.com.pl>

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -214,7 +214,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     if (!rst_ni) begin
       // First cycle out of reset there is no last stall, use valid bit to deal with this case
       id_stall_type_last_valid <= 1'b0;
-      id_stall_type_last       <= 1'b0;
+      id_stall_type_last       <= IdStallTypeNone;
       instr_unstalled_last     <= 1'b0;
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -214,7 +214,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     if (!rst_ni) begin
       // First cycle out of reset there is no last stall, use valid bit to deal with this case
       id_stall_type_last_valid <= 1'b0;
-      id_stall_type_last       <= 1'b0;
+      id_stall_type_last       <= id_stall_type_e'(1'b0);
```
rswarbrick, line 217: Thanks for the improvement. Could you change this to use the enum value (`IdStallTypeNone`)? I think that would probably be clearer than using a cast (and matches what we do for `id_instr_category_last` below).

rswarbrick (approved):
Looks good to me. Unless @GregAC can see anything terrible with this, I'll merge when it passes CI.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
dawidzim: sure, now should be ok

GregAC (approved):

PR #1345, opened 2021-04-20
[ibex] Add reset bypass controls

tjaychen:
Signed-off-by: Timothy Chen <timothytim@google.com>

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -106,7 +106,10 @@ module ibex_top #(
     input  logic                         fetch_enable_i,
     output logic                         alert_minor_o,
     output logic                         alert_major_o,
-    output logic                         core_sleep_o
+    output logic                         core_sleep_o,
+
+    // DFT bypass controls
+    input logic                          scan_rst_ni
 );
 
   import ibex_pkg::*;
@@ -716,7 +719,9 @@ module ibex_top #(
 
       .alert_minor_o     (lockstep_alert_minor_local),
       .alert_major_o     (lockstep_alert_major_local),
-      .core_busy_i       (core_busy_local)
+      .core_busy_i       (core_busy_local),
+      .test_en_i         (test_en_i),
+      .scan_rst_ni       (scan_rst_ni)
     );
 
     // Manually buffer the output signals.
@@ -733,6 +738,8 @@ module ibex_top #(
   end else begin : gen_no_lockstep
     assign lockstep_alert_major = 1'b0;
     assign lockstep_alert_minor = 1'b0;
+    logic unused_scan;
+    assign unused_scan = scan_rst_ni;
   end
 
   // TODO - need a config to reset all registers before the lockstep alert can be used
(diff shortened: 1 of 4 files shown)


GregAC:
If we're adding new top-level ports the documentation will need updating: https://github.com/lowRISC/ibex/blob/master/doc/02_user/integration.rst (https://ibex-core.readthedocs.io/en/latest/02_user/integration.html).

It doesn't need a huge amount of explanation 'Allows direct reset control of lockstep core for DFT. Tie to 0 if not using the lockstep core or you aren't employing DFT'.

rtl/ibex_top.sv
```diff
@@ -716,7 +720,10 @@ module ibex_top #(
 
       .alert_minor_o     (lockstep_alert_minor_local),
       .alert_major_o     (lockstep_alert_major_local),
-      .core_busy_i       (core_busy_local)
+      .core_busy_i       (core_busy_local),
+
+      .scan_rst_ni       (scan_rst_ni),
+      .scanmode_i        (scanmode_i)
```
tomroberts-lowrisc, line 726: We already have a test_en pin (for enabling clock gates) which is not actually connected at the moment in top_earlgrey.
Should we replace that with scanmode maybe?

Also just wondering what the difference is between scan_en and scanmode in top_earlgrey.

tjaychen: ah good point. I did not see `test_en`, we should use that one. I'll switch over.
I think scanmode is basically like a `mux select`, whereas `scan_en` becomes the actual enable.  So you may be in scan mode, but not actually be shifting anything yet.

tjaychen:
> If we're adding new top-level ports the documentation will need updating: https://github.com/lowRISC/ibex/blob/master/doc/02_user/integration.rst (https://ibex-core.readthedocs.io/en/latest/02_user/integration.html).
> 
> It doesn't need a huge amount of explanation 'Allows direct reset control of lockstep core for DFT. Tie to 0 if not using the lockstep core or you aren't employing DFT' or similar should suffice.

thanks, let me have a look.

rtl/ibex_top.sv
msfschaffner: yeah `scanmode` is equivalent to `test_en`. the actual `scan_en` (i.e. the shift enable) is inserted automatically by the tool, afaik.

tomroberts-lowrisc (approved):
LGTM - I'll add the scan_rst wiring though ibex_top_tracing as a follow up PR

rtl/ibex_top_tracing.sv
```diff
@@ -132,6 +132,7 @@ module ibex_top_tracing #(
     .rst_ni,
 
     .test_en_i,
+    .scan_rst_ni('0),
```
imphil, line 135: The tracing top is meant to pass all signals through, can either @tomroberts-lowrisc or @tjaychen do a follow-up?

tomroberts-lowrisc:
Way ahead of you! #1347

imphil:
Oh, and you even commented on it. Sorry for the noise!

rtl/ibex_top_tracing.sv
tjaychen: ah thanks for clarifying.
i'll make sure to do that next time.
sorry i should have taken the hint when all else signals are just directly connected to ports..

PR #1348, opened 2021-04-21
[rtl] Break long lines in icache

rswarbrick:
No functional change.

This is needed to get the code AscentLint clean when vendored into OpenTitan.


tomroberts-lowrisc (approved):
LGTM thanks!

PR #1356, opened 2021-05-18
[dv] Add initial coverage plan

GregAC:
We may want to directly name coverpoints in this document so you can easily tie what's implemented to what's written here but for now I've left this out.

### DIFF
--- a/doc/03_reference/coverage_plan.rst
+++ b/doc/03_reference/coverage_plan.rst
@@ -1,0 +1,314 @@
+.. _coverage-plan:
+
+Coverage Plan
+=============
+
+.. note::
+  Work to implement the functional coverage described in this plan is on-going and the plan itself is not yet complete.
+
+.. todo::
+  Branch prediction hasn't yet been considered, this will add more coverage points and alter some others
+
+Introduction
+------------
+Ibex functional coverage is split into two major categories:
+
+* Architectural coverage - which is concerned with instructions being executed and exercising various features of the RISC-V architecture (e.g. PMP) and does not consider the details of how this execution occurs.
+* Microarchitectural coverage - which is concerned with the specifics of the RTL operation, ensuring interesting corner cases are seen along with various micro-architectural events (e.g. the different kinds of stall) and combinations of them.
+
+Architectural coverage is not Ibex specific. It can be determined directly from a trace of executed instructions and is handled by RISCV-DV, details can be found in the `RISCV-DV documentation <https://htmlpreview.github.io/?https://github.com/google/riscv-dv/blob/master/docs/build/singlehtml/index.html#document-coverage_model>`_.
+
+Microarchitectural coverage will probe the Ibex RTL directly and is described here.
+There is some inevitable overlap between architectural and microarchitectural coverage but we aim to minimise it.
+
+Microarchitectural Events and Behaviour
+---------------------------------------
+Below are lists of specific things from the microarchitecture that will be included in functional coverage.
+Each of the points listed below must be covered.
+This will be further combined using cross coverage which is described in the section below.
+
+Instructions
+^^^^^^^^^^^^
+
+Categories
+""""""""""
+``cp_instr_category_id``
+
+Instructions can be grouped into a number of categories.
+Each category exercises different data and control paths in the core.
+For example the ``ADD`` and ``SUB`` instructions are in the same category as they are almost identical for the microarchitecture (both read two registers and write to one, both feed operands to the ALU and take their result from it, both have the same response to interrupts etc; the only difference is the ALU operation).
+
+Instructions can be compressed or uncompressed but that isn't factored into the instruction categories below (excepting for illegal instructions).
+The decompression occurs in the IF stage and is invisible to the ID/EX stage so isn't relevant for instruction execution.
+A separate set of category-agnostic compressed instruction behaviour is considered instead.
+
+An instruction category is sampled at the ID/EX stage (which is where all the varying behaviours actually occur).
+Some categories are just a single instruction, which is named without further description.
+
+
+* **ALU** - All of the reg/reg reg/imm instructions that use the ALU.
+  This is any RISC-V instruction with an opcode of ``7'b0010011`` or ``7'b0110011`` (``ibex_pkg::OPCODE_OP`` and ``ibex_pkg::OPCODE_OP_IMM``) other than the ``MUL*`` and ``DIV*`` family of instructions (from RV32M).
+* **Mul** - Any ``MUL*`` instruction (from RV32M).
+* **Div** - Any ``DIV*`` instruction (from RV32M).
+* **Branch** - Any ``B*`` family branch instruction.
+* **Jump** - ``JAL``/``JALR``
+* **Load** - Any ``L*`` family load instruction.
+* **Store** - Any ``S*`` family load instruction.
+* **CSRAccess** - Any instruction from Zicsr.
+* **EBreakDbg**/**EBreakExc** - An ``EBREAK`` instruction that either enters debug mode (Dbg) or causes an exception (Exc).
+  Which occurs depends upon the setting of ``dcsr.ebreakm`` / ``dcsr.ebreaku`` combined with the privilege level of executed instruction.
+* ``ECALL``
+* ``MRET``
+* ``DRET``
+* ``WFI``
+* ``FENCE.I``
+* **FetchError** - Any instruction that saw a fetch error.
+* **CompressedIllegal** - Any compressed instruction with an illegal encoding.
+* **UncompressedIllegal** - Any uncompressed instruction with an illegal encoding.
+* **CSRIllegal** - Any instruction attempting a CSR access that is not allowed.
+* **PrivIllegal** - Illegal due to privilege level or being in/out of debug mode.
+* **OtherIllegal** - Any other instruction that raises an Illegal instruction exception that isn't in the other categories.
+* **None** - No instruction in ID/EX stage.
+
+Stalls
+""""""
+``cp_stall_type_id``
+
+Not all instructions can see all kinds of stalls.
+A stall category is sampled at the ID/EX stage only (as stalls in IF and WB don't break down into categories).
+
+* **Instr** - A stall caused by a multi-cycle instruction.
+  This can be seen by instructions from categories:
+
+  * **MUL**
+  * **DIV**
+  * **Branch**
+  * **Jump**
+
+* **LdHz** - A load hazard, the instruction in ID/EX depends upon the result of a load that is awaiting a response in writeback.
+  This can be seen by instructions from categories:
+
+  * **ALU**
+  * **Mul**
+  * **Div**
+  * **Branch**
+  * **Jump**
+  * **Load**
+  * **Store**
+  * **CSRAccess**
+
+* **Mem** - Memory stall, the instruction in ID/EX is awaiting a prior memory request to complete before it can begin (to allow precise interrupts on a memory error response). This can be seen for all instruction categories
+
+Privilege Level
+"""""""""""""""
+Ibex can operate at either the M (machine) or U (user) privilege levels.
+Different aspects of the Ibex microarchitecture can be using different privilege levels at once.
+
+* ``cp_priv_mode_if`` - Privilege level of IF stage instruction.
+* ``cp_priv_mode_id`` - Privilege level of ID/EX stage instruction.
+* ``cp_priv_mode_lsu`` - Privilege level of LSU operation (ID/EX privilege level modified by ``mstatus.mprv`` and ``mstatus.mpp`` settings).
+
+Note that the privilege level of the instruction in WB isn't retained by the microarchitecture and is not relevant to coverage.
+Any instruction that reaches WB can be considered bound to retire and any relevant checks and functionality altered by the privilege mode is dealt with at an earlier stage.
+
+Hazards
+"""""""
+Ibex hazards all occur in the interaction between the ID and EX stage.
+
+* RAW Reg - Read after write hazard, instruction in ID/EX reads a register that writeback is writing.
+  Split into two versions:
+
+  * RAW load - Instruction in ID/EX reading from destination of load in writeback.
+    Produces a stall (Category LdHz) and shouldn't forward data.
+    Covered by ``cp_stall_type_id``
+  * ``cp_wb_reg_no_load_hz`` - Instruction in writeback isn't a load.
+    Handled with data forwarding and no stall.
+
+* RAW Load/Store bytes - Load with bytes overlapping a store immediately before it.
+
+State Specific Behaviour
+""""""""""""""""""""""""
+Some instructions will behave differently depending upon the state of the processor (e.g. the privilege level the instruction executes at, CSR settings or whether the processor is in debug mode).
+
+* Instruction illegal in U Mode (cover execution in U and M mode).
+
+  * ``MRET``
+  * ``WFI``
+  * Access to M-mode CSR.
+
+* Debug mode instructions (cover execution in and out of debug mode).
+
+  * ``DRET``
+  * Access to debug CSRs.
+
+    * ``dcsr``
+    * ``dpc``
+    * ``dscratch0``
+    * ``dscratch1``
+
+  * Access to trigger CSRs (also possible in M mode: cover execution in M mode, debug mode and U mode).
+
+    * ``tselect``
+    * ``tdata1``
+    * ``tdata2``
+    * ``tdata3``
+    * ``mcontext``
+    * ``scontext``
+
+* Loads/stores with ``mstatus.mprv`` set and unset.
+* EBreak behaviour in U/M mode with different ``dcsr.ebreakm`` / ``dcsr.ebreaku`` settings.
+
+Pipeline State
+^^^^^^^^^^^^^^
+Each pipeline stage has some associated state.
+
+* ``cp_if_stage_state`` - IF stage full and fetching, full and idle, empty and fetching, or empty and idle.
+  General IF stage full and stalled uninteresting as will only occur when ID stage is full and stalled.
+* ``cp_wb_stage_state`` - WB stage full and stalled, full and unstalled, or empty
+* ``cp_id_stage_state`` - ID stage full and stalled, full and unstalled, or empty.
+* Controller (within ID stage) state machine states
+
+  * Possible transitions between these states.
+    Those marked with a '*' are of particular interest and should be crossed with instruction categories and other coverpoints as appropriate to fully explore the transitions.
+
+    * ``RESET`` -> ``BOOT_SET``
+    * ``BOOT_SET`` -> ``FIRST_FETCH``
+    * ``FIRST_FETCH`` -> ``DECODE``
+    * ``FIRST_FETCH`` -> ``IRQ_TAKEN``
+    * ``FIRST_FETCH`` -> ``DBG_TAKEN_IF``
+    * ``DECODE`` -> ``FLUSH`` *
+    * ``DECODE`` -> ``DBG_TAKEN_IF`` *
+    * ``DECODE`` -> ``IRQ_TAKEN`` *
+    * ``IRQ_TAKEN`` -> ``DECODE``
+    * ``DBG_TAKEN_IF`` -> ``DECODE``
+    * ``DBG_TAKEN_ID`` -> ``DECODE``
+    * ``FLUSH`` -> ``DECODE`` *
+    * ``FLUSH`` -> ``DBG_TAKEN_ID``
+    * ``FLUSH`` -> ``WAIT_SLEEP``
+    * ``FLUSH`` -> ``IRQ_TAKEN`` *
+    * ``FLUSH`` -> ``DBG_TAKEN_IF`` *
+    * ``WAIT_SLEEP`` -> ``SLEEP``
+    * ``SLEEP`` -> ``FIRST_FETCH``
+
+Exceptions/Interrupts/Debug
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Exceptions, interrupts and debug entry can all cause control flow changes combined with CSR writes and privilege level changes and work quite similarly within the controller but not identically.
+Furthermore they can all occur together and must be appropriately prioritised (consider a instruction with hardware trigger point matching it, that causes some exception and an interrupt is raised the cycle it enters the ID/EX stage)
+
+* Exception from instruction fetch error (covered by the **FetchError** instruction category).
+* Exception from instruction PMP violation.
+* Exception from illegal instruction (covered by the illegal instruction categories).
+* Exception from memory fetch error.
+* Exception from memory access PMP violation.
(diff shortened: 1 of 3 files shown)


GregAC:
Direct link to rendered HTML version: https://ibex-core--1356.org.readthedocs.build/en/1356/03_reference/coverage_plan.html

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Introduction
+------------
+
+Ibex functional coverage is split into two major categories:
+
+* Architectural coverage - which is concerned with instructions being executed and exercising various features of the RISC-V architecture (e.g. PMP) and does not consider the details of how this execution occurs.
+* Microarchitectural coverage - which is concerned with the specifics of the RTL operation, ensuring interesting corner cases are seen along with various micro-architectural events (e.g. the different kinds of stall) and combinations of them.
+
+Architectural coverage is not Ibex specific and can be determined directly from a trace of executed instructions and is handled by RISC-V DV, details can be found in the `RISC-V DV documentation <https://htmlpreview.github.io/?https://github.com/google/riscv-dv/blob/master/docs/build/singlehtml/index.html#document-coverage_model>`_.
```
rswarbrick, line 20: Nit: Long sentence with chained "and"s. Maybe "... not Ibex specific. It can be..." ?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+This will be further combined using cross coverage which is described in the section below.
+
+Instructions
+^^^^^^^^^^^^
+
+Categories
+""""""""""
+Instructions can be grouped into a number of categories.
+Each category represents a different kind of execution.
+For example the ``ADD`` and ``SUB`` instructions are in the same category as they are almost identical for the microarchitecture (both read two registers and write to one, both feed operands to the ALU and take their result from it, both have the same response to interrupts etc, the only difference is the ALU operation).
```
rswarbrick, line 39: Nit: The commas look a bit odd here. Maybe replace the last with a semicolon? "(both read two registers and write to one, both feed operands to the ALU and take their result from it, both have the same response to interrupts etc; the only difference is the ALU operation)."

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Instructions can be compressed or uncompressed but that isn’t factored into the instruction categories below (excepting for illegal instructions).
+The decompress occurs in the IF stage and is invisible to the ID/EX stage so isn’t relevant for instruction execution.
```
rswarbrick, line 42: Nit: The noun is decompress**ion**.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+A separate set of category agnostic compressed instruction behaviour is considered instead.
```
rswarbrick, line 43: Nit: Maybe "category-agnostic"? Otherwise it's a bit hard for a reader to parse.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+An instruction category is sampled at the ID/EX stage (which is where all the varying behaviours actually occur).
+Some categories are just a single instruction which is named without further description.
```
rswarbrick, line 46: Nit: Add a comma before "which" (because this is extra information about the instruction, not a condition on which instruction you're talking about)

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+
+* ALU - All of the reg/reg reg/imm instructions that use the ALU.
```
rswarbrick, line 49: I wonder whether it would be clearer to `monospace` or **bold** the category names ("these are terms of art!")?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  This is any RISC-V instruction with an opcode of ``7’b0010011`` or ``7’b0110011`` (``ibex_pkg::OPCODE_OP`` and ``ibex_pkg::OPCODE_OP_IMM``) other than the ``MUL*`` and ``DIV*`` family of instructions (from RV32M).
+* Mul - Any ``MUL*`` instruction (from RV32M).
+* Div - Any ``DIV*`` instruction (from RV32M).
+* Branch - Any ``B*`` family branch instruction.
+* Jump - ``JAL``/``JALR``
+* Load - Any ``L*`` family load instruction.
+* Store - Any ``S*`` family load instruction.
+* CSRAccess - Any instruction from Zicsr.
+* EBreakDbg/EBreakExc- An ``EBREAK`` instruction that either enters debug mode (Dbg) or causes an exception (Exc).
```
rswarbrick, line 58: Nit: Missing space before the hyphen. That said, if you're happy to distinguish the headings with a font change (as suggested above), a colon might be enough to separate things?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* UncompressedIllegal - Any uncompressed instruction with an illegal encoding.
+* CSRIllegal - Any instruction attempting a CSR access that is not allowed.
+* PrivIllegal - Illegal due to privilege level or being in/out of debug mode.
+* OtherIllegal - Any other instruction that raises an Illegal instruction exception that isn’t in the other categories.
+* None - No instruction in ID/EX stage.
+
+Stalls
+""""""
+Not all instructions can see all kinds of stalls.
+A stall category is sampled at the ID/EX stage only (as stalls in IF and WB don’t break down into categories).
```
rswarbrick, line 76: Are we tracking stalls in IF and WB separately? I couldn't see them mentioned elsewhere?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+* Instr - A stall cause by a multi-cycle instruction.
```
rswarbrick, line 78: Nit: cause**d**

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  This can be seen by instructions from categories:
+
+  * MUL
+  * DIV
+  * Branch
+  * Jump
```
rswarbrick, line 84: I wonder whether these lists are actually needed. Presumably we'll have `illegal_bins` lines somewhere: maybe that's a better place to put "hey, this can't happen for any other categories"?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  * Store
+  * CSRAccess
+
+* Mem - Memory stall, the instruction in ID/EX is awaiting a prior memory request to complete before it can begin (to allow precise interrupts on a memory error response). This can be seen for all instruction categories
+
+Privilege Level
+"""""""""""""""
+
+Ibex can operate at either the M (machine) or U (user) privilege levels.
+Different aspects of the Ibex microarchitecture can be using different privilege levels at once.
```
rswarbrick, line 104: I think you mean different stages of the pipeline? Maybe saying that explicitly would be clearer? (Rather than "the ALU is in M mode, but the icache is in U mode"!)

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+* Privilege level of IF stage instruction.
+* Privilege level of ID/EX stage instruction.
```
rswarbrick, line 107: Do we need to track WB separately here? If not, maybe add a note to explain why?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Privilege level of LSU operation (ID/EX privilege level modified by `mstatus.mprv` and `mstatus.mpp` settings).
+
+Hazards
+"""""""
+Ibex hazards all occur in the interaction between the ID and EX stage.
```
rswarbrick, line 112: Nit: You've added an extra line after headings above here. I probably prefer this more condensed style, but it's maybe a good idea to be consistent.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+* RAW Reg - Read after write hazard, instruction in ID/EX reads a register writeback is writing.
```
rswarbrick, line 114: This should probably be "... a register **that** writeback ...", otherwise the sentence is quite hard to read.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  Produces a stall (Category LdHz) and shouldn’t forward data.
+
+State Specific Behaviour
+""""""""""""""""""""""""
+Some instructions will behave differently depending upon the state of the processor (e.g. the privilege level the instruction executes at, CSR settings or whether the processor is in debug mode).
+
+* Instruction illegal in U Mode (cover execution in U and M mode).
+
+  * Mret
+  * Wfi
```
rswarbrick, line 126: Nit: I think you monospaced instruction names above?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  * Access to M-mode CSR.
+
+* Debug mode only instructions (cover execution in and out of debug mode).
```
rswarbrick, line 129: Nit: I don't think you need the "only" here? Reading quickly, it currently looks like "Debug mode only ever \<verb\>".

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+  * Dret
+  * Access to debug CSRs.
+
+    * ``dcsr``
+    * ``dpc``
+    * ``dscratch0``
+    * ``dscratch1``
+
+  * Access to trigger CSRs (also possible in M mode, cover execution in M mode, debug mode and U mode).
```
rswarbrick, line 139: The stuff in parens looks like a list of 4 things. Maybe the first comma should be a colon?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Loads/stores with ``mstatus.mprv`` set and unset.
+* EBreak behaviour in U/M mode with different ``dcsr.ebreakm`` / ``dcsr.ebreaku`` settings.
+
+Pipeline State
+^^^^^^^^^^^^^^
+Each pipeline stage has some associated state.
+
+* IF stage full or empty
+* WB stage full or empty
+* ID stage full or empty -  effectively covered by instruction categories (Category None meaning the stage is empty).
```
rswarbrick, line 157:
```suggestion
* ID stage full or empty. This is effectively covered by instruction categories, since a category of None means the stage is empty.
```

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+    * ``SLEEP`` -> ``FIRST_FETCH``
+
+Exceptions/Interrupts/Debug
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Exceptions, interrupts and debug entry can all cause control flow changes combined with CSR writes and privilege level changes and work quite similarly within the controller but not identically.
+Furthermore they can all occur together and must be appropriately prioritised (consider a instruction with hardware trigger point matching it, that causes some exception and an interrupt is raised the cycle it enters the ID/EX stage)
+
+* Exception from instruction fetch error (covered by FetchError instruction category).
+* Exception from instruction PMP violation.
+* Exception from illegal instruction (covered by Illegal instruction categories).
```
rswarbrick, line 189: Nit: Maybe "covered by **the** FetchError..." and then "covered by the various illegal instruction categories"? I don't think it makes sense to capitalise "Illegal", because that's not the name of a category.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Exception from memory fetch error.
+* Exception from memory access PMP violation.
+* Unaligned access cases (both accesses saw error, first or second only saw error or neither saw error) for both kinds of memory exceptions.
```
rswarbrick, line 192: Nit: I'd use an Oxford comma for this list, otherwise the grouping of the "or"s is ambiguous.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Interrupt raised/taken.
+
+  * Interrupt raised/taken for each available interrupt line (for wider cross coverage which interrupt is raised/taken is not relevant other than NMI vs non NMI).
```
rswarbrick, line 195:
```suggestion
  * Interrupt raised/taken for each available interrupt line.
    For cross coverage, the precise interrupt that's raised/taken is not relevant and only needs to be grouped by NMI versus non-NMI.
```

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  * Interrupt raised/taken first cycle of instruction or not.
```
rswarbrick, line 196: How should I interpret this in a pipelined processor? Do you mean to cross "the stage that an instruction is in" with "an interrupt is raised/taken"?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  * NA4
+  * NAPOT
+
+* Each region configured with all possible permissions including locked/unlocked.
+
+  * Different permissions with MML enabled and disabled, separate cover points for R/W/X/L values with and without MML.
+
+* Access fail & pass.
+
+  * All combinations of unaligned access split across a boundary, both halves pass, neither pass, one does & one doesn’t.
```
rswarbrick, line 218: Maybe the last item should be expanded to "just the first passes, just the second passes"?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+  * Higher priority entry allow access that lower priority entry prevents.
```
rswarbrick, line 219: Nit: allow**s**

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+  * MML - machine mode lockdown policy.
+
+    * Try to disable when enabled.
+
+  * Access close to PMP region modification that allows/disallows that access.
+
+CSRs
+^^^^
+Basic read/write functionality must be tested on all implemented CSRs.
```
rswarbrick, line 242: Also, read/write for an unimplemented CSR?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Stalled awaiting an instruction fetch.
+* Double fault.
+* Awake from sleep.
+* Interrupt/Debug whilst entering sleep.
+
+Cross Coverage
+--------------
+Much of the more complex behaviour lies at the combination of the individual microarchitectural behaviours above.
+Cross coverage is used to capture that.
+There are some broad crosses containing many bins aiming to capture all combinations of some generalised behaviours as well as some more specific ones to capture all combinations of behaviours focussed on a particular area.
```
rswarbrick, line 269: Nit: I think a double 's' in focused is very unusual in technical writing. A quick Google suggests that it's on the way out, even for British English. Maybe just one?

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Cross coverage will be intentionally broad.
+Where it is proving hard to hit particular bins they will be reviewed in more detail to determine if they’re impossible to hit or if simply hard to hit and whether hitting them provides meaningful gains to verification quality.
```
rswarbrick, line 272: Nit: Non-ASCII apostrophe. (And I think there are some more: find-and-replace?)

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+* Instructions Categories x ID/EX Privilege level
+* Instruction Categories x Stall Categories
+
+  * Illegal bins will be used to exclude instruction and stall categories that cannot occur.
+
+* Instruction Categories x Hazards
+* Instruction Categories x Debug Mode
+* Instruction Categories x IF/WB full or empty
+
+  * Covers all possibilities of instruction combinations that could fill the pipeline, full/empty for IF/WB suffices to cover this as all the interesting per instruction behaviour occurs in ID/EX.
```
rswarbrick, line 287: I think you need a sentence break before "full/empty"? I read it as a list at first.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,299 @@
...
+
+* Instruction Categories x Controller state transitions of interest
+* Interrupt taken/Debug mode entry/Pipe flush x instruction unstalled x instruction category
+
+  * Three seperate cross coverage groups one for interrupt, debug and pipe flush.
```
rswarbrick, line 292: sep**a**rate. Also "... coverage groups: one for ..." to sound less breathless.

doc/03_reference/coverage_plan.rst
GregAC: Good point. For IF there's a point under `Miscellaneous` saying 'Stalled waiting an instruction fetch' I think I can move this here and rephrase so it's relative to ID/EX and hence a stall category.

For WB we have the `Mem` and `LdHz` stall categories, again relative to the ID/EX stage but we'll see these if WB is stalled (currently WB will only stall if it's awaiting a memory result).

WB stalled on its own whilst the rest of the pipeline is empty isn't a hugely interesting point and we've already got coverage for WB valid and nothing else valid to cover it. I'll add some further wording here to explain.

GregAC: I guess there's a slight difference between WB alone valid and writing back this cycle vs WB alone valid and stalled this cycle. Maybe I should expand the pipeline state section so we've got stage full & ready/full & stalled/empty and we do a big cross between those for the IF & WB stages and the ID/EX stall categories and maybe instruction categories.

doc/03_reference/coverage_plan.rst
GregAC: The LSU can be using a different privilege level to ID/EX (thanks to the `mstatus.mprv` bit) and isn't strictly associated with a pipeline stage (rather it interacts with both ID/EX and WB).

doc/03_reference/coverage_plan.rst
GregAC: We don't, I'll add a note.

doc/03_reference/coverage_plan.rst
GregAC: I'd prefer to leave them in for now as this will be the document used to finish off the coverage implementation so I don't want to accidentally loose anything. Once the implementation is done we can consider removing this (e.g. if we think comments in the coverage implementation itself are sufficient).

GregAC:
Thanks for the review @rswarbrick I've addressed all of the smaller comments. Still need to do a little work on stalling behaviour coverage.

doc/03_reference/coverage_plan.rst
rswarbrick: Oh, I see. Ok.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,302 @@
...
+  Branch prediction hasn't yet been considered, this will add more coverage points and alter some others
+
+Introduction
+------------
+Ibex functional coverage is split into two major categories:
+
+* Architectural coverage - which is concerned with instructions being executed and exercising various features of the RISC-V architecture (e.g. PMP) and does not consider the details of how this execution occurs.
+* Microarchitectural coverage - which is concerned with the specifics of the RTL operation, ensuring interesting corner cases are seen along with various micro-architectural events (e.g. the different kinds of stall) and combinations of them.
+
+Architectural coverage is not Ibex specific. It can be determined directly from a trace of executed instructions and is handled by RISC-V DV, details can be found in the `RISC-V DV documentation <https://htmlpreview.github.io/?https://github.com/google/riscv-dv/blob/master/docs/build/singlehtml/index.html#document-coverage_model>`_.
```
udinator, line 19: nit: `RISCV-V DV` -> `RISCV-DV`

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,302 @@
...
+Each of the points listed below must be covered.
+This will be further combined using cross coverage which is described in the section below.
+
+Instructions
+^^^^^^^^^^^^
+
+Categories
+""""""""""
+Instructions can be grouped into a number of categories.
+Each category represents a different kind of execution.
```
udinator, line 36: nit: would it be better to say something like `each category of instructions exercises a different datapath in the core` ? It's a little confusing upon first read imo

doc/03_reference/coverage_plan.rst
GregAC: Pipeline state now has each stage with three different states, full & stalled, full & unstalled, empty (for IF instead of full & stalled, it's awaiting new instruction). There's a new cross for these states plus instruction categories.

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,310 @@
...
+    * ``scontext``
+
+* Loads/stores with ``mstatus.mprv`` set and unset.
+* EBreak behaviour in U/M mode with different ``dcsr.ebreakm`` / ``dcsr.ebreaku`` settings.
+
+Pipeline State
+^^^^^^^^^^^^^^
+Each pipeline stage has some associated state.
+
+* IF stage awaiting fetch, full, or idle.
```
GregAC, line 156: I'm going to change this to four categories:
* awaiting fetch
* full and fetching
* full and idle
* empty

doc/03_reference/coverage_plan.rst
```diff
@@ -0,0 +1,310 @@
...
+  * All bins containing instruction categories other than **None** ignored when ID/EX stage is empty.
+
+* Instructions Categories x ID/EX Privilege level
+* Instruction Categories x Stall Categories
+
+  * Illegal bins will be used to exclude instruction and stall categories that cannot occur.
+
+* Instruction Categories x Hazards
+* Instruction Categories x Debug Mode
+* Instruction Categories x IF/WB full or empty
```
GregAC, line 296: This cross is just the pipeline stage state x instruction categories cross above but with less stuff, I'll remove it.

vogelpi:
Hey @GregAC and @udinator ,
it has been a while. Do you know what the current status of this PR is? Does it need more review? Or are there further changes required first?

GregAC:
This is still a work in progress, various changes made in response to @rswarbrick's first review have been pushed plus some more stuff I'm working on. The work is paused at the minute whilst I work on the cosim.

GregAC:
@rswarbrick can we get this merged?

rswarbrick (approved):
Good idea. Let's merge this as-is. It's clearly better than nothing and we can polish as needed as a follow-up.

PR #1363, opened 2021-05-26
[config] Add an Opentitan configuration

### LINKED ISSUE #1362: Produce a configuration matching OpenTitan
We currently don't have a configuration for Ibex that matches what we're using in OpenTitan experimental-maxperf-pmp-bmfull-icache is close, but we don't enable the B extension. And we probably need to add the SecureIbex flag to the config as well.

tomroberts-lowrisc:
This adds the configuration instantiated in Opentitan and adds
missing prim support to allow the TB to run that configuration.

Resolves #1362


ibex_configs.yaml
```diff
@@ -20,6 +20,23 @@ small:
   PMPEnable                : 0
   PMPGranularity           : 0
   PMPNumRegions            : 4
+  SecureIbex               : 0
+
+# Configuration to match that used in the Opentitan project
```
rswarbrick, line 25: Nit: Open**T**itan

rswarbrick (approved):
Looks good. I wonder whether we should change our CI targets? I'd have thought that the OT configuration is pretty high priority?

imphil:
For CI, please add the config to https://github.com/lowRISC/ibex/blob/master/azure-pipelines.yml#L94. (I'll take care of the private CI config after the merge here, I have some changes for it anyways.)

imphil (approved):
Thanks Tom!

tomroberts-lowrisc:
I have replaced the most similar config in the list since I didn't think we needed to run loads more at the moment. Happy to put it back if anyone objects.

imphil:
I'm fine with reducing the number of options, but if we have a named config, we should also test it.

So I'd either replace experimental-maxperf-pmp-bmfull-icache with opentitan, or add both of them to the testing.

If you go for replacing them, also change the default config in dv/uvm/core_ibex/Makefile (variable IBEX_CONFIG).

tomroberts-lowrisc:
Good point! amended the default config to `opentitan`

Issue #1365, opened 2021-05-27
I git clone the riscv-gnu- toolchain and configured it and then i got the error as follows


AkhilBR98:
akhil@uhdct3-OptiPlex-3040:~$ cd risc/lowrisc/ibex
akhil@uhdct3-OptiPlex-3040:~/risc/lowrisc/ibex$ make build-arty-35 program-arty cd examples/sw/led && make
make[1]: Entering directory '/home/akhil/risc/lowrisc/ibex/examples/sw/led'
riscv32-unknown-elf-gcc -march=rv32imc -mabi=ilp32 -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -Wall -g -Os -MMD -c  -o led.o led.c
make[1]: riscv32-unknown-elf-gcc: Command not found
Makefile:50: recipe for target 'led.o' failed
make[1]: *** [led.o] Error 127
make[1]: Leaving directory '/home/akhil/risc/lowrisc/ibex/examples/sw/led'
Makefile:65: recipe for target 'examples/sw/led/led.vmem' failed
make: *** [examples/sw/led/led.vmem] Error 2
akhil@uhdct3-OptiPlex-3040:~/risc/lowrisc/ibex$

AkhilBR98:
i have used https://github.com/llowRISC/risc-gnu-toolchain this

rswarbrick:
This message means that `riscv32-unknown-elf-gcc` (part of the toolchain) is not on your path. You'll need to fix that by setting up the PATH environment variable. To check that you've done so, run `which riscv32-unknown-elf-gcc`. If it prints out the path to the binary, you're good. If not, you'll need to do more (non-Ibex-related) setup.

AkhilBR98:
But this toolchain doesn't have riscv32-unknown-elf-gcc in bin and  it has riscv64-unknown-elf-gcc so i could not add this to the path

AkhilBR98:
Can tell me which toolchain to use

rswarbrick:
Use a release from here: https://github.com/lowRISC/lowrisc-toolchains. Looks like we've got a missing item in our "System and Tool Requirements" page: I'll get to that in a minute.

AkhilBR98:
I configured riscv toolchain and then i ran the build command i got the below error
akhil@akhil-VirtualBox:~/tbex$ make build-arty-100 program-arty
cd examples/sw/led && make
[...]

PR #1366, opened 2021-05-27
Describe where to find a RISC-V toolchain

rswarbrick:
Triggered by the question at #1365.


rswarbrick:
Oh, wait a sec. This is described in great detail in the verification section. I'll change this to point there.

imphil:
If we want to go down that route we need to give more information on a "matching" toolchain. 

And probably add something that the lowrisc toolchains are really unsupported internal ones, nothing people should rely on.

rswarbrick:
> And probably add something that the lowrisc toolchains are really unsupported internal ones, nothing people should rely on.

Well, fair enough, but we don't support anything else at the moment :-) I'll just change this to point at another part of our existing documentation. Maybe that, in turn, could be tidied up in a further PR.

GregAC (approved):

PR #1370, opened 2021-06-01
Added Width parameter to prim_buf module

dawidzim:
Hey!
In this pull request I added `Width` parameter in `prim_buf`.


rswarbrick:
Could you explain what this is for? These files are going to be dropped in the near(ish) future anyway (see the URL in the comment just above your change).

dawidzim:
I know this file is temporary, but these changes are needed to avoid problem in simulator

rswarbrick:
Sorry, but you're going to need to explain this a bit further before we can merge it. What problems do you see? Is the existing code incorrect? Why does this patch fix them?

dawidzim:
In current version we have problem with assign wire->[0:0]wire via wildcard(.*). There are 2 paths to avoid this problem: change wildcard to explicit port connection or add parameter to this wire (then will be [0:0]wire->[0:0]wire)

rswarbrick:
I don't think this answers my questions (or, if it does, I'm being dense!). Is the existing code buggy or are you trying to work around a limitation in Riviera? If the former, can you explain which part of the IEEE 1800 spec we've violated?

dawidzim:
Yes, this is work around for limitation in Riviera

imphil:
@dawidzim Thanks for taking the time for this pull request. I took the liberty of updating it with a slightly longer commit message, explaining where the prim_buf.sv file is actually coming from.

I did remove the explicit port assignments and left the wildcard assignment in place after re-reading your comment https://github.com/lowRISC/ibex/pull/1370#issuecomment-852764136. Can you try with Riviera-PRO if only adding the Width parameter is sufficient, since you then don't have a wire -> [0:0] wire connection any more?

dawidzim:
Works fine :) Thanks!

imphil (approved):

imphil:
Great, thanks for checking!

PR #1371, opened 2021-06-01
Changes simulation command for Riviera 2021.04

dawidzim:
Hey!
Few changes in Riviera-PRO compilation and simulation command

### DIFF
--- a/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
+++ b/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
@@ -107,14 +107,12 @@
       - "vlib <out>/work"
       - "vlog -work <out>/work
         <cmp_opts>
-        +incdir+<ALDEC_PATH>/vlib/uvm-1.2/src
+        -uvmver 1.2
         +define+UVM
-        -l uvm_1_2
         -f ibex_dv.f"
-
   sim:
     cmd: >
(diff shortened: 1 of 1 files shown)


rswarbrick:
Please can you explain why these changes are needed? Was the previous command wrong? Or is it to match a newer version of Riviera? Or something else?

dawidzim:
I am making these changes to be compatible with the new Riviera-PRO version

rswarbrick:
Ok. Do the new command lines also work for older versions of Riviera? Can we update some documentation to explain what this works with?

rswarbrick:
Well, we can't merge this commit at the moment because we, the maintainers, don't understand what's going on. Please could you explain how these command line changes relate to different Riviera versions? In particular, if someone has an older version that worked with the existing code, will this change break things for them?

dawidzim:
At this point verification of Ibex works only  with version 2021.04. Older version have limitations and code will compile with errors.

The command which I added has worked since version 2020.04

rswarbrick:
Ok, great. Please could you update the commit message with that information (so that people looking through the history in the future know what's going on)? Then we can merge this.

dawidzim:
@rswarbrick Is it ok now?

rswarbrick:
I mean: please force-push to update the commit message in the git commit.

rswarbrick:
Thanks for the update. I've just force-pushed something more like what I meant (with a standard git commit message format, plus a long-form explanation that will make sense to a future reader). Let me know if you don't want this going in. Otherwise, I'll merge once this passes CI.

dawidzim:
That looks great :) Thanks!

rswarbrick (approved):

rswarbrick:
Actually, no point in waiting for CI: we don't do anything that uses this bit of the file. Merging now!

PR #1372, opened 2021-06-02
Teach core_ibex testbench to run the comparison step in parallel

rswarbrick:
The interesting commit is the second last. Here's its message:

This patch teaches Make which tests we're actually running (via the
`list_tests.py` script), which means that we can compare the ISS and RTL
results in parallel rather than serially.

There's a bit of duplicated code (both `list_tests.py` and `sim.py`
currently contain the code to get a list of tests and then filter by
Ibex configuration), but this should go away with a later patch that
runs the RTL simulations in parallel in a similar way.

Note: This might seem a little silly: trace comparison takes way less
time than the RTL simulation! The point is that it's probably easier
to work "from the bottom" than to start by parallelising the
simulations themselves.

@GregAC: This is in draft because it probably clashes quite badly with the stuff you're thinking about as well. I'm very happy to rebase onto your changes (it's my fault for stepping on your toes!). However, it would be helpful if the comparison stuff could work "per-file", passing the pass/fail info to the code that gathers things up through the filesystem, rather than relying on passing a tuple around in the Python. Does that make sense? *(EDIT: We've chatted about it and agreed that it makes sense to merge this first)*

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -211,6 +211,18 @@ FORCE:
 vars-prereq = $(if $(call vars-differ,$(1),$(2),$(3)),FORCE,)
 
 ###############################################################################
+# Get a list of tests and seeds
+#
+# Run list_tests.py to list the things we need to run in the format
+# TESTNAME.SEED and store it in a variable.
+tests-and-seeds := \
+  $(shell ./list_tests.py \
+	--start_seed $(SEED) \
+	--test "$(TEST)" \
+	--iterations $(ITERATIONS) \
+	--ibex-config $(IBEX_CONFIG))
+
+###############################################################################
 # Generate random instructions
 #
 # This depends on the vendored in code in $(GEN_DIR). It also depends on the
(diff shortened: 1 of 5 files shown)


rswarbrick:
(Wow, I've been rather inconsistent with my commit message tags! When I rebase, I'll rename them to match each other)

GregAC:
The general concept behind what I'm doing is you get a bunch of test results in some data structure and output them in whatever format you wish. This should fit into what you've done the issue is how to elegantly implement `collect_results.py`. When you're just dealing with plain text output it's easy enough for `compare.py` to output a bunch of individual logs then stitch them together later. When you throw multiple formats into the mix it's more awkward. E.g. to add junit XML output you'd need to output a junit XML fragment per test then `collect_results.py` needs to gather them together and organise them into test suites. You end up with the details of how you output your junit XML spread across two different programs and `collect_results.py` may need to do some limited parsing of the XML fragments to do it's job (e.g. finding the test name to identify a test suite).

It can also make use of support libraries difficult, I was looking at https://pypi.org/project/junit-xml/ for junit XML output. It doesn't do separate testcase fragment output itself so we couldn't use it.

Maybe `compare.py` should write some serialised version of the `_CompareResult` object, then `collect_results.py` reads it back in and all of the `_CompareResult` objects are processed together at once for the final result?

It'd be easy to pickle `_CompareResult` then unpickle but I think we'd be better off with a YAML serialization so it's easily understandable. `_CompareResult` should remain pretty simple so YAML serialize/deserialize won't be too hard, especially if we're happy with just a few asserts checking the format is what we expect rather than good error handling (we're not expecting hand-written YAML to process).

rswarbrick:
Sounds sensible to me. I'm not sure that combining junit files is all that hard: it looks like you can just parse them all and then dump them in the output as sibling children of a new tree. For example, https://gist.github.com/cgoldberg/4320815 (although the comments below it suggest it's not perfect!)

GregAC:
I guess it's not too hard I think you're just better off with your final step being list of `_CompareResult` -> desired report so everything to do with the report format is in one file without need to parse intermediate fragments. Would also allow easy filtering.

rswarbrick:
I'm not sure that's quite right: if you want to run the different comparisons at different times, you will need to write their results to separate files. I guess we're discussing which of these approaches to take?
1. Each comparison generates a junit file. The collation step loads these junit files and combines them in some way.
2. Each comparison generates a YAML file (format TBD). The collation step loads these YAML files and serialises to junit XML.
3. Each comparison generates a Python pickle file. The collation step loads these blobs and serialises to junit XML.
Option (1) seems the least mysterious to me.

GregAC:
I'd prefer option 2. If you're fixed on junit only then 1. works well enough but if you want multiple formats (and I think we want at least two, junit and plain text) then I think 1. looks more complex. In particular if you want generic filtering that can be applied to any report format you need to do some junit/text -> `_CompareResult` like structure that your generic filtering code can work with.

rswarbrick:
Yep, makes sense. Especially if you care about aggregating the plain text stuff as well.

GregAC (approved):
LGTM!

I do wonder if `collect_results.py` should output a list of failing tests at least, perhaps with their failure message. Though as I'm planning to do a follow up that reworks that part anyway probably wasted work.

rswarbrick:
Good idea about the list of failing tests. Let's do a follow-up with that (possibly as a PR to follow your one?)

PR #1376, opened 2021-06-03
Testing producing junit XML

GregAC:
Modifications to sim.py will still need some work, this is primarily to test the integration with azure JUnit reporting.

### DIFF
--- a/ci/azp-private.yml
+++ b/ci/azp-private.yml
@@ -26,6 +26,7 @@ resources:
     type: github
     endpoint: lowRISC
     name: lowrisc/lowrisc-private-ci
+    ref: refs/heads/junit_test
 
 extends:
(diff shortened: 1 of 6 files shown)


GregAC:
@rswarbrick this gives a sneak peak at the modifications I've done so far. I would anticipate keeping my `compare_test_run` (modified to deal with the changed arguments) with an extra bit of python to serialise/deserialise the `TestRunResult` to and from yaml to do the results gathering for the final report.

rswarbrick:
Sounds good. Do you have strong feelings about what order things are merged? I'm happy either way.

GregAC:
Might be easiest for you to merge first, assuming you don't need any further work on the reporting stuff you've already done (no sense in you doing further work there only for me to remove it).

PR #1383, opened 2021-06-17
Various fixes for known simulation issues

tomroberts-lowrisc:
Small fixes to the RTL and testbench to fix issues seen in nightly runs

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -318,6 +318,7 @@ $(metadata)/instr_gen.iss.stamp: \
      ${TEST_OPTS} \
      --iss="${ISS}" \
      --iss_opts="${ISS_OPTS}" \
+     --isa="${ISA}" \
      ${RISCV_DV_OPTS}
 	$(call dump-vars,$(metadata)/iss-vars.mk,iss,$(iss-var-deps))
 	@touch $@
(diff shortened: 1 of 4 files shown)


rtl/ibex_decoder.sv
```diff
@@ -748,7 +748,7 @@ module ibex_decoder #(
           // First evaluate the branch condition
           alu_op_a_mux_sel_o  = OP_A_REG_A;
           alu_op_b_mux_sel_o  = OP_B_REG_B;
-        end else begin
+        end else if (!BranchTargetALU) begin
```
GregAC, line 751: Just to confirm did this fix a bug where we have the Branch Target ALU and data independent timing is enabled? Otherwise with `BranchTargetALU` set if we've got a branch we should never see `instr_first_cycle_i` unset whilst the instruction remains in the ID stage.

GregAC (approved):

dv/uvm/core_ibex/Makefile
```diff
@@ -318,6 +318,7 @@ $(metadata)/instr_gen.iss.stamp: \
      ${TEST_OPTS} \
      --iss="${ISS}" \
      --iss_opts="${ISS_OPTS}" \
+     --isa="${ISA}" \
```
rswarbrick, line 321: Oops, thanks!

rswarbrick (approved):

rtl/ibex_decoder.sv
tomroberts-lowrisc: Correct

PR #1386, opened 2021-06-18
Append logs for tests to regr.log

rswarbrick:
Commit 700f29b changed things so that details of the various tests
that ran ended up in separate files. This is nice (and important for
running things in parallel), but isn't massively helpful if you use
regr.log to understand what happened from a CI run!

This patch adds the logs again, splitting them up so that failing
tests come before passing ones (since you usually just care about the
failures).

### DIFF
--- a/dv/uvm/core_ibex/collect_results.py
+++ b/dv/uvm/core_ibex/collect_results.py
@@ -5,6 +5,7 @@
 
 import argparse
 import sys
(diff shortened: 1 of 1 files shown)


rswarbrick:
An example of the output:
image

imphil (approved):
Thanks!

PR #1388, opened 2021-06-22
Update the meaning of fetch_en_i

tomroberts-lowrisc:
Minor rtl and testbench updates to extend the functionality of fetch_en_i

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -128,6 +128,7 @@ module ibex_core import ibex_pkg::*; #(
 `endif
 
     // CPU Control Signals
+    input  logic                         fetch_enable_i,
     output logic                         alert_minor_o,
     output logic                         alert_major_o,
     output logic                         core_busy_o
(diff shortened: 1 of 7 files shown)


rswarbrick (approved):

rtl/ibex_core.sv
```diff
@@ -475,6 +477,9 @@ module ibex_core import ibex_pkg::*; #(
   // Qualify the instruction request with PMP error
   assign instr_req_o = instr_req_out & ~pmp_req_err[PMP_I];
 
+  // fetch_enable_i can be used to stop the core fetching new instructions
+  assign instr_req_gated = instr_req_int & fetch_enable_i;
```
GregAC, line 481: Is it worth flopping fetch_enable_i? This introduces a new feed-through path from `fetch_enable_i` to `instr_req_o` (and maybe `instr_addr_o` too) otherwise.

GregAC (approved):
LGTM, we should ensure we do some testing with `fetch_enable_i` randomly toggled.

rtl/ibex_core.sv
GregAC: Also not introduced by this PR but instr_req is a bit overloaded in this file. Meaning both something related to the request into the IF stage as well as the actual outbound imem request and they are not the same.

tomroberts-lowrisc: Yeah I wondered about flopping it but left it up to the user in the end. In the OT use-case the external input comes straight out of a 2 flop synchronizer so should be ok there.

PR #1389, opened 2021-06-24
Ibex dv report format

GregAC:
Here's what the intermediate YAML files look like:

```
name:            riscv_arithmetic_basic_test
idx:             0
seed:            10547
binary:          out/seed-10547/instr_gen/asm_test/riscv_arithmetic_basic_test_0.o
uvm_log:         out/seed-10547/rtl_sim/riscv_arithmetic_basic_test.10547/sim.log
rtl_trace:       out/seed-10547/rtl_sim/riscv_arithmetic_basic_test.10547/trace_core_00000000.log
rtl_trace_csv:   out/seed-10547/rtl_sim/riscv_arithmetic_basic_test.10547/trace_core_00000000.csv
iss_trace:       out/seed-10547/instr_gen/spike_sim/riscv_arithmetic_basic_test.0.log
iss_trace_csv:   out/seed-10547/instr_gen/spike_sim/riscv_arithmetic_basic_test.0.csv
comparison_log:  
passed:          False
failure_message: |-
  [FAILED]: Log processing failed: [Errno 2] No such file or directory: 'out/seed-10547/instr_gen/spike_sim/riscv_arithmetic_basic_test.0.log'
```

Note this PR has a testing commit to enable azure YAML reporting, when we actually merge that commit will disappear and we'll make a change in the private-ci master branch to enable it.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -426,17 +426,18 @@ rtl_sim: $(rtl-sim-logs)
 #
 # The comparison script compares these and writes to a result file at
 #
-#   $(OUT-SEED)/rtl_sim/$(TEST).$(SEED)/comparison-result.txt
+#   $(OUT-SEED)/rtl_sim/$(TEST).$(SEED)/test-result.yml
 #
 # with PASSED or FAILED, depending.
 
-comp-results := $(addsuffix /comparison-result.txt,$(rtl-sim-dirs))
+comp-results := $(addsuffix /test-result.yml,$(rtl-sim-dirs))
 
 $(comp-results): \
-  %/comparison-result.txt: \
+  %/test-result.yml: \
   compare.py $(metadata)/instr_gen.iss.stamp $(rtl-sim-logs)
 	@echo Comparing traces for $*
 	$(verb)./compare.py \
+      --instr-gen-bin-dir $(OUT-SEED)/instr_gen/asm_test \
       --iss $(ISS) \
       --iss-log-dir $(OUT-SEED)/instr_gen/$(ISS)_sim \
       --start-seed $(SEED) \
@@ -446,7 +447,7 @@ $(comp-results): \
 
 $(OUT-SEED)/regr.log: collect_results.py $(comp-results)
 	@echo "Collecting up results (report at $@)"
-	$(verb)./collect_results.py -o $@ $(comp-results)
+	$(verb)./collect_results.py -o $(@D) $(comp-results)
 
 .PHONY: post_compare
 post_compare: $(OUT-SEED)/regr.log
--- a/dv/uvm/core_ibex/collect_results.py
+++ b/dv/uvm/core_ibex/collect_results.py
@@ -4,30 +4,50 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
+import junit_xml
+import os.path
 import sys
-from typing import TextIO
+import yaml
+from test_run_result import (TestRunResult, test_run_result_fields,
+                             check_test_run_result)
+from typing import List, TextIO
 
 
-def parse_log(path: str) -> bool:
-    first_line = '(empty file)'
-    with open(path, 'r', encoding='UTF-8') as log:
-        for line in log:
-            first_line = line.rstrip()
-            break
+def parse_test_run_result(path: str) -> TestRunResult:
+    try:
+        with open(path) as yaml_file:
+            test_run_result_dict = yaml.load(yaml_file, Loader=yaml.SafeLoader)
+            loaded_fields = test_run_result_dict.keys()
+            if set(loaded_fields) != set(test_run_result_fields):
+                raise RuntimeError(f'Error loading YAML at {path}: does not '
+                                   'contain the correct set of fields')
 
-    if first_line == 'PASS':
-        return True
-    if first_line.startswith('FAIL'):
-        return False
+            trr = TestRunResult(**test_run_result_dict)
+            try:
+                check_test_run_result(trr)
+            except AssertionError:
+                raise RuntimeError(f'Error loading YAML at path {path}: '
+                                   'field types were incorrect')
+            return trr
+    except (IOError, yaml.YAMLError) as e:
+        raise RuntimeError(f'Error loading YAML at path {path}: {e}')
 
-    raise RuntimeError('Strange first line ({!r})'.format(first_line))
 
-
-def dump_log(path: str, dest: TextIO) -> None:
-    print('\nLog at {}:'.format(path), file=dest)
-    with open(path, 'r') as fd:
-        for line in fd:
-            dest.write('> ' + line)
+def build_broken_test_run_result(err: str) -> TestRunResult:
+    return TestRunResult(
+            name='unknown',
+            idx=0,
+            seed=0,
+            binary=None,
+            uvm_log=None,
+            rtl_trace=None,
+            rtl_trace_csv=None,
+            iss_trace=None,
+            iss_trace_csv=None,
+            comparison_log=None,
+            passed=False,
+            failure_message=err
+    )
 
 
 def box_comment(line: str) -> str:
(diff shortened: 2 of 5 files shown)


dv/uvm/core_ibex/Makefile
```diff
@@ -446,7 +447,7 @@ $(comp-results): \
 
 $(OUT-SEED)/regr.log: collect_results.py $(comp-results)
 	@echo "Collecting up results (report at $@)"
-	$(verb)./collect_results.py -o $@ $(comp-results)
+	$(verb)./collect_results.py -o $(dir $@) $(comp-results)
```
rswarbrick, line 450: Nit: This could be `$(@D)`, which is a bit more idiomatic.

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
 
 import argparse
+import collections
+import junit_xml
+import os.path
 import sys
-from typing import TextIO
+import yaml
+from test_run_result import TestRunResult, test_run_result_fields, \
+    check_test_run_result
```
rswarbrick, line 13: I think it's probably nicer to use parens than backslashes to delimit things here:
```suggestion
from test_run_result import (TestRunResult, test_run_result_fields,
                             check_test_run_result)
```

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+            try:
+                check_test_run_result(trr)
+            except AssertionError as e:
+                raise RuntimeError(f'Error loading YAML at path {path}: ' \
+                    'field types were incorrect')
+            return trr
+    except IOError as e:
+        raise RuntimeError(f'Error loading YAML at path {path}: {e}')
+    except yaml.YAMLError as e:
+        raise RuntimeError(f'Error parsing YAML at path {path}: {e}')
```
rswarbrick, line 36: There's a neat syntax for catching two different types of errors:
```suggestion
    except (IOError, yaml.YAMLError) as e:
        raise RuntimeError(f'Error parsing YAML at path {path}: {e}')
```

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
-        return True
-    if first_line.startswith('FAIL'):
-        return False
+    return f'{pass_pct:0.2f}% PASS {len(passing_tests)} PASSED, ' \
+           f'{len(failing_tests)} FAILED'
 
-    raise RuntimeError('Strange first line ({!r})'.format(first_line))
+def gen_test_run_result_text(test_run_result: TestRunResult) -> str:
+    '''Generate a string describing a TestRunResult including the failure
+    message if the test did not pass'''
```
rswarbrick, line 69: The usual format for Python docstrings that are longer than a line is like Git commit messages: one short line header, then a blank line, then paragraphs.

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+    test_name_idx = f'{test_run_result.name}.{test_run_result.seed}'
+    test_underline = '-' * len(test_name_idx)
 
+    info_text = f'{test_name_idx}\n{test_underline}\n'
```
rswarbrick, line 73: Rather than faffing around with `\n`'s here, it's probably easier to gather a list and then join at the end. So this would be something like
```python3
    info_lines = [test_name_idx, test_underline]
```
then you'd call `info_lines.append(blah_blah_blah)` below. Finally, return `'\n'.join(info_text) + '\n'` (assuming you still want the trailing newline).

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+
+    return info_text
+
+def output_results_text(passing_tests: List[TestRunResult], failing_tests:
+        List[TestRunResult], dest: TextIO):
+    '''Write results in text form to dest'''
+
+    if failing_tests:
+        dest.write(box_comment('Details of failing tests'))
+        dest.write('\n')
```
rswarbrick, line 103: It might be simpler to use `print(something, dest=dest)` instead of the double `write` calls here. (Something that Philipp taught me!)

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+
+    dest.write('\n')
+    dest.write(gen_summary_line(passing_tests, failing_tests))
+    dest.write('\n')
+
+def output_run_results_junit_xml(passing_tests: List[TestRunResult], failing_tests:
+        List[TestRunResult], junit_dest: TextIO, junit_merged_dest: TextIO):
+    '''Write results to JUnit XML
+
+    Two versions are produced a normal version and a merged version. In the
```
rswarbrick, line 123: Nit: Missing colon after produced.

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+    test case per seed run. In the merged version there is a single test case
+    under the test suite with information for the individual runs merged
+    together. This is to aid use of the Azure Pipelines JUnit dashboard, which
+    doesn't neatly handle the test suite/test case hierarchy
+    '''
+
+    all_tests = passing_tests + failing_tests
 
+    test_suite_cases = {}
+    merged_test_suite_info = {}
```
rswarbrick, line 134: Since these are keyed by the same thing, I'd suggest generating a single dict here with tuples as values. Then the loop below gets a bit nicer: I think the whole lot can be replaced by something like
```python3
    by_name = {}
    for trr in all_tests:
        merged, unmerged = \
            by_name.setdefault(trr.name, ([], {'stdout': '', 'failures': ''}))

        result_text = gen_test_run_result_text(trr)

        # Create a test case for the TestRunResult. stdout holds the text
        # describing the run. Add the same text to failures if the test failed.
        test_case = junit_xml.TestCase(f'{trr.name}.{trr.seed}')
        test_case.stdout = result_text

        merged['stdout'] += result_text

        if not trr.passed:
            test_case.add_failure_info(output=trr.failure_message)
            merged['failures'] += result_text

        unmerged.append(test_case)
```

Following uses look like this:
```python3
    test_suites = [junit_xml.TestSuite(name, unmerged) for
                   (name, (merged, unmerged)) in by_name.items()]
```

dv/uvm/core_ibex/collect_results.py
```diff
@@ -4,78 +4,217 @@
...
+    for test_suite_name, test_suite_info in merged_test_suite_info.items():
+        test_case = junit_xml.TestCase(test_suite_name)
+        test_case.stdout = test_suite_info['stdout']
+        test_case.add_failure_info(output = test_suite_info['failures'])
+
+        merged_test_suites.append(junit_xml.TestSuite(test_suite_name,
+            [test_case]))
+
+    junit_merged_string = junit_xml.to_xml_report_string(merged_test_suites)
+    junit_merged_dest.write(junit_merged_string)
```
rswarbrick, line 181: Probably no need for the intermediate variable here:
```suggestion
    junit_merged_dest.write(junit_xml.to_xml_report_string(merged_test_suites))
```

rswarbrick (changes requested):
I've gone through the `collect_results.py` script and have a few comments. Also, please run this patch through flake8 and fix all the warnings.

I'll review the other two Python scripts in a follow-up review.

GregAC:
Thanks for the review, python is now lint8 clean.

One thing I didn't address in this PR but needs looking at is how we report tests that had some failure executing. `compare.py` will produce a test result giving a failure if the various log files aren't available but it needs to get run to do this. Currently the makefile uses the presence of `sim.log` to trigger the `compare.py` step. Should there be some failure meaning `sim.log` isn't produced we won't run `compare.py`. Perhaps the `collect_results.py` stage won't run at all as not all of the test result YAML files will be available?

Ultimately this should produce a failure, but we may not get a report telling us what failed. Ideally we'd do something such that `compare.py` gets run for every test `list_tests.py` says we should run and `collect_results.py` always runs after everything else is done so we get our report.

Any thoughts on the above @rswarbrick?

GregAC:
For motivation of the above. I've definitely seen failures in the past around the random program generation/spike ISS stage meaning the final report didn't get produced. When you run with `make -k` everything that can get done does get done but you had to manually run `sim.py` to produce a final report.

rswarbrick:
About the dependencies: That's a good point! Firstly, there's a mistake in the Makefile where the run of `compare.py` depends on all the simulation logs, not just one. Oops! I think it should probably depend on `%/sim.log` instead.

To allow things to percolate further, we basically need to add some `|| true` lines to the commands that run, delete output files before they run, and teach downstream scripts to cope when things are missing. It's not massively complicated, but probably belongs in a separate PR.

rswarbrick (approved):
Looks good. The code no longer runs cleanly through mypy (`mypy --strict dv/uvm/core_ibex/compare.py`) but maybe we should address that sometime later. We don't enforce it in CI, at any rate.

GregAC:
> The code no longer runs cleanly through mypy 

Yes I took a brief look at that, it seemed to be getting a bit confused with types in the named tuple. I guess some extra annotations are required?

PR #1391, opened 2021-06-29
[dv] Newline tweaks for regression logging

GregAC:


dv/uvm/core_ibex/collect_results.py
```diff
@@ -92,6 +92,7 @@ def gen_test_run_result_text(test_run_result: TestRunResult) -> str:
     if (test_run_result.passed):
         info_lines.append('[PASSED]')
     else:
+        info_lines.append('')
         info_lines.append(f'{test_run_result.failure_message}')
```
rswarbrick, line 96: This means we get an extra newline before failures, but not `[PASSED]` lines. Is that what you intend? Also, if you're touching this anyway, the following line could be just `info_lines.append(test_run_result.failure_message)`.

rswarbrick (approved):

dv/uvm/core_ibex/collect_results.py
GregAC: The issue was the info dump of log paths etc needed some separation from the failure message to make it easier to read. I think the `[PASSED]` alone reads well enough that it doesn't need it though could add one for consistency. Still there is no linebreak between the `[FAILED]` line and any failure message above it, so no linebreak before `[PASSED]` could also be seen as consistent.

PR #1395, opened 2021-07-11
[sw] Allow CC=/my-path/my-gcc make and fix toolchain substitution.

likewise:


likewise:
Disregard due to a reverted commit.

PR #1396, opened 2021-07-12
[sw] Fix toolchain substitution for file paths.

likewise:
This invocation would break:

CC=/opt/lowrisc-toolchain-gcc-rv32imc-20210412-1/bin/riscv32-unknown-elf-gcc make -C examples/sw/led/

because the "-gcc" occurence inside the directory name would also be replaced.

Fix by using pattern substitution.

Signed-off-by: Leon Woestenberg <leon@sidebranch.com>


likewise:
Fix commit message.

PR #1397, opened 2021-07-12
Fix toolchain substitution. Was broken for specific file paths.

likewise:
This invocation would break:

make -C examples/sw/led/ CC=/opt/lowrisc-toolchain-gcc-rv32imc-20210412-1/bin/riscv32-unknown-elf-gcc 

because the "-gcc" occurence inside the directory name would also be replaced.

Fix by using pattern substitution.

Signed-off-by: Leon Woestenberg <leon@sidebranch.com>

### DIFF
--- a/examples/sw/led/Makefile
+++ b/examples/sw/led/Makefile
@@ -12,8 +12,9 @@ SRCS = $(PROGRAM).c
 
 CC = riscv32-unknown-elf-gcc
 
-OBJCOPY ?= $(subst gcc,objcopy,$(wordlist 1,1,$(CC)))
-OBJDUMP ?= $(subst gcc,objdump,$(wordlist 1,1,$(CC)))
+CROSS_COMPILE = $(patsubst %-gcc,%-,$(CC))
+OBJCOPY ?= $(CROSS_COMPILE)objcopy
+OBJDUMP ?= $(CROSS_COMPILE)objdump
 
 LINKER_SCRIPT ?= link.ld
 CRT ?= crt0.S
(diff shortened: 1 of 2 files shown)


examples/sw/led/Makefile
```diff
@@ -12,8 +12,8 @@ SRCS = $(PROGRAM).c
 
 CC = riscv32-unknown-elf-gcc
 
-OBJCOPY ?= $(subst gcc,objcopy,$(wordlist 1,1,$(CC)))
-OBJDUMP ?= $(subst gcc,objdump,$(wordlist 1,1,$(CC)))
+OBJCOPY ?= $(patsubst %gcc, %objcopy, $(CC))
+OBJDUMP ?= $(patsubst %gcc, %objdump, $(CC))
```
rswarbrick, line 16: I don't think the new spaces are right. This should probably be e.g. `$(patsubst %gcc,%objcopy,$(CC))`.

Also, while this is rather neat, it's a bit hard to understand how it works. Would you mind refactoring this so that we extract the common prefix as a separate variable with something like `$(patsubst %gcc,%,$(CC))`? In your example, it would be `/opt/lowrisc-toolchain-gcc-rv32imc-20210412-1/bin/riscv32-unknown-elf-`. Plonk a comment above that, showing an example of how the paths change. Then you could append `objcopy` and `objdump` to that common prefix. I think the result might be easier to understand and it should also be easier to debug if something goes wrong.

likewise: Agreed, for the separate variable I would strongly suggest CROSS_COMPILE, as this is the defacto standard for the prefix, including the dash just before the toolchain component.

I have a fix pending.

Do you want me to force rebase my branch so that only once commit is visible when you merge, or is this step done in the merge itself?

rswarbrick: Oh, cool. I didn't know there was a standard name for this. The change looks good: please could you squash the two and then I'm very happy to merge.

rswarbrick (approved):
Looks good to me! Will merge when CI goes through.

PR #1399, opened 2021-07-12
Get riscv-formal flow working again

rswarbrick:
No guarantees that this actually does anything useful, but at least
the Makefile works again.

### DIFF
--- a/formal/riscv-formal/Makefile
+++ b/formal/riscv-formal/Makefile
@@ -15,6 +15,9 @@ OUTDIR   := build
 # Source directory relative to this Makefile
 SRC_DIR  := ../../rtl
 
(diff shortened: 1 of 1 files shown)


towoe:
Thanks Rupert for cleaning this up.
I like the addprefix part :)

Not sure if it makes sense to use `ibex_top` instead of `ibex_core` here. What do you think?
I added those changes to https://github.com/rswarbrick/ibex/pull/1 if use want to squash it into this PR.

rswarbrick:
That seems sensible. I'll add those patches to this PR. @imphil, would you mind taking a look?

imphil (approved):

PR #1404, opened 2021-07-15
Fix the verible-format CI job

imphil:
The toplevel core file got renamed, but the experimental verible-format
CI job wasn't updated in sync. Fix that.


imphil:
Looks good, the job is back. (We still need to pass Verible a bit more configuration to get formatting closer to what we want, but that's for another PR.)

rswarbrick (approved):

vogelpi (approved):
LGTM, thanks @imphil !

PR #1406, opened 2021-07-21
[rtl] Add ResetAll parameter

tomroberts-lowrisc:
This parameter forces a reset of all registers inside the core. This is
required to guarantee a common starting point for lockstep and thus
prevent spurious lockstep failure alerts.

Another minor change in this commit rearranges the writeback stage
multiplexing to gate incoming lsu write data when not valid. This stops
any X values from the data bus propagating to the register file
signalling (and thus to the lockstep comparison) which would cause the
lockstep alert to be X. It has the side effect of possibly reducing
power consumption in the register file.


rtl/ibex_fetch_fifo.sv
```diff
@@ -149,9 +150,19 @@ module ibex_fetch_fifo #(
   assign instr_addr_d = clear_i ? in_addr_i[31:1] :
                                   instr_addr_next;
 
-  always_ff @(posedge clk_i) begin
-    if (instr_addr_en) begin
-      instr_addr_q <= instr_addr_d;
+  if (ResetAll) begin : g_instr_addr_ra
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        instr_addr_q <= '0;
```
rswarbrick, line 156: I assume that something like `if (ResetAll && !rst_ni)` (which would allow us to lose the generate block) is too fancy for some synthesis tools?

rswarbrick (approved):
Wow, this can't have been fun to write! Nitty question about whether there's an easier way to do it but, if not, looks good to me.

rtl/ibex_fetch_fifo.sv
tomroberts-lowrisc: I wondered about that, and decided to "assume the worst". It would be MUCH nicer to write it that way though!

imphil: Yeah, better not play with the reset condition too much. I wouldn't be surprised if otherwise a DFT insertion tool or something like that complains.

PR #1407, opened 2021-07-22
Copy verible-format configuration from OpenTitan

imphil:
This configuration is the closest we can get to our style guide right
now, so let's go with that.


imphil:
Verible-format runs in CI with this config, but we only show the resulting diff and don't take any action on it yet.

tomroberts-lowrisc (approved):

PR #1409, opened 2021-07-23
[rtl, script] FPGA Detailed Power Analysis Support

ctopal:


examples/sw/simple_system/common/link.ld
```diff
@@ -6,9 +6,9 @@ OUTPUT_ARCH(riscv)
 
 MEMORY
 {
-    /* 992 kB should be enough for anybody... */
-    ram         : ORIGIN = 0x00100000, LENGTH = 0xF8000 /* 992 kB */
-    stack       : ORIGIN = 0x001F8000, LENGTH = 0x8000  /* 32 kB */
+    /* Change this if you'd like different sizes */
```
vogelpi, line 9: I think it would be more useful if the comment would say something like: "The Arty A7 FPGA board offers 256 kB of BRAM."

examples/fpga/artya7/top_artya7.core
```diff
@@ -18,6 +18,11 @@ filesets:
       - data/pins_artya7.xdc
     file_type: xdc
 
+  files_tcl:
+    files:
+      - util/vivado_setup_hooks.tcl: { file_type: tclSource }
+      - util/vivado_hook_write_bitstream_post.tcl: { file_type: user, copyto: vivado_hook_write_bitstream_post.tcl }
```
vogelpi: Have these two scripts already been part of the repository before? Otherwise, you have to add them to this very commit, too. Every commit should be atomic. Reverting a single commit shouldn't break the repo.

examples/fpga/artya7/top_artya7.core
```diff
@@ -28,7 +33,7 @@ parameters:
   SRAMInitFile:
     datatype: str
     description: SRAM initialization file in vmem hex format
-    default: "../../../../../examples/sw/led/led.vmem"
+    default: "../../../../../examples/sw/benchmarks/coremark/coremark.vmem"
```
vogelpi, line 36: Do you really want to change the default? If yes, you should also adapt the `README.md` accordingly: https://github.com/lowRISC/ibex/blob/master/examples/fpga/artya7/README.md (it still mentions `led.vmem`)

examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -90,54 +90,40 @@ module top_artya7 (
...
-    .rst_ni    ( rst_sys_n      ),
-    .req_i     ( mem_req        ),
-    .we_i      ( mem_write      ),
-    .be_i      ( mem_be         ),
-    .addr_i    ( mem_addr       ),
-    .wdata_i   ( mem_wdata      ),
-    .rvalid_o  ( mem_rvalid     ),
-    .rdata_o   ( mem_rdata      )
-  );
+      ) u_ram (
```
vogelpi, line 97: Please correct the indentation here: Basically, everything needs -2 spaces. `) u_ram (` needs -4 spaces.

examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -90,54 +90,40 @@ module top_artya7 (
...
+      .clk_i       (clk_sys),
+      .rst_ni      (rst_sys_n),
+
+      .a_req_i     (data_req    ),
+      .a_we_i      (data_we     ),
+      .a_be_i      (data_be    ),
+      .a_addr_i    (data_addr   ),
+      .a_wdata_i   (data_wdata  ),
+      .a_rvalid_o  (data_rvalid ),
+      .a_rdata_o   (data_rdata   ),
```
vogelpi, line 107: Please remove the spaces between signal name and `),`.

examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -90,54 +90,40 @@ module top_artya7 (
...
+      .b_wdata_i   (32'b0),
+      .b_rvalid_o  (instr_rvalid),
+      .b_rdata_o   (instr_rdata)
+    );
+
 
   // SRAM to Ibex
-  assign instr_rdata    = mem_rdata;
-  assign data_rdata     = mem_rdata;
-  assign instr_rvalid   = mem_rvalid;
```
vogelpi, line 131: Oh, it looks like we used to signal back an `instr_rvalid` on every memory read, not just instruction reads. So assume we were handling a data read followed by an instruction read. The instruction read would already see the data rvalid before the grant following in the next cycle. I am not 100% sure but think this could be a reason for the issues you noted when using the 1-port memory.

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
```diff
@@ -0,0 +1,27 @@
+open_project ../../lowrisc_ibex_top_artya7_0.1.xpr
+set saif_name "detailed_power.saif"
+
+regexp -- {Vivado v([0-9]{4})\.[0-9]} [version] -> year
+create_run impl_2 -parent_run synth_1 -flow "Vivado Implementation $year"
+set_property STEPS.OPT_DESIGN.IS_ENABLED false [get_runs impl_2]
+launch_runs impl_2
+wait_on_run impl_2
```
vogelpi, line 8: Can you please add comments here explaining why you need to run another implementation and why you need to disable optimization of the design for this?

vogelpi: Also, why do you need to extract the year of the Vivado release?

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
```diff
@@ -0,0 +1,27 @@
...
+
+open_run impl_2
+if {![file exists "$saif_name"]} {
+set_property top top_artya7 [current_fileset sim_1]
+launch_simulation -mode post-implementation -type functional
+open_saif "$saif_name"
+log_saif [get_objects -r *]
+add_force {/top_artya7/IO_CLK} -radix bin {1 0ns} {0 5000ps} -repeat_every 10000ps
```
vogelpi, line 16: Is it necessary to specify the clock waveform in picoseconds?

vogelpi (commented):
Hey @ctopal ,
thanks for your first Ibex PR, this looks pretty good :+1: 

I've added a couple comments. Please have a look and let me know if something isn't clear or if you want to discuss in person. This is sometimes easier.

A general comment: please make sure to structure your commit messages as follows (see https://chris.beams.io/posts/git-commit/ for more details):
1. Separate subject from body with a blank line
1. Limit the subject line to 50 characters
1. Capitalize the subject line
1. Do not end the subject line with a period
1. Use the imperative mood in the subject line
1. Wrap the body at 72 characters
1. Use the body to explain what and why vs. how

As an example, your message would look something like this:
```
[fpga] Add power analysis scripts to Arty A7 example

This commit adds power analysis scripts to the Arty A7
example design. They can be used by specifying the newly
added `power_analysis` FuseSoc flag. 

Signed-off-by: ctopal <ctopal@lowrisc.org>
```

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
ctopal: I tried to leave a detailed explanation, can you please check it? Thanks

examples/fpga/artya7/top_artya7.core
ctopal: Hi Pirmin, you are right this was a mistake. I tried to fix it now, could you check it please? Thanks

examples/sw/simple_system/common/link.ld
ctopal: I agree, thanks for the comment. I tried to update it with more detail.

examples/fpga/artya7/top_artya7.core
ctopal: I think it is better to specify it when calling fusesoc. Where should I explain how to use the power analysis option?

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
ctopal: It was a copy paste from Vivado gui, that's why :) I fixed it now.

examples/fpga/artya7/top_artya7.core
vogelpi: This is good now, thanks Chanberk.

examples/fpga/artya7/top_artya7.core
vogelpi: I think you could add a new section to the README.md under https://github.com/lowRISC/ibex/blob/master/examples/fpga/artya7/README.md

examples/fpga/artya7/rtl/top_artya7.sv
vogelpi: This looks better but it's still not fully right. The indentation should be as follows:
```sv
  ram_2p #(
    .Depth(MEM_SIZE / 4),
    .MemInitFile(SRAMInitFile)
  ) u_ram (
    .clk_i (clk_sys),
    .rst_ni(rst_sys_n),

    .a_req_i   (data_req),
    .a_we_i    (data_we),
    .a_be_i    (data_be),
    .a_addr_i  (data_addr),
    .a_wdata_i (data_wdata),
    .a_rvalid_o(data_rvalid),
    .a_rdata_o (data_rdata),

    .b_req_i   (instr_req),
    .b_we_i    (1'b0),
    .b_be_i    (4'b0),
    .b_addr_i  (instr_addr),
    .b_wdata_i (32'b0),
    .b_rvalid_o(instr_rvalid),
    .b_rdata_o (instr_rdata)
  );
```
More precisely: the indentation is 2 spaces per level (not 3) and there are no spaces between the longest port name and the parenthesis, other parenthesis are aligned to the longest port name in the block. The latter rule is relatively new. For more details, please refer to our SV style guide: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#indentation

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
vogelpi: Perfect, this is very useful. Thanks!

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
vogelpi: Great!

vogelpi (commented):
Hey @ctopal ,
this is going in the right direction. I've resolved some of the comments. Some comments need some further work.

Please also correct the commit messages of the first and second commit in the series. Thanks!

ctopal:
Hi @vogelpi,

I fixed commit messages, added couple of lines to README about power analysis option and changed the instantiation to comply with SV style guide. Thanks for your review, it is very helpful!

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
```diff
@@ -0,0 +1,34 @@
...
+add_force {/top_artya7/IO_RST_N} -radix bin {1 0ns}
+run 3ms
+close_saif
+}
+
+# Reporting power using .saif generated above
+open_run impl_2
+set_operating_conditions -process maximum
+read_saif "../../lowrisc_ibex_top_artya7_0.1.sim/sim_1/impl/func/xsim/$saif_name"  -out_file saif_detailed.log
+read_saif "../../lowrisc_ibex_top_artya7_0.1.sim/sim_1/impl/func/xsim/$saif_name"  -strip_path top_artya7 -out_file saif_detailed1.log
```
rswarbrick, line 32: This looks a little odd to me. Do we need both of these log files? If so, maybe the names could be more informative?

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
```diff
@@ -0,0 +1,34 @@
...
+set_units -power uW
```
rswarbrick, line 33: Wooo!!! :grin:

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
```diff
@@ -0,0 +1,34 @@
...
+regexp -- {Vivado v([0-9]{4})\.[0-9]} [version] -> year
+create_run impl_2 -parent_run synth_1 -flow "Vivado Implementation $year"
+set_property STEPS.OPT_DESIGN.IS_ENABLED false [get_runs impl_2]
+launch_runs impl_2
+wait_on_run impl_2
+open_run impl_2
+
+# Runs a post implementation functional simulation with the memory initialized with SRAMInitFile.
+# Feeds clock (100mhz) and reset switch and records switching activity for 3ms.
+if {![file exists "$saif_name"]} {
```
rswarbrick, line 17: I'm not sure this is the right thing to do: everything else gets rebuilt from scratch on a fusesoc run. Probably we should do the same for SAIF simulation? (@imphil: You have deep knowledge of fusesoc! What would you expect here?)

examples/fpga/artya7/top_artya7.core
```diff
@@ -43,6 +48,7 @@ targets:
     filesets:
       - files_rtl_artya7
       - files_constraints
+      - power_analysis ? (files_tcl)
```
rswarbrick, line 51: This is clever! But is it possible to run `vivado_setup_hooks.tcl` unconditionally and then check for the `power_analysis` parameter from there? Otherwise, we're going to struggle to add another mode in the future.

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
ctopal: It was for debugging why read_saif was not reading logged `.saif` file properly and it left there. I will remove it, thanks for noticing! Two read_saif commands was for stripping the hierarchy so that almost all the nets could be matched.

examples/fpga/artya7/top_artya7.core
ctopal: Thanks for checking the code it really helps me! 

I am not sure how/if we can use the user defined FuseSoc flag inside `vivado_setup_hooks.tcl.tcl` file. Another solution would be defining a vlogparam and then use it in the tcl as you suggested. I am going to look into it further for sure.

rswarbrick: Ah: I hadn't twigged about the difference between a flag and a parameter. @imphil: Do you have an opinion on the best way to do this?

ctopal: I updated the script to do the deciding after sourcing. I think this looks cleaner and creates no worries about future additions.

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
ctopal: I removed because you are right, there would be now way to have a .saif file under build directory before running this one. Thanks for noticing it, line was there because I was first testing the script from an already built project!

examples/fpga/artya7/README.md
```diff
@@ -64,15 +64,26 @@ This will create a directory `build` which contains the output files, including
...
 
 ```
 fusesoc --cores-root=. run --target=synth --setup --build lowrisc:ibex:top_artya7 --part xc7a100tcsg324-1 --SRAMInitFile=examples/sw/benchmarks/coremark/coremark.vmem
 ```
 
+#### Power Analysis Using Vivado
+
+Setting `FPGAPowerAnalysis` parameter to 1 allows user to run a power analysis using Vivado.
+It uses a post-implementation functional simulation on Vivado to log switching activity.
+Switching activity then get used to generate a detailed power report.
```
rswarbrick, line 80: Grammar nit: This should be "then get**s**" (because "activity" is singular), but I'd probably write "This switching activity is then used...", which is a little more idiomatic and ties the sentence to the previous one.

rswarbrick (approved):
Thanks, this looks much better.

ctopal:
I believe now this PR is in a good shape for merging since now it doesn't even need another implementation to run, what do you think @vogelpi ?

examples/fpga/artya7/rtl/top_artya7.sv
vogelpi: This looks good now.

examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -148,7 +124,7 @@ module top_artya7 (
     if (!rst_sys_n) begin
       leds <= 4'b0;
     end else begin
-      if (mem_req && data_req && data_we) begin
```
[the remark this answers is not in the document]
vogelpi: Ah, interesting. Good that you found the issue!

examples/fpga/artya7/util/vivado_hook_route_design_post.tcl
vogelpi: This got resolved as well. The second implementation run without optimization is no longer needed.

examples/fpga/artya7/README.md
```diff
@@ -64,15 +64,26 @@ This will create a directory `build` which contains the output files, including
 the bitstream.
 
 
-Initial memory parameter SRAMInitFile can be given to FuseSoc to specify which .vmem file to load the design with.
-Example use case includes loading coremark.vmem which can be used for performance/power analysis.
+Initial memory parameter `SRAMInitFile` can be given to FuseSoc to specify which .vmem file to load the design with.
+Example use case includes loading `coremark.vmem` which can be used for performance/power analysis.
```
vogelpi: Nit: this change should go into the first commit where you add the two lines for the first time.

vogelpi (approved):
This looks good now. Well done!

There is one minor comment left, some change in the README should be moved to the first commit.

PR #1411, opened 2021-07-26
Added cast to enum

dawidzim:
Hey!
In this PR I added cast to enum, because we can't assign non-enum to enum LRM 6.19.3


rswarbrick (approved):

rswarbrick:
Looks good. For future PRs, would you mind expanding the commit message a bit? Something like "Converting from an integer to an enum needs an explicit cast in the IEEE 1800 spec; see section 1.2.3.4.". It might make it easier for people reading the codebase in future.

PR #1412, opened 2021-07-28
[rtl] Minor lockstep-related updates

tomroberts-lowrisc:
There are two patches here - one is a minor cleanup, the other adds a register slice on the outputs from the lockstep shadow core.

@tjaychen @eunchan The register slice is something we discussed a long time ago when some timing issues showed up in the shadow core. LMK if you think this is worth doing or not.

### DIFF
--- a/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
+++ b/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
@@ -120,7 +120,7 @@ module core_ibex_tb_top;
   );
 
   // We should never see any alerts triggered in normal testing
-  `ASSERT(NoAlertsTriggered, clk, !rst_n, !dut_if.alert_minor && !dut_if.alert_major)
+  `ASSERT(NoAlertsTriggered, !dut_if.alert_minor && !dut_if.alert_major, clk, !rst_n)
 
   // Data load/store vif connection
   assign data_mem_vif.reset                   = ~rst_n;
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -120,7 +120,7 @@ module core_ibex_tb_top;
   );
 
   // We should never see any alerts triggered in normal testing
-  `ASSERT(NoAlertsTriggered, clk, !rst_n, !dut_if.alert_minor && !dut_if.alert_major)
+  `ASSERT(NoAlertsTriggered, !dut_if.alert_minor && !dut_if.alert_major, clk, !rst_n)
```
msfschaffner, line 123: QQ: Do we plan to exercise the alert paths in DV? In that case it may make sense to remove this assertion.
I.e., we do not place assertions on fatal alerts in OT peripherals either, since they are a "function" of the IPs.

Nit: `NoAlertsTriggered` -> `NoAlertsTriggered_A`

tomroberts-lowrisc: Yep we will have testing of injected errors at some point and this assertion will be replaced with scoreboard checking

vogelpi (approved):
LGTM, thanks for the PR @tomroberts-lowrisc !

Issue #1357, opened 2021-05-19
[dv] Ibex co-simulation proposal


GregAC:
Our current Ibex DV environment uses a post simulation trace comparison methodology. This checks that the traces match from a particular test binary that is run on both an ISS (such as Spike or OVPSim) and an Ibex simulation. It has a weakness around interrupts, exceptions and debug related behaviour. These aren't simulated on the ISS (with the exception of some precise exceptions such as an illegal instruction, we don't simulate precise exceptions from imem or dmem error responsess). This will cause trace mismatches so trace matching is disabled on tests that use these with final state comparison and some extra checking during the simulation run to compensate. This can suffice for finding bugs but it can require more tests to ensure a particular bug actually produces a failure.  When failures are seen debugging back from a final state comparison failure can be hard work.

This is a proposal to switch to a co-simulation method where Spike is run along with the Ibex DV simulation in lock step, cross checking between the two on every retired instruction. The DV environment will probe events such as interrupts, debug requests and memory error responses. These will be notified to Spike to keep instruction execution in sync.

Note this proposal specifically concerns using Spike as the ISS for co-simulation. Currently OVPSim is also supported for the post simulation trace comparison method but due to the higher degree of coupling we should focus on supporting a single ISS for co-simulation. The co-simulation framework should be designed such that support for multiple simulators in future is possible without significant refactoring.

I have done some initial investigation around how to directly control the Spike 'core' and inserting interrupts/memory errors/debug requests at appropriate points so the instruction by instruction trace matches an Ibex simulation. This has demonstrated that co-simulation with Spike is feasible.

Spike Modifications
-------------------
The initial investigation has shown minimal Spike source modifications are required. Two changes have been identified:

- Alter the debug entry point which is specified by the `DEBUG_ROM_ENTRY` define in `debug_rom_defines.h`. We may wish to alter this to a run-time parameter or at least provide compile-time configurability (given it needs to match an Ibex parameter `DmHaltAddr` which can be altered in the Ibex instantiation).
- Make register writes visible to the public interface of `processor_t`. When Spike is built with `RISCV_ENABLE_COMMITLOG` register writes are already logged however this happens internally to the `step` function and the publically visible state that would contain register writes is wiped (having been output to the textual log) before `step` returns. An alternative would be to ignore this functionality all together and simply look at the state of the register file after every step and check for changes.

[...]

PR #1416, opened 2021-08-03
[dv] Initial co-simulation framework

GregAC:
As discussed this is just a draft to illustrate how things are shaping up. There's a few things that definitely still need doing

1. Some more documentation (e.g. a README.md in `dv/cosim`). In particular describing how you build spike
2. Decide how the spike build is dealt with. Do we vendor in the riscv-isa-sim repository or just tell people to grab it and build themselves if they want to run DV? How do we feed the relevant paths into the UVM/Simple System build (you'll note it's just hardcoded paths from my system right now)
3. Decide how/if this should be integrated in with simple system for what we actually merge. I was thinking of leaving simple system itself alone and having a separate `dv/simple_system_cosim` that pulls in the simple system components and adds the cosim on top (e.g. via injecting the RTL with a bind). At the very least it shouldn't be a mandatory part of simple system and in keeping with the name I think it's better if it (`examples/simple_system`) stands alone without extra cosim baggage.

The initial proposal was here: https://github.com/lowRISC/ibex/issues/1357

The main departure from that is dropping the `riscv_cosim_init`/`riscv_cosim_destory` functions. I felt it was better to leave it up to the simulation environment to set things up how it chooses. In particular it may want backdoor access for various things and I think it's better placed to setup the memory map than trying to write some generic verilog that will handle all situations.

So far a couple of spike modifications have been required:
* adding the `last_inst_pc` field to the CPU state
* modifying the hard-coded debug ROM address (think we'll want to make this run-time configurable so the sim can snoop the `DmHaltAddr` parameter and set it at start of sim).

This has had some testing. I've got successful runs doing a 'plain' execution (including running all of core mark) as well as basic interrupt/exception/debug examples matching up.

Oh and I suspect it breaks with the writeback stage enabled right now, this is down to how we generate the `rvfi_trap` signal (which is currently incorrect in master, partially fixed here).


imphil:
> Some more documentation (e.g. a README.md in dv/cosim). In particular describing how you build spike

Can you add the documentation to the rst documentation instead? Finding the right README somewhere in the tree isn't a beautiful user experience.

rswarbrick (commented):
Looks good to me.

I'd probably consider using `MemArea` directly in the `Cosim` class, which might save a bit of packing/unpacking.

GregAC:
This is no longer relevant (cosim has long been integrated via a different PR), so closing

PR #1417, opened 2021-08-04
[rtl/doc] Update ePMP CSR addresses and documentation

GregAC:
mseccfg and mseccfgh have changed their addresses. This updates to the
newly allocated values.

The ePMP specification is now available as a versioned PDF,
documentation is updated to point to that removing the local PDF copy.


GregAC:
For the avoidance of doubt I have read the linked v0.9.3 specification and confirmed our implementation still matches what is specified.

imphil (approved):
"This document is in the Stable state. Assume anything could still change, but limited change should be expected"

I love the definition of "stable" on the front page of the PMP spec :smile:

alistair23 (approved):

GregAC:
> Assume anything could still change

"After some discussion ePMP is now a ML acceleration extension, we hope these minor changes do not delay the ratification process"

PR #1418, opened 2021-08-05
[rtl] Add bus integrity checking

tomroberts-lowrisc:
Extra bits are added alongside read/write data for the instruction and
data buses to facilitate data integrity checking.

Ibex testbench extended to generate the expected bits.

All other top-levels modified to add the new signals (which are mostly
ignored).


vogelpi (approved):
Thanks @tomroberts-lowrisc for the PR, the implementation looks good to me.

I have two questions:
1. Command integrity: are we considering to do this as well? This would probably happen inside OpenTitan, after the conversion to TL-UL, right?
2. When using lockstep, the response integrity could probably be computed based on the shadow core only (after registers) which might help timing and reduce area (2 vs. 4 decoders, less flops). For the write data, we could just do it on the data of the main core (no timing improvement but minus 1 encoder). However, I don't know how these savings would compare to the overall complexity.

tomroberts-lowrisc:
Thanks @vogelpi 
1. Yes that's right, it has to go there due to the address translation stuff
2. Yes that's a good point. It was specified to calculate the integrity twice so as to avoid the single point of failure, but you're right that calculating it once in the shadow core (rather than the main core) also achieves that. Agree about the write data, can't see much benefit in calculating that twice. I'll investigate those enhancements and follow up.

doc/03_reference/load_store_unit.rst
```diff
@@ -14,38 +14,53 @@ Data-Side Memory Interface
...
+|                            |           | reads. Sent together with ``data_req_o``      |
++----------------------------+-----------+-----------------------------------------------+
+| ``data_be_o[3:0]``         | output    | Byte Enable. Is set for the bytes to          |
+|                            |           | write/read, sent together with ``data_req_o`` |
++----------------------------+-----------+-----------------------------------------------+
+| ``data_wdata_o[31:0]``     | output    | Data to be written to memory, sent together   |
+|                            |           | with ``data_req_o``                           |
++----------------------------+-----------+-----------------------------------------------+
+| ``data_wdata_intg_o[6:0]`` | output    | Integrity bits to be written to memory, sent  |
+|                            |           | together with ``data_req_o``                  |
```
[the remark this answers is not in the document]
GregAC: oh and similar for the read data integrity port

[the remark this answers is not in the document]
tomroberts-lowrisc: Agreed - now fixed

tomroberts-lowrisc:
@vogelpi I have pushed a new version with the bus integrity checking duplication removed, please take a look

rtl/ibex_lockstep.sv
```diff
@@ -197,6 +204,38 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  );
+
+  assign bus_intg_err = (shadow_inputs_q[0].instr_rvalid & |instr_intg_err) |
+                        (shadow_inputs_q[0].data_rvalid  & |data_intg_err);
+
+  // Generate integrity bits
+  prim_secded_39_32_enc u_data_gen (
+    .data_i (data_wdata_i),
+    .data_o ({data_wdata_intg_o, unused_wdata})
+  );
```
vogelpi, line 237: At first I thought it's a bit counterintuitive so generate the integrity bits here, as these are based on the outputs of the actual core and not the shadow core. But I can also see that it makes sense to do it here to do all the integrity stuff in the same place.

rtl/ibex_lockstep.sv
```diff
@@ -197,6 +204,38 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  // Bus integrity checking //
+  ////////////////////////////
+
+  logic        bus_intg_err;
+  logic [1:0]  instr_intg_err, data_intg_err;
+  logic [31:0] unused_wdata;
+
+  // Checks on incoming data
+  prim_secded_39_32_dec u_instr_intg_dec (
+    .data_i     ({shadow_inputs_q[0].instr_rdata_intg, shadow_inputs_q[0].instr_rdata}),
```
vogelpi, line 217: Sorry to bother you again, but I just noticed some more optimization potential here. I think we don't really care when a potential error is detected. So we could just use `shadow_inputs_q[LockstepOffset-1].x_rvalid` here and remove the other flops.

But even as it is now I think it is okay.

vogelpi (approved):
Thanks @tomroberts-lowrisc for investigating my suggestion and optimizing the design. I think there is some more potential but even as it is now, it looks good.

rtl/ibex_lockstep.sv
```diff
@@ -156,38 +159,43 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  assign shadow_inputs_in.data_err         = data_err_i;
+  assign shadow_inputs_in.rf_rdata_a_ecc   = rf_rdata_a_ecc_i;
+  assign shadow_inputs_in.rf_rdata_b_ecc   = rf_rdata_b_ecc_i;
+  assign shadow_inputs_in.irq_software     = irq_software_i;
+  assign shadow_inputs_in.irq_timer        = irq_timer_i;
+  assign shadow_inputs_in.irq_external     = irq_external_i;
+  assign shadow_inputs_in.irq_fast         = irq_fast_i;
+  assign shadow_inputs_in.irq_nm           = irq_nm_i;
+  assign shadow_inputs_in.debug_req        = debug_req_i;
+  assign shadow_inputs_in.fetch_enable     = fetch_enable_i;
```
vogelpi, line 184: Nit: these whitespace changes are no longer needed.

rtl/ibex_lockstep.sv
```diff
@@ -303,89 +343,89 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+    .rvfi_rs3_rdata     (),
+    .rvfi_rd_addr       (),
+    .rvfi_rd_wdata      (),
+    .rvfi_pc_rdata      (),
+    .rvfi_pc_wdata      (),
+    .rvfi_mem_addr      (),
+    .rvfi_mem_rmask     (),
+    .rvfi_mem_wmask     (),
+    .rvfi_mem_rdata     (),
+    .rvfi_mem_wdata     (),
```
vogelpi, line 422: Nit: these whitespace changes are no longer needed.

rtl/ibex_lockstep.sv
```diff
@@ -303,89 +343,89 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+
+    .irq_software_i     (shadow_inputs_q[0].irq_software),
+    .irq_timer_i        (shadow_inputs_q[0].irq_timer),
+    .irq_external_i     (shadow_inputs_q[0].irq_external),
+    .irq_fast_i         (shadow_inputs_q[0].irq_fast),
+    .irq_nm_i           (shadow_inputs_q[0].irq_nm),
+    .irq_pending_o      (shadow_outputs_d.irq_pending),
+
+    .debug_req_i        (shadow_inputs_q[0].debug_req),
+    .crash_dump_o       (shadow_outputs_d.crash_dump),
```
vogelpi, line 397: Nit: these whitespace changes are no longer needed.

rtl/ibex_lockstep.sv
```diff
@@ -303,89 +343,89 @@ module ibex_lockstep import ibex_pkg::*; #(
...
 `endif
 
-    .fetch_enable_i    (shadow_inputs_q[0].fetch_enable),
-    .alert_minor_o     (shadow_alert_minor),
-    .alert_major_o     (shadow_alert_major),
-    .core_busy_o       (shadow_outputs_d.core_busy)
+    .fetch_enable_i     (shadow_inputs_q[0].fetch_enable),
+    .alert_minor_o      (shadow_alert_minor),
+    .alert_major_o      (shadow_alert_major),
+    .core_busy_o        (shadow_outputs_d.core_busy)
```
vogelpi, line 428: Nit: these whitespace changes are no longer needed.

vogelpi (approved):
Thanks @tomroberts-lowrisc for having another look at this. This looks good.

With the latest version, some of whitespace changes related to alignment are no longer needed (see comments). You can theoretically remove them.

PR #1420, opened 2021-08-09
[rtl] Add LFSR permutation option

tomroberts-lowrisc:
Random constants are sent through the hierarchy as parameters in-line
with other OpenTitan modules.

Further detail on this mechanism can be found in lowrisc/opentitan#2229


vogelpi (approved):
LGTM, thanks @tomroberts-lowrisc !

PR #1421, opened 2021-08-13
[rtl] Fix mtval for unaligned accesses

GregAC:
Previously the raw incremented address was used which is the calculated
address + 4. This is confusing as it refers to a byte that wouldn't be
accessed (e.g. a lw at 0x8000009e which faults on the access to
0x8000000a0, would report an mtval of 0x8000000a2). With this change
mtval will refer to the first byte on the other half of the word
boundary the unaligned access crosses.

Relevant bits from the spec:

> If mtval is not hardwired to zero, then when a breakpoint, address-misaligned, access-fault, or page-
fault exception occurs on an instruction fetch, load, or store, mtval is written with the faulting
virtual address.

> For misaligned loads and stores that cause access-fault or page-fault exceptions, mtval will contain
the virtual address of the portion of the access that caused the fault

So I believe this matches the spec, certainly what we had before wasn't correct. It does match Spike behaviour for unaligned accesses.

### DIFF
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -68,7 +68,7 @@ module ibex_load_store_unit
 
   logic [31:0]  data_addr;
   logic [31:0]  data_addr_w_aligned;
-  logic [31:0]  addr_last_q;
+  logic [31:0]  addr_last_q, addr_last_d;
 
   logic         addr_update;
   logic         ctrl_update;
@@ -206,13 +206,17 @@ module ibex_load_store_unit
     end
   end
 
-  // Store last address for mtval + AGU for misaligned transactions.
-  // Do not update in case of errors, mtval needs the (first) failing address
+  // Store last address for mtval + AGU for misaligned transactions.  Do not update in case of
+  // errors, mtval needs the (first) failing address.  Where an aligned access or the first half of
+  // a misaligned access sees an error provide the calculated access address. For the second half of
+  // a misaligned access provide the word aligned address of the second half.
+  assign addr_last_d = addr_incr_req_o ? data_addr_w_aligned : data_addr;
+
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
       addr_last_q <= '0;
     end else if (addr_update) begin
-      addr_last_q <= data_addr;
+      addr_last_q <= addr_last_d;
     end
   end
 


rtl/ibex_load_store_unit.sv
```diff
@@ -206,13 +206,17 @@ module ibex_load_store_unit
     end
   end
 
-  // Store last address for mtval + AGU for misaligned transactions.
-  // Do not update in case of errors, mtval needs the (first) failing address
+  // Store last address for mtval + AGU for misaligned transactions.  Do not update in case of
+  // errors, mtval needs the (first) failing address.  Where an aligned access or the first half of
+  // a misaligned access sees an error provide the calculated access address. For the second half of
+  // a misaligned access provide the word aligned address of the second half.
+  assign addr_last_d = addr_incr_req_o ? data_addr_w_aligned : data_addr;
```
vogelpi, line 213: I think you don't actually need this mux here. Because on line 485 we have
```sv
assign data_addr_w_aligned = {data_addr[31:2], 2'b00};
```
So, it should be possible to just change from `data_addr` to `data_addr_w_aligned` below.

vogelpi (commented):
Thanks @GregAC for catching and fixing this bug! I think you can simplify the patch somewhat which will avoid the additional mux, please see my comment.

rtl/ibex_load_store_unit.sv
GregAC: If we always use `data_addr_w_aligned`, we'll always get a 32-bit aligned mtval. So taking the lw to `0x8000009e` example we'd get an mtval of `0x8000009c` if the first half of the load faulted. Plus for aligned byte and half-word accesses you won't always get the accessed address.

tomroberts-lowrisc: Agreed, I think you need the mux here (and it's only a mux on 2 bits technically, or actually an AND gate on 2 bits...)

vogelpi: Yes you are right. But as you cited from the spec:

> For misaligned loads and stores that cause access-fault or page-fault exceptions, mtval will contain
the virtual address of the portion of the access that caused the fault

IMO the portion of the access that causes the fault is the word-aligned address and not the misaligned one. The aligned address is what PMP checks and what the bus sees. Ibex only ever outputs word-aligned addresses. I agree that when debugging software it may not be obvious where the aligned address comes from. But if people then check the PMP configuration or the bus, they will only see aligned addresses anyway.

Maybe the spec is somewhat ambiguous here? Anyway, I am open for either interpretation but I would like to hear also other people's opinions.

GregAC: I'd say certainly for aligned accesses you must give `data_addr`.

For the misaligned accesses I was interpreting the spec to state that where the fault occurs over some alignment boundary from the calculated address you should provide the address of a byte accessed over that boundary. The first one on the other side seems reasonable.

So going back to the `lw 0x8000009e` example, this splits into two accesses

1. `0x8000009c`
2. `0x800000a0`

If 1. faults we should give `0x8000009e` as we'd do for an aligned access, if 2. faults we should give  `0x800000a0` to conform with 'mtval will contain the virtual address of the portion of the access that caused the fault'

We could argue aligning in both cases makes sense but then you get the odd behavior that an aligned access always gets a byte precise mtval (and crucially an address of a byte it tried to access) where an unaligned access always gets a 32-bit mtval (and may adderss a byte it didn't try to access).

tomroberts-lowrisc: I guess @vogelpi is saying that we could write the word aligned value for any access, aligned or unaligned. The spec is vague on what would be acceptable here. I guess that might make the spike comparisons harder though, since it will probably give the address of the first byte accessed?

vogelpi: Yes, you're right @tomroberts-lowrisc . This is always how I understood this. And you're totally right about the complexity ;-)

Anyway, back to the question. Your reasoning @GregAC makes sense, you've convinced me. Thanks for bearing with me. Let's get this in.

GregAC: Yeah it would break the spike comparison, though modifying it so it always aligns wouldn't be hard.

I just think for the aligned case the spec is saying you should use the computed address:

> If mtval is not hardwired to zero, then when a breakpoint, address-misaligned, access-fault, or page-fault exception occurs on an instruction fetch, load, or store, mtval is written with the faulting
virtual address.

vogelpi (approved):

PR #1422, opened 2021-08-20
[bitmanip][zba] Add support for Zba (address calculation) extension

mundaym:
Add support for the Zba extension added in v0.93 of the bit manipulation
specification (unchanged in v1.0.0). The new instructions added are:

  - sh1add: rd = (rs1 << 1) + rs2
  - sh2add: rd = (rs1 << 2) + rs2
  - sh3add: rd = (rs1 << 3) + rs2

The instructions are single cycle and have been implemented using the
adder in the ALU.

Signed-off-by: Michael Munday <mike.munday@lowrisc.org>

### DIFF
--- a/rtl/ibex_alu.sv
+++ b/rtl/ibex_alu.sv
@@ -45,11 +45,17 @@ module ibex_alu #(
   // Adder //
   ///////////
 
+  logic        adder_op_a_shift1;
+  logic        adder_op_a_shift2;
+  logic        adder_op_a_shift3;
   logic        adder_op_b_negate;
   logic [32:0] adder_in_a, adder_in_b;
   logic [31:0] adder_result;
 
   always_comb begin
+    adder_op_a_shift1 = 1'b0;
+    adder_op_a_shift2 = 1'b0;
+    adder_op_a_shift3 = 1'b0;
     adder_op_b_negate = 1'b0;
     unique case (operator_i)
       // Adder OPs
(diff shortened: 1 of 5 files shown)


mundaym:
Just to be clear: this isn't a priority for anything, just something I put together for fun.

I think Zba still needs support in https://github.com/google/riscv-dv before this is testable (though I did run some test code to verify it).

rtl/ibex_alu.sv
```diff
@@ -65,12 +71,25 @@ module ibex_alu #(
...
 
   // prepare operand a
-  assign adder_in_a    = multdiv_sel_i ? multdiv_operand_a_i : {operand_a_i,1'b1};
+  always_comb begin
+    unique case(1'b1)
+      multdiv_sel_i:   adder_in_a = multdiv_operand_a_i;
+      adder_op_shift1: adder_in_a = {operand_a_i[30:0],2'b01};
+      adder_op_shift2: adder_in_a = {operand_a_i[29:0],3'b001};
+      adder_op_shift3: adder_in_a = {operand_a_i[28:0],4'b0001};
+      default:         adder_in_a = {operand_a_i,1'b1};
```
GregAC, line 90: Making the mux here bigger does make me a little nervous as chances are the adder is on the longer timing paths (e.g. when used to calculate a load address). When bitmanip is disabled we should never get the `adder_op_shiftN` signals asserted so the synthesis tool should be able to produce the same mux as we had but I'm unsure if it'll be clever enough as it'll have to realise the alu op is never the `ALU_SH1ADD` etc values.

As the previous mux is a one liner this may be worth putting behind a generate statement to make sure.

mundaym: I added `if (RV32B != RV32BNone)` guards that should make it clear to the synthesis tool that the `adder_op_shiftN` signals cannot be asserted. Do you think that is enough or should the mux be guarded instead (or as well as)?

https://github.com/lowRISC/ibex/commit/e05effde61b2a167749edfe74f28ffb0dc438d2a#diff-d1197c7f39a2ce7728c1ae853493090d8713ba169935ba0e754a9463ae101d17R75-R77

Issue #1425, opened 2021-08-25
Vivado 2018.3 doesn't recognize SIM_DEVICE param on BUFGCE


nicolast0604:
[Synth 8-3438] module 'BUFGCE' declared at '/tools/Xilinx/Vivado/2018.3/scripts/rt/data/unisim_comp.v:619' does not have any parameter 'SIM_DEVICE' used as named parameter override ["/home/nioclast0604/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_prim_xilinx_clock_gating_0/rtl/prim_xilinx_clock_gating.sv":18]
[Synth 8-6156] failed synthesizing module 'prim_xilinx_clock_gating' ["/home/nioclast0604/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_prim_xilinx_clock_gating_0/rtl/prim_xilinx_clock_gating.sv":5]
[Vivado_Tcl 4-5] Elaboration failed - please see the console for details

## Observed Behavior

Vivado synthesis & implementation for Arty A7 FPGA board
## Expected Behavior

bitstream finished
## Steps to reproduce the issue
make build-arty-35 program-arty

## My Environment

**EDA tool and version:**
Vivado 2018.3

**Operating system:**

Ubuntu 18.04.5
**Version of the Ibex source code:**

master fork 246

vogelpi:
Hi @nicolast0604 ,

thanks for reporting this issue. It seems that Vivado 2018.3 doesn't support the `SIM_DEVICE` parameter on the `BUFGCE` instance. We should probably recommend the use of a newer Vivado version now.

As a local workaround, you can do the following change inside `ibex/vendor/lowrisc_ip/ip/prim_xilinx/rtl/prim_xilinx_clock_gating.sv`:
```sv
    BUFGCE #(
      .SIM_DEVICE("7SERIES")
    ) u_bufgce (
```
to
```sv
    BUFGCE u_bufgce (
```

vogelpi:
I am closing this issue as we now require Vivado 2020.2 which understands this parameter.

PR #1426, opened 2021-08-25
[util] Document minimal requirement for Xilinx Vivado

vogelpi:
I am not sure if we really need 2020.2 or if also 2020.1 would be sufficient. For now, I just used the same as for OpenTitan. @ctopal can you please check which version you used for the power simulation work you've done?

This is related to lowRISC/Ibex#1425.


ctopal:
I believe power simulation scripts are not dependent on the Vivado version (tcl hooks and `.saif` related operations are mentioned back in Vivado 2013 versions). I was using 2021.1 while testing the code.

ctopal (approved):

tool_requirements.py
```diff
@@ -11,4 +11,8 @@
         'min_version': '2020.03-SP2',
         'as_needed': True
     }
```
ctopal, line 13: I think you might have missed a comma here

vogelpi: Absolutely, thanks for noting that. I was already debugging the documentation build but had no clue what was wrong ;-)

vogelpi:
> I believe power simulation scripts are not dependent on the Vivado version (tcl hooks and `.saif` related operations are mentioned back in Vivado 2013 versions). I was using 2021.1 while testing the code.

Alright, so it's probably best if we just recommend the same version we are using for OpenTitan.

PR #1428, opened 2021-08-31
Ibex fixes

GregAC:
A couple of small fixes that came up whilst working on the cosim DV

### DIFF
--- a/rtl/ibex_counter.sv
+++ b/rtl/ibex_counter.sv
@@ -1,5 +1,9 @@
 module ibex_counter #(
-  parameter int CounterWidth = 32
+  parameter int CounterWidth = 32,
+  // When set `counter_val_upd_o` provides an incremented version of the counter value, otherwise
+  // the output is hard-wired to 0. This is required to allow Xilinx DSP inference to work
+  // correctly. When `ProvideValUpd` is set no DSPs are inferred.
+  parameter bit ProvideValUpd = 0
 ) (
   input  logic        clk_i,
   input  logic        rst_ni,
@@ -8,7 +12,8 @@ module ibex_counter #(
   input  logic        counterh_we_i,
   input  logic        counter_we_i,
   input  logic [31:0] counter_val_i,
-  output logic [63:0] counter_val_o
+  output logic [63:0] counter_val_o,
+  output logic [63:0] counter_val_upd_o
 );
 
   logic [63:0]             counter;
@@ -17,9 +22,11 @@ module ibex_counter #(
   logic                    we;
   logic [CounterWidth-1:0] counter_d;
 
+  // Increment
+  assign counter_upd = counter[CounterWidth-1:0] + {{CounterWidth - 1{1'b0}}, 1'b1};
+
   // Update
   always_comb begin
-
     // Write
     we = counter_we_i | counterh_we_i;
     counter_load[63:32] = counter[63:32];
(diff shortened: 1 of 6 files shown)


rtl/ibex_counter.sv
```diff
@@ -8,7 +8,8 @@ module ibex_counter #(
   input  logic        counterh_we_i,
   input  logic        counter_we_i,
   input  logic [31:0] counter_val_i,
-  output logic [63:0] counter_val_o
+  output logic [63:0] counter_val_o,
+  output logic [63:0] counter_val_upd_o
```
vogelpi, line 12: I don't question that providing `counter_val_upd_o` is required to fix the issues around `minstret` but I am wondering what will happen on FPGA. Without this, all counters are implemented using DSP slices and if the counter width is below 48 bits, the output registers of the DSP slice can be used - so both flops and LUTs can be saved.

Now, for `minstret` (64 bits wide by spec) I think we weren't using the output registers of the DSP slice anyway, so it's not a big deal FF wise. For the compressed instruction counter (40 bits wide by default), we might now use separate FFs, too. I don't think this is a big deal but it's probably worth to check that at least, the DSP slices can still be used. Would you mind to have a quick look @GregAC ?

vogelpi (approved):
This looks good, thanks @GregAC - also for the clear comments! :+1: 

I just have one comment around the use of DSP slices and output regs on FPGA. And the PR needs to be rebased.

rtl/ibex_counter.sv
GregAC: A good point! Turns out this broke the DSP inference.

I've altered `ibex_counter` so there's a parameter that controls whether or not the `counter_val_upd_o` output is driven.  Not hugely elegant but necessary to keep inferring DSP where we can.

GregAC:
@vogelpi have requested a re-review as it's not just a simple fix change.

rtl/ibex_counter.sv
vogelpi: Thanks @GregAC for checking this out! Interesting that just the fact that there is an output signal (even if unused) broke DSP inference. I think the solution you came up with is perfectly fine. LGTM!

vogelpi (approved):

rtl/ibex_counter.sv
GregAC: I guess the inference occurs before they flatten the hierarchy, or haven't run the optimisation pass to realise the output is unused.

GregAC: oh and for compressed instructions we don't get an inferred DSP counter but I don't think that's a big deal.

vogelpi: I think for both minstret and compressed instructions we don't get a DSP counter right?

GregAC: Yes, though we didn't before with minstret as it was always 64 bit

vogelpi: Ah I see. Well I would say it's better to have a counter that is correct than an optimized one that isn't ;-)
One thing that you could do is to set `ProvideValUpd` to 1 if and only if the write-back stage is enabled.

GregAC: I did think about that, decided against it as currently `ibex_cs_registers` doesn't need to care if the write-back stage is enabled so preferred not to add that as a parameter for the sake of a pretty tiny FPGA specific area optimisation.

PR #1433, opened 2021-09-10
Misc fixes

GregAC:


examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -8,7 +8,7 @@ module top_artya7 (
     output [3:0]        LED
 );
 
-  parameter int          MEM_SIZE     = 256 * 1024; // 256 kB
+  parameter int          MEM_SIZE     = 64 * 1024; // 64 kB
```
vogelpi, line 11: This change was done by @ctopal for doing power simulations with CoreMark. Is there a way to keep supporting both options?

GregAC: I think we could do a parameter override to set MEM_SIZE from the fusesoc core file. Probably best to leave it at 64 kB as a default then override when building for power simulation. Is there a specific build flow for the power simulation somewhere?

vogelpi: This sounds good to me. @ctopal should be able to give the details on the build flow. IIRC there was a flag for fusesoc that triggered the power sim.

ctopal: If we set the flag, `vivado_setup_hooks.tcl` under `ibex/examples/fpga/artya7/util/` sources the power simulation related script under the same directory. Other than that, we don't have a specific build flow for power simulation.

vogelpi: It would be good if we could merge this. Someone from the community stumbled across the same thing, see #1500 .

examples/fpga/artya7/rtl/top_artya7.sv
```diff
@@ -8,10 +8,13 @@ module top_artya7 (
...
 );
 
-  parameter int          MEM_SIZE     = 256 * 1024; // 256 kB
-  parameter logic [31:0] MEM_START    = 32'h00000000;
-  parameter logic [31:0] MEM_MASK     = MEM_SIZE-1;
-  parameter              SRAMInitFile = "";
+  parameter int          FPGAPowerAnalysis = 0;
+  // Choose 64kb memory for normal builds and 256kb for FPGAPowerAnalysis builds. The latter will
+  // not fit in the Arty A7-35 Board FPGA.
+  parameter int          MEM_SIZE          = FPGAPowerAnalysis == 0 ? 64 * 1024 : 256 * 1024;
```
vogelpi, line 14: Awesome, thanks for adding this parameter :+1:

vogelpi (approved):
LGTM, thanks @GregAC !

vogelpi:
This PR also fixes #1506. Let's thus merge it.

PR #1434, opened 2021-09-16
[TEST] Fail the linter

imphil:


imphil:
@wsipak somehow the review comments don't show up. https://github.com/lowRISC/ibex/runs/3620371102?check_suite_focus=true has the log, which says at the beginning:

> fatal: couldn't find remote ref lint-test-failing
error: pathspec 'lint-test-failing' did not match any file(s) known to git
the PR branch is lint-test-failing

Do you have an idea what's going on? I don't see a similar error in the runs you've done in your branch (but that doesn't need to mean anything).

imphil:
Or do I need to configure reviewdog manually?

wsipak:
@imphil Thank you so much for creating this test PR. Surely there's no need for you to configure anything. I'm going to prepare a fix for this and will update here when I figure it out.

imphil:
@wsipak a quick update: https://github.com/lowRISC/ibex/pull/1435 was opened recently and has seen messages from the bot, unfortunately they aren't the right messages or in the right lines (the lines mentioned don't exceed 100 characters).

wsipak:
@imphil Thank you for letting me know. I've run Verible locally on the branch in the PR and it looks like the messages are reported for wrong lines (lines 163-168, 175-177, 180, 197 exceed the limit, not the ones that are reported).
I'm working on this, thanks a lot.

imphil:
Thanks @wsipak! And no worries or pressure, it's not blocking anything on our side.

imphil:
Hm. https://github.com/lowRISC/ibex/runs/3665088172?check_suite_focus=true shows that the the new version runs, but doesn't produce any output. Running the same command the CI shows locally shows a huge amount of lint violations. Do you have an idea what's going on @wsipak?

imphil:
Close/reopen to see how the updated action copes.

rtl/ibex_multdiv_slow.sv
```diff
@@ -13,13 +13,13 @@
 
 module ibex_multdiv_slow
 (
-  input  logic             clk_i,
+      input  logic             clk_i,
```
tgorochowik, line 16: This would have been detected by the formatter. We have an action for the formatter too, but it's much more experimental than the linter, that's why we wanted to start with linting only.

rtl/ibex_multdiv_slow.sv
```diff
@@ -82,10 +82,10 @@ module ibex_multdiv_slow
   // ALU Operand MUX //
   /////////////////////
 
-  // Intermediate value register shared with ALU
-  assign imd_val_d_o[0]  = {1'b0,accum_window_d};
+  // Intermediate value register shared with ALU. This is a super long line because I don't have anything else to say I'm just going to full it with random characters that my keyboard is producing. It could do even more but I think we've already exceeded the 100 character threshold, haven't we, Verible?
```
tgorochowik, line 85: This should be detected and annotated though, we'll take a look at this

wsipak: https://github.com/chipsalliance/verible/blob/master/verilog/analysis/checkers/line_length_rule_test.cc#L85-L109
We can expect Verible to forgive this, as the long line is a comment.

rtl/ibex_multdiv_slow.sv
imphil: Yeah, I wasn't sure which problems the linter is detecting currently, I just tried to mess up the formatting in various ways :-)

rtl/ibex_multdiv_slow.sv
tgorochowik: It's something to discuss and solve in Verible then, not in the action itself :) Let's move the discussion here: https://github.com/chipsalliance/verible/issues/941

imphil: I didn't know that (and it's unexpected behavior at least to me). Thanks for filing an issue @tgorochowik.

I can check, but I think our commercial sign-off lint in OpenTitan would have flagged this line.

imphil:
The verible lint action is now in active use, no need to keep this TEST PR open any longer.

PR #1435, opened 2021-09-17
Co-simulation framework

GregAC:
You can also look at the spike changes here: https://github.com/lowRISC/riscv-isa-sim/commit/d3cd47117e47bef6864d8e5c84f38a9cde0b42ae (it's in a branch in our Spike fork) you can comment directly on that commit.

This doesn't include the simple system integration by previous PR did as I've just been working on the UVM environment recently. I do intend to add a simple system integration once we've got this PR merged.

I've run a full regression with a good pass rate (over 90%) though I've been fiddling with it since so may have broken a few more things. In particular it seems turning on the icache causes issues with instruction errors and a recent RTL change I did seems to be hitting an assertion in ibex_controller.

I'll be looking at these issues in parallel (and doing another full regression) but wanted to get this PR out finally!


dv/cosim/cosim_dpi.cc
```diff
@@ -0,0 +1,92 @@
...
+void riscv_cosim_clear_errors(void *cosim_handle) {
+  auto cosim = static_cast<Cosim *>(cosim_handle);
+
+  cosim->clear_errors();
+}
+
+void riscv_cosim_write_mem_byte(void *cosim_handle, const svBitVecVal *addr,
+                                const svBitVecVal *d) {
+  auto cosim = static_cast<Cosim *>(cosim_handle);
+  cosim->backdoor_write_mem(addr[0], 1, (const uint8_t *)(&d[0]));
```
rswarbrick, line 91: The type-cast here is a bit magic. Maybe it's better to explicitly extract 8 bits with a `unt8_t foo = d[0] & 0xff;` and then pass `&foo`?

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+  FILE *log_file = nullptr;
+  if (trace_log_path.length() != 0) {
+    log = std::make_unique<log_file_t>(trace_log_path.c_str());
+    log_file = log->get();
+  }
+
+  processor =
+      std::make_unique<processor_t>("RV32IMC", "MU", DEFAULT_VARCH, this, 0,
+                                    false, log_file, nullptr, secure_ibex,
+                                    icache_en);
```
rswarbrick, line 30: Is there any way to pass `log_file` after constructing the object? If so, you could avoid the `unique_ptr` here, I think?

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+
+  // TODO: Uncomment when rebasing on up to date spike
```
rswarbrick, line 32: Stale TODO message? Nothing seems to be commented out...

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+  } else if (aligned_pc != aligned_addr && (aligned_pc + 4) != aligned_addr) {
+    // Otherwise check if the aligned PC matches with the aligned address or an
+    // incremented aligned PC (to capture the unaligned 4-byte instruction
+    // case). Assume a successful iside access if either of these checks are
+    // true, otherwise assume a dside access and check against DUT dside
+    // accesses.
+    dut_error = check_mem_access(false, addr, len, bytes);
+  }
+
+  return !(bus_error | dut_error);
```
rswarbrick, line 70: Nit: I think `||` is probably more common for booleans in C++ (and similarly below).

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+      if (write_reg != 0) {
+        std::stringstream err_str;
+        err_str << "Synchronous trap occurred at PC: " << std::hex << pc
+                << "but DUT wrote to register: x" << std::dec << write_reg;
+        errors.emplace_back(err_str.str());
+
+        return false;
+      }
+
+      return errors.size() == 0;
```
rswarbrick, line 152: Is there any way for `errors.size()` to be nonzero if we didn't return `false` above? If not, maybe an assertion and then an explicit `return true` would be better?

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+
+    if ((reg_change.first & 0xf) == 0) {
+      // register is GPR
+      // should never see more than one GPR write per step
+      assert(!gpr_write_seen);
+      int cosim_write_reg = reg_change.first >> 4;
+
+      if (write_reg == 0) {
+        std::stringstream err_str;
+        err_str << "DUT didn't write register, but one was expected to x"
```
rswarbrick, line 190: I think this message has been partially reworded?

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+                << " but DUT didn't report one at PC " << pc;
+        errors.emplace_back(err_str.str());
+
+        return false;
+      }
+
+      if (!initial_pc_match) {
+        std::stringstream err_str;
+        err_str << "PC mismatch at synchronous trap, DUT: " << std::hex << pc
+                << " expected: " << std::hex << initial_pc;
```
rswarbrick, line 137: std::hex is sticky, so you only need it when you're switching from decimal.

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+    if ((reg_change.first & 0xf) == 0) {
+      // register is GPR
+      // should never see more than one GPR write per step
+      assert(!gpr_write_seen);
+      int cosim_write_reg = reg_change.first >> 4;
+
+      if (write_reg == 0) {
+        std::stringstream err_str;
+        err_str << "DUT didn't write register, but one was expected to x"
+                << std::dec << cosim_write_reg;
```
rswarbrick, line 191: Nit: You don't actually need the std::dec here (and similarly below).

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+                << cosim_write_reg << " DUT: " << std::hex << write_reg_data
+                << " expected: " << std::hex << cosim_write_reg_data;
+        errors.emplace_back(err_str.str());
+
+        return false;
+      }
+
+      gpr_write_seen = true;
+    } else if ((reg_change.first & 0xf) == 4) {
+      int cosim_write_csr = (reg_change.first >> 4) & 0xfff;
```
rswarbrick, line 221: Maybe a note here saying that this is the CSR case would be handy.

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+
+  for (auto reg_change : reg_changes) {
+    // reg_change.first provides register type in bottom 4 bits, then register
+    // index above that
+
+    // Ignore writes to x0
+    if (reg_change.first == 0)
+      continue;
+
+    if ((reg_change.first & 0xf) == 0) {
```
rswarbrick, line 182: I'd consider moving the `on_gpr_write` / `on_csr_write` functionality into separate helper functions. It might make the (rather mammoth) `step()` method a bit easier to understand.

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+    err_str << "DUT generated an iside error for address: " << std::hex
+            << pending_iside_err_addr << "but the ISS didn't produce one";
+    errors.emplace_back(err_str.str());
+
+    return false;
+  }
+
+  pending_iside_error = false;
+
+  return errors.size() == 0;
```
rswarbrick, line 255: As above: could this just be an assertion and a `return true`?

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    pending_access_done = true;
+  }
+
+  // Check data from expected access matches pending DUT access.
+  // Data is ignored on error responses to loads so don't check it. Always check
+  // store data.
+  if (store || !top_pending_access_info.error) {
+    // Combine bytes into signal word
```
rswarbrick, line 423: Single word?

rswarbrick (commented):
Some comments on the first 2 commits, but this looks really neat!

tomroberts-lowrisc (approved):
LGTM - we can fix up any issues as we gradually enable it as our main testing methodology

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_agent.sv
```diff
@@ -0,0 +1,41 @@
...
+
+    dmem_port = new("dmem_port", this);
+    imem_port = new("imem_port", this);
+  endfunction
+
+  virtual function void build_phase(uvm_phase phase);
+    super.build_phase(phase);
+
+    rvfi_monitor = ibex_rvfi_monitor::type_id::create("rvfi_monitor", this);
+    scoreboard = ibex_cosim_scoreboard::type_id::create("scoreboard", this);
```
udinator, line 27: nit: alignment of `=`

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
```diff
@@ -0,0 +1,189 @@
...
+    dmem_port = new("dmem_port", this);
+    imem_port = new("imem_port", this);
+    cosim_handle = null;
+  endfunction
+
+  function void build_phase(uvm_phase phase);
+    super.build_phase(phase);
+
+    if (!uvm_config_db#(core_ibex_cosim_cfg)::get(this, "", "cosim_cfg", cfg)) begin
+      `uvm_fatal(get_full_name(), "Cannot get cosim configuration")
```
udinator, line 43: `get_full_name()` -> ``` `gfn ``` ?

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
```diff
@@ -0,0 +1,189 @@
...
+
+    forever begin
+      fork
+        run_cosim_rvfi();
+        run_cosim_dmem();
+        run_cosim_imem();
+        run_cosim_imem_errors();
+        wait (instr_vif.instr_cb.reset === 1'b1);
+      join_any
+      disable fork;
```
udinator, line 70: nit: since this is a generic `disable fork`, maybe wrap it in an isolation process?
```systemverilog
forever begin
  fork begin : isolation_fork
    fork
      run_cosim_rvfi();
      ...
    join_any
    disable fork;
  end join
  if (instr_vif.instr_cb.reset === 1'b1) handle_reset();
end
```

or something along those line, perhaps

imphil:
Closing/reopening to see if the linter errors are now shown on the right line.

dv/cosim/cosim.h
```diff
@@ -0,0 +1,131 @@
...
+  virtual bool backdoor_read_mem(uint32_t addr, size_t len,
+                                 uint8_t *data_out) = 0;
+
+  // Step the co-simulator, checking register write and PC of executed
+  // instruction match the supplied values. A `write_reg` of 0 indicates no
+  // register write occurred.
+  //
+  // `sync_trap` is set to true when the instruction caused a synchronous trap.
+  // In this case the instruction doesn't retire so no register write occurs (so
+  // `write_reg` must be 0).
```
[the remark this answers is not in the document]
GregAC: It'd be easier semantics but potentially worse verification. The idea was if everything is working correctly `write_reg` will be 0 when `sync_trap` is 0. If they differ something has gone wrong (either an RTL issue or the tracking logic that drives the RVFI is messed up).

dv/cosim/cosim.h
```diff
@@ -0,0 +1,131 @@
...
+
+  // Set the state of the NMI (non-maskable interrupt) line.
+  //
+  // The NMI signal is a level triggered interrupt. When the NMI is taken the
+  // CPU ignores the NMI line until an `mret` instruction is executed. If the
+  // NMI is asserted following the `mret` (regardless of whether it has been
+  // deasserted or not whilst the first NMI is being handled) an new NMI is
+  // taken.
+  //
+  // When an NMI is due to be taken that will occur at the next call of `step`.
```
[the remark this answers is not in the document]
GregAC: I've left this unspecified as it's architecturally defined, so effectively up to the ISS to choose the priority. I'll add some comments to this effect.

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+    pending_iside_error = false;
+    dut_error = true;
+  } else if (aligned_pc != aligned_addr && (aligned_pc + 4) != aligned_addr) {
+    // Otherwise check if the aligned PC matches with the aligned address or an
+    // incremented aligned PC (to capture the unaligned 4-byte instruction
+    // case). Assume a successful iside access if either of these checks are
+    // true, otherwise assume a dside access and check against DUT dside
+    // accesses.
+    dut_error = check_mem_access(false, addr, len, bytes);
+  }
```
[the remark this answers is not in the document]
GregAC: It does, practically we can ensure our testbench won't generate a data and instruction fetch to the same address close to each other. As the two buses are separate from Ibex's point of view and there's no internal hazarding or anything else between the two this shouldn't be an issue for our verification work.

It's done like this as there's no way to determine in `mmio_load` if an access is coming from the dside or the iside without something like this or a spike modification. The modification is fairly trivial but I felt it was less likely to be upstreamable.

I shall add some documentation to make this restriction clear.

dv/cosim/spike_cosim.cc
```diff
@@ -0,0 +1,516 @@
...
+
+  return !(bus_error | dut_error);
+}
+
+bool SpikeCosim::mmio_store(reg_t addr, size_t len, const uint8_t *bytes) {
+  bool bus_error = !bus.store(addr, len, bytes);
+  bool dut_error = check_mem_access(true, addr, len, bytes);
```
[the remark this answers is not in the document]
GregAC: `dut_error & !bus_error` means the access saw a fault in the RTL so we have to ensure the same fault is produced in the ISS.

`!dut_error & bus_error` means the memory we're trying to access has produced an error response in the ISS framework, probably because we've failed to map some address within the cosim that the RTL accesses.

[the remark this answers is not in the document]
GregAC: Just realised this is a little confusing due to the overloading of the meaning of the return value for `check_mem_access`. A `false` from there either means a fault was generated on the access or the check has failed. I felt generating a fault on the ISS version of the access when it doesn't match what the RTL produced made more sense but I could make the error value easier to understand and return `true` on any check failures and we'd still see a mismatch.

Better yet I can switch to a enum and use a 3-valued return.

dv/cosim/spike_cosim.cc
GregAC: Yes, `check_mem_accesses` adds to `errors`. In particular for stores if there's a bad address/be/data that won't be checked by anything in `step` itself.

In general the idea was anything can add to `errors` and `step` will report failure if so. For now it's just `check_mem_accesses` that does this.

dv/cosim/spike_cosim.cc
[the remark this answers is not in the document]
rswarbrick: Hmm, understood. I think that teaching Spike about Harvard architectures seems like an obvious thing to do, though. Maybe we should include a FIXME and open up an issue on Spike?

dv/cosim/spike_cosim.cc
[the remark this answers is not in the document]
rswarbrick: The enum sounds like a good plan to me: it'll probably save some careful documentation wordsmithing :-)

dv/cosim/spike_cosim.cc
rswarbrick: Ah! Makes sense. Maybe it would be worth adding a comment that explains how this works?

dv/cosim/spike_cosim.cc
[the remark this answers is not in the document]
GregAC: Sounds good, lets see how upstreaming goes and if spike are receptive to the changes we'll give the one needed for this a go.

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
```diff
@@ -0,0 +1,189 @@
...
+    forever begin
+      // Take stream of transaction from imem monitor. Where an imem access has an error record it
+      // in failed_iside_accesses. If an access has succeeded remove it from failed_imem_accesses if
+      // it's there.
+      // Note all transactions are 32-bit aligned.
+      imem_port.get(mem_op);
+      if (mem_op.error) begin
+        failed_iside_accesses[mem_op.addr] = 1'b1;
+      end else begin
+        if (failed_iside_accesses.exists(mem_op.addr)) begin
```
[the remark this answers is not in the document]
GregAC: Potentially, in particular if you're using the prefetch buffer it could contain a failed iside access that then gets flushed due to a branch and then when it's tried again it succeeds. Maybe the same can occur with the icache, though could require some explicit icache flushes.

PR #1442, opened 2021-09-23
Set Verible action version to 'main'

wsipak:
Yet again we're solving the problems discussed here:
https://github.com/lowRISC/ibex/pull/1434

Turns out we had updated Verible to a version in which it switched from using `stdout` to `stderr` for printing rule violations, so the output wasn't passed to Reviewdog as expected.
Right now the action uses `stderr`, as it should.

EDIT:
The things above remain true, but we're changing the version to 'main' to make it easier to apply fixes without incrementing the version each time.

### DIFF
--- a/.github/workflows/pr_lint_review.yml
+++ b/.github/workflows/pr_lint_review.yml
@@ -42,7 +42,7 @@ jobs:
       - run: |
           unzip event.json.zip
       - name: Run Verible linter action
-        uses: chipsalliance/verible-linter-action@v1.2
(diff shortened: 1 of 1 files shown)


imphil:
Thanks!

(To keep a record here: We just discussed in a call to "pin" the action to a rolling "version" for now instead of a specific release to avoid regular update PRs like this one while the action is experimental on Ibex and in development. We'll revisit this once we've done more testing and are happy with the action as-is.)

imphil (approved):

PR #1448, opened 2021-10-05
Update lowrisc_ip to lowRISC/opentitan@ad629e3e6

rswarbrick:
(This pulls in a fix for the annoying "unknown item: Formal" messages: see https://github.com/lowRISC/ibex/issues/1445)

Update code from upstream repository
https://github.com/lowRISC/opentitan to revision
ad629e3e6e70c5eaa3c2dd68457b0a020448b35f

* [dvsim] Introduce {self_dir} as variable (Philipp Wagner)
* [dvsim] Small cleanups (Philipp Wagner)
* [prim_lfsr] Minor lint fix (Michael Schaffner)
* [dv] Update sec_cm testplan (Weicai Yang)
* [prim/lint] Move waiver to correct waiver file (Michael Schaffner)
* [prim_assert] Relocate waivers to dedicated prim_assert.waiver file
  (Michael Schaffner)
* [alert_handler] Lint fixes and waiver updates (Michael Schaffner)
* [prim_lc_receiver] Add parameter to select reset value (Michael
  Schaffner)
* [lint] Add lint waiver for IP regfiles with shadow resets (Michael
  Schaffner)
* [fpv] Fix Verible lint errors (Philipp Wagner)
* [prim_lfsr] Minor lint fixes (Timothy Chen)
* [clkmgr] Fix measurement control CDC (Timothy Chen)
* [fpv/prim_counter] Pad one bit to include overflow case (Cindy Chen)
* [fpv] Fix issue lowRISC#8371 (Zeeshan Rafique)
* [flash_ctrl] Flash ctrl security hardening (Timothy Chen)
* [dv] Fix CI error (Cindy Chen)
* [prim_alert_*] Extend SVAs for FPV (Michael Schaffner)
* [prim_alert_*] Update DV TB to respect initialization timing
  (Michael Schaffner)
* [prim_alert_rxtx_fpv] Update FPV environment and fix SVAs (Michael
  Schaffner)
* [prim_alert_sender] Update sender to support in-band reset mechanism
  (Michael Schaffner)
* [prim_alert_sender] Simplify sender and clear ping req upon sigint
  (Michael Schaffner)
* [prim_lc_sender] Add option to select reset value (Michael
  Schaffner)
* [prim] Correct assertion valid term (Timothy Chen)
* [prim_lc_combine] Align behavior of lc combine with mubi functions
  (Michael Schaffner)
* [fpv/tool] Support GUI mode on dvsim (Cindy Chen)
[...]

PR #1460, opened 2021-10-19
[icache] Add Support for ICache Scrambling

ctopal:
This commit includes switching to a scrambling RAM primitive for
ICache data and tag RAMs. Also includes various testbench changes on ICache DV environment

Note: This PR includes #1448 as its first commit, which will be removed when that PR is merged.


ctopal:
Latest changes include a logic inside icache which:

- Registers the incoming invalidation request 
- Locks look-ups, writes or invalidation process until a valid scrambling key is provided

I did some basic smoke tests and it seems like there is no problem with the overall workings but I saw a slight downtick on the usage of the cache itself (fetch ratio being 3-4% higher). I am investigating why that might be the case but I think the current version can still be a candidate for review.

vendor/lowrisc_ip/ip/prim/prim.core
```diff
@@ -21,7 +21,7 @@ filesets:
       - lowrisc:prim:flop_2sync
       - lowrisc:prim:arbiter
       - lowrisc:prim:fifo
-      - lowrisc:prim:alert
+#      - lowrisc:prim:alert
```
tomroberts-lowrisc, line 24: Is this intentional? If so it will need to go in via a vendoring patch

vendor/lowrisc_ip/ip/prim/prim_ram_1p_scr.core
```diff
@@ -31,4 +29,4 @@ targets:
   default:
     filesets:
       - tool_verilator ? (files_verilator_waiver)
-      - files_rtl
+      - files_rtl
```
tomroberts-lowrisc, line 32: Same above for this vendored file (needs to be fixed upstream or with a vendoring patch) 
Also missing new line at end of file

tomroberts-lowrisc (commented):
This looks like it's getting there! A couple of questions to address and we should be able to merge.

vendor/lowrisc_ip/ip/prim/prim.core
ctopal: This was because alert primitive is dependent on `lc_ctrl_pkg` (from OpenTitan). Problem comes up specifically with this PR because `ram_1p_scr` module is dependent on `prim:all` and up to this point not many `.core` files were dependent on `prim:all` in Ibex.

ctopal: https://github.com/lowRISC/opentitan/pull/9306 and https://github.com/lowRISC/opentitan/pull/9304 does the trick for this PR. Thanks @rswarbrick for the help!

dv/verilator/simple_system_cosim/ibex_simple_system_cosim.core
```diff
@@ -100,6 +100,12 @@ parameters:
     paramtype: vlogparam
     description: "Number of PMP regions"
 
+  Scramble:
+    datatype: int
+    default: 0
+    paramtype: vlogparam
+    description: "Enables security hardening features (EXPERIMENTAL) [0/1]"
```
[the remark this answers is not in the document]
ctopal: Yup, fixed it sorry!

ctopal:
Thanks @tomroberts-lowrisc for the review and comments. Vendor side of things are waiting for CI and merge at OpenTitan side, for the other comments:

I have now included `icache_inval_i` back to the logic with the addition of `inval_lock` so that the design would also consider waiting for scrambling key part.

rtl/ibex_icache.sv
```diff
@@ -799,19 +802,11 @@ module ibex_icache import ibex_pkg::*; #(
...
-        if (fill_way_en[fb]) begin
-          fill_way_q[fb]  <= sel_way_ic1;
-        end
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        fill_way_q[fb]  <= '0;
+      end else if (fill_way_en[fb]) begin
+        fill_way_q[fb]  <= sel_way_ic1;
       end
     end
```
ctopal, line 811: This last push also includes a minor bug fix here. I tried to make it clear in the commit message, please let me know if further clarification is needed for this change.

dv/uvm/icache/dv/tb/ic_top.sv
```diff
@@ -89,43 +104,114 @@ module ic_top import ibex_pkg::*; #(parameter bit ICacheECC = 1'b0) (
...
+  ///////////////////////////////
+  // Scrambling Infrastructure //
+  ///////////////////////////////
+
+  if (Scramble) begin : gen_scramble
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        scramble_key   <= 128'h14e8cecae3040d5e12286bb3cc113298;
+        scramble_nonce <= 64'hf79780bc735f3843;
+        scramble_nonce_buf <= {64'hf79780bc735f3843, 64'hf79780bc735f3843};
```
rswarbrick, line 123: Can you change these to be more obviously "dummy" values (0123456 or similar)? We clearly need to do this properly, but it's probably a good idea to make it super-clear in the waves that we haven't yet!

dv/uvm/icache/dv/tb/ic_top.sv
```diff
@@ -89,43 +104,114 @@ module ic_top import ibex_pkg::*; #(parameter bit ICacheECC = 1'b0) (
...
+                                   (|scramble_nonce_i) & icache_inval & (|scramble_nonce_buf);
+
+    assign scramble_key = '0;
+    assign scramble_nonce = '0;
+    assign scramble_nonce_buf = '0;
+    assign scramble_ack = '0;
   end
 
+  // RAMs
+    for (genvar way = 0; way < IC_NUM_WAYS; way++) begin : gen_rams
```
rswarbrick, line 150: Nit: I think this might all be indented by an extra 2 spaces?

dv/uvm/icache/dv/tb/tb.sv
```diff
@@ -21,9 +23,14 @@ module tb #(parameter bit ICacheECC = 1'b0);
   ibex_icache_core_if core_if (.clk(clk), .rst_n(rst_n));
   ibex_icache_mem_if  mem_if  (.clk(clk), .rst_n(rst_n));
 
+  logic scramble_ack;
+  logic [127:0]   scramble_key;
+  logic [63:0] scramble_nonce;
```
rswarbrick, line 28: Nit: Spacing is a bit off here

dv/uvm/icache/dv/tb/tb.sv
```diff
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 //
-module tb #(parameter bit ICacheECC = 1'b0);
+module tb #(
+ parameter bit ICacheECC = 1'b0,
+ parameter bit Scramble = 1'b0);
```
rswarbrick, line 7: Nit: The `);` should be on the following line.

dv/uvm/icache/dv/tb/tb.sv
```diff
@@ -83,6 +98,19 @@ module tb #(parameter bit ICacheECC = 1'b0);
...
+    scramble_nonce = {$urandom(),$urandom()};
+    scramble_ack   = 1'b1;
+    fork begin
+      forever @(posedge core_if.invalidate) begin
+          scramble_ack   = 1'b0;
+          #(20 * $urandom_range(1, 10))
+          scramble_key   = {$urandom(),$urandom(),$urandom(),$urandom()};
+          scramble_nonce = {$urandom(),$urandom()};
+          scramble_ack   = 1'b1;
+      end
```
rswarbrick, line 111: Wow! That's super-hacky! :-D But fine for now: I or someone else can come through and sort this out when we do the DV work proper.

dv/uvm/icache/dv/tb/tb.sv
```diff
@@ -83,6 +98,19 @@ module tb #(parameter bit ICacheECC = 1'b0);
   initial begin
     // drive clk and rst_n from clk_if
     clk_rst_if.set_active();
+    scramble_key   = {$urandom(),$urandom(),$urandom(),$urandom()};
+    scramble_nonce = {$urandom(),$urandom()};
+    scramble_ack   = 1'b1;
+    fork begin
+      forever @(posedge core_if.invalidate) begin
```
rswarbrick, line 105: Would it be easier to take this out of the initial block and just make it `always @(posedge core_if.invalidate) ...` ? It avoids the fork/join stuff.

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -107,19 +108,49 @@ module ibex_simple_system (
   logic [31:0] instr_rdata;
   logic instr_err;
 
+  logic scramble_req, scramble_ack;
+  logic [127:0]   scramble_key;
+  logic [63:0] scramble_nonce;
+
+
```
rswarbrick, line 115: Maybe tidy up the spacing here. Also, I don't think we use double blank lines elsewhere. Maybe drop them?

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -107,19 +108,49 @@ module ibex_simple_system (
...
   assign instr_gnt = instr_req;
   assign instr_err = '0;
 
+
   `ifdef VERILATOR
     assign clk_sys = IO_CLK;
     assign rst_sys_n = IO_RST_N;
+    always @(posedge clk_sys) begin
+      // TODO: Scrambling interface implementation in Verilator sim
```
rswarbrick, line 124: Is this going to support reseeding? (The stuff you do with the fork/join below)

examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -107,19 +108,49 @@ module ibex_simple_system (
...
+            scramble_key   = {$urandom(),$urandom(),$urandom(),$urandom()};
+            scramble_nonce = {$urandom(),$urandom()};
+            scramble_ack   = 1'b1;
+          end
+        end
+      end
+      join_none
     end
     always begin
+
```
rswarbrick, line 153: Probably drop this newline

ibex_configs.yaml
```diff
@@ -37,6 +38,7 @@ opentitan:
   PMPGranularity           : 0
   PMPNumRegions            : 16
   SecureIbex               : 1
+  Scramble                 : 0
```
rswarbrick, line 41: Sensible, but could you file an issue (on the Ibex repo) to turn this on once we think it works?

rswarbrick (commented):
Just looked at the DV side. Some nits, but it basically looks good to go!

dv/uvm/icache/dv/tb/tb.sv
ctopal: As much as I like fork/join stuff, you are right :-)

dv/uvm/icache/dv/tb/tb.sv
ctopal: I am not sure what would be the proper scrambling key protocol so had to improvise, sorry! (It was really useful when I was testing the functionality though)

examples/simple_system/rtl/ibex_simple_system.sv
ctopal: Yes I believe so, it is essentially the same structure inside ICache block level testbench and it works there.

ctopal:
I happened to see an interesting failure in scrambling logic at `prim_subst_perm` level. Even though input `data_i` is same, we are getting two different `data_o` signals at different times. You can reproduce it by running the command below while using this PR.

```vendor/lowrisc_ip/util/dvsim/dvsim.py dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson -i ibex_icache_ecc --reseed 1 --fixed-seed=1879374868 -w```

File: examples/simple_system/rtl/ibex_simple_system.sv (hunk 12 above)
Comment (rswarbrick) [reply examples/simple_system/rtl/ibex_simple_system.sv:122]: I'm just a bit confused that with non-Verilator targets there's some extra code to handle re-seeding but it doesn't exist for Verilator. Can you explain that?

Comment (ctopal) [reply examples/simple_system/rtl/ibex_simple_system.sv:122]: It can handle reseeding but the difference is not being able to insert some delay between the new key and the valid signal for verilator sims. I can do that with a counter here but not sure if that would be helpful given that we are planning to do some proper DV work for this scrambling key connections in the near future. What do you think, should I implement the counter stuff?

Comment (rswarbrick) [reply examples/simple_system/rtl/ibex_simple_system.sv:122]: Oh, I see. Nah, I'd do the stupidest, simplest thing possible!

**Comment by ctopal** [comment]
The way we test ECC inside ICache causes faulty bits (one or two) inside the scrambled RAMs. However, scrambled RAMs get unscrambled and then sent inside ICache to do the ECC checking. When we unscramble, it might propagate the faulty bits in a way that we might end up inserting 4+ faulty bits to the original memory. Thus, cause ECC to not catch the faulty situation and generate a false negative. 

A proper DV work should include fixing this bit insertion scheme so that we would not see 4+ bits change (or not check when it happens) between original memory content and unscrambled memory when trying to insert bad bits.

In OTBN, we are also using scrambled memories and in there we are first unscrambling the memory, changing the content and scrambling it back to put inside the memory. In order to do that, we need to use the functions from `sram_scrambling_pkg` (which is not inside vendor)

My work-around is much more hacky then this. Seeing this error is not much probable (although the default seed 123 catches it!) so I just changed the default seed to 1234.

File: dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson [hunk 15]
```diff
@@ -23,17 +23,19 @@
...
     {
       name: default
-      en_build_modes: ["{tool}_icache_ecc"]
+      en_build_modes: ["{tool}_icache_ecc",
+                       "{tool}_memutil_dpi_scrambled_build_opts"]
     }
     {
       name: vcs_icache_ecc
-      build_opts: ["-gv tb.ICacheECC=1"]
+      build_opts: ["-gv tb.ICacheECC=1 -gv tb.Scramble=1"]
```
rswarbrick, line 38: This doesn't look quite right to me (because we're not reflecting it in the Riviera config below). But should we just leave it disabled for now anyway, until the DV work is done? Looking at the top-level TB, I think you're never passing a scrambling key, which means that turning on scrambling will essentially disable the cache, no?

rswarbrick (commented):
Other than the config file tweak, the DV side looks good to me.

dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson
ctopal: I agree! I disabled it in ICache block level testbench. In others, it was already turned off too.

rswarbrick: Yep, I was being stupid and thought this was controlling things for the core_uvm tests. Ignore me :-)

rswarbrick (approved):
DV and tooling side looks good to me (although you'll have to rebase over a change to `ibex_dv.f`, it seems).

@GregAC: Please could you take a look at the design side?

rswarbrick:
@ctopal: Would you mind rebasing?
@GregAC: Ping! I think your review is all that's needed for this to go in.

rtl/ibex_top.sv
```diff
@@ -377,42 +402,113 @@ module ibex_top import ibex_pkg::*; #(
...
+
+    logic unused_scramble_inputs = scramble_ack_i & (|scramble_key_i) & (|RndCnstIbexKey) &
+                                   (|scramble_nonce_i) & (|RndCnstIbexNonce) &
+                                   icache_inval & (|scramble_nonce_buf);
+
+    assign scramble_req_o      = 1'b0;
+    assign scramble_key = '0;
+    assign scramble_nonce = '0;
+    assign scramble_nonce_buf = '0;
+    assign scramble_ack = 1'b1;
```
GregAC, line 437: nit: `=` alignment

rtl/ibex_top.sv
```diff
@@ -377,42 +402,113 @@ module ibex_top import ibex_pkg::*; #(
...
+        scramble_nonce <= RndCnstIbexNonce;
+        scramble_nonce_buf <= {RndCnstIbexNonce, RndCnstIbexNonce};
+        scramble_ack   <= 1'b1;
+      end else if (scramble_ack_i) begin
+        scramble_key   <= scramble_key_i;
+        scramble_nonce <= scramble_nonce_i;
+        scramble_nonce_buf <= {scramble_nonce, scramble_nonce_i};
+        scramble_ack   <= scramble_ack_i;
+      end
+      scramble_req_o <= icache_inval;
```
GregAC, line 422: The scramble key interface to the OTP in OpenTitan requires `req` to be held until an `ack` is seen. At least that's what the timing diagrams indicate (see 'Interfaces to SRAM and OTBN Scramblers' in https://docs.opentitan.org/hw/ip/otp_ctrl/doc/index.html#hardware-interfaces). We could say the protocol is a bit different at the Ibex top-level as we already need to do some wrapping for the CDC stuff but it's probably simpler to use the same protocol here.

rtl/ibex_top.sv
```diff
@@ -159,6 +173,15 @@ module ibex_top import ibex_pkg::*; #(
   // Alert signals
   logic                        core_alert_major, core_alert_minor;
   logic                        lockstep_alert_major, lockstep_alert_minor;
+  // Scramble signals
+  logic                        icache_inval;
+  logic [SCRAMBLE_KEY_W-1:0]   scramble_key;
+  logic [SCRAMBLE_NONCE_W-1:0] scramble_nonce;
+  logic [127:0]                scramble_nonce_buf;
+  logic scramble_ack;
```
GregAC, line 181: From it's usage this needs renaming, `scramble_key_valid` or similar looks more suitable.

rtl/ibex_top.sv
```diff
@@ -377,42 +402,113 @@ module ibex_top import ibex_pkg::*; #(
...
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        scramble_key   <= RndCnstIbexKey;
+        scramble_nonce <= RndCnstIbexNonce;
+        scramble_nonce_buf <= {RndCnstIbexNonce, RndCnstIbexNonce};
+        scramble_ack   <= 1'b1;
+      end else if (scramble_ack_i) begin
+        scramble_key   <= scramble_key_i;
+        scramble_nonce <= scramble_nonce_i;
+        scramble_nonce_buf <= {scramble_nonce, scramble_nonce_i};
```
GregAC, line 419: If I've understood correctly the intent is to reuse existing nonce bits as we're not getting enough from OTP to cover both the tag and data memories? Has it been agreed this still meets security requirements?

When acting this way the `scramble_nonce` and `scramble_nonce_buf` names don't make much sense. `scramble_tag_nonce` and `scramble_data_nonce` or similar would be better.

rtl/ibex_top.sv
```diff
@@ -159,6 +173,15 @@ module ibex_top import ibex_pkg::*; #(
   // Alert signals
   logic                        core_alert_major, core_alert_minor;
   logic                        lockstep_alert_major, lockstep_alert_minor;
+  // Scramble signals
+  logic                        icache_inval;
+  logic [SCRAMBLE_KEY_W-1:0]   scramble_key;
+  logic [SCRAMBLE_NONCE_W-1:0] scramble_nonce;
+  logic [127:0]                scramble_nonce_buf;
```
GregAC, line 180: We should use a parameter here too, `SCRAMBLE_NONCE_W * 2` would be better than nothing. Though we could write something that calculates a suitable nonce width from any arbitrary data RAM width (you should be able to copy it from the scrambling primitive itself).

ctopal:
Hi @GregAC thanks for your review. I did the changes that you suggested with the naming of the parameter and nonce signal. Also added a little snippet of documentation. Sorry, I did not notice replicating of nonce signal could cause a security concern.

About the the point with scrambling nonce width of data RAM: We are using the same interface for both tag RAM and data RAM and the width only changes when we are enabling `ICacheECC` parameter. Therefore, it looks like we have two options:

1. Replicate the incoming scrambling nonce if we are enabling `ICacheECC`. (As you said it introduces a possible security concern, whom do you think I should ask?)
2. Separate tag RAM scrambling and data RAM scrambling so that both keys and nonces would be different. (They only need different widths when we are enabling `ICacheECC`, so maybe it's wasteful to do this? Obviously if we are planning to use `ICacheScramble` with `ICacheECC` enabled all the time, this makes more sense.)

rtl/ibex_top.sv
```diff
@@ -373,42 +394,115 @@ module ibex_top import ibex_pkg::*; #(
...
+  if (ICacheScramble) begin : gen_scramble
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        scramble_key       <= RndCnstIbexKey;
+        scramble_nonce     <= RndCnstIbexNonce;
+        scramble_key_valid <= 1'b1;
+      end else if (scramble_key_valid_i) begin
+        scramble_key       <= scramble_key_i;
+        scramble_nonce     <= scramble_nonce_i;
+        scramble_key_valid <= scramble_key_valid_i;
```
GregAC, line 410: With the current logic `scramble_key_valid` will never be 0,

You should rewrite and use an `always_comb` block to specify the control logic for `scramble_key_valid` and `scramble_req_o`. Separate `scramble_key_valid` into `scramble_key_q` and `scramble_key_d` so the `always_ff` block doesn't contain any logic.

`scramble_key_valid_q` won't have the `scramble_key_valid_i` enable that `scramble_key` and `scramble_nonce` have. So I'd have a separate `always_ff` block for it.

dv/uvm/icache/dv/tb/ic_top.sv
```diff
@@ -85,43 +102,116 @@ module ic_top import ibex_pkg::*; #(parameter bit ICacheECC = 1'b0) (
...
+        scramble_key        <= 128'hDDDDDDDDEEEEEEEEAAAAAAAADDDDDDDD;
+        scramble_nonce      <= 64'hBBBBEEEEEEEEFFFF;
+        scramble_key_valid  <= 1'b1;
+      end else if (scramble_key_valid_i) begin
+        scramble_key        <= scramble_key_i;
+        scramble_nonce      <= scramble_nonce_i;
+        scramble_key_valid  <= scramble_key_valid_i;
+      end
+      scramble_req_q <= scramble_req_o;
+    end
```
GregAC, line 128: This has the issues as the logic in ibex_top

GregAC:
Let's go with option 1. for now (replicate nonce and reuse key). We can always review this decision (maybe as part of the D2S review) and that allows us to get this PR merged.

rtl/ibex_top.sv
```diff
@@ -373,42 +394,134 @@ module ibex_top import ibex_pkg::*; #(
...
+        scramble_nonce_q     <= scramble_nonce_d;
+      end
+    end
+
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        scramble_key_valid <= 1'b1;
+        scramble_req_q     <= '0;
+      end else begin
+        scramble_key_valid <= scramble_key_valid_i;
```
GregAC, line 428: `scramble_key_valid_i` will only be asserted for the cycle the OTP returns a new valid key. This needs more logic to set `scramble_key_valid` when a new key comes in, leave it set until a new key is requested, unsetting it when the new key request goes out.

rtl/ibex_top.sv
```diff
@@ -373,42 +394,134 @@ module ibex_top import ibex_pkg::*; #(
...
+
+  if (ICacheScramble) begin : gen_scramble
+
+    always_comb begin
+      if (scramble_key_valid_i) begin
+        scramble_key_d       = scramble_key_i;
+        scramble_nonce_d     = scramble_nonce_i;
+      end else begin
+        scramble_key_d       = '0;
+        scramble_nonce_d     = '0;
```
GregAC, line 409: As we've already got the `if(scramble_key_valid_i)` in the `always_ff` we don't need this `always_comb` block at all. A `always_comb` block would be a clean way to do the logic for `scramble_key_valid`, separating into `scramble_key_valid_q` and `scramble_key_valid_d`.

ctopal:
Hi Greg, sorry about the delayed change. I think it works according to your comments now. I am attaching the screenshot of a req/valid interaction as an example.

Screenshot from 2022-01-17 10-19-58

rtl/ibex_top.sv
```diff
@@ -373,42 +394,129 @@ module ibex_top import ibex_pkg::*; #(
...
 
+  ///////////////////////////////
+  // Scrambling Infrastructure //
+  ///////////////////////////////
+
+  if (ICacheScramble) begin : gen_scramble
+
+  // Scramble key valid starts with OTP returning new valid key and stays high
+  // until we request a new valid key.
+    assign scramble_key_valid_d = scramble_req_q ? scramble_key_valid_i : 1'b1;
```
GregAC, line 405: I think this would be better as something like:

```systemverilog
assign scramble_key_valid_d = scramble_req_q ? scramble_key_valid_i : 
                              icache_inval   ? 1'b0                 :
                                               scramble_key_valid_q;
```

It ensures `scramble_key_valid_q` will be unset the cycle after the invalidate (currently there's one cycle after the invalidate the key is still marked as valid, I think we'd get away with it but this is cleaner). Otherwise logically it will end up doing the same thing but it's better to just say we keep the same value when there's no request in flight rather than forcing it to high.

Issue #1462, opened 2021-10-20
[Bug] Branch prediction on predicted branches causes bad behaviour


sam-shahrestani:
## Steps to reproduce the issue

Consider the following code:
```
int main(int argc, char **argv) {
    static int x = 25;
    int* x_addr = (int*)&x;
    __asm__ (
        "j skip\n\t"

        "predict_to_here:\n\t" // we shouldn't branch here but even if we do, just finish
        "j finish\n\t"
        "wfi\n\t" // this will kill things if hit

        "skip:\n\t"
        "lw t0, 0(%0)\n\t"
        "beqz t0, predict_to_here\n\t" // clearly this branch is not meant to be taken but is predicted taken
        "j finish\n\t"

        "finish:\n\t"
        "nop\n\t"
        :
        :"r" (x_addr)
        : "t0"
    );
    return 0;

}
```
We would expect the following branch pattern in general: `-> skip -> finish`
Building ibex_simple_system with the following parameters works correctly:
`fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast --BranchTargetALU=1 --WritebackStage=1 --BranchPredictor=0`

Building with these parameters causes the core to seemingly "lock up":
`fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast --BranchTargetALU=1 --WritebackStage=1 --BranchPredictor=1`

## Observed Behavior

Ibex overrides `branch_mispredict_addr` by predicting on a predicted branch. This means when a mispredict fires on `beqz t0, predict_to_here` the PC fetched next is at `wfi`.

[waves.zip](https://github.com/lowRISC/ibex/files/7384667/waves.zip)

## Expected Behavior

I honestly don't know. In the prefetch buffer configuration, there is only one `branch_mispredict_addr_q` present so there is no practical way to nest branch predictions as far as I can tell. It seems like branches should just be predicted not taken when we are already on a predicted path.

## My Environment

**Compiler:**
riscv64-unknown-elf-gcc (GCC) 8.2.0

**EDA tool and version:**

Verilator 4.210 2021-07-07 rev v4.210

**Operating system:**

Ubuntu 21.04

**Version of the Ibex source code:**

cfeef7e864926cf4d30e271a3e2aa5d95e0c141f

[...]

PR #1469, opened 2021-11-06
Move not-taken, predicted taken branch address calculation to ID stage

sam-shahrestani:
Currently, the not-taken address of a branch is registered in flops in the prefetch buffer or instruction cache when ever a branch prediction is lodged with these modules. However, as discussed in https://github.com/lowRISC/ibex/issues/1462, there is an issue with this address being overridden by a second predicted branch when the an instruction moves from the prefetch to the IF stage skid buffer. If the instruction in the skid buffer was mispredicted taken, the prefetch will replay from the stored address of the second predicted branch, not the first one.

To solve this without adding extra flops in the IF stage to store the not-taken address of the instruction in the skid buffer, we can get rid of these flops entirely and calculate the not-taken address in the ID stage. This would be fed back along with the existing not-taken mispredict signal `nt_branch_mispredict_i` used to clear the prefetch and allows it to resume fetching on the correct path.

As a code refactoring bonus, we do not need to consider the existence of the branch predictor in the prefetch buffer, instruction cache and fetch FIFO modules any more. Previously, the next address of the fetch FIFO would be registered in flops as the not-taken address using an `if generate` block based on the `BranchPredictor` parameter, but these flops do not exist anymore as of this PR.

Though not specifically relevant to this change, we now also suppress the validity of the skid buffer content by `nt_branch_mispredict_i` which prevents mispredicted instructions in the skid buffer from entering the ID stage. This is required as part of fixing the aforementioned bug.


sam-shahrestani:
This is my first contribution to ibex, so I apologise if anything is out of shape. All feedback is appreciated.

sam-shahrestani:
Looks like Private CI is failing but public CI is passing. Any thoughts?

tomroberts-lowrisc:
I think there are some minor changes required in the icache testbench - should be fairly straightforward

rtl/ibex_if_stage.sv
```diff
@@ -485,6 +483,7 @@ module ibex_if_stage import ibex_pkg::*; #(
     logic        instr_skid_en;
     logic        instr_bp_taken_q, instr_bp_taken_d;
 
+    logic        predicted_branch;
```
tomroberts-lowrisc, line 486: This is causing an error as it is declared inside this generate block, but used later on inside another generate block.
Since it is only used for assertions now, it probably makes sense to move the declaration and assignment down into the block with the assertions (around line 634).

tomroberts-lowrisc (commented):
From a first pass this looks good to me. If you fix the signal declaration issue and the icache testbench portlist we should be able to get this approved and merged.

tomroberts-lowrisc:
Please also squash the commits into one once you think everything is done
Thanks!

sam-shahrestani:
> Please also squash the commits into one once you think everything is done Thanks!

Hmmm, should I edit my Git history to make my forked master branch have a single commit, or did you mean to squash and merge at the end of everything?

tomroberts-lowrisc (approved):
Thanks for doing these changes (and raising the issue in the first place!). This looks good to go to me.

PR #1473, opened 2021-11-18
[dv/icache] Add missing window reset call

tomroberts-lowrisc:
The cache hit-rate tracking logic needs to be reset on every
invalidation.


ctopal (approved):
It solves the issues around scrambling PR too, thanks!

rswarbrick (approved):
Oops, sorry!

PR #1477, opened 2021-11-29
Update lowrisc_ip and add memory backdoor package

ctopal:
First commit is about adding memory backdoor package for Scrambling RAMs. Second commit is updating the lowrisc_ip with the latest version from OpenTitan.


rswarbrick (commented):
Wow! That's a lot of code!

Looks good to me, but could you squash the two commits and add the "added a new package" comment to the auto-generated commit message of the second? That way, we don't get a "partially applied" state in the history.

PR #1476, opened 2021-11-26
[bitmanip] Re-align bitmanip implementation with v.1.0.0 and v.0.93 of the specs - Part 1

vogelpi:
This PR contains the first chunk of work to re-align the bitmanip implementation inside Ibex with the ratified v.1.0.0 spec (covering Zba, Zbb, Zbc, and Zbs only) and draft version v.0.93 (covering also Zbe, Zbf, Zbp, Zbr). In particular, this PR takes care of aligning the ratified sub-extensions Zba, Zbb, Zbc and Zbs. The changes were successfully tested using RISCV-DV. However, to keep verifying this in CI and hence before merging this PR a couple of other things need to happen. For details, see #1470.

PR #1483, opened 2021-12-03
[bitmanip] Re-align bitmanip implementation with v.1.0.0 and v.0.93 of the specs - Part 2

vogelpi:
This PR contains the second part of the work to re-align the bitmanip implementation inside Ibex with the ratified v.1.0.0 spec (covering Zba, Zbb, Zbc, and Zbs only) and draft version v.0.93 (covering also Zbe, Zbf, Zbp, Zbr, Zbt).

In particular, this takes care of aligning the non-ratified sub-extensions Zbe, Zbf, Zbp, Zbr, Zbt with draft v.0.93. It turns out that for RV32, the only difference is the addition of the `xperm.[nbh]` crossbar permutation instructions. This PR implements these instructions as single-cycle instructions by introducing 8 parallel 4-bit-wide, 8-input crossbars to perform the permutation of nibbles. The cost for this implementation is roughly 0.97 kGE when targeting 250 MHz in nangate45 (single cycle mult, WB stage, BT ALU).

Note this PR is based on #1476. Please review the last commit only.


rtl/ibex_alu.sv
```diff
@@ -724,6 +725,96 @@ module ibex_alu #(
...
 
+      //////////////
+      // Crossbar //
+      //////////////
+      // The crossbar permutation instructions xperm.[nbh] (Zbp) can be implemented using 8
+      // parallel 4-bit-wide, 8-input crossbars. Basically, we permute the 8 nibbles of operand_a_i
+      // based on operand_b_i.
+
+      // Generate selector indices and valid signals.
+      // - sel_n[x] indicates which nibble of val is selected for position x.
```
rswarbrick, line 736: Maybe change "position" to "output nibble"? (I had to do some head scratching to work out what you meant!)

Also, could you explain here what the `_n`, `_b` and `_h` suffixes mean? Or maybe it makes more sense to move the byte / halfword declarations to appear just before their respective for loops? (I'm not sure which one looks better!)

rtl/ibex_alu.sv
```diff
@@ -724,6 +725,96 @@ module ibex_alu #(
...
+      logic  [7:0][2:0] sel_n;
+      logic  [3:0]      vld_b;
+      logic  [3:0][1:0] sel_b;
+      logic  [1:0]      vld_h;
+      logic  [1:0][0:0] sel_h;
+
+      // Per nibble, 3 bits are needed for the selection. Other bits must be zero.
+      // sel_n bit mask: 32'b0111_0111_0111_0111_0111_0111_0111_0111
+      // vld_n bit mask: 32'b1000_1000_1000_1000_1000_1000_1000_1000
+      for(genvar i = 0; i < 8; i++) begin : gen_sel_vld_n
```
rswarbrick, line 748: Nit: Space after `for` and similarly below.

rtl/ibex_alu.sv
```diff
@@ -724,6 +725,96 @@ module ibex_alu #(
...
+            // No conversion needed.
+            sel = sel_n;
+            vld = vld_n;
+          end
+
+          ALU_XPERM_B: begin
+            // Convert byte to nibble indicies.
+            for (int b = 0; b < 4; b++) begin
+              sel[b*2 +  0] =    sel_b[b] << 1;
+              sel[b*2 +  1] =    sel_b[b] << 1 | 3'b001;
```
rswarbrick, line 785: This is a bit awkward because it makes use of Verilog's implicit up-conversion to avoid the left-shift overflowing. Maybe it would be less error-prone as `{sel_b[b], 1'b0}` and `{sel_b[b], 1'b1}`, respectively? Similarly for the half-word case below.

rtl/ibex_alu.sv
vogelpi: That's changed. The `_n`, `_b`, `_h` suffixes here and in the instruction name stand for nibbles, bytes, half words. If added this to the comments. Leaving the declaration together is better IMHO as it helps understanding the different conversions below.

rtl/ibex_alu.sv
vogelpi: Done, thanks!

rtl/ibex_alu.sv
vogelpi: Good point Rupert, I honestly don't know how I ended up with the shift and OR in the first place if the concatenation is so simple...

vogelpi:
Thanks for your review @rswarbrick , I've implemented your suggestions.

rswarbrick (approved):
Looks great!

PR #1486, opened 2021-12-08
[secded] Update ECC primitives

msfschaffner:
This is update vendors in a recent update in the secded primitives.
They have been enhanced so that in addition to the all-zero code word, the all-one code word is now also invalid when an "inverted" coding scheme is selected.


msfschaffner:
This will have to be vendored back into OpenTitan in https://github.com/lowRISC/opentitan/pull/9472.

GregAC (approved):

PR #1491, opened 2021-12-10
[dv] Remove riscv_reset_test from regression temporarily

GregAC:
The test seems to be killing full regression runs. Disable for now until
a fix can be identified.

Will do a full run in private CI first to see if this fixes things. We've had lots of recent runs outright fail without results and they all seem to be running this test at the time.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -553,19 +553,20 @@
     compare_final_value_only: 1
     verbose: 1
 
-- test: riscv_reset_test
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Looks reasonable as a quick hack (assuming it works!). If we do commit this, can we also open an issue to remind us to re-enable?

GregAC:
Yes will do,

However I've been debugging in parallel and may be an easy fix, looks like some UVM logging is being far too verbose and giving us multi-gb log files. I would guess they're filling up the disks of the runners causing them to fall over.

GregAC:
No longer required as I've found a proper fix

Issue #1481, opened 2021-12-02
[ci] Let's try to integrate a Yosys synthesis flow into CI


rswarbrick:
The guys at Antmicro have been working on their Surelog+UHDM+Yosys flow and it seems that it can now build Ibex with vanilla Yosys plus their stuff as a plugin.

It would be neat if we could try running this in CI. We'd get early warning if we made a change that broke their frontend (possibly a sign of a genuine Verilog error) and we'd also get instant kGE estimates.

We'll need to decide which version of the software to track and how to "install" it. Maybe the right approach would be for Antmicro to generate a docker image containing the tool, so we could consume that?

We'll also need to figure out how to run stuff. A GitHub action? Some sort of Azure integration?

rswarbrick:
@tgorochowik: Please correct me if I've mangled something here! :-)

tgorochowik:
Sure, thanks! If you want to go with the docker image approach we could create an action (similar to the action with verible) and use it. 
This way it should be possible to control the version of yosys/uhdm plugin we're using.
We're of course open to doing it differently too (be it compiling everything in the CI or providing binaries in some way). Thanks again!

johngt:
I believe that this effort is still ongoing but will likely not make it before V2, so bringing it into the Project to be tracked as a future effort.

GregAC:
Removed from DV project as this is not DV related

PR #1501, opened 2022-01-04
[ci] Add Yosys synthesis flow

rkapuscik:
Opening a draft PR to get some early input and suggestions.

As proposed in #1481, this PR adds a Github action workflow that synthesizes Ibex via Yosys, using Surelog-UHDM frontend.

It obtains a release from [Antmicro's plugin repository](https://github.com/antmicro/yosys-uhdm-plugin-integration/) and [edalize fork](https://github.com/antmicro/edalize) (until UHDM support is merged in https://github.com/lowRISC/edalize/pull/3).

Each flow generates an EDIF artifact that can be downloaded and processed further locally.

### DIFF
--- a/.github/workflows/pr_trigger.yml
+++ b/.github/workflows/pr_trigger.yml
@@ -4,7 +4,10 @@
 name: pr-trigger
 
 on:
-  pull_request:
+  # FIXME: Restore to all PRs
+  push:
+    branches:
+       - master
 
 jobs:
   upload:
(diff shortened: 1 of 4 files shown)


rswarbrick (commented):
Wow, it's cool that this works!

One question: The first commit adds a special core file for the UHDM flow. Eventually, I assume we'd want to combine this with the existing core files (since we're just specifying a different tool). Is there a reason that's difficult at the moment?

rkapuscik:
I based this on a patch we use to add other targets, but I don't think there will be any problem with rolling that into the existing core. I agree that it would be cleaner.

PR #1503, opened 2022-01-07
Add clarifying comment

vogelpi:


GregAC (approved):

vogelpi:
Thanks @GregAC for the quick review. It turns out changing the loop variable to unsigned causes the loop to become infinite. We could also change the check from `i >= 0` to `i < 15` but it's less readable and we don't know if AscentLint will be happy about that. So let's just waive it in AscentLint directly.

PR #1505, opened 2022-01-11
[rtl] Use prim_flop/clock_mux2 primitives for lockstep reset generation

vogelpi:
These primitives can serve as anchor points for constraining backend tools.

As recommended by Michael in https://github.com/lowRISC/opentitan/pull/9935#discussion_r781582710 .


dv/uvm/core_ibex/common/prim/prim_clock_mux2.sv
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Abstract primitives wrapper.
+//
+// This file is a stop-gap until the DV file list is generated by FuseSoC.
+// Its contents are taken from the file which would be generated by FuseSoC.
+// https://github.com/lowRISC/ibex/issues/893
+
+module prim_clock_mux2 #(
```
msfschaffner, line 11: am I assuming correctly that this won't matter on the OT side, because this part is not vendored back in?

msfschaffner (approved):
Cool, thanks for doing this!

dv/uvm/core_ibex/common/prim/prim_clock_mux2.sv
vogelpi: Yes, that's correct. The full `dv` subfolder is excluded when vendoring in Ibex into OT.

vogelpi:
Thanks @msfschaffner for the review. I am merging this now.

Issue #1506, opened 2022-01-13
Instruction fetch can behave incorrectly with delayed grant


zyedidia:
## Observed Behavior
I am observing a bug in which instructions are being fetched/executed that shouldn't be, specifically after commit 72a892d. I used Verilator to run a simulation of Ibex that runs a small test program ([located here](https://github.com/zyedidia/ocibex/blob/ibex_bug/mem/riscvtest.s)) which should ultimately store 25 at memory address 100 and then enter an infinite loop. When Ibex runs the program it stores 7 into memory address 100, before later storing 25 (it should not store 7 first), and when it executes the infinite loop it incorrectly also executes the instruction after the jump, which is an illegal instruction. This causes the core to take an exception and jump to address 0, the exception address for the instantiated core. The start of the program is also located at address 0, which causes it to re-execute the test.

## Expected Behavior
Ibex should only perform one store to address 100: a store of 25, and it should properly jump during the infinite loop without trying to execute an illegal instruction past the end of the program.

## Steps to reproduce the issue
I have uploaded my code to https://github.com/zyedidia/ocibex. The branch `ibex_bug` has Ibex pointing at the buggy commit, while master works properly and points Ibex to the commit just before the buggy one.

To run the example:
```
git clone https://github.com/zyedidia/ocibex
cd ocibex
git checkout ibex_bug
git submodule update --init
make verilator
./obj_dir/test
```

This will print the current instruction for 1000 clock cycles. Note that it keeps repeating the program (due to taking an illegal instruction exception) instead of entering an infinite loop and repeatedly executing `0x0000006f` (`j 0`). This will also generate `trace.vcd` which you can inspect with a waveform viewer. You can run the test on `master` vs. on `ibex_bug` to see the expected vs. observed behavior. The master branch pins an older version of ibex.

## Solution

I believe the problem arises because my connection between the SRAM and Ibex core uses a one-cycle delay before `instr_gnt` is assigned. See the block "SRAM to Ibex" in [soc_top.sv](https://github.com/zyedidia/ocibex/blob/1bf77e1af95a477aa010c614925fe1976f738a86/rtl/soc_top.sv#L111). I copied this from the Arty7 FPGA example, which I assume has a correct connection between SRAM and processor.

I dug into this more, and found that the following line was removed in commit 72a892d, which is suspicious:

```
assign rvalid_or_pmp_err = rdata_outstanding_q[0] & (instr_rvalid_i | rdata_pmp_err_q[0]);
```

[...]

PR #1507, opened 2022-01-13
Fix memory protocol in FPGA example

zyedidia:
The FPGA example does not follow the memory protocol, which causes it to behave incorrectly. This PR fixes that. See the discussion at #1506.


vogelpi:
Thanks @zyedidia for your bug report and the PR. This looks good. However, it's a duplicate. @GregAC already included the fix in #1433. It happened just about an hour before you pushed this PR. I am thus closing this one. Sorry for the inconvenience.

PR #1511, opened 2022-01-19
[ibex_tracer] Void cast function calls

sriyerg:
...whose return value is ignored. This cleans up a compile time warning.

Signed-off-by: Srikrishna Iyer <sriyer@google.com>

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -112,7 +112,7 @@ module ibex_tracer (
 
     if (file_handle == 32'h0) begin
       string file_name_base = "trace_core";
-      $value$plusargs("ibex_tracer_file_base=%s", file_name_base);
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
Wouldn't it be better to actually check the plusarg exists? That's probably less mysterious than just writing to `_0.log` :-)

sriyerg:
> Wouldn't it be better to actually check the plusarg exists? That's probably less mysterious than just writing to `_0.log` :-)

I agree - I just 

> Wouldn't it be better to actually check the plusarg exists? That's probably less mysterious than just writing to `_0.log` :-)

If the plusarg does not exist, it will default to `trace_core_0.log`, which seems fine?

rswarbrick (approved):
Aaand, now I re-read the code! Yes, you're completely right!

PR #1512, opened 2022-01-21
[rtl,doc] Add customisable PMP reset values

### LINKED ISSUE #1423: Implenting implementer specified reset value for ePMP
OpenTitan will want a PMP region configured on reset, along with a non-zero reset value of `mseccfg`: https://github.com/lowRISC/opentitan/issues/7834

How do we want to support this in Ibex? I don't think we should use top-level parameters as it'll be a rarely used feature and would require many parameters. We could use `define`. Perhaps if a single `IBEX_CUSTOM_PMP_RESET_VAL` is defined it then uses a whole set of `IBEX_CUSTOM_PMPCFG0_RESET_VAL` `IBEX_CUSTOM_PMPADDR0_RESET_VAL` etc then specify all of the reset values.

Another possibility is an `ibex_pmp_reset.svh` include. It'd have a bunch of `localparam` giving all of the reset values. We'd supply a default one with Ibex (that just resets to 0) and implementers, such as OpenTitan, would supply a different include. This may mean a patch against Ibex when we vendor into OT, or perhaps we could use a `define` to specify the name of the include file? Maybe there's something we can do with fusesoc to override the include file used in the build (@imphil any cunning ideas)?

GregAC:
Fixes #1423

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -986,6 +986,13 @@ module ibex_cs_registers #(
   // -----------------
 
   if (PMPEnable) begin : g_pmp_registers
+    // PMP reset values
+    `ifdef IBEX_CUSTOM_PMP_RESET_VALUES
+      `include "ibex_pmp_reset.svh"
+    `else
+      `include "ibex_pmp_reset_default.svh"
+    `endif
+
     pmp_mseccfg_t                pmp_mseccfg_q, pmp_mseccfg_d;
     logic                        pmp_mseccfg_we;
     logic                        pmp_mseccfg_err;
@@ -1072,7 +1079,7 @@ module ibex_cs_registers #(
       ibex_csr #(
         .Width     ($bits(pmp_cfg_t)),
         .ShadowCopy(ShadowCSR),
-        .ResetValue('0)
+        .ResetValue(pmp_cfg_rst[i])
       ) u_pmp_cfg_csr (
         .clk_i     (clk_i),
         .rst_ni    (rst_ni),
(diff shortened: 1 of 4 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -1101,7 +1108,7 @@ module ibex_cs_registers #(
       ibex_csr #(
         .Width     (PMPAddrWidth),
         .ShadowCopy(ShadowCSR),
-        .ResetValue('0)
+        .ResetValue(pmp_addr_rst[i][33-:PMPAddrWidth])
```
rswarbrick, line 1111: Would it make sense to add an initial assertion saying that the bottom bits are zero?

rswarbrick (approved):

rtl/ibex_cs_registers.sv
vogelpi: This sounds like a good idea.

rtl/ibex_pmp_reset_default.svh
```diff
@@ -0,0 +1,53 @@
...
+  '{lock: 1'b0, mode: PMP_MODE_OFF, exec: 1'b0, write: 1'b0, read: 1'b0}, // region 11
+  '{lock: 1'b0, mode: PMP_MODE_OFF, exec: 1'b0, write: 1'b0, read: 1'b0}, // region 12
+  '{lock: 1'b0, mode: PMP_MODE_OFF, exec: 1'b0, write: 1'b0, read: 1'b0}, // region 13
+  '{lock: 1'b0, mode: PMP_MODE_OFF, exec: 1'b0, write: 1'b0, read: 1'b0}, // region 14
+  '{lock: 1'b0, mode: PMP_MODE_OFF, exec: 1'b0, write: 1'b0, read: 1'b0}  // region 15
+};
+
+// Addresses are given in byte granularity for readibility. A minimum of two
+// bits will be stripped off the bottom (PMPGranularity == 0) with more stripped
+// off at coarser granularities
```
vogelpi, line 33: Nit: missing full stop at the end.

vogelpi (approved):
LGTM!

PR #1515, opened 2022-01-25
[rtl] Add SEC_CM markers for security features

GregAC:
These are used by the OpenTitan project to track the implementation of
security countermeasures

### DIFF
--- a/rtl/ibex_lockstep.sv
+++ b/rtl/ibex_lockstep.sv
@@ -6,6 +6,8 @@
 // This module instantiates a second copy of the core logic, and compares it's outputs against
 // those from the main core. The second core runs synchronously with the main core, delayed by
 // LockstepOffset cycles.
+
+// SEC_CM: LOGIC.SHADOW
 module ibex_lockstep import ibex_pkg::*; #(
   parameter int unsigned LockstepOffset    = 2,
   parameter bit          PMPEnable         = 1'b0,
@@ -241,6 +243,7 @@ module ibex_lockstep import ibex_pkg::*; #(
   // Bus integrity checking //
   ////////////////////////////
 
+  // SEC_CM: BUS.INTEGRITY
   logic        bus_intg_err;
   logic [1:0]  instr_intg_err, data_intg_err;
   logic [31:0] unused_wdata;
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -422,8 +422,9 @@ module ibex_top import ibex_pkg::*; #(
 
   if (ICacheScramble) begin : gen_scramble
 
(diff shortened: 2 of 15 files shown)


rswarbrick (commented):
This isn't going to work without some changes, I don't think. Each countermeasure listed needs to be of a specific form ("ASSET.CM" or "PLACE.ASSET.CM"). Also the assets and countermeasures need selecting from a specific list. See here: https://docs.opentitan.org/doc/rm/comportability_specification/index.html#countermeasures.

You should be able to try it out with the reggen tooling by copying the code across to the OT repo, adding stuff to the hjson and then running regtool on it.

GregAC:
The counter measures don't map particularly well to the current set of assets/countermeasures documented there. Perhaps this is something to discuss in the D2S meeting.

GregAC:
After a second pass I've mostly managed to map to the existing categories, will need to introduce a couple of new ones.

GregAC:
OT PR to add new categories and list countermeasures in the rv_core_ibex hjson is here: https://github.com/lowRISC/opentitan/pull/10341

GregAC:
@rswarbrick @vogelpi The OT CM list has been updated and approved (https://github.com/lowRISC/opentitan/pull/10341). I've updated this PR to use the new naming.

There's also a new commit to wire up the icache ecc error as a minor alert, as a TODO specified should be added.

I'd appreciate getting this reviewed ASAP to allow us to close Ibex D2S.

rtl/ibex_lockstep.sv
```diff
@@ -241,6 +243,8 @@ module ibex_lockstep import ibex_pkg::*; #(
   // Bus integrity checking //
   ////////////////////////////
 
+  // SEC_CM: BUS.INTEGRITY
+
```
vogelpi, line 247: Nit: empty line

rtl/ibex_top.sv
```diff
@@ -563,6 +565,7 @@ module ibex_top import ibex_pkg::*; #(
 
   // Redundant lockstep core implementation
   if (Lockstep) begin : gen_lockstep
+    // SEC_CM: LOGIC.SHADOW
```
vogelpi, line 568: Nit: would it be worth to add this label also somewhere in the lockstep module?

vogelpi (approved):
LGTM @GregAC !

I would suggest to vendor this PR into OpenTitan and run the register tool (checks the labels) before actually merging. Just to avoid that we miss something obvious.

rtl/ibex_top.sv
GregAC: It is, right at the top just before the module definition

rswarbrick (approved):

PR #1524, opened 2022-02-02
[simple_system, sw] Add PMP example tests

silvestrst:
# Introduction

These tests and support libraries have been ported from the OpenTitan repository.

I have stripped some of the less relevant dependencies, with the exception of:
- Bit manipulation API: makes the `pmp.{h,c}` a bit tidier, but also factoring it out would take some effort
- CSR access helpers: same as above, however, can be factored out more easily (if required)

This PR is split into smaller logical commits, so should be relatively easy to review.


silvestrst:
I should probably add a README file.

PR #1529, opened 2022-02-15
[ci] Bump RISC-V toolchain version to get bitmanip support

rswarbrick:
This version should have support for bitmanip 1.00+0.93, the version
that we target in the RTL.


GregAC:
We'll also need to rebase our spike implementation as that doesn't have up to date bitmanip support. I had been planning on just doing that to the cosim version to save doing the work twice.

GregAC (approved):

rswarbrick:
Ah, good point about the Spike version. I'll do that as a follow-up and then switch the ISA in (yet another) follow-up afterwards. I *think* that should give us maximally bisectable changes without breaking anything on the way :-)

rswarbrick:
@vogelpi: About the nightlies, I've just taken a look and this is controlled by a crontab-like file in the lowrisc-private-ci repo. When it fires, we run the stuff in `ibex-regression-template.yml` (with a "not smoke" flag). As far as I can tell, this should just pull in the bitmanip tests once #1532 is merged.

vogelpi:
> @vogelpi: About the nightlies, I've just taken a look and this is controlled by a crontab-like file in the lowrisc-private-ci repo. When it fires, we run the stuff in `ibex-regression-template.yml` (with a "not smoke" flag). As far as I can tell, this should just pull in the bitmanip tests once #1532 is merged.

Awesome!

PR #1530, opened 2022-02-15
[ci] Bump Spike version to get cosim implementation

rswarbrick:
This will only have an effect on our private CI, which picks up this
spike build from the toolnas. The build is the ibex_cosim branch,
which contains the stuff we need for the recent cosim support. It's
also new enough to support the v1.0+0.93 bitmanip flavour that we
support in the RTL.


vogelpi (approved):
LGTM, let's see if CI likes it.

PR #1534, opened 2022-02-17
[rtl, doc] Seperate major alert into internal and bus

GregAC:
This is to allow more consistent signalling in systems that integrate
Ibex (e.g. OpenTitan) so bus integrity errors external to Ibex and one's
detected within Ibex can be fed into the same alert whilst seperating
out Ibex's various internal alert causes.


vogelpi (approved):
This is as discussed in the D2S meeting. LGTM!

PR #1538, opened 2022-02-17
[rtl] Switch to multi-bit fetch enable

GregAC:
The multi-bit enable aids security hardening. For non secure Ibex all
but the bottom bit is ignored so it is effectively a single bit enable.


rtl/ibex_pkg.sv
```diff
@@ -606,4 +606,15 @@ package ibex_pkg;
       128'h14e8cecae3040d5e12286bb3cc113298;
   parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonceDefault =
       64'hf79780bc735f3843;
+
+  // Fetch enable. Mult-bit signal used for security hardening. For non-secure implementation all
+  // bits other than the bottom bit are ignored.
+  typedef logic [3:0] fetch_enable_t;
```
tjaychen, line 612: is it difficult from ibex perspective to make use of the mubi primitive under `opentitan/hw/ip/prim`?

vogelpi: IIRC, we decided to define a new type with the exact same encoding as the LC signals to not add the LC dependency in the Ibex repo (the signal is used only in one place anyway). Inside the `rv_core_ibex` wrapper an SVA will be added to ensure the encodings are the same.

rtl/ibex_top.sv
```diff
@@ -123,7 +123,7 @@ module ibex_top import ibex_pkg::*; #(
 `endif
 
   // CPU Control Signals
-  input  logic                         fetch_enable_i,
+  input  fetch_enable_t                fetch_enable_i,
```
vogelpi, line 126: I think we should add a set of `prim_buf` cells for the `fetch_enable_i` at this hierarchy level. Otherwise, there are only buffers added to the copy fed to the shadow core but not to the signal fed to the real core. Thus, the synthesis tool will likely optimize away the redundant bits when feeding the signal into the real core. WDYT?

vogelpi (approved):
LGTM with one question/comment regarding the addition of buffers on the copy fed to the real core. Thanks @GregAC !

rtl/ibex_top.sv
GregAC: I did think about it and decided that signals coming in from the lifecycle controller would presumably have some buffering anyway so it didn't need it here.

However you're right, there's no harm in adding extra and systems other than OpenTitan will do things differently so buffering will be required here.

PR #1539, opened 2022-02-18
[ibex, dv] Added new base, interrupt, debug and mem error sequences

prajwalaputtappa:
Added an independent base test with following capabilities:
* chooses between single run, multiples runs or infinite runs (existing sequence
does this via the `num_of_interations` variable).
* interval between runs can be fixed or random, with 0 delay between runs possible.
For random intervals there should be a way to bias them more towards 0 delay
(e.g. specify 75% of delays should 0 with the rest randomly chosen).

Added an interrupt sequence that inherits from the above base sequence.
It has following capabilities:
* chooses the number of interrupts to raise
* specifies the interval between interrupt being raised and dropped
* a mask to specify interrupts that shouldn't be raised.

Added a debug sequence with the only functionality to specify the interval between
the debug request being raised and dropped

Added a sequence to corrupt instruction and data memory.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
@@ -1,0 +1,227 @@
+typedef enum bit [1:0] {
+    SingleRun, // Singl iteration
+    InfiniteRuns, // Run forever until stop is specified
+    MultipleRuns, // Multiple runs with configurable or randomizable iteration count
+    InvalidRuns // Default state
+  } run_type_e;
+
+typedef enum bit [1:0] {
+  IsideErr, // Inject error in instruction side memory.
+  DsideErr, // Inject error in data side memory.
+  PickErr // Pick which memory to inject error in.
+  } error_type_e;
+
+class core_base_new_seq #(type REQ = uvm_sequence_item) extends uvm_sequence #(REQ);
+  // Default set to 50% for zero delay to be picked. Set to 100 if zero delay is required always.
+  int unsigned                     zero_delay_pct = 50;
+  rand bit                         zero_delays;
+  rand int                         delay = 0;
+  int unsigned                     iteration_cnt = 0;
+  int unsigned                     max_delay = 500;
+  virtual clk_rst_if               clk_vif;
+  bit                              stop_seq;
+  bit                              seq_finished;
+  rand run_type_e                  num_of_iterations = InvalidRuns;
+  virtual core_ibex_dut_probe_if   dut_vif;
+  // Use this bit to start any unique sequence once
+  bit                              start_seq = 0;
+
+  `uvm_object_param_utils(core_base_new_seq#(REQ))
+
+  function new (string name = "");
+    super.new(name);
+    if(!uvm_config_db#(virtual clk_rst_if)::get(null, "", "clk_if", clk_vif)) begin
+       `uvm_fatal(get_full_name(), "Cannot get clk_if")
+    end
+    if (!uvm_config_db#(virtual core_ibex_dut_probe_if)::get(null, "", "dut_if", dut_vif)) begin
+      `uvm_fatal(get_full_name(), "Cannot get dut_if")
+    end
+  endfunction
+
+  constraint zero_delays_c {
+    zero_delays dist {1 :/ zero_delay_pct,
+                      0 :/ 100 - zero_delay_pct};
+  }
+  constraint reasonable_delay_c {
+    delay inside {[0 : max_delay]};
+  }
+
+  virtual task body();
+    if(num_of_iterations == InvalidRuns) begin
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(num_of_iterations, num_of_iterations != InvalidRuns;)
+    end
+    `uvm_info(`gfn, $sformatf("Doing %s", num_of_iterations.name()), UVM_LOW)
+    case (num_of_iterations)
+      SingleRun: begin
+        drive_stimulus(delay);
+      end
+      MultipleRuns: begin
+        if(iteration_cnt == 0) begin
+          `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(iteration_cnt,
+                                                iteration_cnt > 0 && iteration_cnt <= 20; )
+        end
+        for (int i = 0; i <= iteration_cnt; i++) begin
+          drive_stimulus(delay);
+        end
+      end
+      InfiniteRuns: begin
+        while (!stop_seq) begin
+          drive_stimulus(delay);
+        end
+        seq_finished = 1'b1;
+      end
+      default: begin
+        `uvm_fatal(`gfn, "Type of run not set")
+      end
+    endcase
+  endtask: body
+
+  task drive_stimulus(int unsigned delay);
+    if(delay == 0) begin
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(zero_delays)
+      if(zero_delays) begin
+        delay = 0;
+      end else begin
+        `DV_CHECK_MEMBER_RANDOMIZE_FATAL(delay)
+      end
+    end
+    clk_vif.wait_clks(delay);
+    `uvm_info(get_full_name(), "Starting sequence...", UVM_MEDIUM)
+    send_req();
+    `uvm_info(get_full_name(), "Exiting sequence", UVM_MEDIUM)
+  endtask: drive_stimulus
+
+  virtual task send_req();
+    `uvm_fatal(get_full_name(), "This task must be implemented in the extended class")
+  endtask
+
+  virtual task stop();
+    stop_seq = 1'b1;
+    `uvm_info(get_full_name(), "Stopping sequence", UVM_MEDIUM)
+    wait (seq_finished == 1'b1);
+  endtask
+
+endclass
+
+class irq_new_seq extends core_base_new_seq #(irq_seq_item);
+
+  `uvm_object_utils(irq_new_seq)
+  `uvm_object_new
+
+  bit no_nmi;
+  bit no_fast;
+  bit no_external;
+  bit no_timer;
+  bit no_software;
+  int max_delay = 500;
+  int min_delay = 50;
+
+  rand int interval = min_delay;
+
+  constraint reasonable_interval_c {
+    interval inside {[min_delay : max_delay]};
+  }
+
+  virtual task send_req();
+    irq_seq_item irq;
+    irq = irq_seq_item::type_id::create("irq");
+    // Raise interrupts
+    start_item(irq);
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(irq, num_of_interrupt > 1;
+                                        if (no_nmi) {
+                                          irq_nm == 0;
+                                        }
+                                        if (no_fast) {
+                                          irq_fast == '0;
+                                        }
+                                        if (no_external) {
+                                          irq_external == 0;
+                                        }
+                                        if (no_timer) {
+                                          irq_timer == 0;
+                                        }
+                                        if (no_software) {
+                                          irq_software == 0;
+                                        })
+    finish_item(irq);
+    get_response(irq);
+    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(interval)
+    clk_vif.wait_clks(interval);
+    // Drop interrupts
+    start_item(irq);
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(irq, num_of_interrupt == 0;)
+    finish_item(irq);
+    get_response(irq);
+  endtask: send_req
+
+endclass: irq_new_seq
+
+// Simple debug sequence
+// debug_req is just a single bit sideband signal, use the interface to drive it directly
+class debug_new_seq extends core_base_new_seq#(irq_seq_item);
+
+  `uvm_object_utils(debug_new_seq)
+  `uvm_object_new
+
+  int max_delay = 500;
+  int min_delay = 75;
+  rand int unsigned drop_delay = 0;
+
+  virtual task body();
+    dut_vif.dut_cb.debug_req <= 1'b0;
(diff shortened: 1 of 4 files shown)


dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+  rand int           delay;
+  int unsigned       iteration_cnt;
+  int unsigned       max_delay = 500;
+  virtual clk_rst_if clk_vif;
+  bit                is_started;
+  bit                stop_seq;
+  bit                seq_finished;
+  rand run_type_e    num_of_iterations;
+
+  `uvm_object_param_utils(core_base_seq#(REQ))
```
GregAC, line 19: Presume this should be `core_base_new_seq`?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+  constraint reasonable_delay_c {
+    delay inside {[0 : max_delay]};
+  }
+
+  virtual task body();
+    if(!uvm_config_db#(virtual clk_rst_if)::get(null, "", "clk_if", clk_vif)) begin
+       `uvm_fatal(get_full_name(), "Cannot get clk_if")
+    end
+    if(!uvm_config_db#(run_type_e)::get(null, "", "num_of_iterations", num_of_iterations)) begin
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(num_of_iterations)
```
GregAC, line 35: Having to specify via the config DB seems a little awkward, but then I can't come up with a clean solution. We could have an extra bit `randomise_num_of_interations` that tells the sequence whether it should randomise it or not though this also feels a bit messy.

@rswarbrick any thoughts?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+  virtual task stop();
+    stop_seq = 1'b1;
+    `uvm_info(get_full_name(), "Stopping sequence", UVM_LOW)
+    wait (seq_finished == 1'b1);
+    is_started = 1'b0;
+  endtask
+
+endclass
+
+class irq_raise_drop_new_seq extends core_base_new_seq #(irq_seq_item);
```
GregAC, line 91: Let's just call this `irq_new_seq`

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+  bit no_nmi;
+  bit no_fast;
+  bit no_external;
+  bit no_timer;
+  bit no_software;
+
+  rand int interval = 50;
+
+  constraint reasonable_interval_c {
+    interval inside {[50 : 500]};
```
GregAC, line 105: Would be useful to make the min and max interval configurable here

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+
+// Simple debug sequence
+// debug_req is just a single bit sideband signal, use the interface to drive it directly
+class debug_new_seq extends core_base_new_seq#(irq_seq_item);
+  virtual core_ibex_dut_probe_if dut_vif;
+
+  `uvm_object_utils(debug_new_seq)
+  `uvm_object_new
+
+  int unsigned drop_delay = 75;
```
GregAC, line 150: Would be useful to allow randomising this also.

GregAC (commented):
Thanks for this @prajwalaputtappa. Mostly there just a couple of tweaks to make.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
prajwalaputtappa: I could change the code to have an extra bit if the two of you think that's the best approach. Let me know.

rswarbrick: I'd be lazy: initialise to zero (an invalid iteration count). Then something that wants to customise the number of iterations can set it to something sensible and the code here can randomise if it's still zero?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
+typedef enum bit [1:0] {
+    SingleRun,
+    InfiniteRuns,
+    MultipleRuns
+  } run_type_e;
+
+class core_base_new_seq #(type REQ = uvm_sequence_item) extends uvm_sequence #(REQ);
+  int unsigned       zero_interval = 50; // Default set to 50% for zero delay to be picked
```
rswarbrick, line 8: Could you rename this to something like `zero_delay_pct`? At the moment, the name doesn't make it clear that this is a percentage chance.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+      InfiniteRuns: begin
+        if (!is_started) is_started = 1'b1;
+        while (!stop_seq) begin
+          drive_stimulus();
+        end
+        seq_finished = 1'b1;
+      end
+      default: begin
+        drive_stimulus();
+      end
```
rswarbrick, line 59: Since this case shouldn't be possible, could you change it to a fatal error?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+      MultipleRuns: begin
+        if(!uvm_config_db#(int)::get(null, "", "iteration_cnt", iteration_cnt)) begin
+          `DV_CHECK_MEMBER_RANDOMIZE_FATAL(iteration_cnt)
+        end
+        for (int i = 0; i <= iteration_cnt; i++) begin
+          drive_stimulus();
+        end
+      end
+      InfiniteRuns: begin
+        if (!is_started) is_started = 1'b1;
```
rswarbrick, line 51: What does this flag do, exactly? I thought it might have been something to prevent re-entrancy, but that doesn't seem to be how you're using it?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+
+  rand int interval = 50;
+
+  constraint reasonable_interval_c {
+    interval inside {[50 : 500]};
+  }
+
+  virtual task send_req();
+    irq_seq_item irq;
+    irq = irq_seq_item::type_id::create($sformatf("irq_seq_item[%0d]", iteration_cnt));
```
rswarbrick, line 110: I think that `iteration_cnt` is being used in the base class for the total number of iterations to do, so I'm not sure this is the right name?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,168 @@
...
+
+  `uvm_object_utils(debug_new_seq)
+  `uvm_object_new
+
+  int unsigned drop_delay = 75;
+
+  virtual task body();
+    if (!uvm_config_db#(virtual core_ibex_dut_probe_if)::get(null, "", "dut_if", dut_vif)) begin
+      `uvm_fatal(get_full_name(), "Cannot get dut_if")
+    end
```
rswarbrick, line 155: Maybe it would be worth doing this in the base class?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
prajwalaputtappa: This flag is not required. I'm removing it from the code.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
prajwalaputtappa: This was nit intentional. Looks like I made a mistake while pasting some code. I've corrected it now.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
prajwalaputtappa: Moved to base sequence.

prajwalaputtappa:
@GregAC and @rswarbrick  I've also added the memory error sequence in the same PR. Please review the last sequence in cire_ibex_new_seq_lib.sv and the last test in core_ibex_test_lib.sv as they're the newest changes. Thanks.

GregAC:
Private CI failure is due to a flaky test, may be worth just disabling it for it or using another test as one of the three for the Private CI smoke. Once we're using co-sim and drop the current checking we can use that test again. @rswarbrick any thoughts?

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,222 @@
...
+  int min_delay = 75;
+  rand int unsigned drop_delay = min_delay;
+
+  constraint reasonable_drop_delay_c {
+    drop_delay inside {[min_delay : max_delay]};
+  }
+
+  virtual task body();
+    dut_vif.dut_cb.debug_req <= 1'b0;
+    if(!uvm_config_db#(int)::get(null, "", "drop_delay", drop_delay)) begin
```
GregAC, line 171: I'd go with the same method we chose for 'num_of_interations' in base sequence. If set to some invalid value randomise, otherwise use the value as given.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,222 @@
...
+    if(delay == 0) begin
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(zero_delays)
+      if(zero_delays) begin
+        delay = 0;
+      end else begin
+        `DV_CHECK_MEMBER_RANDOMIZE_FATAL(delay)
+      end
+    end
+    clk_vif.wait_clks(delay);
+    `uvm_info(get_full_name(), "Starting sequence...", UVM_LOW)
```
GregAC, line 84: Perhaps switch these sequence start/stop/exit messages to UVM_MEDIUM? I don't think they'll be too useful in normal sim output.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,222 @@
...
+    InvalidRuns
+  } run_type_e;
+
+typedef enum bit [1:0] {
+  IsideErr,
+  DsideErr,
+  PickErr
+  } error_type_e;
+
+class core_base_new_seq #(type REQ = uvm_sequence_item) extends uvm_sequence #(REQ);
```
GregAC, line 14: Let's have some documentation on the various setting/features of the sequence. It can live here in a comment for now.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,222 @@
...
+      end
+      default: begin
+        `uvm_fatal(`gfn, "Type of run not set")
+      end
+    endcase
+  endtask: body
+
+  task drive_stimulus(int unsigned delay);
+    if(delay == 0) begin
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(zero_delays)
```
GregAC, line 76: With this setup to always produce zero delays I think we'd need to set `zero_delay_pct` to 100? That should be fine but worth making explicit mention of it in the documentation when that is added.

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -0,0 +1,222 @@
...
+
+  `uvm_object_utils(memory_error_seq)
+  `uvm_declare_p_sequencer(core_ibex_vseqr)
+
+  function new (string name = "");
+    super.new(name);
+    vseq = core_ibex_vseq::type_id::create("vseq");
+  endfunction
+
+  virtual task body();
```
GregAC, line 199: Will this work properly? As it does it overrides `body` we won't get any of the functionality from the base sequence. Plus as it's creating a new `core_ibex_vseq` rather than referring to the existing one will it correctly inject errors? It is possible there's something about UVM virtual sequences I'm not understanding.

prajwalaputtappa: I meant to override send_seq task. I've corrected this now.

rswarbrick:
This is currently not compiling. @prajwalaputtappa, could you take a look?
```
Error-[SE] Syntax error
  Following verilog source has syntax error :
  "/azp/agent/_work/1/s/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv", 11: 
  token is 'PickErr'
    PickErr // Pick which memory to inject error in.
           ^
```

prajwalaputtappa:
> This is currently not compiling. @prajwalaputtappa, could you take a look?
> 
> ```
> Error-[SE] Syntax error
>   Following verilog source has syntax error :
>   "/azp/agent/_work/1/s/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv", 11: 
>   token is 'PickErr'
>     PickErr // Pick which memory to inject error in.
>            ^
> ```

I've resolved this and pushed it again.

GregAC (approved):
Think we're good to go, nice work @prajwalaputtappa 

Still getting a private-ci failure, pretty sure it's a flaky test. I think I'll just remove it from the smoke regression for now.

PR #1540, opened 2022-02-18
Refer to a specific tag for the ibex-cosim version of Spike

rswarbrick:
We're going to want to make a couple more releases of Spike, cleaving
a bit closer to the upstream repository. Let's be explicit about which
version people should get.


vogelpi (approved):
LGTM, thanks @rswarbrick !

GregAC (approved):

PR #1541, opened 2022-02-21
[rtl] Add prim_buf for security critical signals

GregAC:
A sufficiently agressive optimiser may optimise these away as under
normal functioning they effectively don't do anything. They are purely
to detect the presence of induced faults.


vogelpi (approved):
LGTM, as discussed in the D2S review. Thanks @GregAC !

PR #1543, opened 2022-02-23
[ibex, dv] Added agent configuration for ibex_mem_intf_response_agent

prajwalaputtappa:
Defining agent configuration for any agent is a standard UVM flow and is
a cleaner flow for defining delay between driving sequence items,
passing virtual interface etc.

Agent configuration has been added to the existing agent to make delay
configuration more flexible in the future.

Signed-off-by: Prajwala Puttappa <prajwalaputtappa@lowrisc.org>

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
@@ -10,7 +10,8 @@ class ibex_mem_intf_response_agent extends uvm_agent;
 
   ibex_mem_intf_response_driver     driver;
   ibex_mem_intf_response_sequencer  sequencer;
-  ibex_mem_intf_monitor          monitor;
+  ibex_mem_intf_monitor             monitor;
+  ibex_mem_intf_response_agent_cfg  cfg;
 
   `uvm_component_utils(ibex_mem_intf_response_agent)
   `uvm_component_new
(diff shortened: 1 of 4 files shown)


dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
```diff
@@ -30,6 +34,7 @@ class ibex_mem_intf_response_agent extends uvm_agent;
       driver.seq_item_port.connect(sequencer.seq_item_export);
       monitor.addr_ph_port.connect(sequencer.addr_ph_port.analysis_export);
     end
+    driver.vif = cfg.vif;
```
rswarbrick, line 37: In the OpenTitan setup, drivers and monitors have references to the cfg object. I don't have strong feelings, but maybe that would be cleaner? (In particular, it makes it easier to put simple config knobs for the driver into the config object)

rswarbrick (approved):
This looks good (assuming that CI passes). Happy to merge this, but maybe worth thinking about passing a reference to `cfg` instead of `vif` to the driver?

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
prajwalaputtappa: I've added cfg knobs in the next PR after realising that is the better way to do it.

PR #1544, opened 2022-02-24
[dv] Add X assertions for top-level IO

GregAC:
Ensure all top-level inputs and outputs are known when they are
expected to be known.

Seems to work fine with some quick tests. I'll do a full regression once this is merged in.

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -935,8 +935,41 @@ module ibex_top import ibex_pkg::*; #(
   assign alert_major_bus_o      = lockstep_alert_major_bus;
   assign alert_minor_o          = core_alert_minor | lockstep_alert_minor;
 
+  // X checks for top-level outputs
+  `ASSERT_KNOWN(IbexInstrReqX, instr_req_o)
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):

GregAC:
Did have an x-check for the crash dump but that was causing failures in non OT configs as it outputs a PC from a non-reset flop (the flop itself is fine being X as it's paired with a pipeline valid signal). Just ditched that x-check for now. Maybe we should add one into rv_core_ibex in OpenTitan instead (where it should pass as all flops are reset)?

rswarbrick:
For the crash dump, would it make sense to copy the flop across AND'ing with the valid signal? Seems like a very small design change for a more uniform interface.

GregAC:
> For the crash dump, would it make sense to copy the flop across, AND'ing with the valid signal? Seems like a very small design change for a more uniform interface.

Guess it can't hurt. Indeed we might want to separately capture it anyway so it's definitely the last PC that we attempted to execute rather than just whatever PC happens to be in the IF -> ID stage registers at the time the crash dump is sampled.

PR #1545, opened 2022-02-24
[ibex, dv] Makes delays between req, gnt and rvalid configurable

prajwalaputtappa:
This commit adds functionalty to the memory response agent to make delays more
    configurable.
    There are two delays
    - Delay between req and gnt
    - Delay between gnt and rvalid
    
    For each of these delays we have three modes:
    * Fully random delay
    * Fixed delay
    * Biased delay. Randomised delays but allow biasing towards 0 delay, to give a mix of runs with back
    to back transfers with no delay and some with delays.

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
@@ -34,7 +34,8 @@ class ibex_mem_intf_response_agent extends uvm_agent;
       driver.seq_item_port.connect(sequencer.seq_item_export);
       monitor.addr_ph_port.connect(sequencer.addr_ph_port.analysis_export);
     end
-    driver.vif = cfg.vif;
+    driver.cfg = cfg;
+    sequencer.cfg = cfg;
   endfunction : connect_phase
 
   function void reset();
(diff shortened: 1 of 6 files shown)


dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
```diff
@@ -35,6 +35,8 @@ class ibex_mem_intf_response_agent extends uvm_agent;
       monitor.addr_ph_port.connect(sequencer.addr_ph_port.analysis_export);
     end
     driver.vif = cfg.vif;
+    driver.cfg = cfg;
```
rswarbrick, line 38: I definitely don't think we should do this: probably just pass the `cfg` object in the first commit (which is in another PR, I know)

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
```diff
@@ -8,15 +8,13 @@
 
 class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
 
-  int                     max_rvalid_delay = 20;
-  int                     min_rvalid_delay = 0;
-  ibex_mem_intf_seq_item  item;
-  mem_model               m_mem;
-  bit                     enable_error = 1'b0;
+  ibex_mem_intf_seq_item           item;
```
rswarbrick, line 11: Nit: We seem to have gained a lot of whitespace here. Maybe drop it again?

prajwalaputtappa: I've added spaces to align them to one line. Hence the increase in white space.

rswarbrick: What I mean is that instead of
```
  foo_t           a;
  bar_baz_t       b;
```
we could use:
```
  foo_t     a;
  bar_baz_t b;
```

rswarbrick (approved):

PR #1552, opened 2022-03-08
[icache, dv] Made changes required to make TB compatible with Xcelium

prajwalaputtappa:
Fixed some trivial issues regarding hierarchical accesses and removed setting of some parameters from ibex_icache_sim_cfg.hjson file.


rswarbrick (approved):
We'll need to figure out how to plumb this into dvsim properly at some point, but let's just get this in for now.

PR #1558, opened 2022-03-14
[icache, dv] Added scrambling agent to verify scrambling in RAMs

prajwalaputtappa:
This commit adds a new scrambling agent to drive scrambling key and valid to the data and tag memory interfaces.

### DIFF
--- a/dv/uvm/icache/dv/env/ibex_icache_env.sv
+++ b/dv/uvm/icache/dv/env/ibex_icache_env.sv
@@ -12,6 +12,8 @@ class ibex_icache_env extends dv_base_env #(
 
   ibex_icache_core_agent core_agent;
   ibex_icache_mem_agent  mem_agent;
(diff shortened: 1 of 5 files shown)


rswarbrick:
This looks fine, except for the vendoring part, where I think you copied the files across manually? The right thing to do is to update `lowrisc_ip.vendor.hjson` and then re-run the vendoring tool to grab the files.

prajwalaputtappa:
> This looks fine, except for the vendoring part, where I think you copied the files across manually? The right thing to do is to update `lowrisc_ip.vendor.hjson` and then re-run the vendoring tool to grab the files.

I've generated the push pull agent using the script.

rswarbrick:
Ah, could you re-do this so that you run the vendoring script first? At the moment, I think the first commit has the manually copied files, which get overwritten by the vendoring commit: it should probably go the other way around.

rswarbrick (approved):
Looks great, thanks!

PR #1561, opened 2022-03-15
[icache, dv] Removed support for single clock cycle PMP error response

prajwalaputtappa:
Earlier the design supported single clock cycle error responses from PMP
block whenever a read was done from blocked memory. Now there is at
least one clock cycle delay after the request has been granted for the
error to be asserted. Therefore, this commit removes the support for
single clock cycle PMP error response.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
+++ b/dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
@@ -54,11 +54,7 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
       // will still be passed from the monitor to any later sequence.
       p_sequencer.request_fifo.get_peek_export.peek(req_item);
 
-      if (!req_item.is_grant) begin
-        take_req(resp_item, req_item);
-      end else begin
-        take_gnt(resp_item, req_item);
-      end
+      take_gnt(resp_item, req_item);
 
       // Get and drop the request item now that we've dealt with it. To check that nothing has gone
       // wrong, make sure that the two items match.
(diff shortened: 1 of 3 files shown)


dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
```diff
@@ -54,9 +54,7 @@ class ibex_icache_mem_resp_seq extends ibex_icache_mem_base_seq;
       // will still be passed from the monitor to any later sequence.
       p_sequencer.request_fifo.get_peek_export.peek(req_item);
 
-      if (!req_item.is_grant) begin
-        take_req(resp_item, req_item);
-      end else begin
+      if (req_item.is_grant) begin
```
rswarbrick, line 57: Do we actually need the `is_grant` field any more? I think we can probably remove it, right?

rswarbrick (commented):
Assuming the tests pass, this looks good to me.

I think we can probably rip out the `is_grant` field from the item class, though. Would you mind adding that to this PR? (Either as part of the same commit or a follow-up: either is fine with me).

dv/uvm/icache/dv/ibex_icache_mem_agent/seq_lib/ibex_icache_mem_resp_seq.sv
prajwalaputtappa: Removed is_grant from ibex_icache_mem_req_item.sv and any references to it from ibex_icache_mem_monitor.sv and ibex_icache_mem_resp_seq.sv

rswarbrick (commented):
Sorry I hadn't noticed, but I think we also want to get rid of `ibex_icache_mem_monitor.new_request` (it's no longer being called because you deleted `collect_requests`, but we should probably expunge it entirely!)

Otherwise, this looks great!

prajwalaputtappa:
> new_request

Removed the new_request function and pushed it again. Please review.

rswarbrick:
Looks great, thanks!

PR #1562, opened 2022-03-16
[icache, dv] Fixed regression failure in ibex_icache_back_line

prajwalaputtappa:
There was issue with rtespect to calculating number of instructions per
word and this commit fixes that issue.

Number of instructions per word = 1/4*1 + 3/4(1/4*3/2 + 3/4*2) = 53/32.
Earlier th5s was calculated as 7/4.
Ideal window length needed to calculate fetch ratio percentage is
calculated as 53/32*C*2 = 848. Earlier it was calculated to be 300.


rswarbrick (approved):
Brill, thanks!

PR #1576, opened 2022-03-17
Update ISA strings from Xbitmanip to XZb*

rswarbrick:
This allows us to model stuff more closely. This depends on Spike
version ibex-cosim-v0.2 (which is rebased onto a master branch commit
supporting these more specific ISA strings).


vogelpi (approved):
This looks good to me, but CI currently fails and should be fixed before merging this.

rswarbrick:
@hcallahan-lowrisc: Some of what I said this morning might make a bit more sense when you've seen this PR: I'd completely forgotten that it wasn't merged!

PR #1581, opened 2022-03-21
[rtl] Small refactors to improve code coverage

GregAC:


rtl/ibex_if_stage.sv
```diff
@@ -178,12 +178,14 @@ module ibex_if_stage import ibex_pkg::*; #(
...
   always_comb begin : exc_pc_mux
+    exc_pc = {csr_mtvec_i[31:8], 8'h00};
+
     unique case (exc_pc_mux_i)
       EXC_PC_EXC:     exc_pc = { csr_mtvec_i[31:8], 8'h00                    };
       EXC_PC_IRQ:     exc_pc = { csr_mtvec_i[31:8], 1'b0, irq_id[4:0], 2'b00 };
       EXC_PC_DBD:     exc_pc = DmHaltAddr;
       EXC_PC_DBG_EXC: exc_pc = DmExceptionAddr;
-      default:        exc_pc = { csr_mtvec_i[31:8], 8'h00                    };
+      default: ;
```
rswarbrick, line 188: I don't think I understand this: why does this affect the code coverage? (It feels like cheating, so I'm not a huge fan, but maybe I've misunderstood?)

rtl/ibex_load_store_unit.sv
```diff
@@ -115,6 +115,8 @@ module ibex_load_store_unit #(
   ///////////////////
 
   always_comb begin
+    data_be = 4'b0000;
```
rswarbrick, line 118: The defaults below are `4'b1111`: do you really mean zero here?

rtl/ibex_if_stage.sv
vogelpi: This is indeed strange and unexpected.

rtl/ibex_load_store_unit.sv
vogelpi: `4'b0000` is marked as `// this is not used, but included for completeness` in the code. I think we should instead use a default value that can actually occur such as `4'b0001`. It may not make a big difference here but imagine you had say a 128-bit wide mux.

If the all-zero default byte strobe is motivated by security, a comment would be useful to explain that.

rtl/ibex_if_stage.sv
GregAC: The default cases are never being hit as the case covers all values that can occur. So I moved the default to become an initial write before the case statement in these changes. However I guess the empty `default:` can still appear as a coverage hole in code coverage so this may not fix much.

We have to keep the empty `default:` as per the style guide (https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#case-statements) so we'd need to waiver. Still I think it's better for us to have empty `default:` as it's maybe more obvious that it's ok to waive.

rtl/ibex_load_store_unit.sv
GregAC: I chose the all zero as 'do nothing' felt like the safest fallback. So in a sense security related, I'll add a comment to explain.

rtl/ibex_if_stage.sv
rswarbrick: Another way to do this would be to make e.g. the `EXC_PC_DBG_EXC` case the default:
```systemverilog
    unique case (exc_pc_mux_i)
      EXC_PC_EXC:     exc_pc = { csr_mtvec_i[31:8], 8'h00                    };
      EXC_PC_IRQ:     exc_pc = { csr_mtvec_i[31:8], 1'b0, irq_id[4:0], 2'b00 };
      EXC_PC_DBD:     exc_pc = DmHaltAddr;
      // EXC_PC_DBG_EXC
      default:        exc_pc = DmExceptionAddr;
    endcase
```

GregAC:
I'm going to abandon this, the empty default doesn't effect the coverage and how we want to deal with coverage around this unused defaults looks to be an open question still.

rtl/ibex_if_stage.sv
rswarbrick: > The issue with the 5 possible values is there is no test that will give you an invalid value (i.e. one of the 3 others), unless it's a fault injection style test where we force a signal going into the IF.

Yes, exactly. But I think that's exactly what we should be doing. Or, if we don't care what value things take in the fall-through case, we can use the `default` trick above.

> I'm not really a fan of the default: being used where we're intending it to be for a specific case label. Even though you have the comment I think it's less readable and more error prone when modifying things.

Hmm, I guess maybe. But I quite like the way you get (in the example above) a very clear 4-way mux. With the extra default case, it looks like it might have 5 cases and you have to go and look at the enum encoding to be sure. Yuck!

Issue #1571, opened 2022-03-17
Get core_ibex UVM tests working with Xcelium


rswarbrick:

ctopal:
I believe this one relates with some common problem that Rupert solved in an [ OTBN commit ](https://github.com/lowRISC/opentitan/commit/4b7fe88d6b46854dba3b391291dd0518e15daefc) regarding the DPI calls. I am leaving the slightly cryptic error message if it might make things easier.

@ sim_riscv_arithmetic_basic_test_0.log
```
xmsim: *E,IMPDLL: Unable to load the implicit shared object.
OSDLERROR: (null)/riscv_instr_gen_tb_top/sv/_sv_export.so: cannot open shared object file: No such file or directory or file is not valid ELFCLASS64 library..
```

@ compile.log
```
xmelab: *W,DPIEXP: DPI export function in  _sv_export.so not available.
```

PR #1588, opened 2022-03-28
Remove logfile param in Ibex RTL Sim for Xcelium

hcallahan-lowrisc:
Fixes a bug where both Xcelium and Python open the same sim.log file and race to
write the simulation results into it. This change makes Python the sole writer of this
file using the captured stdout/stderr from the subprocess.run call in
run_rtl.py.

This bug was also previously present for VCS but was fixed in 90ff7ca6c.
Goes towards #1571


rswarbrick (approved):
Looks great! Thanks!

PR #1604, opened 2022-04-11
Core ibex dv parallelism improvements

hcallahan-lowrisc:
Improves the core-ibex dv Makefile by adding parallelism to the running of the riscv-dv instruction generator, and also the running of the test binaries using the iss.
This saves approximately 45 mins on a private-ci job with SIMULATOR=xlm ITERATIONS=20 SEED=1 TEST=all COV=1. This particularly helps to better scale to the number of Xcelium licenses available. 
The .stamp files are also renamed through the file to be more consistent.

This PR also makes the whitespace in this makefile more consistent.


hcallahan-lowrisc:
Updated the whitespace commit to consistent as discussed offline. See commit comment for details.

rswarbrick (approved):
Looks great! I think the variable dependencies list for the instruction generator build might need a tweak, but let's get this in now.

PR #1607, opened 2022-04-12
Update ibex_controller.sv

mbaykenar:
Vivado gives synthesis error complaining about assignment of irq_nm_int_cause to '0 by saying it is an enum type.

Change the assignment to NMI_INT_CAUSE_ECC, which is defined to 5'b0 in ibex_pkg.sv

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -357,7 +357,7 @@ module ibex_controller #(
 
     // No integrity checking on incoming load data so no internal interrupts
     assign irq_nm_int       = 1'b0;
-    assign irq_nm_int_cause = '0;
+    assign irq_nm_int_cause = NMI_INT_CAUSE_ECC;
     assign irq_nm_int_mtval = '0;
   end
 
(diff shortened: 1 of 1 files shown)


rtl/ibex_controller.sv
```diff
@@ -357,7 +357,7 @@ module ibex_controller #(
 
     // No integrity checking on incoming load data so no internal interrupts
     assign irq_nm_int       = 1'b0;
-    assign irq_nm_int_cause = '0;
+    assign irq_nm_int_cause = NMI_INT_CAUSE_ECC;
```
rswarbrick, line 360: Please could you do this with a cast instead? So it's `assign irq_nm_int_cause = nmi_int_cause_e'(0)`.

The point is that there's nothing special about ECC here, so it's probably better to just assign to "whatever happens to be zero".

mbaykenar: Yes thanks you are right. I just saw NMI_INT_CAUSE_ECC is also zeros and that was the reason.
I just committed

rswarbrick (approved):
Looks good to me. Thanks!

PR #1609, opened 2022-04-13
[dv] Tweak ibex_dv.f so we can use it in OpenTitan

rswarbrick:
It turns out that there aren't many changes needed, but we need to be
able to point at a prim directory that might not be inside
`vendor/lowrisc_ip`: put that path in an environment variable and pass
it from `core_ibex/Makefile`. That way, a caller on the OpenTitan side
can just override the corresponding variable and everything works.


hcallahan-lowrisc (approved):
Makes sense!

PR #1610, opened 2022-04-13
Add better dependencies for the coverage gen and merge targets in core-ibex dv

hcallahan-lowrisc:
This PR also reworks some comments for clarity around creating dependencies on Makefile variables.


rswarbrick (approved):
Looks great, thanks!

PR #1611, opened 2022-04-13
[dv] Wrap commands in core_ibex/Makefile with scripts

rswarbrick:
This (sadly) adds quite a lot of lines but the point is that all the magic to figure out what we're doing now lies in Python scripts: the Makefile will eventually just be used for scheduling. No functional change intended.

This is a preparatory step towards splitting stuff up differently, to get something in a shape that OpenTitan's "dvsim" tool will understand.


GregAC (approved):

rswarbrick:
Hmm. This seems to have broken the jobserver plumbing. I'll try to fix that up before merging.

PR #1612, opened 2022-04-13
Fixup .rtl.tb_compile vars.mk filename

hcallahan-lowrisc:
This was forcing rebuilds continuously


rswarbrick (approved):
Ah, excellent! This was on my list to debug :-)

PR #1613, opened 2022-04-13
Xcelium cosim support

hcallahan-lowrisc:
This builds on the earlier work in #1575 to get closer to running a cosimulation workflow for core-ibex dv by default.

This PR enables both VCS and Xcelium to run with the cosimualation workflow using COSIM=1 and SIMULATOR=xlm/vcs.
However, the ISS is still setup to run standalone and the output test success critera is based upon the existing comparison mechanisms. Switching over to the cosim workflow by default is left for a later task.

This depends upon using a recent version of spike where pkg-config is supported.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -6,16 +6,6 @@ GEN_DIR             := $(realpath ../../../vendor/google_riscv-dv)
 TOOLCHAIN           := ${RISCV_TOOLCHAIN}
 export IBEX_ROOT    := $(realpath ../../../)
 
-ifeq ($(COSIM),1)
-ifndef IBEX_COSIM_ISS_ROOT
-$(error IBEX_COSIM_ISS_ROOT must be set to the root of a suitable spike build if COSIM=1)
-else
-# Spike builds a libsoftfloat.so shared library that the simulator binary needs.
-# Set LD_LIBRARY_PATH so it can be found.
-export LD_LIBRARY_PATH := $(IBEX_COSIM_ISS_ROOT)/lib/:${LD_LIBRARY_PATH}
-endif
-endif
-
 # Explicitly ask for the bash shell
 SHELL                := bash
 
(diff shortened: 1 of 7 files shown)


GregAC (approved):

dv/uvm/core_ibex/Makefile
```diff
@@ -480,9 +492,9 @@ $(rtl-sim-logs): \
 	@echo Running RTL simulation at $@
 	$(verb)mkdir -p $(@D)
 	$(verb)env \
-	  SPIKE_CFLAGS="$(SPIKE_CFLAGS)" \
-	  SPIKE_LDFLAGS="$(SPIKE_LDFLAGS)" \
-	  SPIKE_LIBS="$(SPIKE_LIBS)" \
+	  SPIKE_CFLAGS="$(SPIKE_CFLAGS_I) $(SPIKE_CFLAGS_OTHER)" \
+	  SPIKE_LIBS="$(SPIKE_LIBS_lowerL) $(SPIKE_LIBS_upperL)" \
+	  SPIKE_LDFLAGS="$(SPIKE_LIBS_FMT)" \
```
rswarbrick, line 497: I realise this isn't new to this commit, but I don't *think* we should need to pass these flags when running (as opposed to compiling) the testbench? Similarly for `gen_cov` below.

dv/uvm/core_ibex/Makefile
```diff
@@ -414,30 +414,42 @@ wave-arg := $(if $(call equal,$(WAVES),1),--en_wave,)
...
+# to format them appropriately for each simulator.
+# - VCS expects most things to be supplied with -LDFLAGS arguments, but
+# unfortunately these come before the object files that are being linked. To add
+# libraries themselves, you're supposed to pass them directly (-lfoo). Split
+# things out here to make sense of them.
+# - Xcelium expects CFLAGS, includes, and libs to be passed directly to xrun.
+# However, linker flags are to be passed using -Wld. This can be finnicky, so
+# using the alternate -Xlinker style simplifies things.
+SPIKE_PCS          := riscv-riscv riscv-disasm riscv-fdt
+ifeq ($(COSIM),1)
```
rswarbrick, line 429: Eek! This is getting a bit crazy! If I'm right and this is only needed when compiling the testbench, I wonder whether it would be better to move this into the `sim.py` wrapper? It might be a bit less horrible than expressing everything in Makefile syntax.

dv/uvm/core_ibex/tests/core_ibex_base_test.sv
```diff
@@ -104,8 +104,8 @@ class core_ibex_base_test extends uvm_test;
     cosim_cfg = core_ibex_cosim_cfg::type_id::create("cosim_cfg", this);
 
     cosim_cfg.isa_string = get_isa_string();
-    cosim_cfg.start_pc = {{32'h`BOOT_ADDR}[31:8], 8'h80};
-    cosim_cfg.start_mtvec = {{32'h`BOOT_ADDR}[31:8], 8'h1};
+    cosim_cfg.start_pc =    ((32'h`BOOT_ADDR & ~(32'h0000_00FF)) | 8'h80);
+    cosim_cfg.start_mtvec = ((32'h`BOOT_ADDR & ~(32'h0000_00FF)) | 8'h1);
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Agreed.

dv/uvm/core_ibex/Makefile
hcallahan-lowrisc: Good point! They're only needed for compiling/building the tb. Removed in other steps.

dv/uvm/core_ibex/Makefile
hcallahan-lowrisc: I agree!! The Makefile level is very tricky for these sorts of ops.
Can I suggest merging as-is then migrating this logic to python-land in a follow-up PR?

hcallahan-lowrisc:
Rebased

PR #1614, opened 2022-04-13
Fix spurious rebuilds from dumped vars

rswarbrick:
Commit f49f452f2 greatly tidied up the file, which is nice, but also
introduced some whitespace around the first argument to `vars-prereq`.
In Make, this causes chaos!

In particular, we ended up checking whether a variable called
something like "`$(last-   gen  -vars-loaded)`" was defined. It isn't,
but `$(last-gen-vars-loaded)` is.

Call strip in the `vars-prereq` function to get rid of the whitespace at
the "entry point" for all this machinery.

### DIFF
--- a/dv/uvm/core_ibex/Makefile
+++ b/dv/uvm/core_ibex/Makefile
@@ -226,7 +226,7 @@ FORCE:
 # might print a message to the console explaining why we're rebuilding TGT).
 #
 # Call it as $(call vars-prereq,X,TGT,VS)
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Good catch! Your make-foo is sharp!

rswarbrick:
Ha! Not that sharp: you should have seen how many `$(info ...)` lines I had to add before I figured out what was going on...

PR #1618, opened 2022-04-14
Finish up the wrapper script migration work I was doing in the core_ibex testsuite

rswarbrick:


hcallahan-lowrisc (approved):
Yep all makes sense. Seems clean. -37 lines is always nice too.

rswarbrick:
Ah drat. These two commits:
image

weren't supposed to survive (they break the way we plumb configs through). I'd dropped them from #1623, but hadn't realised they were still here. I'm going to revert the two and prepend the reversion commits to that PR.

PR #1623, opened 2022-04-20
Split rules in the dv/uvm/core_ibex Makefile

rswarbrick:
This is a rather long series of patches, but the point is that at the end we've got to a state where all the scripts that run stuff run once per test+iteration, rather than iterating over the tests internally.

I plan to follow up with a few more cleanups, but this is the bare minimum needed for OpenTitan integration.


rswarbrick:
Hi @GregAC and @hcallahan-lowrisc! Any chance one of you could take a look at this PR?

rswarbrick:
Oops! I hadn't realised that this depends on #1618. Please look at that first :-)

GregAC (approved):

PR #1631, opened 2022-04-22
Rationalize paths in core_ibex Makefile and scripts

rswarbrick:
The first commit (*"Delete some Makefile variables that we're not using any more"*) is a minor cleanup. The second commit is more interesting and rejigs the paths that we use to look a bit more like the ones in OpenTitan's dvsim (and, honestly, to just look a bit less silly!)

Now we have the following structure (for an example where we've just
run seed 123 of `riscv_arithmetic_basic_test`):

    out/
    +- build/
    |  +- tb/
    |  |    (built testbench)
    |  +- instr-gen/
    |  |    (built instruction generator)
    +- run/
       +- riscv_arithmetic_basic_test.123/
       |  +- test.S
       |  +- test.o
       |  +- test.bin
       |  +- rtl.log
       |  +- trace_core_00000000.log
       |  +- spike.log
       |  ... (some more log files etc. here)
       +- fcov/
       |    (architectural fcov from riscv-dv)
       +- coverage/
       |    (merged coverage data)
       +- regr.log
       +- regr_junit.xml
       +- regr_junit_merged.xml


hcallahan-lowrisc (approved):
LGTM. Just needs a few conflicts resolved now @rswarbrick

PR #1636, opened 2022-04-26
[dv] Various functional coverage point additions

ctopal:
PMP related coverage points are not included in this PR.

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_csr_categories.svh
+++ b/dv/uvm/core_ibex/fcov/core_ibex_csr_categories.svh
@@ -99,12 +99,7 @@
   CSR_MHPMCOUNTER28H, \
   CSR_MHPMCOUNTER29H, \
   CSR_MHPMCOUNTER30H, \
-  CSR_MHPMCOUNTER31H, \
-  // Must exist when implementing hardware triggers (breakpoints), but read as 0 and ignore \
-  // writes. Unused/Uneeded by debugger infrastructure. \
-  // TODO: Don't ignore these? \
-  CSR_MCONTEXT, \
-  CSR_SCONTEXT
+  CSR_MHPMCOUNTER31H
 
 // Debug related CSRs
 `define DEBUG_CSRS \
@@ -117,3 +112,9 @@
   CSR_TDATA2, \
   CSR_TDATA3
 
+// Must exist when implementing hardware triggers (breakpoints), but read as 0 and ignore \
+// writes. Unused/unneeded by debugger infrastructure. \
+`define NOT_IMPLEMENTED_CSRS \
+  CSR_MCONTEXT, \
+  CSR_SCONTEXT, \
+  CSR_TDATA3
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -460,12 +460,20 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     cp_controller_fsm: coverpoint id_stage_i.controller_i.ctrl_fsm_cs {
       bins out_of_reset = (RESET => BOOT_SET);
       bins out_of_boot_set = (BOOT_SET => FIRST_FETCH);
-      bins out_of_first_fetch[] = (FIRST_FETCH => DECODE, IRQ_TAKEN, DBG_TAKEN_IF);
-      bins out_of_decode[] = (DECODE => FLUSH, DBG_TAKEN_IF, IRQ_TAKEN);
+      bins out_of_first_fetch0 = (FIRST_FETCH => DECODE);
+      bins out_of_first_fetch1 = (FIRST_FETCH => IRQ_TAKEN);
+      bins out_of_first_fetch2 = (FIRST_FETCH => DBG_TAKEN_IF);
+      bins out_of_decode0 = (DECODE => FLUSH);
+      bins out_of_decode1 = (DECODE => DBG_TAKEN_IF);
+      bins out_of_decode2 = (DECODE => IRQ_TAKEN);
       bins out_of_irq_taken = (IRQ_TAKEN => DECODE);
       bins out_of_debug_taken_if = (DBG_TAKEN_IF => DECODE);
       bins out_of_debug_taken_id = (DBG_TAKEN_ID => DECODE);
-      bins out_of_flush[] = (FLUSH => DECODE, DBG_TAKEN_ID, WAIT_SLEEP, IRQ_TAKEN, DBG_TAKEN_IF);
+      bins out_of_flush0 = (FLUSH => DECODE);
+      bins out_of_flush1 = (FLUSH => DBG_TAKEN_ID);
+      bins out_of_flush2 = (FLUSH => WAIT_SLEEP);
+      bins out_of_flush3 = (FLUSH => IRQ_TAKEN);
+      bins out_of_flush4 = (FLUSH => DBG_TAKEN_IF);
       bins out_of_wait_sleep = (WAIT_SLEEP => SLEEP);
       bins out_of_sleep = (SLEEP => FIRST_FETCH);
       // TODO: VCS does not implement default sequence so illegal_bins will be empty
@@ -483,6 +491,35 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     `DV_FCOV_EXPR_SEEN(irq_continue_sleep, id_stage_i.controller_i.ctrl_fsm_cs == SLEEP &&
                                            id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
                                            (|cs_registers_i.mip))
+
+    controller_instr_cross: cross cp_controller_fsm, cp_id_instr_category {
+    // Only expecting DECODE => FLUSH when we have the instruction categories constrained below.
+      bins decode_to_flush =
+        binsof(cp_controller_fsm.out_of_decode0) &&
+        binsof(cp_id_instr_category) intersect {InstrCategoryMRet, InstrCategoryDRet,
+                                                InstrCategoryEBreakDbg, InstrCategoryEBreakExc,
+                                                InstrCategoryECall, InstrCategoryFetchError,
+                                                InstrCategoryCSRAccess,
+                                                InstrCategoryCSRIllegal,
+                                                InstrCategoryUncompressedIllegal,
+                                                InstrCategoryCompressedIllegal,
+                                                InstrCategoryPrivIllegal,
+                                                InstrCategoryOtherIllegal};
+      bins decode_to_dbg = binsof(cp_controller_fsm.out_of_decode1);
+      bins decode_to_irq = binsof(cp_controller_fsm.out_of_decode2);
+    // Only expecting FLUSH => DECODE when we have the instruction categories constrained below.
+      bins flush_to_decode =
+        binsof(cp_controller_fsm.out_of_flush0) &&
+        !binsof(cp_id_instr_category) intersect {InstrCategoryEBreakDbg, InstrCategoryWFI};
+    // Only expecting FLUSH => IRQ_TAKEN when we have InstrCategoryCSRAccess in ID/EX
+      bins flush_to_irq_taken =
+        binsof(cp_controller_fsm.out_of_flush3) &&
+        binsof(cp_id_instr_category) intersect {InstrCategoryCSRAccess};
+    // Only expecting FLUSH => DEBUG_IF when we have InstrCategoryEBreakDbg in ID/EX
+      bins flush_to_dbg_if =
+        binsof(cp_controller_fsm.out_of_flush4) &&
+        !binsof(cp_id_instr_category) intersect {InstrCategoryEBreakDbg};
+    }
 
     // Include both mstatus.mie enabled/disabled because it should not affect wakeup condition
     irq_wfi_cross: cross cp_controller_fsm_sleep, cs_registers_i.mstatus_q.mie iff
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -349,6 +349,45 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     `DV_FCOV_EXPR_SEEN(mret_in_umode, id_stage_i.mret_insn_dec && priv_mode_id == PRIV_LVL_U)
     `DV_FCOV_EXPR_SEEN(wfi_in_umode, id_stage_i.wfi_insn_dec && priv_mode_id == PRIV_LVL_U)
 
+    // Unsupported writes to WARL type CSRs
+    `DV_FCOV_EXPR_SEEN(warl_check_mstatus,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_mstatus_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_mie,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_mie_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_mtvec,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_mtvec_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_mepc,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_mepc_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_mtval,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_mtval_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_dcsr,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_dcsr_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(warl_check_cpuctrl,
+                       fcov_csr_write &&
+                       (cs_registers_i.u_cpuctrl_csr.wr_data_i !=
+                       cs_registers_i.csr_wdata_int))
+
+    `DV_FCOV_EXPR_SEEN(double_fault, cs_registers_i.cpuctrl_d.double_fault_seen)
+    `DV_FCOV_EXPR_SEEN(icache_enable, cs_registers_i.cpuctrl_d.icache_enable)
+
     cp_irq_pending: coverpoint id_stage_i.irq_pending_i | id_stage_i.irq_nm_i;
     cp_debug_req: coverpoint id_stage_i.controller_i.fcov_debug_req;
 
@@ -358,6 +397,15 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
 
     cp_csr_write: coverpoint cs_registers_i.csr_addr_i iff (fcov_csr_write) {
       ignore_bins ignore = {`IGNORED_CSRS};
+    }
+
+    // All CSR operations perform a read so we don't need to specify the CSR operation.
+    cp_ignored_csrs: coverpoint cs_registers_i.csr_addr_i iff (id_stage_i.csr_access_o) {
+      bins unimplemented_csrs_read = {`NOT_IMPLEMENTED_CSRS};
+    }
+
+    cp_ignored_csrs_w: coverpoint cs_registers_i.csr_addr_i iff (fcov_csr_write) {
+      bins unimplemented_csrs_written = {`NOT_IMPLEMENTED_CSRS};
     }
 
     `DV_FCOV_EXPR_SEEN(csr_invalid_read_only, fcov_csr_read_only && cs_registers_i.illegal_csr)
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -413,6 +413,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
 
     cp_debug_mode: coverpoint debug_mode;
 
+    `DV_FCOV_EXPR_SEEN(debug_wakeup, id_stage_i.controller_i.fcov_debug_wakeup)
     `DV_FCOV_EXPR_SEEN(interrupt_taken, id_stage_i.controller_i.fcov_interrupt_taken)
     `DV_FCOV_EXPR_SEEN(debug_entry_if, id_stage_i.controller_i.fcov_debug_entry_if)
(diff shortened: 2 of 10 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -385,6 +391,11 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
       illegal_bins illegal_transitions = default sequence;
     }
 
+    mprv_load_store_cross: cross cp_mprv, cp_id_instr_category {
```
rswarbrick, line 456: I know that this is all @GregAC wrote in the plan, but... I think this only has any effect if we're in U-mode? So maybe we need to change `cp_mprv` to be "privilege bit set when we're in U-mode"? And I think the MPP field also has an effect on what it means? (This is from a cursory glance through the privileged spec: it's possible I've misunderstood!)

Also, I think we need to strengthen things a bit here. In particular, I think the point is that we want to see things like "This access wouldn't have been allowed in U-mode, but the privilege bit is set so it was", right?

dv/uvm/core_ibex/fcov/core_ibex_csr_categories.svh
```diff
@@ -99,12 +99,10 @@
   CSR_MHPMCOUNTER28H, \
   CSR_MHPMCOUNTER29H, \
   CSR_MHPMCOUNTER30H, \
-  CSR_MHPMCOUNTER31H, \
+  CSR_MHPMCOUNTER31H
   // Must exist when implementing hardware triggers (breakpoints), but read as 0 and ignore \
   // writes. Unused/Uneeded by debugger infrastructure. \
   // TODO: Don't ignore these? \
-  CSR_MCONTEXT, \
```
rswarbrick, line 106: I think we need to drop the comment above too? Or maybe move (most of) it to live above the new NOT_IMPLEMENTED_CSRS define?

vendor/lowrisc_ip/dv/sv/dv_utils/dv_fcov_macros.svh
```diff
@@ -110,3 +110,16 @@
...
+`ifndef DV_FCOV_WALR_UNSUPPORTED
+`ifdef DV_FCOV_DISABLE_CP
+  `define DV_FCOV_WALR_UNSUPPORTED(CSR_NAME_)
+`else
+  `define DV_FCOV_WALR_UNSUPPORTED(CSR_NAME_) \
+    cp_unsupported_w``CSR_NAME_: coverpoint cs_registers_i.u_``CSR_NAME_``_csr.wr_en_i && \
+                                            cs_registers_i.u_``CSR_NAME_``_csr.wr_data_i != \
+                                            cs_registers_i.csr_wdata_int { bins seen = {1}; }
+`endif
+`endif
```
rswarbrick, line 125: Since this macro is specific to `core_ibex_fcov_if.sv`, I think it should probably live there rather than in a general header. You can also drop the DV_FCOV_DISABLE_CP support, I think (since we unconditionally add coverpoints in that file otherwise). Also, I'm not sure that the comment above is quite right?

vendor/lowrisc_ip/dv/sv/dv_utils/dv_fcov_macros.svh
```diff
@@ -110,3 +110,16 @@
...
+// Creates a coverpoint for an expression where only the expression true case is of interest for
+// coverage (e.g. where the expression indicates an event has occured).
+`ifndef DV_FCOV_WALR_UNSUPPORTED
+`ifdef DV_FCOV_DISABLE_CP
+  `define DV_FCOV_WALR_UNSUPPORTED(CSR_NAME_)
+`else
+  `define DV_FCOV_WALR_UNSUPPORTED(CSR_NAME_) \
+    cp_unsupported_w``CSR_NAME_: coverpoint cs_registers_i.u_``CSR_NAME_``_csr.wr_en_i && \
+                                            cs_registers_i.u_``CSR_NAME_``_csr.wr_data_i != \
+                                            cs_registers_i.csr_wdata_int { bins seen = {1}; }
```
rswarbrick, line 123: I wonder if it would be worth defining and using a macro like `DEF_SEEN_CP` that we use in OTBN:
```systemverilog
  // A macro to define a coverpoint for a condition that should be seen: EXPR should be a single bit
  // and there's just one bin (with expected value 1'b1).
`define DEF_SEEN_CP(NAME, EXPR) NAME: coverpoint (EXPR) { bins seen = {1'b1}; }
```

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -355,6 +355,36 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     `DV_FCOV_EXPR_SEEN(mret_in_umode, id_stage_i.mret_insn_dec && priv_mode_id == PRIV_LVL_U)
     `DV_FCOV_EXPR_SEEN(wfi_in_umode, id_stage_i.wfi_insn_dec && priv_mode_id == PRIV_LVL_U)
 
+    `DV_FCOV_WALR_UNSUPPORTED(mstatus)
+    // Hardcoded
```
rswarbrick, line 359: I wonder: would it make sense to at least require that we see writes to these registers? That makes sure that the writes don't get routed to some other random CSR!

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -355,6 +355,36 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    //`DV_FCOV_WALR_UNSUPPORTED(misa)
+    //`DV_FCOV_WALR_UNSUPPORTED(mseccfg)
+    //`DV_FCOV_WALR_UNSUPPORTED(mseccfgh)
+    //`DV_FCOV_WALR_UNSUPPORTED(secureseed)
+
+    // Not implemented
+    //`DV_FCOV_WALR_UNSUPPORTED(scontext)
```
rswarbrick, line 366: What happens if someone tries to write to `scontext`? Is this covered by `cp_ignored_csrs_w`? If so, maybe a comment here pointing at that would help.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -355,6 +355,36 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    //`DV_FCOV_WALR_UNSUPPORTED(mcontext)
+    //`DV_FCOV_WALR_UNSUPPORTED(tdata3)
+
+    `DV_FCOV_WALR_UNSUPPORTED(mie)
+    `DV_FCOV_WALR_UNSUPPORTED(mtvec)
+    `DV_FCOV_WALR_UNSUPPORTED(mepc)
+    `DV_FCOV_WALR_UNSUPPORTED(mtval)
+    // Array
+    //`DV_FCOV_WALR_UNSUPPORTED(pmpcfg0)
```
rswarbrick, line 375: Is the plan to handle this in a follow-up commit? (If so, could you add a TODO?). Similarly with the "Debugs" entries below.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -429,6 +433,26 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    }
+
+    // This will only be seen when specific interrupt is disabled by MIE CSR
+    `DV_FCOV_EXPR_SEEN(irq_continue_sleep, id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
+                                           (|cs_registers_i.mip))
+
+    // Include both mstatus.mie enabled/disabled because it should not affect wakeup condition
+    irq_wfi_cross: cross cp_controller_fsm_sleep, cp_irq_pending, cs_registers_i.mstatus_q.mie{
+    ignore_bins ignore = !binsof(cp_irq_pending) intersect {1'b1};
+    }
```
rswarbrick, line 450: Nit: Missing space before the first `{` and then indentation of 2nd line (and for `debug_wfi_cross` below)

rswarbrick (commented):
I think there are a few tweaks and rethinks needed. More generally, please could you update `coverage_plan.rst` in each commit that implements coverage so that the two stay in sync?

rtl/ibex_controller.sv
```diff
@@ -886,6 +886,8 @@ module ibex_controller #(
   // FCOV //
   //////////
 
+  `DV_FCOV_SIGNAL(logic, debug_wakeup, (ctrl_fsm_cs == SLEEP) &
+                                        (debug_req_i || debug_mode_q || debug_single_step_i))
```
GregAC, line 890: It's tempting to add a `ctrl_fsm_ns == FIRST_FETCH` in here or similar, just so we're covering the actual wakeup. Really we should ensure the wakeup actually occurs but I don't know if we've got any checking for that, the cosim might just cover it.

Let's leave it for now and open an issue to remind us to check if this wakeup behaviour is checked anywhere.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -438,6 +439,26 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
 
+    cp_controller_fsm_sleep: coverpoint id_stage_i.controller_i.ctrl_fsm_cs {
+      bins out_of_sleep = (SLEEP => FIRST_FETCH);
+      bins enter_sleep = (WAIT_SLEEP => SLEEP);
+      // TODO: VCS does not implement default sequence so illegal_bins will be empty
+      illegal_bins illegal_transitions = default sequence;
+    }
+
+    // This will only be seen when specific interrupt is disabled by MIE CSR
+    `DV_FCOV_EXPR_SEEN(irq_continue_sleep, id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
```
GregAC, line 450: Lets have a `id_stage_i.controller_i.ctrl_fsm_cs == SLEEP` too as otherwise this will cover the case when we're entering sleep too.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -438,6 +439,26 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+                                           (|cs_registers_i.mip))
+
+    // Include both mstatus.mie enabled/disabled because it should not affect wakeup condition
+    irq_wfi_cross: cross cp_controller_fsm_sleep, cp_irq_pending, cs_registers_i.mstatus_q.mie {
+      ignore_bins ignore = !binsof(cp_irq_pending) intersect {1'b1};
```
GregAC, line 455: Rather than the ignore bin you can use an `iff` expression for the cross

```systemverilog
irq_wfi_cross: cross cp_controller_fsm_sleep, cs_registers_i.mstatus_q.mie iff (cp_irq_pending) {
```

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -438,6 +439,26 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    }
+
+    debug_wfi_cross: cross cp_controller_fsm_sleep, cp_debug_wakeup {
+      ignore_bins ignore = !binsof(cp_debug_wakeup) intersect {1'b1};
```
GregAC, line 459: As above instead of the ignore bin you can use an `iff`

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -326,6 +326,12 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     cp_wb_reg_no_load_hz: coverpoint id_stage_i.fcov_rf_rd_wb_hz &&
                                      !wb_stage_i.outstanding_load_wb_o;
 
+    cp_store_raw_hz: coverpoint id_stage_i.instr_type_wb_o == WB_INSTR_STORE &&
+                                id_instr_category == InstrCategoryLoad &&
+                                load_store_unit_i.addr_last_d == load_store_unit_i.addr_last_o;
```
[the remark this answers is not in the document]
ctopal: I thought by checking `addr_last_d` and `addr_last_o` we could have been able to catch both misaligned and not-misaligned conditions. In other words, I aimed for a condition where checking the addresses would happen after the misalignment corrections are done (comment above line 216 at `ibex_load_store_unit.sv`). The coverpoint wouldn't specify if the RAW occurred with a misaligned access or not but I can add explicit bins for that if it makes sense to you?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: I tried to implement it like the way you suggested. It's now in `core_ibex_pmp_fcov_if` and is named `mprv_effect_cross`. In order to get there I reshaped the RTL in a way that utilises functions wherever we use privilege levels. Would love to hear your feedback about it.

rtl/ibex_core.sv
```diff
@@ -1641,6 +1641,13 @@ module ibex_core import ibex_pkg::*; #(
           g_pmp.pmp_i.region_match_all[PMP_I2][i_region] & if_stage_i.if_id_pipe_reg_we)
       `DV_FCOV_SIGNAL(logic, pmp_region_dchan_access,
           g_pmp.pmp_i.region_match_all[PMP_D][i_region] & data_req_out)
+      // pmp_cfg[5:6] is reserved and because of that the width of it inside cs_registers module
+      // is 6-bit.
+      `DV_FCOV_SIGNAL(logic, warl_check_pmpcfg,
+          fcov_csr_write &&
+          (cs_registers_i.g_pmp_registers.g_pmp_csrs[i_region].u_pmp_cfg_csr.wr_data_i !=
+          {cs_registers_i.csr_wdata_int[(i_region%4)*PMP_CFG_W+:5],
+           cs_registers_i.csr_wdata_int[(i_region%4)*PMP_CFG_W+7]}))
```
GregAC, line 1648: Might be worth another coverpoint to cover writes to the reserved [5:6] bits. Though it's not a priority so let's make a note of it somewhere (e.g. just a TODO comment here) and sort it out in another PR.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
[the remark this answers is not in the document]
ctopal: Oh thanks for noticing that I now understand your point. This one is just catching when we have 
`mis/AlignedStore -> AlignedLoad` but we also want to catch the situations where we would have `MisalignedLoad` after either type of `Store`. I will try to come up with a solution to catch that.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -360,6 +399,14 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
       ignore_bins ignore = {`IGNORED_CSRS};
     }
 
+    cp_ignored_csrs_ro: coverpoint cs_registers_i.csr_addr_i iff (fcov_csr_read_only) {
+      bins unimplemented_csrs_read = {`NOT_IMPLEMENTED_CSRS};
+    }
```
rswarbrick, line 443: I think this is a bit more narrow than necessary: it only picks up CSR_OP_READ but not CSR_OP_SET / CSR_OP_CLEAR, which both also imply a read of the CSR.

@GregAC: Have I missed something here, or should we also be defining an `fcov_csr_read` signal? (In the latter case, happy to do it as a follow-up: this code is "sound" but not as permissive as it could be)

rtl/ibex_pmp.sv
```diff
@@ -36,10 +36,112 @@ module ibex_pmp #(
...
+  // region_match_all --------------------------------> access_fault_check <----------
+  //                                                            |
+  //                                                            \--> pmp_req_err_o
+
+  // A wrapper function in which its decided which form of permission check function gets called
+  function automatic logic perm_check_wrapper(logic csr_pmp_mseccfg_mml,
+                                      ibex_pkg::pmp_cfg_t csr_pmp_cfg,
+                                      ibex_pkg::pmp_req_e pmp_req_type,
+                                      ibex_pkg::priv_lvl_e priv_mode,
+                                      logic permission_check);
```
rswarbrick, line 58: Nit: Indentation

rtl/ibex_pmp.sv
```diff
@@ -36,10 +36,112 @@ module ibex_pmp #(
...
+      if (csr_pmp_cfg.read & csr_pmp_cfg.write & csr_pmp_cfg.exec
+          & csr_pmp_cfg.lock) begin
+        // Special-case shared read only region when R = 1, W = 1, X = 1, L = 1
+        result = pmp_req_type == PMP_ACC_READ;
+      end else begin
+        // Otherwise use basic permission check. Permission is always denied if in S/U mode and
+        // L is set or if in M mode and L is unset.
+        result =
+          priv_mode == PRIV_LVL_M ? csr_pmp_cfg.lock & permission_check :
+                                     ~csr_pmp_cfg.lock & permission_check;
```
rswarbrick, line 107: Nit: Maybe better to pull `permission_check` out as `permission_check & (PRIV_LVL_M ? csr_pmp_cfg.lock : ~csr_pmp_cfg.lock)` ?

rtl/ibex_pmp.sv
```diff
@@ -36,10 +36,112 @@ module ibex_pmp #(
...
+    // When MSECCFG.MMWP is set default deny always, otherwise allow for M-mode, deny for other
+    // modes
+    logic access_fail = csr_pmp_mseccfg_mmwp | (priv_mode != PRIV_LVL_M);
+
+    // PMP entries are statically prioritized, from 0 to N-1
+    // The lowest-numbered PMP entry which matches an address determines accessibility
+    for (int r = PMPNumRegions - 1; r >= 0; r--) begin
+      if (match_all[r])
+        access_fail = ~final_perm_check[r];
+    end
```
rswarbrick, line 142: Nit: Wouldn't this be easier to write counting upwards and using `break` on a match?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -276,6 +276,41 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+    end
+    else begin
+      curr_data_addr = load_store_unit_i.data_addr;
+    end
+  end
+
+  // If we have LOAD at ID/EX stage and STORE at WB stage, compare the calculated address for LOAD
+  // and the saved STORE address. If they are matching we would have RAW hazard.
+  assign raw_hz = id_stage_i.instr_type_wb_o == WB_INSTR_STORE &&
+                  id_instr_category == InstrCategoryLoad &&
```
rswarbrick, line 311: Micro-arch question (for @ctopal or @GregAC?): Is there a possible hazard for "fetch after write" that we should be tracking as well?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -232,6 +233,16 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
       `DV_FCOV_INSTANTIATE_CG(pmp_region_cg, en_pmp_fcov)
     end
 
+    // Current Priv means the privilege level of Instruction channels and Ibex in general. Note that
+    // the privilege level of PMP data channel can be set something different than this using MSTATUS.MPRV
+    // and MSTATUS.MPP.
+    logic pmp_current_priv_req_err;
+    assign pmp_current_priv_req_err =
+      g_pmp.pmp_i.access_fault_check(csr_pmp_mseccfg.mmwp,
```
rswarbrick, line 241: I like the way we get to re-use the functions from the RTL here! :-)

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -246,6 +257,22 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+      cp_mprv: coverpoint cs_registers_i.mstatus_q.mprv;
+
+      mprv_effect_cross: cross cp_mprv, cs_registers_i.mstatus_q.mpp,
+                               cs_registers_i.priv_mode_id_o, pmp_current_priv_req_err,
+                               pmp_dside_req_err iff
+                                 (id_stage_i.instr_rdata_i[6:0] inside
+                                    {ibex_pkg::OPCODE_LOAD, ibex_pkg::OPCODE_STORE}){
+        // If MPRV is set to 0, system priv lvl and lsu priv lvl has to be same.
+        illegal_bins illegal_mprv =
+        binsof(cp_mprv) intersect {1'b0} with (pmp_current_priv_req_err != pmp_dside_req_err);
```
rswarbrick, line 270: Nit: This line needs a couple of extra spaces.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -484,6 +492,32 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
                                            id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
                                            (|cs_registers_i.mip))
 
+    controller_instr_cross: cross cp_controller_fsm, cp_id_instr_category {
+      bins decode_to_flush =
+        binsof(cp_controller_fsm.out_of_decode0) &&
+        binsof(cp_id_instr_category) intersect {InstrCategoryMRet, InstrCategoryDRet,
```
rswarbrick, line 498: Just because I don't understand: Why do we restrict the instruction category here?

ctopal: Because the other instruction types won't result with the transaction indicated by `out_of_decode0`. I checked the highlighted controller FSM transitions and looked at if they're dependent on the decoded instruction.

rtl/ibex_pmp.sv
```diff
@@ -36,10 +36,112 @@ module ibex_pmp #(
...
+
+  // Compute permissions checks that apply when MSECCFG.MML is unset. This is the original PMP
+  // behaviour before Smepmp was added.
+  function automatic logic orig_perm_check(logic pmp_cfg_lock,
+                                          ibex_pkg::priv_lvl_e priv_mode,
+                                          logic permission_check);
+      return (priv_mode == PRIV_LVL_M) ?
+          // For M-mode, any region which matches with the L-bit clear, or with sufficient
+          // access permissions will be allowed
+          (~pmp_cfg_lock | permission_check) :
```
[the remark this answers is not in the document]
ctopal: I believe we talked about this one already but just for the record:

2nd paragraph of the "Priority and Matching Logic" in the privileged spec v.20211203 talks about this.
```
If the L bit is clear and the privilege mode of the access is M, the access
succeeds. Otherwise, if the L bit is set or the privilege mode of the access is S or U, then the access
succeeds only if the R, W, or X bit corresponding to the access type is set.
```
So, from this we can say that only the access of highest priority region matters in PMP checks and if it's unlocked we should grant the access.

ctopal:
Thanks Rupert for the review! I updated the parts you pointed out.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: Changed this one to be constrained by `csr_access` signal instead since as you said the read is ubiquitous in CSR operations.

PR #1640, opened 2022-04-27
Cosim rebase

hcallahan-lowrisc:
Updated version of pr #1613 , as the build scripts have been significantly refactored underneath.

### DIFF
--- a/dv/uvm/core_ibex/scripts/compile-tb.py
+++ b/dv/uvm/core_ibex/scripts/compile-tb.py
@@ -7,10 +7,27 @@
 import argparse
 import os
 import sys
+import subprocess
 
 from ibex_cmd import get_compile_opts
 from scripts_lib import THIS_DIR, run_one, subst_vars
 from sim_cmd import get_simulator_cmd
+
+
+def _get_iss_pkgconfig_flags(specifiers, iss_pc, simulator):
+    _flags = subprocess.check_output(
+        args=(['pkg-config'] + specifiers + iss_pc),
+        universal_newlines=True,
+        ).strip()
+    if simulator == 'xlm':
+        # See xcelium documentation for the -Wld syntax for passing
+        # flags to the linker. Passing -rpath,<path> options is tricky
+        # because commas are parsed strangely between xrun and the xmsc
+        # tool, and its easy for the options to arrive malformed. Use
+        # the following hack to get it through.
+        if '-Wl' in _flags:  # This should be in LDFLAGS only
+            _flags = "'-Xlinker {}'".format(_flags.replace('-Wl,', ''))
+    return _flags
 
 
 def main() -> int:
@@ -36,6 +53,36 @@ def main() -> int:
 
     os.makedirs(args.output, exist_ok=True)
 
+    subst_vars_dict = {
+        'core_ibex': core_ibex,
+        'out': args.output,
+        'cmp_opts': get_compile_opts(args.ibex_config,
+                                     args.simulator)
+    }
+
+    # Find the correct flags for the tb to link against the compiled ISS
+    spike_iss_pc = ['riscv-riscv', 'riscv-disasm', 'riscv-fdt']
+    iss_pkgconfig_dict = {
+        'ISS_CFLAGS'  : ['--cflags'],
+        'ISS_LDFLAGS' : ['--libs-only-other'],
+        'ISS_LIBS'    : ['--libs-only-l', '--libs-only-L'],
+    }
+    if args.en_cosim:
+        try:
+            subprocess.check_output(['pkg-config', '--exists'] + spike_iss_pc)
+        except subprocess.CalledProcessError as err:
+            raise RuntimeError(
+                f'Failed to find {spike_iss_pc} pkg-config packages. '
+                f'Did you set the PKG_CONFIG_PATH correctly?') from err
+        subst_vars_dict.update(
+            {k: _get_iss_pkgconfig_flags(v,
+                                         spike_iss_pc,
+                                         args.simulator)
+             for k, v in iss_pkgconfig_dict.items()})
+
+    if args.en_cov:
+        subst_vars_dict.update({'shared_cov_dir': args.shared_cov_dir})
+
     enables = {
         'cov_opts': args.en_cov,
         'wave_opts': args.en_wave,
@@ -44,14 +91,7 @@ def main() -> int:
     compile_cmds, _ = get_simulator_cmd(args.simulator, enables)
 
     for pre_cmd in compile_cmds:
-        cmd = subst_vars(pre_cmd,
-                         {
-                             'core_ibex': core_ibex,
-                             'out': args.output,
-                             'shared_cov_dir': args.shared_cov_dir,
-                             'cmp_opts': get_compile_opts(args.ibex_config,
-                                                          args.simulator)
-                         })
+        cmd = subst_vars(pre_cmd, subst_vars_dict)
         retcode = run_one(args.verbose, ['sh', '-c', cmd],
                           redirect_stdstreams='/dev/null')
         if retcode:
--- a/ci/ibex-rtl-ci-steps.yml
+++ b/ci/ibex-rtl-ci-steps.yml
@@ -67,10 +67,6 @@ steps:
       displayName: Run RISC-V Compliance test for Ibex RV32IMC for ${{ config }}
 
     - bash: |
-        # Setup environment to use cosim with Simple System
-        export IBEX_COSIM_ISS_ROOT=/opt/spike-cosim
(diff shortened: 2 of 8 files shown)


dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
 
     os.makedirs(args.output, exist_ok=True)
 
+    # Setup environment variables for the tb to link against the compiled ISS
+    if args.en_cosim:
+        _spike_iss_pc = 'riscv-riscv riscv-disasm riscv-fdt'
+
+        if not subprocess.run(shlex.split(f'pkg-config --exists {_spike_iss_pc}'),
```
rswarbrick, line 44: Probably easier to make `_spike_iss_pc` a list of strings and then call something like `subprocess.run(['pkg-config', '--exists'] + _spike_iss_pc, ...)`?

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
 
     os.makedirs(args.output, exist_ok=True)
 
+    # Setup environment variables for the tb to link against the compiled ISS
+    if args.en_cosim:
+        _spike_iss_pc = 'riscv-riscv riscv-disasm riscv-fdt'
```
rswarbrick, line 42: Not that I feel strongly, but why the underscore prefix?

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
...
+
+        if not subprocess.run(shlex.split(f'pkg-config --exists {_spike_iss_pc}'),
+                              check=True):
+            raise EnvironmentError(f'Failed to find "{_spike_iss_pc}" pkg-config\
+            packages. Did you set the PKG_CONFIG_PATH correctly?')
```
rswarbrick, line 47: I think this sort of thing is easier to read if you use Python's C-style string pasting (e.g. `"a" "b"` is the same as "ab") and stick newlines between the strings.

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
...
+
+        def _get_iss_pkgconfig_flags(cmd, iss_pc):
+            _flag = subprocess.check_output(
+                args=shlex.split(f'pkg-config {cmd} {iss_pc}'),
+                universal_newlines=True)
+            if args.simulator == 'xlm':
+                return _flag.replace('-Wl,', '')  # Don't ask
```
rswarbrick, line 54: I like the comment :-) But we should probably say something a little more informative for the next poor soul that has to make sense of this!

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
...
+    # Setup environment variables for the tb to link against the compiled ISS
+    if args.en_cosim:
+        _spike_iss_pc = 'riscv-riscv riscv-disasm riscv-fdt'
+
+        if not subprocess.run(shlex.split(f'pkg-config --exists {_spike_iss_pc}'),
+                              check=True):
+            raise EnvironmentError(f'Failed to find "{_spike_iss_pc}" pkg-config\
+            packages. Did you set the PKG_CONFIG_PATH correctly?')
+
+        def _get_iss_pkgconfig_flags(cmd, iss_pc):
```
rswarbrick, line 49: I don't think this is closing over anything? If not, I'd probably suggest moving it out of the parent function to keep everything reasonably short.

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +37,32 @@ def main() -> int:
...
+            else:
+                return _flag
+
+        _iss_pkgconfig_envvars_dict = {
+            'ISS_CFLAGS'  : '--cflags',
+            'ISS_LDFLAGS' : '--libs-only-other',
+            'ISS_LIBS'    : '--libs-only-l --libs-only-L',
+        }
+        for k, v in _iss_pkgconfig_envvars_dict.items():
+            os.environ[k] = _get_iss_pkgconfig_flags(v, _spike_iss_pc)
```
rswarbrick, line 64: The usual approach here is to take a copy of the environment and modify that, then pass it to subprocess (I think `run_one` has an `env` argument that you can use)

rswarbrick (commented):
Please can you squash the last commit in with the previous one (so that we don't have an intermediate state where COSIM = 1 by default)?

Also, I think some of the commit messages could be a bit more informative: once this is merged, no one will know or care about a rebase :-)

hcallahan-lowrisc:
Changes addressing the comments are pushed now. 
I think I've addressed everything, but stuff has moved around a little bit now. I've changed to using the subst_vars() method rather than environment variables for populating the yaml template.

dv/uvm/core_ibex/scripts/compile-tb.py
hcallahan-lowrisc: This is the only comment I didn't fully address in the new changes, since it does close over args.simulator, and I like the readability of keeping the function local to its use. But if you feel strongly I don't mind moving it either.

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +36,41 @@ def main() -> int:
...
+    spike_iss_pc = ['riscv-riscv', 'riscv-disasm', 'riscv-fdt']
+    iss_pkgconfig_dict = {
+        'ISS_CFLAGS'  : ['--cflags'],
+        'ISS_LDFLAGS' : ['--libs-only-other'],
+        'ISS_LIBS'    : ['--libs-only-l', '--libs-only-L'],
+    }
+    if args.en_cosim:
+
+        try:
+            subprocess.check_output(['pkg-config'] + ['--exists'] + spike_iss_pc)
```
rswarbrick, line 50: Nit: I'd probably suggest `['pkg-config', '--exists']` here

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +36,41 @@ def main() -> int:
...
+        except subprocess.CalledProcessError as err:
+            raise RuntimeError(
+                'Failed to find ' + str(spike_iss_pc) + ' pkg-config packages. '
+                'Did you set the PKG_CONFIG_PATH correctly?') from err
```
rswarbrick, line 54: This is easier with format strings:
```
                f'Failed to find {spike_iss_pc} pkg-config packages. '
                f'Did you set the PKG_CONFIG_PATH correctly?') from err
```
(the second `f'...'` is optional, but I tend to do it to make everything line up!)

You also don't need the `from err` bit: that's the default behaviour.

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -35,6 +36,41 @@ def main() -> int:
...
+    # Find the correct flags for the tb to link against the compiled ISS
+    cosim_subst_vars = {}
+    spike_iss_pc = ['riscv-riscv', 'riscv-disasm', 'riscv-fdt']
+    iss_pkgconfig_dict = {
+        'ISS_CFLAGS'  : ['--cflags'],
+        'ISS_LDFLAGS' : ['--libs-only-other'],
+        'ISS_LIBS'    : ['--libs-only-l', '--libs-only-L'],
+    }
+    if args.en_cosim:
+
```
rswarbrick, line 48: Nit: I think the conventional layout for these things wouldn't have the newline here.

dv/uvm/core_ibex/scripts/compile-tb.py
rswarbrick: Ah, I see. I'd probably prefer moving it out: that way, it's more explicit what we're passing to the helper function (and it's `args.simulator`, instead of `args`) and it also avoids the surrounding function getting too enormous.

dv/uvm/core_ibex/scripts/compile-tb.py
```diff
@@ -49,7 +85,8 @@ def main() -> int:
                              'out': args.output,
                              'cmp_opts': get_compile_opts(args.ibex_config,
                                                           args.simulator)
-                         })
+                         } | cosim_subst_vars
+                         )
```
rswarbrick, line 89: Ha! As Canberk and I discovered a couple of days ago, this operator was only added in Python 3.9. Since we only require Python 3.6, you'll have to do the ugly stateful update. (But you can, at least, hoist it out of the loop)

hcallahan-lowrisc:
Comments and nits addressed.

PR #1641, opened 2022-04-27
Updating parameters for OpenTitan option

ctopal:
Updated the parameters with respect to top_earlgrey.hjson in OpenTitan
repository. For other builds, kept the previously undeclared parameters
as their default values.


rswarbrick (approved):

rswarbrick:
Looks like the `lowrisc_ibex_ibex_riscv_compliance` core needs updating too?

PR #1642, opened 2022-04-28
Infer PMP config from Ibex config in core_ibex scripts

rswarbrick:
There are a couple of other tidy-up commits, but the reason for the change is the 2nd commit in the PR.


ctopal (approved):
LGTM thanks for the cleanup!

Issue #1514, opened 2022-01-25
Add option to for non-scrambled ICache


GregAC:
The recent PR to add scrambling support to the ICache lacks an option to allow the use of non-scrambled RAMs. Scrambling is only of interest to applications requiring security hardening, other applications may simply want an ICache without scrambled memories.

GregAC:
Ah turns out this already exists (The `ICacheScramble` parameter)!

However it's implemented by continuing to use the scrambled ram primitives just with no scrambling rounds (so turning off the scrambling).

It may be preferable to instead use the plain memory primitive when `ICacheScramble` is 0 as this will make life easier for Ibex users that don't want to drag in the OT primitive library.

I'll leave this issue open to track this as a potential enhancement.

PR #1643, opened 2022-04-29
Instantiate plain memory primitives for ICache when ICache scrambling is disabled

jamesrbailey:
Opening a PR to resolve issue #1514. Plain (ie. not scrambled) memory primitives are instantiated conditionally when parameter ICacheScramble=0. This change eliminates some unused logic when scrambling is disabled.

I have successfully run riscv_machine_mode_rand_test with and without ICache enabled, as well as with and without ICache scrambling enabled. I have also run synthesis and see a minor reduction in cell area about 10-15% when ICache is enabled without scrambling, not including memory area itself.

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -514,73 +514,117 @@ module ibex_top import ibex_pkg::*; #(
 
     for (genvar way = 0; way < IC_NUM_WAYS; way++) begin : gen_rams_inner
 
-      // SEC_CM: ICACHE.MEM.SCRAMBLE
-      // Tag RAM instantiation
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Thanks for this: looks great to me.

jamesrbailey:
One item to note is that it changes the instance names of the memories because they now sit inside an additional conditional. If users are relying on their instance names (eg. testbench references, timing constraints, PD scripts, etc.) this may cause some minor disruption.

GregAC (approved):

GregAC:
Thanks for the PR!

PR #1644, opened 2022-04-30
Sim: Simple system: handle unused parameter

kgugala:
Verilator complains about MHPMCounterNum being not used. Passing it to the instantiated core solves the issue

Signed-off-by: Karol Gugala <kgugala@antmicro.com>


rswarbrick (approved):

rswarbrick:
Thanks!

PR #1653, opened 2022-05-12
[dv] Fix bitmanip test building

GregAC:


GregAC:
I did briefly experiment with trying the ISS ISA string (that has the individual components e.g. _Zba_Zbb_Xbitmanip) but our GCC build didn't like it. I figure it doesn't actually matter as we're building assembly files so the toolchain being capable of emitting bitmanip instructions we don't want for a config isn't an issue (it would only be a problem if we were compiling .c).

@luismarques - Any comments on bitmanip ISA strings for GCC/LLVM toolchain builds? Will there be a switch to using the Zxx and Xbitmanip strings (e.g. `rv32imc_Zba_Zbb_Zbs_Xbitmanip`)?

rswarbrick (commented):
Sorry for messing this up! This patch won't quite work because `base_isa` is used for "isa_iss" as well. Maybe you just need to add something like `toolchain_isa = base_isa + ('b' if has_bitmanip else '')`?

rswarbrick (approved):

GregAC:
Ignoring the CI failure as it's a known issue and not to do with this PR

luismarques:
> @luismarques - Any comments on bitmanip ISA strings for GCC/LLVM toolchain builds? Will there be a switch to using the Zxx and Xbitmanip strings (e.g. `rv32imc_Zba_Zbb_Zbs_Xbitmanip`)?

Using Zba Zbb Zbc Zbs in OpenTitan is a policy decision, there's no toolchain impediment. I'm not sure what the holdup is. The non-ratified bitmanip extensions are less likely to be used, at least for automatic (compiler-generated) code generation.

GregAC:
> Using Zba Zbb Zbc Zbs in OpenTitan is a policy decision, there's no toolchain impediment. I'm not sure what the holdup is.

There's no holdup, I was just wondering if we'll always be using the `rv32imcb` ISA string for the toolchain or if they'll be a more granular `rv32imc_Zba_Zbb_Zbs_Xbitmanip` style possibility instead at some point.

luismarques:
> There's no holdup, I was just wondering if we'll always be using the `rv32imcb` ISA string for the toolchain or if they'll be a more granular `rv32imc_Zba_Zbb_Zbs_Xbitmanip` style possibility instead at some point.

IIRC you *can* already use `rv32imc_Zba_Zbb_Zbs_Zbp`. There's no `Xbitmanip`, but the other subsets were extended/adapted to include the modifications from the non-ratified specs, based on Pirmin's work. IIRC Clang did not include the same Zbp instructions but if you're using GCC then it should work with those `_Zb*` extensions. But my memory could be betraying me...

PR #1657, opened 2022-05-13
[dv] PMP related functional coverage points

ctopal:
Adding MSECCFG CSR related functionality also some write checks etc.

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
@@ -21,7 +21,7 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
 
   input logic data_req_out,
 
-  input logic lsu_req_done
+  input fcov_csr_write
 );
   localparam int unsigned PMPNumChan = 3;
 
@@ -31,40 +31,39 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   // Enum to give more readable coverage results for privilege bits. 4 bits are from the pmpcfg CSF
   // (L, X, W, R) and the 5th is the MML bit. Setting the MML bit changes the meaning of the other
   // 4 bits
-  // TODO: Better MML names?
   typedef enum logic [4:0] {
-    PMP_PRIV_NONE     = 5'b00000,
-    PMP_PRIV_R        = 5'b00001,
-    PMP_PRIV_W        = 5'b00010,
-    PMP_PRIV_WR       = 5'b00011,
-    PMP_PRIV_X        = 5'b00100,
-    PMP_PRIV_XR       = 5'b00101,
-    PMP_PRIV_XW       = 5'b00110,
-    PMP_PRIV_XWR      = 5'b00111,
-    PMP_PRIV_L        = 5'b01000,
-    PMP_PRIV_LR       = 5'b01001,
-    PMP_PRIV_LW       = 5'b01010,
-    PMP_PRIV_LWR      = 5'b01011,
-    PMP_PRIV_LX       = 5'b01100,
-    PMP_PRIV_LXR      = 5'b01101,
-    PMP_PRIV_LXW      = 5'b01110,
-    PMP_PRIV_LXWR     = 5'b01111,
-    PMP_PRIV_MML_NONE = 5'b10000,
-    PMP_PRIV_MML_R    = 5'b10001,
-    PMP_PRIV_MML_W    = 5'b10010,
-    PMP_PRIV_MML_WR   = 5'b10011,
-    PMP_PRIV_MML_X    = 5'b10100,
-    PMP_PRIV_MML_XR   = 5'b10101,
-    PMP_PRIV_MML_XW   = 5'b10110,
-    PMP_PRIV_MML_XWR  = 5'b10111,
-    PMP_PRIV_MML_L    = 5'b11000,
-    PMP_PRIV_MML_LR   = 5'b11001,
-    PMP_PRIV_MML_LW   = 5'b11010,
-    PMP_PRIV_MML_LWR  = 5'b11011,
-    PMP_PRIV_MML_LX   = 5'b11100,
-    PMP_PRIV_MML_LXR  = 5'b11101,
-    PMP_PRIV_MML_LXW  = 5'b11110,
-    PMP_PRIV_MML_LXWR = 5'b11111
+    NONE        = 5'b00000,
+    R           = 5'b00001,
+    W           = 5'b00010,
+    WR          = 5'b00011,
+    X           = 5'b00100,
+    XR          = 5'b00101,
+    XW          = 5'b00110,
+    XWR         = 5'b00111,
+    L           = 5'b01000,
+    LR          = 5'b01001,
+    LW          = 5'b01010,
+    LWR         = 5'b01011,
+    LX          = 5'b01100,
+    LXR         = 5'b01101,
+    LXW         = 5'b01110,
+    LXWR        = 5'b01111,
+    MML_NONE    = 5'b10000,
+    MML_RU      = 5'b10001,
+    MML_WRM_RU  = 5'b10010,
+    MML_WRU     = 5'b10011,
+    MML_XU      = 5'b10100,
+    MML_XRU     = 5'b10101,
+    MML_WRM_WRU = 5'b10110,
+    MML_XWRU    = 5'b10111,
+    MML_L       = 5'b11000,
+    MML_RM      = 5'b11001,
+    MML_XM_XU   = 5'b11010,
+    MML_WRM     = 5'b11011,
+    MML_XM      = 5'b11100,
+    MML_XRM     = 5'b11101,
+    MML_XRM_XU  = 5'b11110,
+    MML_RM_RU   = 5'b11111
   } pmp_priv_bits_e;
 
   // Break out PMP signals into individually named signals for direct use in `cross` as it cannot
@@ -80,6 +79,14 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   logic [PMPNumRegions-1:0] current_priv_perm_check;
   bit en_pmp_fcov;
 
+  logic csr_wdata_mseccfg_rlb;
+  logic csr_wdata_mseccfg_mmwp;
+  logic csr_wdata_mseccfg_mml;
+
+  assign csr_wdata_mseccfg_rlb = cs_registers_i.csr_wdata_int[CSR_MSECCFG_RLB_BIT];
+  assign csr_wdata_mseccfg_mmwp = cs_registers_i.csr_wdata_int[CSR_MSECCFG_MMWP_BIT];
+  assign csr_wdata_mseccfg_mml = cs_registers_i.csr_wdata_int[CSR_MSECCFG_MML_BIT];
+
   initial begin
     if (PMPEnable) begin
       void'($value$plusargs("enable_ibex_fcov=%d", en_pmp_fcov));
@@ -89,6 +96,44 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   end
 
   if (PMPEnable) begin : g_pmp_cgs
+    logic [PMPNumRegions-1:0] pmp_iside_match;
+    logic [PMPNumRegions-1:0] pmp_iside2_match;
+    logic [PMPNumRegions-1:0] pmp_dside_match;
+    logic [PMPNumRegions-1:0] pmp_dside_match_last;
+
+    logic pmp_iside_nomatch;
+    logic pmp_iside2_nomatch;
+    logic pmp_dside_nomatch;
+
+    logic pmp_iside_boundary_cross;
+    logic pmp_dside_boundary_cross;
+
+    logic misaligned_pmp_err_last;
+
+    assign pmp_iside_match      = g_pmp.pmp_i.region_match_all[PMP_I];
+    assign pmp_iside2_match     = g_pmp.pmp_i.region_match_all[PMP_I2];
+    assign pmp_dside_match      = g_pmp.pmp_i.region_match_all[PMP_D];
+
+    assign pmp_iside_nomatch  = ~|pmp_iside_match;
+    assign pmp_iside2_nomatch = ~|pmp_iside2_match;
+    assign pmp_dside_nomatch  = ~|pmp_dside_match;
+
+    assign misaligned_pmp_err_last = load_store_unit_i.fcov_ls_first_req ?
+                                       load_store_unit_i.data_pmp_err_i :
+                                       misaligned_pmp_err_last;
+
+    // Store the Data Channel PMP match info from the first request to calculate boundary cross
+    always @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        pmp_dside_match_last <= '0;
+      end else if (load_store_unit_i.fcov_ls_first_req) begin
+        pmp_dside_match_last <= pmp_dside_match;
+      end
+    end
+
+    assign pmp_iside_boundary_cross = |(pmp_iside_match ^ pmp_iside2_match);
+    assign pmp_dside_boundary_cross = |(pmp_dside_match ^ pmp_dside_match_last);
+
     for (genvar i_region = 0; i_region < PMPNumRegions; i_region += 1) begin : g_pmp_region_fcov
       pmp_priv_bits_e pmp_region_priv_bits;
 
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -96,10 +108,23 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+        cp_mmwp : coverpoint csr_pmp_mseccfg.mmwp;
+
+        cp_mml : coverpoint csr_pmp_mseccfg.mml;
+
+        // TODO: Add a cross for checking the stickyness of the mseccfg registers.
+
         cp_region_mode : coverpoint csr_pmp_cfg[i_region].mode;
 
         cp_region_priv_bits : coverpoint pmp_region_priv_bits {
-          wildcard ignore_bins ignore = {5'b1????};
```
rswarbrick, line 102: It's not that I disagree, but I don't quite understand: why was this combination previously ignored?

rswarbrick, line 102: It's not that I disagree, but I don't quite understand: why was this combination previously ignored?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -136,26 +161,118 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+            wildcard illegal_bins illegal_allow_exec =
+              // In User mode - no match case, we should always deny
+              binsof(cp_priv_lvl_iside) intersect {PRIV_LVL_U} &&
+              binsof(cp_req_type_iside) intersect {PMP_ACC_EXEC} &&
+              binsof(pmp_iside_req_err) intersect {0} ||
+              // In Machine mode - no match case, we should deny always except MML=0 & MMWP=0
+              (binsof(cp_priv_lvl_iside) intersect {PRIV_LVL_M} &&
+               binsof(cp_req_type_iside) intersect {PMP_ACC_EXEC} &&
+               binsof(pmp_iside_req_err) intersect {0}) with
+                ((!cp_mmwp) && !(cp_region_priv_bits & 5'b10000));
```
rswarbrick, line 178: Nit: I think the indentation is off here? I think line 178 should align with line 175? (I'd probably move the "with" to this line too, to make it clear that all of this isn't nested within the parens on line 175)

Also, I think that the stuff on lines 171-173 is supposed to be at the same level as 175-178. I think the code is correct, but could you add parens around that stuff too to make it look more obviously symmetrical?

Similarly with the crosses below.

rswarbrick, line 178: Nit: I think the indentation is off here? I think line 178 should align with line 175?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -136,26 +161,118 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+            }
+
+        pmp_iside2_nomatch_cross :
+          cross cp_region_priv_bits, cp_req_type_iside2, cp_priv_lvl_iside2, pmp_iside2_req_err,
+                cp_mmwp
+            iff (pmp_iside2_match && csr_pmp_cfg[i_region].mode != PMP_MODE_OFF) {
+            // Will never see a succesful exec access when execute is disallowed
+            wildcard illegal_bins illegal_allow_exec =
+              // In User mode - no match case, we should always deny
+              binsof(cp_priv_lvl_iside2) intersect {PRIV_LVL_U} &&
```
rswarbrick, line 195: Looking at these three crosses, I wonder whether there's a way we could use a macro to collapse stuff down a bit? It looks rather repetitive. But don't stress too much: it might be more difficult than it looks!

rswarbrick, line 195: Looking at these three crosses, I wonder whether there's a way we could use a macro to collapse stuff down a bit? It looks rather repetitive. But don't stress too much: it might be more difficult than it looks!

rswarbrick (approved):
Oof, this looks like it was painful to write! This all looks good to me, with a couple of nitty questions.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
ctopal: The cross rules are tiny bit different for the data channel and same for `PC`/`PC+2` PMP checks. If there would be a macro it would only include two of the examples out of three and I am not sure if that would be prettier with the overhead from the macros.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
ctopal: I am not exactly sure, maybe because it was unimplemented yet. @GregAC what do you think?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -21,48 +21,51 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
   input logic data_req_out,
 
-  input logic lsu_req_done
+  input fcov_csr_write
 );
   `include "dv_fcov_macros.svh"
   import uvm_pkg::*;
 
+  `ASSERT(IbexRLBWriteAfterLocked, cs_registers_i.g_pmp_registers.any_pmp_entry_locked |-> ##[0:$]
+      $stable(csr_pmp_mseccfg.rlb))
```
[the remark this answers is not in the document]
ctopal: I removed this and instead added an illegal bin cause to `rlb_csr_cross` about it (named "illegal_locked"), in which we are checking the change (or not there of!) in the register.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
GregAC: It was ignored because we weren't stimulating any of the epmp modes (top bit is set when `MML` is turned on).

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -185,31 +341,125 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+            (binsof(cp_priv_lvl_dside) intersect {PRIV_LVL_U} &&
+             binsof(cp_req_type_dside) intersect {PMP_ACC_WRITE} &&
+             binsof(pmp_dside_req_err) intersect {1})
+            with (cp_region_priv_bits inside {MML_WRM_WRU, MML_WRU, MML_XWRU});
         }
+
+        rlb_csr_cross : cross cp_rlb, csr_wdata_mseccfg_rlb
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
+              (binsof(csr_wdata_mseccfg_rlb) intersect {0} &&
```
GregAC, line 429: This is simply disabling `RLB` (writing it to 0 when it's set to 1). It's not illegal and doesn't need any specific coverage bin as it'll just get included in the generic cross bins.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -185,31 +341,125 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+            // Trying to enable RLB when RLB is disabled and locked regions present is illegal
+            illegal_bins illegal_locked =
+              (binsof(csr_wdata_mseccfg_rlb) intersect {1} &&
+               binsof(cp_rlb) intersect {0})
+              iff (cs_registers_i.g_pmp_registers.any_pmp_entry_locked);
+          }
+
+        mmwp_csr_cross : cross cp_mmwp, csr_wdata_mseccfg_mmwp
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
```
GregAC, line 440: Same thing applies here as to `illegal_locked` for RLB

GregAC (commented):
Thanks for the work on this @ctopal think we're almost there just a couple of things to address. The PMP fcov was always going to be fiddly so it's no surprise we need to go through multiple iterations here!

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -136,26 +163,128 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
         pmp_dside_mode_cross : cross cp_region_mode, pmp_dside_req_err
           iff (g_pmp_fcov_signals.g_pmp_region_fcov[i_region].fcov_pmp_region_dchan_access);
 
-        // TODO: Coverage for accesses that don't match any regions
+        pmp_iside_nomatch_cross :
+          cross cp_region_priv_bits, cp_req_type_iside, cp_priv_lvl_iside, pmp_iside_req_err,
+                cp_mmwp
+            iff (pmp_iside_match && csr_pmp_cfg[i_region].mode != PMP_MODE_OFF) {
```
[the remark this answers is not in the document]
ctopal: Sorry this was a mistake on my end, now fixed the `iff` clause with `~|match`. I am not sure if it's worth it to carry the cross outside of this region since it is using coverpoints from `pmp_region_cg`.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -136,26 +163,128 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
-
-          // Will never see an exec access denied when executed is allowed
-          illegal_bins illegal_deny_exec = binsof(cp_region_priv_bits) intersect {5'b00100, 5'b00101,
-            5'b00110, 5'b00111, 5'b01100, 5'b01101, 5'b01110, 5'b01111} &&
-            binsof(cp_req_type_iside) intersect {PMP_ACC_EXEC} &&
-            binsof(pmp_iside_req_err) intersect {1};
+            // Will never see a succesful exec access when execute is disallowed
+            wildcard illegal_bins illegal_allow_exec =
+              // Non-MML disallowance conditions
+              (binsof(cp_region_priv_bits) intersect {5'b0?0??} &&
```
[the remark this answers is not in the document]
ctopal: You are right it does not compile with VCS. I now fixed the wildcards and added a comment explaining the problem.

ctopal:
Hi Greg, thanks for the review I corrected most of things that you mentioned. About the last two comments, I tried to do the thing that you asked but not sure if it is sound enough. I would really appreciate if you could have another look. Thanks!

rswarbrick:
Hi @GregAC! Any chance of another look? It would be nice to get some of these closed.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -31,38 +31,38 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   // 4 bits
   // TODO: Better MML names?
```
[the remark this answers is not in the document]
GregAC: TODO is still here

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
GregAC: Was something missed off the PR update? I don't think this has changed.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -185,31 +341,125 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
         }
+
+        rlb_csr_cross : cross cp_rlb, csr_wdata_mseccfg_rlb
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
+              (binsof(csr_wdata_mseccfg_rlb) intersect {0} &&
+               binsof(cp_rlb) intersect {1});
+            // Trying to enable RLB when RLB is disabled and locked regions present is illegal
+            illegal_bins illegal_locked =
+              (binsof(csr_wdata_mseccfg_rlb) intersect {1} &&
```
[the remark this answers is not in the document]
GregAC: Has this changed?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
GregAC: Has this changed?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -185,31 +341,125 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+        mmwp_csr_cross : cross cp_mmwp, csr_wdata_mseccfg_mmwp
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
+              (binsof(csr_wdata_mseccfg_mmwp) intersect {0} &&
+               binsof(cp_mmwp) intersect {1});
+          }
+
+        mml_sticky_cross : cross cp_mml, csr_wdata_mseccfg_mml
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
```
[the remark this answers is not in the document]
GregAC: Has this changed?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -89,6 +97,38 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+    logic pmp_dside_boundary_cross;
+
+    logic misaligned_pmp_err_last;
+
+    assign pmp_iside_match      = g_pmp.pmp_i.region_match_all[PMP_I];
+    assign pmp_iside2_match     = g_pmp.pmp_i.region_match_all[PMP_I2];
+    assign pmp_dside_match      = g_pmp.pmp_i.region_match_all[PMP_D];
+    assign pmp_dside_match_last = load_store_unit_i.fcov_ls_first_req ?
+                                    pmp_dside_match :
+                                    pmp_dside_match_last;
```
GregAC, line 119: Did you want a flop here? As is you've got a circular reference

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -203,31 +309,207 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+          }
+
+        mml_sticky_cross : cross cp_mml, csr_wdata_mseccfg_mml
+          iff (fcov_csr_write && cs_registers_i.csr_addr_i == CSR_MSECCFG) {
+            illegal_bins illegal_sticky =
+              (binsof(csr_wdata_mseccfg_mml) intersect {0} &&
+               binsof(cp_mml) intersect {1});
+          }
+
+        pmp_iside_nomatch_cross :
```
GregAC, line 424: As non of this is per region it would be nice if it could live outside of the per-region cover groups, though I think this means you need to duplicate some coverpoints which is a bit messy. I guess having the same bins replicated doesn't really matter from a coverage number perspective though potentially artificially skews the % covered ratio. Any thoughts @rswarbrick?

GregAC (commented):
The region boundary cross stuff looks good, other than I think the dside boundary cross detection logic needs a flop somewhere.

Potentially some updates missing from the latest push as well? Or is it not clear what I mean from the original comments?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
ctopal: Added the flop and a comment explaining why we are saving `dside_match` for the `first_req`. (Although I am not sure why keeping the old version would cause a trouble considering it's not going to get synthesised.)

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
rswarbrick: Eh, I'm not too concerned about that: honestly, the only numbers that matter are 100% and not-100% :-)

GregAC (approved):
Think we're good to go, great work @ctopal!

Issue #1579, opened 2022-03-18
[dv] Implement ePMP tests


GregAC:
RISC-DV only generates code to test normal PMP, this needs to be extended to stimulate ePMP functionality.

Original estimate 8.
2022-07-28 based on @marnovandermaas comment increasing estimate to accommodate time spent on issue
> estimate 9
> remaining 2022-07-22 8
> remaining 2022-08-01 2

haroonshafique-10xe:
Hi @GregAC 

This issue is for adding/extending the support for ePMP functionality tests in riscv-dv. Right?
Also, which ISS you are using for ePMP? 
I think mainstream spike does not support ePMP. I know a fork which has ePMP support https://github.com/joxie/riscv-isa-sim.

Thanks.

GregAC:
The majority of the work is likely to be in riscv-dv yes.

We're using Spike as our simulator. It is unfortunate ePMP hasn't reached the upstream yet. However as it's fairly simple we should be able to rebase the relevant patches from the fork on upstream or possibly implement from scratch.

haroonshafique-10xe:
Right.

> rebase the relevant patches from the fork on upstream

Does it mean that the changes from the fork (joxie/riscv-isa-sim) will be rebased on upstream (riscv-software-src/riscv-isa-sim) and then make a PR from their into the upstream? 

> possibly implement from scratch.

Does it mean to implement in the fork (lowrisc/riscv-isa-sim) only for now?

rswarbrick:
Hi there! Yes, we'll probably try to take those changes if possible. We don't want to diverge from upstream, so will do our best to get the changes upstreamed if we can.

haroonshafique-10xe:
Hi @rswarbrick 

Yes, sounds good.

I am not sure whether you have seen it or not, there is a [test-suite](https://github.com/joxie/riscv-isa-sim/tree/master/tests/mseccfg) already available in the https://github.com/joxie/riscv-isa-sim.That can be used if in immediate need for basic testing. 

However, I plan to add support for ePMP in the riscv-dv.

vogelpi:
Update: We have now a new version of Spike with ePMP support (v.0.3, CI keeps using v.0.2 at the moment) and RISCV-DV with basic ePMP support has been vendored in. To enable ePMP checking, one needs to use Spike v.0.3 and inside `riscv_core_setting.sv` one has to set `support_epmp` to `1`.

marnovandermaas:
[...]

PR #1659, opened 2022-05-16
Switch to updated Spike version

vogelpi:
This commit switches to an updated Spike version with support for ePMP.

~I suspect that this PR won't pass CI because the updated Spike version is not yet available on the tool NAS. I just created a new `ibex-cosim-v0.3` tag for that new version.~ ~Related to that I just realized that I probably made mistake when rebasing as I replaced previous call we had to `get_max_xlen()` with `get_xlen()`. I only discovered now that `get_max_xlen()` has been moved to `isa_parser.h` within Spike. I need fix this.~ This is now done.

This is related to #1579.

### DIFF
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -14,6 +14,23 @@
 #include <iostream>
 #include <sstream>
 
+// For a short time, we're going to support building against version
+// ibex-cosim-v0.2 (20a886c) and also ibex-cosim-v0.3 (9af9730). Unfortunately,
+// they've got different APIs and spike doesn't expose a version string.
+//
+// However, a bit of digging around finds some defines that have been added
+// between the two versions.
+//
+// TODO: Once there's been a bit of a window to avoid a complete flag day,
+//       remove this ugly hack!
+#ifndef HGATP_MODE_SV57X4
+#define OLD_SPIKE
+#endif
+
+#ifndef OLD_SPIKE
+#include "riscv/isa_parser.h"
+#endif
+
 SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
                        uint32_t start_mtvec, const std::string &trace_log_path,
                        bool secure_ibex, bool icache_en)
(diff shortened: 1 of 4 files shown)


dv/cosim/spike_cosim.cc
```diff
@@ -24,8 +25,10 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
     log_file = log->get();
   }
 
+  isa_parser = new isa_parser_t(isa_string.c_str(), "MU");
```
vogelpi, line 28: I am not sure if this is the "correct" way of doing this. I've tried using a `unique_ptr` type for this but if I understand correctly, this doesn't work because we need to pass `isa_parser *` to constructor of `processor` (part of Spike), casting doesn't work and I think it also shouldn't work here. Anyway, some input on this is appreciated.

rswarbrick: There are two clean ways to do this.

First (and most similar to what you've got), turn `isa_parser` into a `unique_ptr<isa_parser_t>` and use `isa_parser.get()` to get the underlying pointer.

Second (and probably less effort), make `isa_parser` an actual `isa_parser_t` as a field and call its constructor in the class initialisation list. Then you can pass it to the `processor_t` constructor with just `&isa_parser`.

vogelpi: Thanks Rupert, I've now switched to the first option :-)

vogelpi:
Thank you both, I've now fixed the lint error. Let's see if CI passes.

vogelpi:
Cool, CI is passing now. Should we merge this?

rswarbrick (commented):
Merging this will cause a "flag day" for anyone who uses Ibex because they'll need to update to version 0.3 to keep working. How difficult would it be to add an ifdef like I did in 96d8aa6c1522eae45b4ec0d03424c9c753544fe7? The idea is that we'd keep both versions working for a week or so to allow things like bisecting across the boundary.

vogelpi:
The argument for enabling bisecting across the version boundary convinced me @rswarbrick . I've now added a similar switch. By default, we continue to use `ibex-cosim-v0.2` for the moment. Feel free to merge this if you're happy with it. I will be OoO for the rest of the week.

rswarbrick:
Thanks for doing this, Pirmin. I'm not completely sure about the `ifdef` though: the `MSTATUSH_GVA` symbol is defined for `ot-v0.2` and not for `ot-v0.1`. To do the same thing, I think you'd need to find a preprocessor symbol that's defined for `ot-v0.3` but not for `ot-v0.2`. If this isn't possible, never mind. But we could always add a version header to our Spike fork for future situations like this.

vogelpi:
Ah, I simply thought `MSTATUS_GVA` was something you added and wasn't aware it's something defined by Spike. I'll need to have a look if there is something similar for `ot-v0.3`. What is definitely working is to add a `-DMSECCFG_GVA` argument for the used simulator in `dv/uvm/core_ibex/yaml/rtl_simulation.yaml`. But it would be much better if this was defined by Spike already.

vogelpi:
I've now identified another define that can be found in v0.3 but not in v0.2 and updated the PR accordingly. Successfully tested in CI (still using v0.2) and locally (using v0.3).

rswarbrick (approved):
Neat! Let's hope the Spike maintainers keep adding defines :-p

PR #1660, opened 2022-05-16
Dump riscv-dv generation messages to a log file

rswarbrick:
I was previously just dumping them to /dev/null because the
code always worked but... predictably I was wrong! Write them
somewhere more useful for debug.


rswarbrick:
Hmm. Looks like `stdstream_dest` is just an fd in some situations. Let me re-push with something that ignores more errors.

hcallahan-lowrisc (approved):

PR #1662, opened 2022-05-17
[dv,core_ibex] Allow instructions near the top of initialised IMEM

rswarbrick:
If you call the `read()` function on the memory model with an
uninitialised word, it generates a UVM error.

This is reasonable for data memory (where we never want to read
something without an architectural value) but is not reasonable for
IMEM, where Ibex runs ahead. Squash the error in this case, but force
bad integrity for the fetch to make sure we see something explode.

This was inspired by trying to debug riscv_csr_test. It still fails after this change, but for different reasons :-) Specifically, it's designed so that if a check fails then it jumps to the end of its instructions and loops forever. Before this change, it then died immediately when Ibex read ahead. Now it hangs (arguably not much of an improvement!)


GregAC (approved):
LGTM though could we also set the error bit on the response? As in non secure configs we won't be looking at the integrity bits.

rswarbrick:
Oh, good idea. I'll force it to all zeros (which is an unimplemented instruction in RISC-V).

rswarbrick:
Hmm. Or maybe something like 0xdeadbeef would be better: I'll feed it through objdump to find out whether it's disallowed :-)

GregAC:
I for one am disappointed we can't use `D00D2BAD`!

PR #1665, opened 2022-05-19
Change makefile default simulator for core_ibex dv to xcelium

hcallahan-lowrisc:


ctopal (approved):
I already use xlm quite heavily without any problem, so it should be fine :-) Thanks!

PR #1666, opened 2022-05-20
Fix Formatting of IcacheScramble Description

nedguthrie:
Description field for the IcacheScramble instruction rendered incorrectly in the [html version of this page](https://ibex-core.readthedocs.io/en/latest/02_user/integration.html). Fixed the issue in the github rendered version (see rich diff) which will presumable also fix the problem in the HTML version.


rswarbrick (approved):
Thank you very much for the fix!

PR #1670, opened 2022-05-30
[fcov] Adding debug related functional coverage

ctopal:
Includes coverpoints for:

- Hardware trigger point matches
- Debug simple step entrance in controller
- Cross for seeing different insns while single stepping

Also updates on coverage plan to fill up missing mentions of
coverpoints/crosses

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -276,6 +276,22 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     end
   end
 
+  logic instr_id_matches_trigger_d, instr_id_matches_trigger_q;
+
+  assign instr_id_matches_trigger_d = id_stage_i.controller_i.trigger_match_i &&
+                                      id_stage_i.controller_i.fcov_debug_entry_if;
+
+  // Delay instruction matching trigger point since it is catched in IF stage.
+  // We would want to cross it with decoded instruction categories and it does not matter
+  // when exactly we are hitting the condition.
+  always @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      instr_id_matches_trigger_q <= 1'b0;
+    end else begin
+      instr_id_matches_trigger_q <= instr_id_matches_trigger_d;
+    end
+  end
+
   // Keep track of previous data addr of Store to catch RAW hazard caused by STORE->LOAD
   logic [31:0]     prev_store_addr;
   logic [31:0]     data_addr_incr;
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -456,6 +472,10 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     `DV_FCOV_EXPR_SEEN(debug_entry_if, id_stage_i.controller_i.fcov_debug_entry_if)
     `DV_FCOV_EXPR_SEEN(debug_entry_id, id_stage_i.controller_i.fcov_debug_entry_id)
     `DV_FCOV_EXPR_SEEN(pipe_flush, id_stage_i.controller_i.fcov_pipe_flush)
+    `DV_FCOV_EXPR_SEEN(single_step_taken, id_stage_i.controller_i.fcov_debug_single_step_taken)
+    `DV_FCOV_EXPR_SEEN(insn_trigger_enter_debug, instr_id_matches_trigger_q)
+    `DV_FCOV_EXPR_SEEN(insn_trigger_exception, instr_id_matches_trigger_q &&
+                                               id_stage_i.controller_i.fcov_pipe_flush)
```
vogelpi, line 478: I was wondering if there would be better way of detecting the exception than using `fcov_pip_flush` just in case the flush behavior changes in the future. But I didn't find some other suitable signal or condition. I think this is okay.

vogelpi (approved):
This looks good to me. Thanks for the PR @ctopal !

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: Yes I thought since any instruction exception would mean a pipe flush this should be fine.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -492,6 +512,12 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
                                            id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
                                            (|cs_registers_i.mip))
 
+
+    single_step_instr_cross: cross id_stage_i.debug_single_step_i, cp_id_instr_category {
```
GregAC, line 516: I think you'd be better off with `fcov_debug_single_step_taken` here rather than `debug_single_step_i`. The issue is the `debug_single_step_i` signal will remain asserted in debug mode so this cross will capture instructions executed in debug mode whilst doing the single step. Instead it should only capture instructions we single step (i.e. execute outside of debug mode with single step activated).

rtl/ibex_controller.sv
```diff
@@ -895,6 +895,7 @@ module ibex_controller #(
       (ctrl_fsm_cs != DBG_TAKEN_ID) & (ctrl_fsm_ns == DBG_TAKEN_ID))
   `DV_FCOV_SIGNAL(logic, pipe_flush, (ctrl_fsm_cs != FLUSH) & (ctrl_fsm_ns == FLUSH))
   `DV_FCOV_SIGNAL(logic, debug_req, debug_req_i & ~debug_mode_q)
+  `DV_FCOV_SIGNAL(logic, debug_single_step_taken, do_single_step_q)
```
GregAC, line 898: I think we'd be better off capturing the edge here, so this is asserted the cycle the controller chooses to single step, so we'd want the expression `do_single_step_d & ~do_single_step_q`.

GregAC (commented):
@ctopal My apologies for taking a long time to review this. Couple of minor tweaks to the single step related stuff but otherwise looks good.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: Oh sorry! I think I got what you mean. Is it because `debug_single_step_i` would also catch what's getting executed inside debug memory? If so maybe a cross would be pointless and I can do it like this:

```systemverilog
    single_step_instr_cp: coverpoint cp_id_instr_category iff (fcov_debug_single_step_taken);
```

GregAC (approved):
Thanks for this @ctopal this is good to go, nice work :)

PR #1672, opened 2022-05-31
[syn] Use sv2v for prim_buf and prim_generic_buf

### LINKED ISSUE #1557: yosys synthesis flow - Module `\prim_buf' is not part of the design.
Hi,

I am trying to run the experimental Yosys/OpenSTA Ibex Synthesis flow according to the README instructions, but I get the following error during synthesis:

```
31.2.3. Executing AST frontend in derive mode using pre-parsed AST for module `\ibex_core'.
ERROR: Module `\prim_buf' referenced in module `\ibex_top' in cell `\u_rf_rdata_b_ecc_buf' is not part of the design.
```
When I look into the generated code in `syn_out`, I see that `ibex_top.v` does reference `prim_buf` multiple times, which is not defined in the generated verilog files.

I am not sure how to solve this, maybe it's necessary to include the `prim_buf`rtl definition in the `sv2v` [call from line 47](https://github.com/lowRISC/ibex/blob/094451a9489d631895101a5de19a26cd4cf4fd7d/syn/syn_yosys.sh#L47) of `syn_yosys.sh`.

## My Environment

I am following the instruction from [syn/README.md](https://github.com/lowRISC/ibex/blob/master/syn/README.md).

- Yosys 0.9 (git sha1 1979e0b) installed from apt
- sv2v v0.0.9-0-gc822d2e binary from GitHub
- OpenSTA installed from GitHub latest source code (git hash 3c3f8f61fde2dd28afec9eca81f5b361f22f0cc4)
- OpenRoad-flow-scripts from latest GitHub version (git hash fbcfb64bc92d6543e5c6c5c7e4c485a7b7709198)

**EDA tool and version:**

**Operating system:**

- Ubuntu 20.04

- Ibex from `master` (git hash 094451a9489d631895101a5de19a26cd4cf4fd7d)

I haven't made any changes to the source code.

ctopal:
Not sure if this is a clean way of doing it but as far as I see
if we are using sources from `rtl/` we should at least convert
`prim_buf` and `prim_generic_buf` to Verilog as well.

Fixes #1557


rswarbrick (approved):

rswarbrick:
Oh, wait: I thought we'd agreed that the right solution was to change uses of `prim_buf` to use `prim_generic_buf`? See, for example, the super-fancy sed scripts that @vogelpi added for AES :-p

https://github.com/lowRISC/opentitan/blob/b4d3dedb2f09a22614f365a0c156aa4afa88e375/hw/ip/aes/pre_syn/syn_yosys.sh#L107-L112

vogelpi:
As we have the `prim_buf` module available in Ibex already (because for OpenTitan these are auto-generated on the fly, thus not available in the archaic AES/KMAC synthesis flows) it's fine to it.

The main reason for switching to the Xilinx primitives for AES and KMAC in the first place was that only these have synthesis attributes to avoid aggressive optimizations. Yosys can then be instructed to translate these attributes to something that is understood internally, see https://github.com/lowRISC/opentitan/blob/master/hw/ip/aes/pre_syn/tcl/yosys_run_synth.tcl#L30-L31

However, later experiments showed that especially for the FI countermeasures, the Yosys `keep` attribute isn't suitable. Instead the `keep_hierarchy` attribute needs to be placed on most `prim_` cells and then removed again before the final flattening step, see  https://github.com/lowRISC/opentitan/blob/master/hw/ip/aes/pre_syn/tcl/yosys_run_synth.tcl#L33-L39

This approach also works with the non-Xilinx but generic primitives.

vogelpi (approved):
From my side, this looks good. Thanks @ctopal for looking into this and preparing a PR.

vogelpi:
Ah, I just noticed that @rswarbrick withdrew his original approval. Please don't merge this without also getting Rupert's approval.

PR #1675, opened 2022-06-03
Update google_riscv-dv to google/riscv-dv@0b2b3d6

vogelpi:
This update is required to enable ePMP verification in Ibex.


andreaskurth (approved):
LGTM, thanks!

ctopal (approved):

Issue #1117, opened 2020-09-24
[rtl] Double fault handling


GregAC:
Consider the case where an exception occurs on the first instruction of an exception handler (e.g. due to a bus fault or PMP access fault), what should happen?

I suspect Ibex currently enters some kind of infinite loop or deadlocks.

We need to define and document a sensible behaviour here.

I would suggest some kind of double fault detection that causes the core to halt and assert a top-level signal e.g. `double_fault_o`. System implementers are then left with the choice of what to do (e.g. immediately reset the core). I don't think Ibex should perform a reset itself, simply halt and let the system decide what to do.

rswarbrick:
Cores I've worked with in the past just infinite loop. Arguably a reasonable behaviour ("don't set up your software like that!")

lenary:
@rswarbrick I want to make sure we don't do anything too silly, but thinking about it, in a double-fault scenario, the watchdog probably would take over, because the core is infinitely looping. That said, a signal probably is a nicer interface than just being entirely unresponsive until the end of time, and having to rely on a watchdog.

rswarbrick:
I think we probably need to figure out what we're doing here. As @lenary said, at the moment we presumably just loop on the address until killed by the watchdog. Seems like a perfectly reasonable behaviour to me.

@imphil, @GregAC, @tomroberts-lowrisc: Do we need to think about this further, or can we just consider this closed (possibly with a bit of explicit documentation)?

GregAC:
We should see if there's any requirements on the OT side but simply documenting the existing behaviour (i.e. infinite loop) should suffice

GregAC:
OT may want some form of double fault detection. We need to define exactly what we consider a double fault (e.g. if you see a synchronous exception in an interrupt handler you may not want to allow the synchronous exception handler to run and wouldn't signal a double fault, but you would for a synchronous exception whilst handling a synchronous exception).

I propose we should build the detector as an independent piece of logic that has no effect on Ibex internally (i.e. no halting on double fault as I initially proposed above). Users of Ibex can observe the double fault and choose to halt the clock, reset the core, raise an NMI etc or ignore entirely, which will result in the logic getting optimised away.

tjaychen:
hey all,
From my perspective (@cfrantz @moidx please weigh in), we do not plan any kind of graceful recovery from double fault. 
[...]

PR #1680, opened 2022-06-16
Update crash dump to contain mtval

tjaychen:
- mtval is a bit more useful for double fault situations
  as on the second exception we can still "remember" the
  data address and PC of the first exception.

see https://github.com/lowRISC/ibex/issues/1117


vogelpi (approved):
LGTM, thanks @tjaychen !

PR #1688, opened 2022-06-30
Commenting UVM testbench code, tidy formatting, minor refactoring

hcallahan-lowrisc:
- Adds comments for explanations of test and library functionality
- Refactor types and naming of control knob signals for clarity
- Move constraints from MEMBER to CLASS for more flexibility
- Add missing license header

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
@@ -1,72 +1,87 @@
-typedef enum bit [1:0] {
-    SingleRun, // Singl iteration
-    InfiniteRuns, // Run forever until stop is specified
-    MultipleRuns, // Multiple runs with configurable or randomizable iteration count
-    InvalidRuns // Default state
-  } run_type_e;
-
-typedef enum bit [1:0] {
-  IsideErr, // Inject error in instruction side memory.
-  DsideErr, // Inject error in data side memory.
-  PickErr // Pick which memory to inject error in.
-  } error_type_e;
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
 
 class core_base_new_seq #(type REQ = uvm_sequence_item) extends uvm_sequence #(REQ);
-  // Default set to 50% for zero delay to be picked. Set to 100 if zero delay is required always.
-  int unsigned                     zero_delay_pct = 50;
-  rand bit                         zero_delays;
-  rand int                         delay = 0;
-  int unsigned                     iteration_cnt = 0;
-  int unsigned                     max_delay = 500;
-  virtual clk_rst_if               clk_vif;
+
+  // Virtual interfaces for driving stimulus directly
+  virtual             clk_rst_if   clk_vif;
+  virtual core_ibex_dut_probe_if   dut_vif;
+
   bit                              stop_seq;
   bit                              seq_finished;
-  rand run_type_e                  num_of_iterations = InvalidRuns;
-  virtual core_ibex_dut_probe_if   dut_vif;
-  // Use this bit to start any unique sequence once
-  bit                              start_seq = 0;
+
+  rand bit                         zero_delays;
+  // CONTROL_KNOB: Randomly override the delay between stimulus items to be zero
+  // Default set to 50% for zero delay to be picked. Set to 100 if zero delay is required always.
+  int unsigned                     zero_delay_pct = 50;
+  constraint zero_delays_c {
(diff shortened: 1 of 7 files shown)


ctopal (commented):
LGTM, thanks for the clean-up work!

I found `num_of_iterations` meaning something different in `new_seq_lib` and `seq_lib` is really confusing. I think either changing `new_seq_lib` one to something like `iteration_modes` would do the trick, WDYT? Still we will be implementing the same idea differently in both cases but at least some variable names (that does different thing in different seq_lib) wouldn't be the same that way.

hcallahan-lowrisc:
> LGTM, thanks for the clean-up work!
> 
> I found `num_of_iterations` meaning something different in `new_seq_lib` and `seq_lib` is really confusing. I think either changing `new_seq_lib` one to something like `iteration_modes` would do the trick, WDYT? Still we will be implementing the same idea differently in both cases but at least some variable names (that does different thing in different seq_lib) wouldn't be the same that way.

@ctopal Good suggestion. Change pushed!

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -1,72 +1,86 @@
...
-        drive_stimulus(delay);
+        drive_stimulus();
       end
       MultipleRuns: begin
+        // Randomize iteration_cnt if not specified externally by test
         if(iteration_cnt == 0) begin
           `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(iteration_cnt,
-                                                iteration_cnt > 0 && iteration_cnt <= 20; )
+            // with {"CONSTRAINTS"}
+            iteration_cnt > 0;)
```
ctopal, line 74: Sorry I missed this one with the previous review but it looks like in here we are randomizing without an upper bound. I think we're never going to end up in this piece of code (because we are correctly randomizing `iteration_cnt` to become greater than 0 at `pre_body`). So I think either something like this below or an assertion to check iteration_cnt is not 0 would be better here.

```systemverilog
`DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(iteration_cnt,
  // with {"CONSTRAINTS"}
  iteration_cnt inside {[0:iteration_cnt_max]};)
```

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -1,72 +1,86 @@
...
-    if(num_of_iterations == InvalidRuns) begin
-      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(num_of_iterations, num_of_iterations != InvalidRuns;)
-    end
-    `uvm_info(`gfn, $sformatf("Doing %s", num_of_iterations.name()), UVM_LOW)
-    case (num_of_iterations)
+    // core_base_new_seq::body() provides a flexible sequence scheduler, where
+    // 'iteration_modes' allows different tests to change the frequency
+    // stimulus is generated.
+
+    `uvm_info(`gfn, $sformatf("Running the \"%s\" schedule for stimulus generation", iteration_modes.name()), UVM_LOW)
```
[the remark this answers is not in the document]
ctopal: Another stray lint error here

ctopal (commented):
Thanks for the change in naming, I'm less confused now! :smile: Added a suggestion about `iteration_cnt` related randomization, also I think some lint warnings went unnoticed, tried to point them out.

GregAC (approved):
LGTM with one small tweak

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
```diff
@@ -1,72 +1,83 @@
...
-        drive_stimulus(delay);
+        drive_stimulus();
       end
       MultipleRuns: begin
-        if(iteration_cnt == 0) begin
-          `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(iteration_cnt,
-                                                iteration_cnt > 0 && iteration_cnt <= 20; )
-        end
+        `ASSERT(IterationCntNotZero, iteration_cnt != 0)
+        `DV_CHECK_MEMBER_RANDOMIZE_FATAL(iteration_cnt)
```
ctopal, line 72: Minor nit: I believe we don't need this since we are already randomising `iteration_cnt` in `pre_body`.

ctopal (approved):
LGTM with a minor nit. Thank you for the work @hcallahan-lowrisc!

dv/uvm/core_ibex/tests/core_ibex_new_seq_lib.sv
ctopal: Also, compile error might be from using `ASSERT` here. Can you try it with `DV_CHECK_FATAL` please?

Issue #1570, opened 2022-03-17
Teach auto-generated interrupt handlers to clear interrupts again


rswarbrick:

johngt:
@GregAC - checking if this is still relevant and if so, to whom we should assign this.

GregAC:
Still relevant but something to do for V3 and it's already assigned to that milestone

Issue #1568, opened 2022-03-16
[test] Add tests to mix up exceptions/interrupts/debug


rswarbrick:

johngt:
With reference to https://ibex-core.readthedocs.io/en/latest/03_reference/verification.html and cosim documentation found there the point here is to create new tests that can now be covered with the cosim tool that were previously not possible.

Quick discussion with @hcallahan-lowrisc identified the following pre-cursor stages.
1. Learn about the test framework
2. Learn about the Ibex architecture
3. Learn the CI framework

@hcallahan-lowrisc to discuss with @ctopal the coverage figures that are generated by CI tools to understand if tests are covering new areas and therefore useful to include in the regression tests.

44 existing tests (grep -ir " test:" ./testlist.yaml | wc -l) in [testlist.yaml](https://github.com/johngt/ibex/blob/master/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml) but want to create new tests.
Two new subissues will be created:
1. Create a single test, with increased coverage, test that it works locally and then commit to CI regression engine. Then get PR accepted.
2. Multiple other tests that increase the test coverage.

Aim is to increase total coverage of Ibex DV to sufficient level. Will update the issue with the coverage metrics that have been agreed upon.

PR #1689, opened 2022-06-30
Create interrupts/debug/exceptions test

hcallahan-lowrisc:
(Ignore first commit, it is #1688 and needs merging before this PR.)

Create a test that uses riscv-dv to create an exception-causing instruction stream, while using the UVM testbench to create external interrupt and debug request stimulus.

The interrupt and debug stimulus are driven as pulses with randomised interval and duration. 
This test would benefit from a way for the interrupts to be de-asserted in response to the ISR's being executed. (See #1570)

Goes towards #1568.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -1424,3 +1424,64 @@ class core_ibex_fetch_en_chk_test extends core_ibex_directed_test;
   endtask
 
 endclass
+
+// Stimulate a combination of traps/debug requests
+// - exceptions are inserted through the instruction generator cfg (testlist.yaml)
+// - interrupts/debug requests are inserted through testbench stimulus
+class core_ibex_assorted_traps_interrupts_debug_test extends core_ibex_directed_test;
+
+   debug_new_seq debug_new_seq_h;
+   irq_new_seq irq_new_seq_h;
+
+   `uvm_component_utils(core_ibex_assorted_traps_interrupts_debug_test)
+   `uvm_component_new
+
+   virtual task send_stimulus();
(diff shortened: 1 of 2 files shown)


dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1420,3 +1420,64 @@ class core_ibex_fetch_en_chk_test extends core_ibex_directed_test;
...
+
+     irq_new_seq_h = irq_new_seq::type_id::create("irq_new_seq_h", this);
+     debug_new_seq_h = debug_new_seq::type_id::create("debug_new_seq_h", this);
+
+     irq_new_seq_h.iteration_modes = InfiniteRuns;
+     irq_new_seq_h.stimulus_delay_cycles_min = 500; // Interval between requests
+     irq_new_seq_h.stimulus_delay_cycles_max = 2000;
+     irq_new_seq_h.zero_delay_pct = 10;
+     debug_new_seq_h.iteration_modes = MultipleRuns;
+     debug_new_seq_h.iteration_cnt = 10; // Limit this or the test will never end. (end = ecall)
```
ctopal, line 1446: Is this a hard limit or can this be also randomised (e.g. `iteration_cnt_max` can be set to 10 maybe?)

ctopal (commented):
LGTM overall, just a minor nit for the `iteration_cnt` thing. Let's get the first PR merged and I'll approve this after that. Thanks a lot for the work!

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
hcallahan-lowrisc: Yep you're right, this should have been `iteration_cnt_max` in the first place. Thanks for the review!

ctopal (approved):

PR #1695, opened 2022-07-08
[dv] Write ePMP tests and enable ePMP in DV

marnovandermaas:
- Vendor in a fix `gen_pmp_instr` to:
   * Set address of region 0 before setting MSECCFG
   * If  MML, initially set MSECCFG to MML=0, MMWP=0, RLB=1
   * If  MML, set the config of region 0 to LXWR=1100 and TOR
   * If MMWP, set the config of region 0 to LXWR=0100 and TOR
   * If MML or MMWP, set requested MSECCFG with RLB hardcoded to 1
   * Don't override region 0 config if `+pmp_region_0` is passed
   * Set default region 0 config in MML mode to shared execute
   * Set all other addresses and configs
   * Set requested MSECCFG (including RLB)
- Enable ePMP in `riscv_core_settings.sv`
- Disable PMP and ePMP for "small" and "experimental-branch-predictor" configurations
- Bump CI and Spike version in `ci/vars.yml`
- Create tests for machine mode lockdown (MML)
    * Code is execute only, the rest is read/write
    * All regions execute only
    * All regions read only
- Create test for machine mode whitelist policy (MMWP)
- Create test for rule lock bypass (RLB)
- Add `README` for ibex DV

Closes: https://github.com/lowRISC/ibex/issues/1579


dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -716,6 +719,64 @@
     #  rtl_params:
     #    PMPEnable: 1
 
+# TODO hit some interesting Smepmp configurations while MML is enabled such as:
```
vogelpi, line 722: Note: if you don't provide the `+mseccfg` Argument and set `pmp_randomize` to 1 instead, RISCV-DV will randomize the `Smepmp` configuration on a per-test basis. I am thus not sure if having three separate tests here is necessary but I also don't have a clear view what the best way would be.

vogelpi (commented):
Thanks for the PR @marnovandermaas . This looks reasonable but somehow CI seems to fail for the small and the branch-predictor configuration. This probably needs to be investigated.

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
marnovandermaas: Thanks Pirmin. Yes, I see that there is a `riscv_pmp_full_random_test` with `pmp_randomize` enabled. It was commented out, but I'll enable it and see what happens. I still think it is useful to have a few tests that enable specific MML, MMWP and RLB configurations, especially for configurations that are unlikely to show up with random testing.

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -693,28 +695,163 @@
...
+  rtl_test: core_ibex_base_test
+  rtl_params:
+    PMPEnable: 1
+
+- test: riscv_epmp_mml_execute_only_test
+  desc: >
+    An enhanced PMP machine mode lockdown test - all PMP regions are set to
+    execute only. Exception is expected on any store or load.
+    Randomize mstatus.mprv.
+  iterations: 10
```
GregAC, line 794: Fewer iterations as above

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -693,28 +695,163 @@
...
+  rtl_test: core_ibex_base_test
+  rtl_params:
+    PMPEnable: 1
+
+- test: riscv_epmp_mmwp_test
+  desc: >
+    An enhanced PMP machine mode whitelist policy - all PMP regions will be configured to default setting,
+    enabling all forms of accesses, expect that an exception when machine mode access memory not in PMP.
+    Randomize mstatus.mprv.
+  iterations: 10
```
GregAC, line 827: Fewer iterations as above

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -693,28 +695,163 @@
...
+  rtl_test: core_ibex_base_test
+  rtl_params:
+    PMPEnable: 1
+
+- test: riscv_epmp_rlb_test
+  desc: >
+    An enhanced PMP rule lock bypass - all PMP regions will be configured to default setting,
+    enabling all forms of accesses, expect that no exception will be thrown.
+    Randomize mstatus.mprv.
+  iterations: 10
```
GregAC, line 844: Fewer iterations as above

GregAC (approved):
LGTM with some tweaks, thanks @marnovandermaas

dv/uvm/core_ibex/README.md
```diff
@@ -0,0 +1,26 @@
+# DV for the ibex core
+
+This design verification setup is supposed to be run on lowRISC's private CI, but this document should give you an idea of how to get it running on your local machine.
+
+## Prerequisites
+You need to have Excelium available on your machine.
+You can check whether you have it available by running: `xrun --verison`
+Internally at lowRISC we load this module by running: `module load cadence/xcelium/latest`
```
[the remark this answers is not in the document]
marnovandermaas: Agreed, should be gone now

dv/uvm/core_ibex/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+You also need Spike to be able to compare to in the cosimulation.
+We use a lowRISC specific Spike which you can find [on its own GitHub page](https://github.com/lowRISC/riscv-isa-sim/tree/ibex_cosim).
+Internally at lowRISC we load this module by running: `module load spike-ibex/latest`
```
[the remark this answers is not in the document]
marnovandermaas: Ack

dv/uvm/core_ibex/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+You will need the [RISC-V toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain).
+You'll need to add this to your path and then also set the following environment variables:
+```bash
+export RISCV_GCC=riscv32-unknown-elf-gcc
+export RISCV_OBJCOPY=riscv32-unknown-elf-objcopy
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, I've added that information now.

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -693,28 +695,163 @@
...
+  rtl_params:
+    PMPEnable: 1
+
+- test: riscv_epmp_mml_test
+  desc: >
+    An enhanced PMP machine mode lockdown test - main region is set to execute
+    only mode and all other regions are set to read/write only. Exceptions when
+    reading/writing code or executing data. No behavior difference between M, S
+    and U expected. Randomize mstatus.mprv.
+  iterations: 10
```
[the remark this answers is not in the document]
marnovandermaas: Agreed, should be fixed now

dv/uvm/core_ibex/README.md
```diff
@@ -0,0 +1,26 @@
+# DV for the ibex core
+
+This design verification setup is supposed to be run on lowRISC's private CI, but this document should give you an idea of how to get it running on your local machine.
```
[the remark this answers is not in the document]
marnovandermaas: Thanks for the pointer, I've added a link to that doc now.

marnovandermaas:
Two things still need to be done for this PR:
- Make a PR to RISCV-DV for the changes necessary in `gen_pmp_instr`
- Investigate why the "small" configuration is failing in CI

marnovandermaas:
I made a pull request for RISCV-DV: https://github.com/google/riscv-dv/pull/873
That PR needs to be merged and vendored into this PR.

marnovandermaas:
I just added a fix that disables PMP and Smepmp in the `riscv_core_setting.sv` file for the small and experimental-branch-predictor configurations. I admit my fix is a bit hacky, but it saves us from having to make a change to RISCV-DV, which when Smepmp is enabled always generates a write to the `MSECCFG` register during initialization. Let me know if anyone sees a better way of doing this.

hcallahan-lowrisc:
@marnovandermaas I believe that there is a mechanism already in place to filter out tests which don't match certain ibex configs. 
If you look [here](https://github.com/lowRISC/ibex/blob/42d92c7c9bb96a63cdaec78cd40f1367d2eadbd3/dv/uvm/core_ibex/list_tests.py#L32) I believe this should be filtering out tests with parameters that don't match the ibex_config? 
So in the testlist.yaml file, each test can specify `rtl_params` which should be matched against the parameters in ibex_configs.yaml?

`testlist.yaml`
```yaml 
- test: riscv_pmp_basic_test
  ........
  rtl_params:
    PMPEnable: 1
```
If tests are not being filtered out then this may be broken. I haven't looked to closely though.
Also this doesn't address the broken experimental configs, though I'm not sure we necessarily want to exclude those.

marnovandermaas:
> @marnovandermaas I believe that there is a mechanism already in place to filter out tests which don't match certain ibex configs. If you look [here](https://github.com/lowRISC/ibex/blob/42d92c7c9bb96a63cdaec78cd40f1367d2eadbd3/dv/uvm/core_ibex/list_tests.py#L32) I believe this should be filtering out tests with parameters that don't match the ibex_config? So in the testlist.yaml file, each test can specify `rtl_params` which should be matched against the parameters in ibex_configs.yaml?
> 
> `testlist.yaml`
> 
> ```yaml
> - test: riscv_pmp_basic_test
>   ........
>   rtl_params:
>     PMPEnable: 1
> ```
> 
> If tests are not being filtered out then this may be broken. I haven't looked to closely though. Also this doesn't address the broken experimental configs, though I'm not sure we necessarily want to exclude those.

@hcallahan-lowrisc The test filtering is not broken, the problem is that when `support_epmp` is enabled, a write to `MSECCFG` is generated for all tests by the following line in RISCV-DV: https://github.com/lowRISC/ibex/blob/42d92c7c9bb96a63cdaec78cd40f1367d2eadbd3/vendor/google_riscv-dv/src/riscv_pmp_cfg.sv#L304
As far as I can tell, we don't have access to the Ibex configuration at that point, or do we? Additionally, I thought it was better to solve this problem on our end rather than changing vendored in code.

hcallahan-lowrisc:
> > a write to `MSECCFG` is generated for all tests
> 
> @marnovandermaas I think the idea is that the 'small' configuration should be filtered out by the function I linked to, so it doesn't matter that the MSECCFG write is always present, because we never actually run a test with that instruction on a build of the cpu that would fail? Or maybe I'm misunderstanding?

@marnovandermaas Yes you are right, this is an issue. As discussed, for the set of tests which don't specify anything about requiring/enabling PMP using `rtl_params` (eg. the three current smoketests), they would still contain MSECCFG write instructions and would not be filtered out for the 'small' config.
After putting our heads together we couldn't come up with an obvious solution, so follow up next week. TBD

GregAC:
A write to `MSECCFG` should be fine even if you don't have PMP going by the v1.12 privileged spec, however they way we've implemented it causes an illegal instruction error (my doing).

It's a bit tricky as v1.11, the version we say we support, does not have the `MSECCFG` CSR so the current behaviour is probably the most correct we can get until we move to v1.12

vendor/google_riscv-dv/src/riscv_pmp_cfg.sv
```diff
@@ -297,37 +297,75 @@ class riscv_pmp_cfg extends uvm_object;
...
-        cfg_byte = {1'b0, pmp_cfg[i].zero, TOR, 1'b1, 1'b1, 1'b1};
+      // In case an argument is give, we want to allow the test writer to change region 0 to
+      // anything they like even if it means that a trap occurs on the next instruction. If no
+      // argument is given, we should set region 0 to have a configuration with execute enabled.
+      if (i == 0 && !inst.get_arg_value("+pmp_region_0=", arg_value)) begin
+        if(riscv_instr_pkg::support_epmp && mseccfg.mml) begin
+          // This value is different from above (M-mode execute only) because we need region 0 to
+          // be executable in both M-mode and U-mode, since RISCV-DV switches privledge before
+          // <main> but after <pmp_setup>. We choose not to use the shared code region that also
+          // allows read in M-mode because that is inconsitente with the execute-only in other
```
ctopal, line 349: nit: inconsistent

vendor/google_riscv-dv/src/riscv_pmp_cfg.sv
```diff
@@ -297,37 +297,75 @@ class riscv_pmp_cfg extends uvm_object;
...
+          // everything before configuration is done.
+          `uvm_info(`gfn, $sformatf("MSECCFG: MML 0, MMWP 0, RLB 1"), UVM_LOW)
+          cfg_byte = {1'b1, 1'b0, 1'b0};
+          instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
+          // This configuration needs to be executable in M-mode both before and after writing to
+          // MSECCFG. It will deny execution for U-Mode, but this is necessary because RWX=111 in
+          // MML means read only, and RW=01 is not allowed before MML is enabled.
+          cfg_byte = {1'b1, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
+        end else begin
+          // We must set pmp region 0 to executable before enabling MMWP. RW=00 to be consistend
```
ctopal, line 323: nit: consistent

dv/uvm/core_ibex/README.md
```diff
@@ -0,0 +1,35 @@
+# DV for the ibex core
+
+For detailed documention on how Ibex's verification works, please have a look at [the dedicated documentation page](https://ibex-core.readthedocs.io/en/latest/03_reference/verification.html).
+This README provides a quick start guide to get things running.
+
+## Prerequisites
+You need to have Excelium available on your machine.
```
ctopal, line 7: I think this one didn't change :grin:

ctopal (approved):
LGTM with really minor nits.

Additionally, can you please remove the part in `riscv_core_setting.sv.default:49` about `FENCE.I`, it's already removed in master branch but I think it might not ask for a rebase. Also, do you think it would make sense if you add `L:1` to all pmp regions for `riscv_pmp_disable_all_regions_test` in `testlist.yaml`? (not that it would change the behaviour but I think if we are disabling all regions we also probably want to lock them, WDYT?)

marnovandermaas:
> LGTM with really minor nits.
> 
> Additionally, can you please remove the part in `riscv_core_setting.sv.default:49` about `FENCE.I`, it's already removed in master branch but I think it might not ask for a rebase. Also, do you think it would make sense if you add `L:1` to all pmp regions for `riscv_pmp_disable_all_regions_test` in `testlist.yaml`? (not that it would change the behaviour but I think if we are disabling all regions we also probably want to lock them, WDYT?

Git seems smart enough and removed the `FENCE.I` automatically. Also, I've added the lock bits to the `riscv_pmp_disable_all_regions_test` as you suggest.

dv/uvm/core_ibex/README.md
marnovandermaas: Thanks, well spotted!

GregAC (approved):
This is looks good to go, thanks @marnovandermaas

PR #1698, opened 2022-07-11
[lint] Remove whitespace from existing source files

marnovandermaas:


marnovandermaas:
@hcallahan-lowrisc told me that it is not desirable to change code in the `vendor` directory, especially for something trivial like whitespace. I should change this pull request to only make whitespace corrections to files that are not vendored in.

hcallahan-lowrisc:
For things like whitespace, if you feel strongly there is always the patching mechanism (see ibex/vendor/google_riscv-dv.vendor.hjson and /ibex/vendor/patches/google_riscv-dv/)

marnovandermaas:
Ok, I have now changed this pull request to only remove the trailing white-space from the source code that is not vendored in.

hcallahan-lowrisc (approved):
Thanks for the cleanup, always welcome! It would be great to setup some more linting / pre-commit checks for this stuff at some point.

marnovandermaas:
> Thanks for the cleanup, always welcome! It would be great to setup some more linting / pre-commit checks for this stuff at some point.

Yes, agreed. I'll have a look at that when I have time.

PR #1699, opened 2022-07-11
[dv] Check privilege after DRET

### LINKED ISSUE #1691: [dv] riscv_dret_test fails privilege mode check
## Observed Behavior

UVM fail with the error message below: 

`Check failed dut_vif.dut_cb.priv_mode == mode (0 [0x0] vs 3 [0x3]) Incorrect privilege mode`

## Expected Behavior

I think this happened because `dret` actually changes our privilege level and if our privilege level is `U` before entering debug, we'll set the privilege as so after receiving `dret`. So I believe the problem comes from us using `check_illegal_insn` task after we have a `dret`.

Run this command while at the repository top.
``` make -C dv/uvm/core_ibex TEST=riscv_dret_test ITERATIONS=1 SIMULATOR=xlm -j2 SEED=22633 ```

ctopal:
Timing fix for dret_test and modelling controller behaviour for `FLUSH` transition.

Resolves #1691

### DIFF
--- a/dv/uvm/core_ibex/env/core_ibex_dut_probe_if.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_dut_probe_if.sv
@@ -18,6 +18,7 @@ interface core_ibex_dut_probe_if(input logic clk);
   logic                    alert_major_bus;
   logic                    debug_req;
   ibex_pkg::priv_lvl_e     priv_mode;
+  ibex_pkg::ctrl_fsm_e     ctrl_fsm_cs;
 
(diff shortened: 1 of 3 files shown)


GregAC (approved):
Thanks for the fix @ctopal 

Please do note the checking in this test which is causing the failure is the kind of thing we're looking to rip out (as the cosim checking covers it all). For future failures like this please try to avoid spending lots of time fixing them as we'll eventually ditch the buggy code. However quick fixes to give us a healthier regression are still useful.

PR #1700, opened 2022-07-12
[dv,fcov] Implement Misaligned Mem Error coverage

### LINKED ISSUE #1697: [dv,fcov] Implement "Memory Fetch Error" coverage point
I believe this ( [unaligned memory access](https://github.com/lowRISC/ibex/blob/master/doc/03_reference/coverage_plan.rst#exceptionsinterruptsdebug) ) is the sole remaining coverpoint to implement, rather similar to `misaligned_lsu_access_cross` but instead of PMP errors we're looking at memory fetch errors.

_Originally posted by @GregAC in https://github.com/lowRISC/ibex/pull/1696#discussion_r916957719_

ctopal:
Adds some signal to the load store unit to catch when we have the
fetch error signals from both first and second part of the misaligned
load/store access cases.

Resolves #1697

### DIFF
--- a/doc/03_reference/coverage_plan.rst
+++ b/doc/03_reference/coverage_plan.rst
@@ -210,8 +210,8 @@ Furthermore they can all occur together and must be appropriately prioritised (c
 * ``pmp_dside_mode_cross`` - Exception from memory access PMP violation.
 * Unaligned memory access
 
-    * Cover all error and no error scenarios for memory fetch error; first access saw error, second
-      access saw error, neither access saw error
+  * ``misaligned_insn_bus_err_cross``, ``misaligned_data_bus_err_cross`` - Cover all error and no error scenarios for memory fetch error; first access saw error, second
+    access saw error, neither access saw error
 
 * Interrupt raised/taken.
 
--- a/rtl/ibex_load_store_unit.sv
+++ b/rtl/ibex_load_store_unit.sv
@@ -555,6 +555,40 @@ module ibex_load_store_unit #(
   //////////
   // FCOV //
   //////////
+`ifndef DV_FCOV_DISABLE
+  // Set when awaiting the response for the second half of a misaligned access
+  logic fcov_mis_2_en_d, fcov_mis_2_en_q;
+
+  // fcov_mis_rvalid_1: Set when the response is received to the first half of a misaligned access,
+  // fcov_mis_rvalid_2: Set when response is received for the second half
+  logic fcov_mis_rvalid_1, fcov_mis_rvalid_2;
+
+  // Set when the first half of a misaligned access saw a bus errror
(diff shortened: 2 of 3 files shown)


rtl/ibex_load_store_unit.sv
```diff
@@ -555,6 +555,39 @@ module ibex_load_store_unit #(
   //////////
   // FCOV //
   //////////
+`ifndef DV_FCOV_DISABLE
+  // To differentiate between the second request of misaligned access and an aligned access.
```
GregAC, line 559: An aligned access only has a single request this may imply it could have two. I also think you should reword these to be an explicit, set when ... e.g. here:

> Set when awaiting the response for the second half of a misaligned access

rtl/ibex_load_store_unit.sv
```diff
@@ -555,6 +555,39 @@ module ibex_load_store_unit #(
...
+  logic fcov_mis_2_en_d, fcov_mis_2_en_q;
+
+  // To catch data phases of misaligned acceses.
+  logic fcov_mis_rvalid_1, fcov_mis_rvalid_2;
```
GregAC, line 563: Following the above, reword to something like

> `fcov_mis_rvalid_1` set when the response is received to the first half of a misaligned access, `fcov_mis_rvalid_2` set when response is received for the second half

rtl/ibex_load_store_unit.sv
```diff
@@ -555,6 +555,39 @@ module ibex_load_store_unit #(
...
+
+  // To catch and hold bus error of the first request
+  logic fcov_mis_bus_err_1_d, fcov_mis_bus_err_1_q;
```
GregAC, line 566: Following the above, reword to something like

> Set when the first half of a misaligned access saw a bus errror

doc/03_reference/coverage_plan.rst
```diff
@@ -352,3 +352,5 @@ There must be a documented reason a particular bin is added to the illegal or ig
 * ``pmp_iside_priv_bits_cross``, ``pmp_iside2_priv_bits_cross``, ``pmp_dside_priv_bits_cross``, PMP regions x permissions x access fail/pass x privilege level
 
   * Three crosses, one for each PMP channel (instruction, instruction 2 and data).
+
+* ``misaligned_insn_bus_err_cross``, ``misaligned_data_bus_err_cross`` - Implemented to check if we see errors from memory interfaces (two different crosses for data and instruction memory) in the case of a misaligned access.
```
GregAC, line 356: No need to add this here, consider the Cross Coverage section to be all those crosses that aren't already mentioned.

GregAC (approved):
Actual coverage looks good, some nits around documentation/comments

ctopal:
Thanks for the review, comments updated I will merge it when CI passes.

Issue #1692, opened 2022-07-05
[dv,cosim] Scoreboard expects 0 for some register writes


ctopal:
## Observed Behavior

`risv_perf_counter_test` fails with every seed. For one example see below:
```
 @ 124850: uvm_test_top.env.cosim_agent.scoreboard [ibex_cosim_agent_pkg::ibex_cosim_scoreboard.run_cosim_rvfi] Cosim mismatch Register write data mismatch to x14 DUT: 3b1 expected: 0
```

For `riscv_debug_branch_jump_test` we see this error in some of the seeds, please see the error message below:
```
@ 9203330: uvm_test_top.env.cosim_agent.scoreboard [ibex_cosim_agent_pkg::ibex_cosim_scoreboard.run_cosim_rvfi] Cosim mismatch Register write data mismatch to x24 DUT: 17f35 expected: 0
```
## Expected Behavior

I think either cosim misses the `wdata` for some reason or our writes to the registers happen in the hardware level (rather than in the software level so cosim couldn't model it. This is a completely wild guess, might make no sense at all! Definitely needs more investigating.
## Steps to reproduce the issue

Please run this command to see both of the failing tests in one go.
`make -C dv/uvm/core_ibex TEST=riscv_perf_counter_test,riscv_debug_branch_jump_test ITERATIONS=1 SIMULATOR=xlm -j2 SEED=22630`

> estimate 2

PR #1706, opened 2022-07-14
[cosim,dv] Add support to set mcount registers

ctopal:
Counts on https://github.com/lowRISC/riscv-isa-sim/pull/5 to work.

Extends RVFI connections further to include mhpmcounterX registers that are implemented to count something in `ibex_cs_registers.sv` Total number of them are 10 (and +10 for their  mhpmcounterhX counterpart).
Sets them up before every cosim step to let Spike know their real values.

Resolves https://github.com/lowRISC/ibex/issues/1692


dv/cosim/cosim_dpi.h
```diff
@@ -19,6 +19,7 @@ void riscv_cosim_set_mip(Cosim *cosim, const svBitVecVal *mip);
 void riscv_cosim_set_nmi(Cosim *cosim, svBit nmi);
 void riscv_cosim_set_debug_req(Cosim *cosim, svBit debug_req);
 void riscv_cosim_set_mcycle(Cosim *cosim, svBitVecVal *mcycle);
+void riscv_cosim_set_csr(Cosim *cosim, int csr_id, svBitVecVal *csr_val);
```
dmcardle, line 22: I may be out of my depth, but is it possible for `csr_val` to be a const pointer? Or is that not expressible in the .svh file?

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_rvfi_seq_item.sv
```diff
@@ -13,6 +13,9 @@ class ibex_rvfi_seq_item extends uvm_sequence_item;
   bit        debug_req;
   bit [63:0] mcycle;
 
+  bit [31:0] mhpmcounters  [10];
+  bit [31:0] mhpmcountersh [10];
```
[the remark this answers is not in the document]
ctopal: The main idea in this PR is to communicate the values inside the counters to the ISS so that in the case of a test using those registers, ISS would know the proper values. I didn't see any easier way of doing that rather than using what is already being done (e.g. communicating `mcycle` info for the same reason). But yes, you are right this will extend RVFI packet to now have 20 counters (yikes!). Would love to hear if you have a practical way out for the problem.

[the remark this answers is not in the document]
marnovandermaas: @ctopal thanks for the discussion and explanation just now. Based on what we talked about, I think instead of synchronizing these counter values, it is probably better for us to configure RISCV-DV to avoid reading from these specific CSRs. Instead, we can create some directed tests to make sure these CSRs have desired properties like being monotonically increasing.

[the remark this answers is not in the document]
marnovandermaas: After a bit more discussion with @GregAC, it seems like it is desirable to synchronize these counter values in case reading from these CSRs causes some unforeseen effect later on.

marnovandermaas (approved):
LGTM, bar the lint errors

ctopal:
Cleaned Lint errors, implemented @dmcardle's suggestion (thanks!)

dv/cosim/spike_cosim.cc
```diff
@@ -366,6 +366,19 @@ void SpikeCosim::set_mcycle(uint64_t mcycle) {
   processor->get_state()->mcycle->write(mcycle + 1);
 }
 
+void SpikeCosim::set_csr(const int csr_num, const uint32_t new_val) {
+  // If the default value, 0, did not change no need to write to the register.
+  // Note that this is tested with ibex-cosim-v0.3 version of Spike. 'set_csr'
+  // method might have a hardwired zero for mhpmcounterX registers.
+  if (new_val > 0) {
```
GregAC, line 373: I think we're better off unconditionally writing. Assuming a value of 0 means no change is dangerous.

dv/cosim/spike_cosim.cc
```diff
@@ -366,6 +366,19 @@ void SpikeCosim::set_mcycle(uint64_t mcycle) {
...
+#ifdef OLD_SPIKE
+    processor->set_csr(csr_num, new_val + 1);
+#else
+    processor->put_csr(csr_num, new_val + 1);
```
GregAC, line 377: Why the +1?

dv/cosim/cosim.h
```diff
@@ -110,6 +110,12 @@ class Cosim {
   // A full 64-bit value is provided setting both the mcycle and mcycleh CSRs.
   virtual void set_mcycle(uint64_t mcycle) = 0;
 
+  // Set the value of a CSR. This is only used for MHPMCOUNTER3..12 registers
```
GregAC, line 113: I'd ditch the stuff explaining it's only used for performance counters and why that's needed as this is meant to be a generic interface. This function can set any CSR if it wants and maybe some future use of this interface will have a need to do it for other CSRs.

Perhaps add a more generic explanation, e.g. 'This is useful to keep CSRs in sync between the DUT and the model where their values cannot be easily modeled, e.g. performance counters'

GregAC (changes requested):
Looks good, though I'd like to understand the `new_val + 1` and ditched the conditional setting in `set_csr`.

dv/cosim/spike_cosim.cc
ctopal: It was because the performance counters in the spike side were the type `wide_counter_csr_t`. So, if we do a `unlogged_write` to them (which both `set_csr` and `put_csr` essentially do) they are getting decremented by one automatically. But you are right, handling it here does not make sense. I'll come up with a better way of doing it inside spike.

ctopal:
Thanks for the feedback and review @GregAC, I agree that solution was not appropriate. I opened up a new PR in our spike fork so this PR now depends on https://github.com/lowRISC/riscv-isa-sim/pull/7. Although that way I got rid of if clause and `+1` hack in this PR.

GregAC (approved):

PR #1713, opened 2022-07-19
[dv] Fix traps in simple system cosim

GregAC:
Previously any traps seen on RVFI were skipped over. This was old
behaviour. With the latest cosim setup traps must be passed to the
`step` function.

Note the UVM environment does do this correctly so this bug isn't related to the cosim exception issues people have been seeing.


ctopal (approved):
LGTM, thanks

PR #1714, opened 2022-07-19
Pmp exception test

GregAC:
As I'd already written it thought it may be worth adding as a test for CI.

I've actually got a far more comprehensive PMP smoke test (and Silvestrs also had something he was working on before he left https://github.com/lowRISC/ibex/pull/1524). I think some form of that would be useful to add but more work to be done.


GregAC:
I think I want to rename this 'pmp_smoke_test' rather than 'pmp_exception_test' but that's a job for tomorrow.

marnovandermaas (approved):
I confirmed the value of `TEST_PMPCFG0`, and the changes to the YAML file look good to me.

PR #1715, opened 2022-07-20
[dv,fcov] IRQ and NMI detection fixes for fcov

ctopal:
This commit makes sure that different IRQs are collected in a single
`fcov_irqs` bus correctly. Also changes nmi_taken coverpoint to catch
interrupt taken case for NMI types from the same bus.

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -327,12 +327,14 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
                   prev_store_addr == curr_data_addr;
 
   // Collect all the interrupts for collecting them in different bins.
-  logic [5:0] fcov_irqs = {id_stage_i.controller_i.irq_nm_ext_i,
-                           id_stage_i.controller_i.irq_nm_int,
-                           (|id_stage_i.controller_i.irqs_i.irq_fast),
-                           id_stage_i.controller_i.irqs_i.irq_external,
-                           id_stage_i.controller_i.irqs_i.irq_software,
-                           id_stage_i.controller_i.irqs_i.irq_timer};
+  logic [5:0] fcov_irqs;
+
+  assign fcov_irqs = {id_stage_i.controller_i.irq_nm_ext_i,
+                      id_stage_i.controller_i.irq_nm_int,
+                      (|id_stage_i.controller_i.irqs_i.irq_fast),
+                      id_stage_i.controller_i.irqs_i.irq_external,
+                      id_stage_i.controller_i.irqs_i.irq_software,
+                      id_stage_i.controller_i.irqs_i.irq_timer};
 
   logic            instr_unstalled;
   logic            instr_unstalled_last;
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -327,12 +327,14 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
-                           id_stage_i.controller_i.irqs_i.irq_software,
-                           id_stage_i.controller_i.irqs_i.irq_timer};
+  logic [5:0] fcov_irqs;
+
+  assign fcov_irqs = {id_stage_i.controller_i.irq_nm_ext_i,
+                      id_stage_i.controller_i.irq_nm_int,
+                      (|id_stage_i.controller_i.irqs_i.irq_fast),
+                      id_stage_i.controller_i.irqs_i.irq_external,
+                      id_stage_i.controller_i.irqs_i.irq_software,
+                      id_stage_i.controller_i.irqs_i.irq_timer};
```
marnovandermaas, line 337: Do I understand correctly that this has the same behavior as before, just that you split out the declaration and the assigning?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -484,8 +486,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     `DV_FCOV_EXPR_SEEN(insn_trigger_exception, instr_id_matches_trigger_q &&
                                                id_stage_i.controller_i.fcov_pipe_flush)
 
-    cp_nmi_taken: coverpoint (id_stage_i.controller_i.nmi_mode_d &
-                              (~id_stage_i.controller_i.nmi_mode_q)) iff
+    cp_nmi_taken: coverpoint ((fcov_irqs[5] || fcov_irqs[4])) iff
```
marnovandermaas, line 489: Why have you changed `nmi_mode_d & nmi_mode_q` to `irqs_i.irq_software || irqs_i.irq_timer`?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: Behaviour is not the same for type `logic` it seems. I also didn't notice it before, please have a look at https://verificationacademy.com/forums/systemverilog/why-cant-we-declare-and-assign-logic-type-one-step#answer-96794

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: No `fcov_irqs[5] =  id_stage_i.controller_i.irq_nm_ext_i` and `fcov_irqs[4] = id_stage_i.controller_i.irq_nm_int`. Please have a look at SystemVerilog LRM at pg 133:
```
When unpacked data are converted to the packed
representation, the order of the data in the packed representation is such that the first field in the structure
occupies the MSBs. The effect is the same as a concatenation of the data items (struct fields or array
elements) in orde
```

 The reason behind not using `nmi_mode_d` and `nmi_mode_q` is that they are updated one cycle after `id_stage_i.controller_i.fcov_interrupt_taken` is high which means we are not being able to hit this bin at all. Since `fcov_interrupt_taken` guarantees that we are having an interrupt and `NMI` types are already have the most priority, this change insures that the functionality that w are trying to catch is the same.

hcallahan-lowrisc (approved):
Makes sense! LGTM

PR #1716, opened 2022-07-21
Rework SpikeCosim::step() to handle exception on 1st ISR instr

### LINKED ISSUE #1687: [dv, cosim] Add support for exceptions on first instruction of interrupt
When an interrupt occurs that causes an exception on it's first instruction co-sim break. This is a missing feature that needs implementing, see the TODO here

https://github.com/lowRISC/ibex/blob/5c49fad9a236351a884f55d44248e4f9a0f7c7e5/dv/cosim/spike_cosim.cc#L145

> estimate 8

hcallahan-lowrisc:
Seperate concerns so the flow of stepping spike and checking against
the ibex RVFI data is clearer. One 'step' of either system produces
different amounts of progress, and this conditional-checking-and-stepping
is needed to tie up the flows.

Closes #1687

### DIFF
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -127,76 +127,156 @@ bool SpikeCosim::backdoor_read_mem(uint32_t addr, size_t len,
   return bus.load(addr, len, data_out);
 }
 
-bool SpikeCosim::step(uint32_t write_reg, uint32_t write_reg_data, uint32_t pc,
+// When we call processor->step(), spike advances to the next pc IFF a trap does
+// not occur. If a trap does occur, state.last_inst_pc is set to PC_INVALID, and
+// we need to call step() again to actually execute the first instruction of the
+// trap handler.
+// This sentinel value PC_INVALID of state.last_inst_pc allows the cosimulation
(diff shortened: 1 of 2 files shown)


dv/cosim/spike_cosim.cc
```diff
@@ -127,78 +127,140 @@ bool SpikeCosim::backdoor_read_mem(uint32_t addr, size_t len,
...
                       bool sync_trap) {
   assert(write_reg < 32);
 
-  uint32_t initial_pc = (processor->get_state()->pc & 0xffffffff);
-  bool initial_pc_match = initial_pc == pc;
+  // The DUT has just produced an RVFI item (parameters of this func is the data of the item).
 
-  // Execute the next instruction
+  uint32_t initial_spike_pc;
+  bool pending_synch_instr_exception = false;
```
GregAC, line 166: nit: Please use sync rather than synch in keeping with the existing `sync_trap`

GregAC (approved):

hcallahan-lowrisc:
Force-pushed new changes which : ran clang-format, addressed nit, reword some longer prose-like comments to be clearer.

PR #1717, opened 2022-07-21
[dv] Various coverage hole fixes

ctopal:
First commit fixes how we collect a certain condition in the case of core not having a clock. Second commit enables generating `FENCE.I` instruction in riscv-dv.

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
@@ -276,6 +276,22 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     end
   end
 
+  // This latch is needed because if we cannot register this condition being true
+  // with a clock. Being in the sleep mode implies that we don't have an active clock.
+  // So, we need to catch the condition, latch it and keep it until we wake up and decode
+  // an instruction (which guarantees we have a clock in the core)
+  logic kept_wfi_with_irq;
+
+  always_latch begin
+    if (id_stage_i.controller_i.ctrl_fsm_cs == DECODE) begin
+      kept_wfi_with_irq = 1'b0;
+    end else if (id_stage_i.controller_i.ctrl_fsm_cs == SLEEP &&
+                 id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
+                 (|cs_registers_i.mip)) begin
+      kept_wfi_with_irq = 1'b1;
+    end
+  end
+
   logic instr_id_matches_trigger_d, instr_id_matches_trigger_q;
 
   assign instr_id_matches_trigger_d = id_stage_i.controller_i.trigger_match_i &&
@@ -528,10 +544,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     }
 
     // This will only be seen when specific interrupt is disabled by MIE CSR
-    `DV_FCOV_EXPR_SEEN(irq_continue_sleep, id_stage_i.controller_i.ctrl_fsm_cs == SLEEP &&
-                                           id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
-                                           (|cs_registers_i.mip))
-
+    `DV_FCOV_EXPR_SEEN(irq_continue_sleep, kept_wfi_with_irq)
 
     cp_single_step_instr: coverpoint id_instr_category iff
                                      (id_stage_i.controller_i.fcov_debug_single_step_taken) {
--- a/dv/uvm/core_ibex/riscv_dv_extension/riscv_core_setting.sv
+++ b/dv/uvm/core_ibex/riscv_dv_extension/riscv_core_setting.sv
@@ -46,7 +46,7 @@ privileged_mode_t supported_privileged_mode[] = {MACHINE_MODE, USER_MODE};
 // Unsupported instructions
 // Avoid generating these instructions in regular regression
 // FENCE.I is intentionally treated as illegal instruction by ibex core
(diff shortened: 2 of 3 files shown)


dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -276,6 +276,18 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     end
   end
 
+  // This latch is needed because if we cannot register this condition being true
+  // with a clock. Being in the sleep mode implies that we don't have an active clock.
+  // So, we need to catch the condition, latch it and keep it until we wake up and decode
```
marnovandermaas, line 281: How can you latch a condition when you don't have a clock?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -528,10 +540,7 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
     }
 
     // This will only be seen when specific interrupt is disabled by MIE CSR
-    `DV_FCOV_EXPR_SEEN(irq_continue_sleep, id_stage_i.controller_i.ctrl_fsm_cs == SLEEP &&
-                                           id_stage_i.controller_i.ctrl_fsm_ns == SLEEP &&
-                                           (|cs_registers_i.mip))
-
```
marnovandermaas, line 534: To me this expression looks similar if not the same as the assigned value of `kept_wfi_with_irq`, is that right?

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: This part of of the code doesn't need to be synthesisable but still, we can have clockless latches so I don't understand the question here. 

image

For our example we can think like

```
rstn = (id_stage_i.controller_i.ctrl_fsm_cs == DECODE)`

d = 1
 
en = (id_stage_i.controller_i.ctrl_fsm_cs == SLEEP &&
       (|cs_registers_i.mip)) 
```

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: In this part of the code, we are only updating this expression when we are at the posedge of clk (which is connected in core level). If we are sleeping, we don't have an active clock, which means we are not going to update this expression. So, as a solution we need to have a latch that keeps the information of this expression being true until we see a clock.

ctopal:
Now added a fix for the failure in CI for `riscv_multiple_interrupt_test` that happens because of a race condition for the signal `dut_if.ecall`

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: Nice catch.

This is a bit of a tricky one as every coverpoint/cross is only sample on the clock, so no clock no sampling.

I can see how this works as it will capture that we've ignored an IRQ wakeup whilst there's no clock running and the latched value will remain set when we see the first clock.

We could instead clock the coverage gathering off an ungated clock. though now we're sampling things that aren't actually happening. It's probably fine as the state is just frozen in place but again doesn't seem great.

We could have an whole new covergroup clocked off an ungated clock, but currently it'd only have this single coverpoint in it.

For now I think we'll keep this latch setup but please rewrite to use `always_latch` rather than the infinite loop with an `always`. It does just about work but explicit latch logic will make more sense

```systemverilog
always_latch begin
  if (latch_en) begin
    latch_q = ....;
  end
end
```

GregAC (approved):
Looks good minus nit around latch construction and commit spelling errors

>  [dv,test] Fix racing condition to catch ecall
>
> We alrady have a clocking block i nsid dut_if. This commit uses it
> to avoid a race condition that happens when `instr_valid_i` goes high
> while `ecall_insn_i` goes low.

Should be

>  [dv,test] Fix race condition to catch ecall
>
> We already have a clocking block inside dut_if. This commit uses it
> to avoid a race condition that happens when `instr_valid_i` goes high
> while `ecall_insn_i` goes low.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
ctopal: I believe already know these stuff but just to record it here, in OTBN (and I think in OpenTitan in general) we have definitions of covergroups outside of the actual design (so they are not binded and they are not interfaces). 

We get the needed information through a `uvm_sequence_item` (rtl_item). Then using what is inside the scoreboard (rtl_item, iss_item, tl_seq_item etc.), we are driving the sampling functions inside `otbn_env_cov` from the scoreboard. That way it is actually through the scoreboard we do the coverage collection and not the `fcov_if` that we bind it inside the design.

I think maybe in the future after cosim integration is fully done, we can do a similar thing in Ibex DV environment, what do you think?

ctopal:
Implemented feedback from Greg, thanks for having a look!

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
GregAC: It may not solve this issue as all it's really doing is divorcing the actual sampling from the coverage gathering. The thing that generates the `uvm_sequence_item` still needs to be clocked. Of course we could generate them from multiple places and maybe have multiple sequence items. So we'd have a special monitor that runs off the un-gated clock. Though this adds extra complexity so is only worth it if we've got lots of things to cover based on the un-gated clock.

PR #1724, opened 2022-07-26
Improve comments around check_mem_access and mmio_load/store

hcallahan-lowrisc:
Added more comments around the function of SpikeCosim::check_mem_access() and SpikeCosim::mmio_load()/mmio_store().
Minor refactoring for clarity and flow with the comments.

I guess it could be argued that this should be in the /doc directory, but I'm indifferent to either. Feedback welcome.


dv/cosim/spike_cosim.cc
```diff
@@ -516,12 +535,76 @@ void SpikeCosim::fixup_csr(int csr_num, uint32_t csr_val) {
...
+  // always run slightly-behind the ibex, and hence we need to store memory
+  // accesses and errors when they are seen by the testbench, and then pop
+  // them from the queues when spike actually executes them.
+  // This delay may be a few cycles in general, but may never happen if
+  // the access was iside-speculative.
+  //
+  // Whenever a load/store is attempted by spike, we check the following
+  // conditions....
+  //
+  // -1) First, if an iside error has been detected and
```
ctopal, line 549: Maybe 0 instead of -1 because it looks like and erroneous `1-)` :)

ctopal (approved):
LGTM and thanks a lot for clarifying how things work! I was wondering if there would be a better way of documenting this rather than in putting it directly into the code as comments though.

dv/cosim/spike_cosim.cc
```diff
@@ -68,40 +68,59 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
...
     // incremented aligned PC (to capture the unaligned 4-byte instruction
-    // case). Assume a successful iside access if either of these checks are
-    // true, otherwise assume a dside access and check against DUT dside
-    // accesses.  If the RTL produced a bus error for the access, or the
-    // checking failed produce a memory fault in spike.
-    dut_error = (check_mem_access(false, addr, len, bytes) != kCheckMemOk);
-  }
+    // case).
+    uint32_t aligned_addr = addr & 0xfffffffc;
+    if ( pending_iside_error &&
```
GregAC, line 102: This does alter the behaviour from what it was previously. Now the address needs to the fail the 'is_probably_dmem_access' heuristic before we check if the address matches a pending iside error. Previously if it matched a pending iside error it automatically assumed it was an iside error.

dv/cosim/spike_cosim.cc
```diff
@@ -68,40 +68,59 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
...
+  // Within Spike (mmu_t::load_slow_path), this causes a load_access_fault.
+
+  // Spike does not differentiate between iside and dside memory accesses
+  // (as ibex does) so we make an educated guess which is which, and
+  // then after fetching the data, check against the recorded DUT access
+  // that should be queued as pending.
+  //
+  // Spike may attempt to access up to 8-bytes from the PC when fetching, so
+  // assume as a dside access when it falls outside that range.
+  // N.B. Heuristic
```
GregAC, line 83: Worth mentioning that is the heuristic doesn't get it right it will produce false failures. In general we don't expect to see the scenarios that generate such false failures in practise and if they prove to be an issue we'll rethink this mechanism.

dv/cosim/spike_cosim.cc
```diff
@@ -68,40 +68,59 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
...
-  }
+    // case).
+    uint32_t aligned_addr = addr & 0xfffffffc;
+    if ( pending_iside_error &&
+        (pending_iside_err_addr == aligned_addr)) {
+      // If the pending_iside_error has been set, produce a dut_error.
+      pending_iside_error = false;
+      dut_error = true;
+    } else {
+      // Probably a valid iside access
```
GregAC, line 108: I can see the argument documentation wise but we shouldn't really have an empty else, so ditch it and move the 'Probably a valid iside access' to an appropriate point.

GregAC (changes requested):
A good improvement here. Though I'd prefer not to alter the behaviour of the checking, was it intentional you altered the behaviour?

dv/cosim/spike_cosim.cc
hcallahan-lowrisc: I've updated this to be an assertion failure if the `is_probably_dmem_access` and `pending_iside_error` are both true. I think it is probably better to be explicit about what happens when something goes wrong here. Does that make sense?

hcallahan-lowrisc:
Rebased

dv/cosim/spike_cosim.cc
```diff
@@ -71,40 +71,56 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
...
+  // that should be queued as pending.
+  //
+  // Spike may attempt to access up to 8-bytes from the PC when fetching, so
+  // assume as a dside access when it falls outside that range.
+  // N.B. Heuristic
+  uint32_t pc = processor->get_state()->pc;
+  bool is_probably_dmem_access = (addr < pc || addr >= (pc + 8));
+  // Check we don't have a heuristic mismatch (this may be a false-positive but
+  // worth failing explicity if it is seen)
+  assert(!(is_probably_dmem_access && pending_iside_error));
```
GregAC, line 91: This also needs to check if the pending iside error address matches the incoming address (`pending_iside_err_addr == aligned_addr` check you can see below).

Maybe add a new `bool` that combines the two as it'll be used in a couple of places?

dv/cosim/spike_cosim.cc
GregAC: Yup sounds good

GregAC (approved):
Thanks for this @hcallahan-lowrisc, looks good once we've fixed up the assertion (see comment).

dv/cosim/spike_cosim.cc
hcallahan-lowrisc: Ah yes good point. Will do!

GregAC:
@hcallahan-lowrisc any progress on this?

hcallahan-lowrisc:
> @hcallahan-lowrisc any progress on this?

Not yet. Will try and get it closed soon.

dv/cosim/spike_cosim.cc
```diff
@@ -519,12 +535,76 @@ void SpikeCosim::fixup_csr(int csr_num, uint32_t csr_val) {
...
+  // Because Spike is only fed instructions that have been made available
+  // on the RVFI by IBEX (retired and synchronously trapping), spike will
+  // always run slightly-behind the ibex, and hence we need to store memory
+  // accesses and errors when they are seen by the testbench, and then pop
+  // them from the queues when spike actually executes them.
+  // This delay may be a few cycles in general, but may never happen if
+  // the access was iside-speculative.
+  //
+  // Whenever a load/store is attempted by spike, we check the following
+  // conditions....
```
marnovandermaas, line 547: Nit: 4 dots instead of 3

dv/cosim/spike_cosim.cc
```diff
@@ -519,12 +535,76 @@ void SpikeCosim::fixup_csr(int csr_num, uint32_t csr_val) {
...
+  //
+  // The correspondence checks are as follows:
+  //
+  // 1) Check there is actually a dmem access in the queue at all.
+  // Pop the top dmem access from the queue, and then...
+  // 2) Check the addr of spike access matches that of the queued dmem access.
+  // 3) Check the type of spike access (load/store) matches that of the queued
+  // dmem access.
+  //
+  // If the dut dmem access was misaligned :
```
marnovandermaas, line 569: Nit: space before colon

dv/cosim/spike_cosim.cc
```diff
@@ -659,15 +745,24 @@ SpikeCosim::check_mem_result_e SpikeCosim::check_mem_access(
...
-    // access
-    if (top_pending_access_info.misaligned_first &&
+    // If requested to make a misaligned access that crosses a 32-bit
+    // boundary (this could be a word or halfword access), the DUT
+    // LSU generates two word-aligned accesses with appropriate
+    // byte-enables set.
+    //
+    // If the top pending access from the DUT was the first half of a misaligned
+    // access which accesses the top byte, it must have crossed the 32-bit
+    // boundary and generated a second access
```
marnovandermaas, line 761: Nit: missing .

marnovandermaas (approved):
Great work! Just a few nits, but otherwise I think it looks good.

PR #1726, opened 2022-07-26
[dv] Make Xcelium wave dumping includes unpacked arrays

GregAC:
In various places within Ibex we use unpacked arrays. We weren't dumping
these in Xcelium.

### DIFF
--- a/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
+++ b/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
@@ -200,5 +200,5 @@
       +enable_ibex_fcov=1
     wave_opts: >
       -input @"database -open <sim_dir>/waves -shm -default"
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
LGTM!

ctopal (approved):
LGTM with some concern about performance. Wouldn't it create some performance slow downs (and potentially big `.shm` files) if we enable ICache since it will instantiate 2 RAMs?

GregAC:
> LGTM with some concern about performance. Wouldn't it create some performance slow downs (and potentially big .shm files) if we enable ICache since it will instantiate 2 RAMs?

They're only 4k a piece I think the simulator and dump file format are intelligent enough to deal with this smoothly without major performance/space impact.

GregAC:
Oh and if the icache RAMs prove to be an issue we can do a finer grained version that only applies within ibex_core + the register file.

PR #1729, opened 2022-07-27
[dv,core_ibex] Add push_pull agent for Scramble IF

### LINKED ISSUE #1728: [dv,core_ibex] Instantiate Scrambling Key Interface in core_ibex DV
If we are starting to issue `FENCE.I` instructions, we definitely need to drive scrambling key ports so that ICache would have a valid key. Currently we don't have a valid key, so we are stuck at an invalidating state.

>estimate 4
>remaining 2022-07-27 2

ctopal:
This allows us to actually use Icache in our tests beecause
before this commit key_valid was tied to 0 which means everytime
we receive a FENCE.I instruction, we wouldn't be able to successfully
flag inval_done. Which means we weren't probably using ICache correctly.

Resolves #1728 


GregAC (approved):
This LGTM to me though where are the keys coming from to supply the push/pull agent? Is there a follow up PR to feed the agent with suitable values?

ctopal:
What do you mean by suitable values? Push pull agent generates random responses when DUT requests it, handling of it is done in `push_pull_device_seq.sv`. Do you think we need to feed custom values to it?

GregAC:
> What do you mean by suitable values? Push pull agent generates random responses when DUT requests it, handling of it is done in `push_pull_device_seq.sv`. Do you think we need to feed custom values to it?

Ah right, I had missed that aspect, yes that's just fine.

PR #1730, opened 2022-07-27
[dv,testlist] Enable irq_timer, allow WFI in Umode

ctopal:
This commit sets two different riscv-dv knob to make sure we hit
some holes in our coverpoints.

Activating `enable_timer_irq` for everytime we enable other
interrupts makes sure that we respond to it just like we respond
to other interrupts.

Setting `tw=0` makes it so that we would allow WFI in user mode. We
were already randomizing it but for some certain tests, we actually
want to be in a sleep state, which wouldn't happen if tw=1 in user
mode.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -226,6 +226,7 @@
     +no_csr_instr=1
     +no_fence=1
     +no_wfi=0
+    +set_mstatus_tw=0
     +randomize_csr=1
     +num_of_sub_program=0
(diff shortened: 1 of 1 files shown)


ctopal:
Have any of you had a chance to look at this one? It's really just a couple of riscv-dv knobs getting enabled and improves our coverage of `cp_interrupt_taken`

hcallahan-lowrisc (approved):

marnovandermaas (approved):
LGTM, as far as I can tell `+enable_timer_irq` enables `XTIE` bits in `mie` register. Also, according to the spec we need to set `mstatus.TW` to zero to execute `WFI`. So the changes look good to me.

PR #1734, opened 2022-07-28
[dv] Enable pmp_full_random test

ctopal:
Tested with 50 seeds, 2 failing 48 passing. Enables us to observe locked regions as well.

(with latest spike-cosim + https://github.com/lowRISC/riscv-isa-sim/pull/10)


hcallahan-lowrisc (approved):
Sounds good to get this back in.
What sort of cp's does this let us hit now the test is reintroduced? Could you just amend the PR message with some examples of what you've seen/expect?
Also, I know this wasn't your test to begin with, but does 100 iterations seem like the right number to hit all the coverpoints we might want to? That does seem like a lot, I imagine there is a significant cost to the time taken for the full regression.

Issue #1720, opened 2022-07-22
[dv] Exception divergence using MML PMP mode


marnovandermaas:
I'm seeing a divergence between Spike and Ibex when in MML mode and setting region 0 to read only. Spike throws a trap and Ibex zeroes out a destination register. I'm not sure if this is an issue with Spike, Ibex or with the UVM comparison.

## Observed Behavior

Cosimulation result:
```
name:            riscv_epmp_mml_read_only_test
seed:            14822
binary:          out/run/riscv_epmp_mml_read_only_test.14822/test.o
uvm_log:         out/run/riscv_epmp_mml_read_only_test.14822/rtl.log
rtl_trace:       out/run/riscv_epmp_mml_read_only_test.14822/trace_core_00000000.log
rtl_trace_csv:   out/run/riscv_epmp_mml_read_only_test.14822/trace_core_00000000.log.csv
iss_trace:       
iss_trace_csv:   
en_cosim:        True
cosim_trace:     
cosim_trace_csv: 
comparison_log:  
passed:          False
failure_message: |-
  UVM_FATAL /home/mvdmaas/ibex/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv(131) @ 7304: uvm_test_top.env.cosim_agent.scoreboard [ibex_cosim_agent_pkg::ibex_cosim_scoreboard.run_cosim_rvfi] Cosim mismatch Synchronous trap was expected at ISS PC: 80001900 but DUT didn't report one at PC 80000118
  [FAILED]: sim error seen
```

Ibex RVFI trace:
```
Time	Cycle	PC	Insn	Decoded instruction	Register and memory contents
           2524	        24	80000080	400015b7	lui	x11,0x40001	 x11=0x40001000
           2864	        41	80000084	10658593	addi	x11,x11,262	 x11:0x40001000 x11=0x40001106
           3004	        48	80000088	30159073	csrrw	x0,misa,x11	 x11:0x40001106  x0=0x00000000
           3024	        49	8000008c	0002c397	auipc	x7,0x2c	  x7=0x8002c08c
           3444	        70	80000090	c5838393	addi	x7,x7,-936	  x7:0x8002c08c  x7=0x8002bce4
           3724	        84	80000094	00002597	auipc	x11,0x2	 x11=0x80002094
           3744	        85	80000098	86c58593	addi	x11,x11,-1940	 x11:0x80002094 x11=0x80001900
[...]

PR #1735, opened 2022-07-29
[rtl] Do not register instruction fetch errors

ctopal:
Registering instruction fetch errors causes problem in the case of seeing
a PMP execution access block. ID/EX stage learns about
the instruction that it executed should be blocked a
cycle late. This commit fixes that.

Found this one while working on fixing https://github.com/lowRISC/ibex/issues/1720. However it does not fully solves the issue because we are still seeing some mismatches because of some self modifying code problems.

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -502,7 +502,7 @@ module ibex_id_stage #(
   );
 
   /////////////////////////////////
-  // CSR-related pipline flushes //
+  // CSR-related pipeline flushes //
   /////////////////////////////////
   always_comb begin : csr_pipeline_flushes
     csr_pipe_flush = 1'b0;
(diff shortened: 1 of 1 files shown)


rtl/ibex_if_stage.sv
```diff
@@ -242,7 +240,7 @@ module ibex_if_stage import ibex_pkg::*; #(
     assign instr_intg_err            = 1'b0;
   end
 
-  assign instr_err        = instr_intg_err | instr_bus_err_i;
+  assign instr_err        = (instr_intg_err | instr_bus_err_i) & instr_rvalid_i;
   assign instr_intg_err_o = instr_intg_err & instr_rvalid_i;
```
ctopal, line 244: If you have a more optimized way of doing this I would really appreciate the feedback

marnovandermaas (approved):
LGTM, thanks!

GregAC (changes requested):
We can't ditch the register stage here. The PC used by the PMP check is from the IF stage. So with change we feed the PMP error of whatever is in the IF stage directly to the ID/EX stage, so we get the PMP check if the PC following the PC we're executing.

The fundamental issue is the timing of PMP register updates. When we have an instruction in ID/EX that writes a PMP register that update gets written to the CSR the same cycle the next instruction moves from IF to ID/EX with it's PMP check done with the old value.

The solution is to flush the pipeline when we get a PMP CSR write. We already have a signal that can do this, `csr_pipe_flush` within `ibex_id_stage`. I think we can just raise it when we have a PMP CSR write.

Alternatively we do nothing. It may acceptable under the specification to say PMP changes aren't guaranteed to take effect until an FLUSH or FLUSH.I, I need to double check the wording. Though as the flush is simple and doesn't cause any other issues I think that's the better option.

GregAC:
From the spec

> If page-based virtual memory is not implemented, memory accesses check the PMP settings synchronously, so no SFENCE.VMA is needed

I interpret that to mean that PMP CSR writes must take immediate effect (i.e. apply to the next instruction execution) where virtual memory isn't implemented. So we do definitely have to fix this.

ctopal:
I implemented the solution you suggested, I think it works but need more time to test  it thoroughly.

ctopal:
I did test the fix in pmp_full_random test, looks like it indeed fixes our problem. @GregAC can you have another look for the new version please?

rtl/ibex_id_stage.sv
```diff
@@ -514,7 +514,9 @@ module ibex_id_stage #(
     // - When modifying debug CSRs - TODO: Check if this is really needed
     if (csr_op_en_o == 1'b1 && (csr_op_o == CSR_OP_WRITE || csr_op_o == CSR_OP_SET)) begin
       if (csr_num_e'(instr_rdata_i[31:20]) == CSR_MSTATUS   ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_MIE) begin
+          csr_num_e'(instr_rdata_i[31:20]) == CSR_MIE       ||
+          // Changes to PMPADDR/PMPCFG registers will cause a flush.
+          instr_rdata_i[31:25] == 7'h1D) begin
```
GregAC, line 519: Try `instr_rdata_i[31:24] == CSR_PMPCFG0[11:5]` makes it more obvious what's going on.

GregAC (approved):
LGTM with a couple of small suggestions

PR #1737, opened 2022-08-01
[rtl] Deny no-match X access in M-Mode while MML=1

### LINKED ISSUE #1736: [rtl,bug] Ibex allows execution in no-match case while MML=1
Header explains the issue. According to Smepmp v1.0 in a no-match case we should block execution in machine mode. However, we are not currently doing that. It's a tricky thing to recreate it in our master branch for now but if you add #1695 and #1735 to master and run: `make ITERATIONS=1 TEST=riscv_pmp_full_random_test SEED=14910` you should be able to re-create the error.

> estimate 4

ctopal:
In ePMP spec, it specifies as:

```
Executing code with Machine mode privileges is only possible from memory
regions with a matching Mmode-only rule or a locked Shared-Region rule
with executable privileges. Executing code from a region without a
matching rule or with a matching S/U-mode-only rule is denied.
```

This change provides that.

Resolves #1736 

### DIFF
--- a/rtl/ibex_pmp.sv
+++ b/rtl/ibex_pmp.sv
@@ -123,14 +123,17 @@ module ibex_pmp #(
 
   // Access fault determination / prioritization
   function automatic logic access_fault_check (logic                     csr_pmp_mseccfg_mmwp,
+                                               logic                     csr_pmp_mseccfg_mml,
+                                               ibex_pkg::pmp_req_e       pmp_req_type,
                                                logic [PMPNumRegions-1:0] match_all,
                                                ibex_pkg::priv_lvl_e      priv_mode,
                                                logic [PMPNumRegions-1:0] final_perm_check);
 
 
     // When MSECCFG.MMWP is set default deny always, otherwise allow for M-mode, deny for other
-    // modes
-    logic access_fail = csr_pmp_mseccfg_mmwp | (priv_mode != PRIV_LVL_M);
+    // modes. Also deny unmatched for M-mode whe MSECCFG.MML is set and request type is EXEC.
+    logic access_fail = csr_pmp_mseccfg_mmwp | (priv_mode != PRIV_LVL_M) |
+                        (csr_pmp_mseccfg_mml && (pmp_req_type == PMP_ACC_EXEC));
 
     // PMP entries are statically prioritized, from 0 to N-1
     // The lowest-numbered PMP entry which matches an address determines accessibility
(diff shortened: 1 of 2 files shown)


GregAC:
After looking at this further it is a rule that makes perfect sense. The idea being that during boot/initialization you setup all of the code regions you want M mode to be able to execute. Once done there's no reason to execute outside of that so this provides you with extra protection without needing some catch all rule to fill in the gaps.

There's also this:

> Adding a rule with executable privileges that either is M-mode-only or a locked Shared-Region is not possible and such pmpcfg writes are ignored, leaving pmpcfg unchanged. This restriction can be temporarily lifted e.g. during the boot process, by setting mseccfg.RLB

Which is not implemented either. I'm a bit unclear on what it's getting at with the 'temporarily lifted' bit. It doesn't seem to be explicitly saying setting `mseccfg.RLB` bypasses this requirement but rather it should be clear that setting `mseccfg.RLB` bypasses the requirement from some other part of the specification and it's just mentioned as an example here.

I think these things must have been late additions, perhaps I just managed to skip over them during my initial implementation. Still an encouraging sign our DV work has already found our lack of compliance with one of the points.

GregAC:
> I think these things must have been late additions, perhaps I just managed to skip over them during my initial implementation. Still an encouraging sign our DV work has already found our lack of compliance with one of the points.

Nope, they were in the original proposal document I used for the implementation. I somehow missed them!

GregAC:
>  I'm a bit unclear on what it's getting at with the 'temporarily lifted' bit. It doesn't seem to be explicitly saying setting mseccfg.RLB bypasses this requirement but rather it should be clear that setting mseccfg.RLB bypasses the requirement from some other part of the specification and it's just mentioned as an example here.

Answering my own question it comes down the comma. What it's saying is setting mseccfg.RLB bypasses this requirement which you may wish to do temporarily, e.g. during the boot process.

rtl/ibex_pmp.sv
```diff
@@ -123,14 +123,17 @@ module ibex_pmp #(
...
+                                               ibex_pkg::pmp_req_e       pmp_req_type,
                                                logic [PMPNumRegions-1:0] match_all,
                                                ibex_pkg::priv_lvl_e      priv_mode,
                                                logic [PMPNumRegions-1:0] final_perm_check);
 
 
     // When MSECCFG.MMWP is set default deny always, otherwise allow for M-mode, deny for other
-    // modes
-    logic access_fail = csr_pmp_mseccfg_mmwp | (priv_mode != PRIV_LVL_M);
+    // modes. Also deny unmatched for M-mode whe MSECCFG.MML is set and request type is EXEC.
```
marnovandermaas, line 134: Nit: whe -> when

marnovandermaas (approved):

GregAC (approved):
LGTM, good job on spotting this @ctopal

PR #1744, opened 2022-08-04
Fix incorrect debug_cause priority against riscv-debug 1.0.0-STABLE

### LINKED ISSUE #1743: Ibex DCSR.cause priority encoding does not match debug-spec 0.13.2 or 1.0.0-STABLE
## Observed Behavior

While testing single-step behaviour, a cosimulation-mismatch was encountered when reading DCSR.cause. It was tracked down to a simultaneous assertion of DCSR.step and debug_req_i when entering debug_mode, and I believe that the spike behaviour is correct in this case.
According to both debug specifications [0.13.2](https://riscv.org/wp-content/uploads/2019/03/riscv-debug-release.pdf) and [1.0.0-STABLE](https://raw.githubusercontent.com/riscv/riscv-debug-spec/master/riscv-debug-stable.pdf), the priority among reporting DCSR.cause has **step** (0) as a lower priority than **haltreq** (1). Ibex [currently reports](https://ibex-core.readthedocs.io/en/latest/03_reference/debug.html#debug-support) debug support in compliance with the RISCV Debug Specification Version 0.13.
Current ibex behaviour is a priority-encoder [which places **step** above **haltreq**](https://github.com/lowRISC/ibex/blob/c4a97e05f5853b25e40fa1ea5b4edcfafba9a370/rtl/ibex_controller.sv#L662).

## Expected Behavior

Ibex should prioritize reporting DCSR.cause as **haltreq** above **step** when entering debug_mode with both stimulus present.

hcallahan-lowrisc:
The relevant page [Debug Spec v1.0.0-STABLE, p.53] gives the following
priorities for resolving multiple concurrent reasons for entering debug mode....

> DCSR.cause : Explains why Debug Mode was entered.
> 
> When there are multiple reasons to enter Debug Mode in a single cycle,
> hardware should set cause to the cause with the highest priority.
> 1: An ebreak instruction was executed.                        (priority 3)
> 2: A Trigger Module trigger fired with action=1.              (priority 4)
> 3: The debugger requested entry to Debug Mode using haltreq.  (priority 1)
> 4: The hart single stepped because step was set.              (priority 0, lowest)
> 5: The hart halted directly out of reset due to resethaltreq. (priority 2)
>    It is also acceptable to report 3 when this happens.
> 6: The hart halted because it’s part of a halt group.         (priority 5, highest)
>    Harts may report 3 for this cause instead.
> 
> Other values are reserved for future use.

Fixes #1743

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -658,11 +658,11 @@ module ibex_controller #(
 
         csr_save_cause_o = 1'b1;
         if (trigger_match_i) begin
-          debug_cause_o = DBG_CAUSE_TRIGGER;
-        end else if (debug_single_step_i) begin
-          debug_cause_o = DBG_CAUSE_STEP;
+          debug_cause_o = DBG_CAUSE_TRIGGER;     // (priority 4)
+        end else if (debug_req_i) begin
+          debug_cause_o = DBG_CAUSE_HALTREQ;     // (priority 1)
         end else begin
-          debug_cause_o = DBG_CAUSE_HALTREQ;
+          debug_cause_o = DBG_CAUSE_STEP;        // (priority 0, lowest)
         end
 
         // enter debug mode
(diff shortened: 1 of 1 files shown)


GregAC (approved):
LGTM. This is a good bug to find! It's a more complex case, demonstrating we're getting good stimulus. It's also very trivial, giving some confidence we haven't got more major bugs in this area (as they would have appeared first).

GregAC:
Looked at the CI failure, it's the simulator complaining it can't create a wave file as one already exists. I'll rerun the job

hcallahan-lowrisc:
Pushed a new change to make a comment about the priority match the spec a bit better.

PR #1745, opened 2022-08-04
[dv,cosim] Add memory region to match UVM env

ctopal:
This extends our memory range to be between 0 and
0xFFFF_FFFF. This is needed because in the case of
no match between UVM memory model and spike memory
model, we see a mismatch.

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_cosim_agent/spike_cosim_dpi.cc
+++ b/dv/uvm/core_ibex/common/ibex_cosim_agent/spike_cosim_dpi.cc
@@ -23,6 +23,7 @@ void *spike_cosim_init(const char *isa_string, svBitVecVal *start_pc,
   SpikeCosim *cosim = new SpikeCosim(isa_string, start_pc[0], start_mtvec[0],
                                      log_file_path, false, true);
   cosim->add_memory(0x80000000, 0x80000000);
+  cosim->add_memory(0x00000000, 0x80000000);
   return static_cast<Cosim *>(cosim);
 }
 


dv/uvm/core_ibex/common/ibex_cosim_agent/spike_cosim_dpi.cc
```diff
@@ -23,6 +23,7 @@ void *spike_cosim_init(const char *isa_string, svBitVecVal *start_pc,
   SpikeCosim *cosim = new SpikeCosim(isa_string, start_pc[0], start_mtvec[0],
                                      log_file_path, false, true);
   cosim->add_memory(0x80000000, 0x80000000);
+  cosim->add_memory(0x00000000, 0x80000000);
```
marnovandermaas, line 26: I don't think this is correct. This would cause any memory address from `0x8000_0000` to `0xffff_ffff` to be out of bounds right?

marnovandermaas (changes requested):

dv/uvm/core_ibex/common/ibex_cosim_agent/spike_cosim_dpi.cc
ctopal: I thought it would create two separate memories. I tested this and haven't seen the behaviour you suggested.

marnovandermaas: Ah yes, I misunderstood. Should work!

marnovandermaas (approved):
LGTM, thanks!

GregAC (approved):

hcallahan-lowrisc:
I think this fix is fine, but it could mask some other issues down the line. Like we have discussed @ctopal, the issue where you can get pmp violations from the [pushing gpr's to stack routine](https://github.com/lowRISC/ibex/commit/6e46b3ce1ec37ee6d8e9860a3fce9b691b14a271) in riscv-dv out of reset is mitigated by this fix but is also a problem in itself that needs addressing. If the ibex should somehow spuriously access memory that we don't expect it to, that should be caught by the testbench as a failure in some cases. Communicating what memory we consider valid to access is I admit a trickier task to solve. In the above case it's really the binary causing us to make strange accesses, but in other cases it's always possible it could be a bug in the ibex.
For example, a normal well-formed program would allocate memory to use as a stack and a heap, and shouldn't otherwise access memory outside of that specified to the linker. It would be nice to constrain the allocated/initialised memory in those cases.

GregAC:
In terms of finding bugs in Ibex the cosim will find any spurious memory acceses Ibex makes but spike does not. Though @hcallahan-lowrisc's point is more around finding bugs in our test generation. For this I think rather than playing around with the base memory map we'd be better off with something configurable hanging off the memory bus monitor that can be told to produce warnings/errors if it sees certain accesses in certain regions. This can be then be configured by the test. I'll create an issue to track the idea, probably a post V2 thing.

PR #1746, opened 2022-08-05
[lint] Minor fixes

msfschaffner:
This is part of the OpenTitan lint cleanup push in order to get the entire design lint clean.

### DIFF
--- a/rtl/ibex_pmp.sv
+++ b/rtl/ibex_pmp.sv
@@ -220,6 +220,11 @@ module ibex_pmp #(
                                                           pmp_req_type_i[c],
                                                           priv_mode_i[c],
                                                           region_basic_perm_check[c][r]);
+
+      // Address bits below PMP granularity (which starts at 4 byte) are deliberately unused.
+      logic unused_sigs;
+      assign unused_sigs = ^{region_start_addr[r][PMPGranularity+2-1:0],
+                             pmp_req_addr_i[c][PMPGranularity+2-1:0]};
     end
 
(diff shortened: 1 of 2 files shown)


msfschaffner:
If this looks good, it would be great if someone could merge this and re-vendor it into OpenTitan so that I can take care of the remaining top-level issues tomorrow and then switch the AscentLint step to be fully blocking.

CC: @moidx

rtl/ibex_pmp.sv
```diff
@@ -220,6 +220,9 @@ module ibex_pmp #(
                                                           pmp_req_type_i[c],
                                                           priv_mode_i[c],
                                                           region_basic_perm_check[c][r]);
+      logic unused_sigs;
+      assign unused_sigs = ^{region_start_addr[r][PMPGranularity+2-1:0],
```
andreaskurth, line 224:
```suggestion
      // Address bits below PMP granularity (which starts at 4 byte) are deliberately unused.
      assign unused_sigs = ^{region_start_addr[r][PMPGranularity+2-1:0],
```

andreaskurth (commented):
LGTM, thanks @msfschaffner!

I suggest to add a comment describing why the signals are deliberately unused.

GregAC (approved):

rtl/ibex_pmp.sv
msfschaffner: Sounds good, I've added that!

PR #1750, opened 2022-08-08
Change test_done mechanism to use riscvdv handshake, not ecall

### LINKED ISSUE #1718: [dv] Allow generic use of ecall
As I understand it RISC-V DV uses `ecall` to the signal the end of a test. The issue with this is we can't have arbitrary `ecall` executions within a generated test.

We should use another mechanism to trigger test end (maybe a write to a specific memory location?) allowing use of `ecall` within the test program body.

I've added this to the V2 milestone as I think we'll need to to close some of the coverage holes (specifically single stepping over an `ecall`).

Depending on how complex it ends up being we could consider waiving the need for this for V2 or do some limited fix that allows us to close the coverage without altering the end of test signalling too radically.

> estimate 8
> remaining 2022-08-08 1

hcallahan-lowrisc:
Use the address (signature_addr - 0x4) for a TEST_PASS handshake.
Create seperate subscriber port for test_done functionality

By creating a new, distinct port and subscribing to all incoming memory items,
the existing wait_for_mem_txn() can be used with minor modifications to be
able to choose the port to wait on as an argument to the task.

Because the wait_for_mem_txn() implementation currently uses get() to pop the
latest item from the item_collected queue, having two different forked processes
that both await on this queue is not possible. The simplest solution is to
create a new, seperate port which also subscribes to the sequence_items
broadcast by the mem_if monitor.

Possible improvements / todo:
- Create a seperate plusarg for this test_done address (don't just assume signature_addr - 0x4 is okay)
- Update documentation
- Consider upstreaming if it proves a good solution.

Closes #1718

Signed-off-by: Harry Callahan <hcallahan@lowrisc.org>

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
@@ -43,4 +43,40 @@ class ibex_asm_program_gen extends riscv_asm_program_gen;
     instr.push_back("csrwi 0x7c0, 1");
   endfunction
 
+  // Generate "test_done" section.
+  // The test is ended from the UVM testbench, which awaits the following write to
+  // the special test-control signature address (signature_addr - 0x4).
+  // The ECALL trap handler will handle the clean up procedure while awaiting the
+  // simulation to be ended.
+  virtual function void gen_test_done();
+    bit [XLEN-1:0] test_control_addr;
+    string str;
+    string i = indent; // lint-hack
+    test_control_addr = cfg.signature_addr - 4'h4;
+
+    str = {str,
+      {format_string("test_done:", LABEL_STR_LEN), "\n"},
+      {i, "li gp, 1", "\n"}
+    };
+
+    if (cfg.bare_program_mode) begin
+      str = {str,
+        {i, "j write_tohost", "\n"}
+      };
+    end else begin
+      // The testbench will await a write of TEST_PASS, and use that to end the test.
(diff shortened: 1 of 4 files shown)


ctopal (commented):
LGTM apart from the lint errors. Although I am not sure if I understand the functional difference between using this approach versus catching `ecall` from a probe. In both cases after we see `ecall` we run for a while and then and then end the test, right? (I guess we are issuing `vseq.stop()` a bit later with the new version?). My main concern is that we probably still couldn't use this to end expected forever loops in our PMP tests but I guess that is not the scope of this PR.

ctopal (approved):
I did misunderstand the code in the first glance, this is not necessarily about generating code after seeing an `ecall` but actually putting some specific snippet to end the execution, which totally makes sense. Thanks @hcallahan-lowrisc for explaining!

hcallahan-lowrisc:
Thanks for the review @ctopal :) I've pushed some minor changes to fix the linting issues.
I'm going to wait for @GregAC to review before merging, just because he was thinking about this problem already in the discussions of #1718. But thanks for your review too, I appreciate it!

hcallahan-lowrisc:
Pushed a new commit handling the special-case that is riscv_csr_test. Easier than modifying riscv_dv is just to pass that test the special test_done address directly as the signature addr, as that particular address does not use it for anything except the test_done handshaking.

GregAC (approved):
LGTM

marnovandermaas (approved):

Issue #1701, opened 2022-07-13
[dv, fcov] Debug related coverage holes


GregAC:
@hcallahan-lowrisc assigned to all of these 

The following debug related coverpoints are not yet being hit:

- [ ] Some categories for single stepping over each instruction category - `cp_single_step_instr` - #1763  Orig Estimate 4. Remain 2022-07-28 2. Remain 4 2022-08-01. Remain 4 2022-08-03. Remain 1 2022-08-10. Remain 0 2022-10-06.
  - [x] `CSRAccess`
  - [ ] `EBreakDbg`
  - [x] `EBreakExc`
  - [x] `ECall`
  - [x] `MRet`
  - [x] `DRet` - Illegal bin?
  - [x] `WFI`
  - [x] `Fence`
  - [x] `FenceI`
  - [x] `FetchError`
  - [x] `CompressedIllegal`
  - [x] `UncompressedIllegal`
  - [x] `CSRIllegal`
  - [x] `PrivIllegal`
- [ ] Hardware breakpoint triggered - `cp_insn_trigger_enter_debug` - #1801  Orig Estimate 8. Remain 2022-08-16 4.  Remain 2022-08-26 1.
- [ ] Hardware breakpoint on instruction that causes an exception `cp_insn_trigger_exception` - #1801  Orig Estimate 8. Remain 2022-08-16 4. Remain 2022-08-26 1.

Note for some of the single step instruction categories it's possible we simply cannot trigger the coverpoint for those. However from an initial look I think it should be possible for all of them (other than `None`) 

Original HW/PM estimate 20
> estimate 30
> remaining 2022-07-27 16
> remaining 2022-08-01 18
> remaining 2022-08-03 18
> remaining 2022-08-10 15
> remaining 2022-08-16 9
> remaining 2022-08-23 5
> remaining 2022-08-26 3
> remaining 2022-10-06 2

hcallahan-lowrisc:
Update for 2022-07-27 : In progress for `cp_single_step_instr` 2 remaining
                                       Also now 6 remaining for `cp_insn_trigger_enter_debug`

hcallahan-lowrisc:
Update for 2022-08-01 : In progress for `cp_single_step_instr`, increased remaining to 4

hcallahan-lowrisc:
Update for 2022-08-03 : In progress for `cp_single_step_isntr`, remaining still at 4, revised estimate up +4

hcallahan-lowrisc:
[...]

PR #1763, opened 2022-08-11
Single step debug test coverage

hcallahan-lowrisc:
Build on the cosimulation environment to greatly simplify the checking needed for this test. This PR greatly simplify the test, while adding more varied stimulus by tweaking the RISCV-DV flags. This allows us to get much more coverage.

I have seen this PR hit all single_step coverpoints except `InstrCategoryEBreakDbg`, and while I haven't yet investigated exactly why I suspect that we aren't randomizing the csr's to hit that case. TODO.

The single-step tests currently almost all fail, because the additional stimulus happens to hit other cosimulation bugs, still yet to be fully categorized and investigated. However, I thought it was good to get this running as part of the regressions for now.

N.B. This PR contains an update to the vendored RISCV-DV to enable ecall generation. I think this should probably be upstreamed.

Goes towards #1701


hcallahan-lowrisc:
Rebased

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -564,28 +564,32 @@
 
 - test: riscv_debug_single_step_test
   description: >
-    Randomly assert debug_req_i, and set dcsr.step to make ibex execute one isntruction and then re-enter debug mode
+    Randomly assert debug_req_i, and set dcsr.step to make ibex execute one insn then re-enter debug mode
+    The riscv-dv plusarg +enable_debug_single_step adds a short sequence of assembly (riscv_debug_rom_gen::gen_single_step_logic) that single-steps for a randomized number of sequential instructions. It stops setting dcsr.step after the counter reaches 0, and then starts the process again the next time an external debug stimulus is encountered.
```
marnovandermaas, line 568: Should we limit the line length here?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1121,133 +1121,11 @@ class core_ibex_debug_single_step_test extends core_ibex_directed_test;
...
-    // stop waiting if a branch is taken or a jump is set (the PC following `cur_pc` will be
-    // available the same cycle this happens). If `cur_pc` is a branch that isn't taken this will
-    // be resolved when the branch is unstalled (i.e. `branch_taken_id` won't be set and the branch
-    // definitely isn't taken).
-    wait (instr_vif.instr_cb.pc_id != cur_pc     &&
-          instr_vif.instr_cb.valid_id            &&
-          (!instr_vif.instr_cb.stall_id       ||
-           instr_vif.instr_cb.branch_taken_id ||
-           instr_vif.instr_cb.jump_set_id));
-  endtask
```
marnovandermaas, line 1144: Why is this `wait_for_pc_chang` nog longer necessary?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1121,133 +1121,11 @@ class core_ibex_debug_single_step_test extends core_ibex_directed_test;
...
-        // log information about this instruction
-        curr_pc = instr_vif.instr_cb.pc_id;
-        step_instr = instr_vif.instr_cb.instr_id;
-        is_compressed = instr_vif.instr_cb.is_compressed_id;
-        branch_taken  = instr_vif.instr_cb.branch_taken_id;
-        // need to zero the bottom bit of branch_target to prevent unaligned addresses.
-        branch_target = instr_vif.instr_cb.branch_target_id;
-        branch_target[0] = 1'b0;
-        counter = next_counter;
-      end
```
marnovandermaas, line 1250: Why is this no longer necessary?

marnovandermaas (commented):
Looks like a good start to me. Until the changes for RISCV-DV are added upstream, I would convert this PR into a draft PR.

hcallahan-lowrisc:
> Until the changes for RISCV-DV are added upstream, I would convert this PR into a draft PR.

Whoops, meant to do this to start with! Thanks :)

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
hcallahan-lowrisc: With the introduction of cosim, stepping the ISS and seeing identical behaviour gives us a lot of confidence that we are doing the correct thing. Hence we can remove a lot of this checking.
You're right in that it's not no longer necessary per se, but this checking is very brittle when you add things like illegal instructions etc. to the test stimulus. I tried to patch it but it quickly became very messy indeed, so easier to remove.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
hcallahan-lowrisc: See [other comment.](https://github.com/lowRISC/ibex/pull/1763/files#r944369402)

marnovandermaas: Thanks!

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
marnovandermaas: Thanks!

GregAC: Longer term I'd like to bring these kinds of checks back, a belt and braces approach where the test is doing some checking to ensure we're seeing the kinds of things we expect. However we need to do this in a far more flexible way, as @hcallahan-lowrisc said these checks are very brittle so better to remove them and rely entirely on cosim for now. That is certainly sufficient for our V2 goals.

vendor/google_riscv-dv/src/riscv_instr_stream.sv
```diff
@@ -237,9 +237,12 @@ class riscv_rand_instr_stream extends riscv_instr_stream;
         ((avail_regs.size() > 0) && !(SP inside {avail_regs}))) begin
       exclude_instr = {C_ADDI4SPN, C_ADDI16SP, C_LWSP, C_LDSP};
     end
-    // Post-process the allowed_instr and exclude_instr lists to handle
-    // adding ebreak instructions to the debug rom.
     if (is_in_debug) begin
+      // [Debug Spec v1.0.0-STABLE, p.50]
+      //  - ecall, mret, sret, uret have UNSPECIFIED behaviour while in debug mode
```
GregAC, line 242: We should probably decide what they do in Ibex, document it and verify it. However as this is strictly related to code run in debug mode it's not a big priority. Could you create an issue to track this? Possibly something we'll do for V3.

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
```diff
@@ -551,7 +551,17 @@ interface core_ibex_fcov_if import ibex_pkg::*; (
...
+      illegal_bins illegal =
+        {InstrCategoryOther, InstrCategoryNone, InstrCategoryOtherIllegal,
+         // [Debug Spec v1.0.0-STABLE, p.95]
+         // > dret is an instruction which only has meaning while Debug Mode
+         // Therefore, we can't single-step over it in normal mode.
+         InstrCategoryDRet,
+         // [Debug Spec v1.0.0-STABLE, p.50]
+         // > If the instruction being stepped over is wfi and would normally stall the hart,
+         // > then instead the instruction is treated as nop.
+         InstrCategoryWFI
```
GregAC, line 563: We definitely want to single step over a `WFI` to see this convert to NOP behaviour. When that happens does it come out as something other then `InstrCategoryWFI`? If so we may need a specific cover point for that scenario or a tweak to the instruction category classification code.

GregAC: We can address this as a follow up PR if it's non trivial.

GregAC (approved):
Looks good, thanks @hcallahan-lowrisc pleasing it's only need a few tweaks to get the coverage we need, though clearly there's been a lot of bug fixing and diagnosing going on elsewhere to get here.

hcallahan-lowrisc:
Rebased

dv/uvm/core_ibex/fcov/core_ibex_fcov_if.sv
hcallahan-lowrisc: I have removed this from the illegal bins in the latest changes, and we can successfully get coverage for it.

vendor/google_riscv-dv/src/riscv_instr_stream.sv
hcallahan-lowrisc: Issue created here : https://github.com/lowRISC/ibex/issues/1840

PR #1764, opened 2022-08-11
[rtl] Fix MaxOutstandingDSideAccessesCorrect assertion

GregAC:


hcallahan-lowrisc (approved):

ctopal (approved):
CI failure is about not being able to upload regression report for that specific config (logs seems fine when I looked manually), thanks for the fix!

GregAC:
Yeah think I'll just override the CI fail

PR #1766, opened 2022-08-11
[cosim] Pass PMP configuration through to spike

GregAC:


dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_cfg.sv
```diff
@@ -8,13 +8,17 @@ class core_ibex_cosim_cfg extends uvm_object;
   bit [31:0] start_mtvec;
   bit        probe_imem_for_errs;
   string     log_file;
+  bit        pmp_num_regions;
```
ctopal, line 11:
```suggestion
  bit  [31:0]      pmp_num_regions;
```

ctopal (approved):
`pmp_num_regions` is set as `bit` in cosim_cfg, that needs fixing but otherwise LGTM.

PR #1767, opened 2022-08-11
Pmp csr updates

GregAC:


ctopal (approved):
LGTM, thanks!

vogelpi (commented):
The `mseccfg` changes look good to me but I am not sure about the unused `pmpcfg/addr` registers: the spec explicitly says:
```
Changes from version 1.10 include:
...
- Made the unused pmpaddr and pmpcfg fields WARL, rather than WIRI.
...
```
AFAIK `WIRI` stands for write illegal read illegal. So actually, Spike might be wrong here.

ctopal:
@vogelpi: 
We also have this quote in privileged spec page 6: `Attempts to access a non-existent CSR raise an illegal instruction exception.`

Considering `PMPEnable=0` would make PMP related CSRs illegal, I thought the implementation was correct here.

GregAC:
> The mseccfg changes look good to me but I am not sure about the unused pmpcfg/addr registers: the spec explicitly says:
>> Changes from version 1.10 include:
>>  ...
>>  - Made the unused pmpaddr and pmpcfg fields WARL, rather than WIRI.
 >> - 

It also states:
> Implementations may implement zero, 16, or 64 PMP CSRs;

My interpretation of all this is you can implement any number (up to 64) of PMP regions but you must implement 0, 16 or 64 CSRs. Where you implement CSRs you just use WARL behaviour for the regions you don't implement of ignore writes read as 0.

rtl/ibex_cs_registers.sv
```diff
@@ -508,6 +504,16 @@ module ibex_cs_registers #(
...
       end
     endcase
+
+    if (!PMPEnable) begin
+      if (csr_addr inside {CSR_PMPCFG0,   CSR_PMPCFG1,   CSR_PMPCFG2,   CSR_PMPCFG3,
+                           CSR_PMPADDR0,  CSR_PMPADDR1,  CSR_PMPADDR2,  CSR_PMPADDR3,
+                           CSR_PMPADDR4,  CSR_PMPADDR5,  CSR_PMPADDR6,  CSR_PMPADDR7,
+                           CSR_PMPADDR8,  CSR_PMPADDR9,  CSR_PMPADDR10, CSR_PMPADDR11,
+                           CSR_PMPADDR12, CSR_PMPADDR13, CSR_PMPADDR14, CSR_PMPADDR15}) begin
+        illegal_csr = 1'b1;
```
marnovandermaas, line 514: Optional CSRs should have a different behavior to non-existent CSRs. Optional CSRs should be read zero and ignore writes according to slide 12 of this presentation: https://riscv.org/wp-content/uploads/2018/05/13.50-14.25pm-07.05.2018-Allen-Baum.pdf
So I think this should be changed to `csr_rdata_int = '0` if I'm not mistaken.

GregAC: Going by the wording of this slide these CSRs are 'non-existent' for configurations without PMP enabled (alternatively they could be treated as 'optional' again going by the wording of that slide which is what we used to do but this shifts things to be compatible with how spike implements things).

I did just have a patch accepted into spike which fixed up it's behaviour (it did one way for pmpaddr and another for pmpcfg): https://github.com/riscv-software-src/riscv-isa-sim/pull/1068 it got approved by Andrew Waterman so given that confident this is sound architecture.

marnovandermaas: Ok, makes sense, but then `mseccfg` should have the same behavior right? This is also an optional CSR.
Also, what about PMP address 16-63 and config 4-15, are we throwing an illegal CSR fault for those as well?

GregAC:
> Ok, makes sense, but then mseccfg should have the same behavior right?

Nope, `mseccfg` is a mandatory CSR, the fields within are optional.

> Also, what about PMP address 16-63 and config 4-15, are we throwing an illegal CSR fault for those as well?

Yes, as with all other unimplemented CSRs (note the default of that case sets `illegal_csr = 1'b1` so if we don't explicitly handle it it becomes illegal). This change does things a bit differently just to make the code more readable. Otherwise the currently nice compact block of code that returns the PMP CSR values ends up spreading out with lots of `begin`/`end` and many repetitions of `if (!PMPEnable) illegal_csr = 1'b1....`

GregAC:
Just had a quick scan through the ratified privleged spec and whilst `mseccfg` is described as optional it doesn't provide a definition of an optional CSR.

GregAC:
Ah and going by the 'what spike does is probably what's intended' where the specification isn't hugely clear `mseccfg` should indeed throw an illegal instruction exception where you don't implement it at all. Which is the behaviour we previously had so I'll drop that commit off this PR.

vogelpi (approved):
Thanks for providing further details @GregAC . I wasn't aware of the difference between unimplemented regions and unimplemented CSRs. Your reasoning is clear to me now and the fact that you were in contact with Andrew Waterman regarding this futher convinces me. LGTM!

marnovandermaas:
> Ah and going by 'what spike does is probably what's intended' where the specification isn't hugely clear `mseccfg` should indeed throw an illegal instruction exception where you don't implement it at all. Which is the behaviour we previously had so I'll drop that commit off this PR.

That sounds good to me. Thanks for the explanation.

rtl/ibex_cs_registers.sv
marnovandermaas: Thanks for the explanation below.

marnovandermaas (commented):
LGTM assuming you revert the MSECCFG behavior back to being illegal when PMP is not enabled.

marnovandermaas (approved):

PR #1771, opened 2022-08-14
[dv] Double timeout for an RTL run in regression flow

GregAC:
This takes it from 15 to 30 minutes

It reduces the total number of timeouts we see. We should still look at fixing up the long running tests but this will give us a healthier regression.

Draft for now whilst we look into impacts on regression run times.


GregAC:
Now we've got a faster runner for nightlies I think it's reasonable to increase this timeout

hcallahan-lowrisc (approved):

PR #1779, opened 2022-08-16
[dv] Add MCAUSE and MSTATUS to the riscv_csr_test

### LINKED ISSUE #1663: riscv_csr_test fails because we don't model MCAUSE properly
*Copying from https://github.com/lowRISC/ibex/issues/1658#issuecomment-1129412757 (a silly place for me to have put this text in the first place)*

Running the `riscv_csr_test` with
```
make -j4 ITERATIONS=1 SEED=1 TEST=riscv_csr_test VERBOSE=1 SIMULATOR=xlm WAVES=1
```
fails. It turns out that this is in the self-checking part, where we're writing and then reading back the MCAUSE register. The problem is that we describe the `mcause` CSR to riscv-dv with the following hunk of YAML:
```
# MCAUSE
- csr: mcause
  description: >
    Indicates trap cause
  address: 0x342
  privilege_mode: M
  rv32:
    - field_name: Interrupt
      description: >
        Indicates if trap caused by interrupt
      type: WARL
      reset_val: 0
      msb: 31
      lsb: 31
    - field_name: Exception Code
      type: WLRL
      reset_val: 0
      msb: 4
      lsb: 0
```

This might have been right once, but is no longer correct since Greg implemented internal interrupts. I don't think that `gen_csr_test.py` supports the behaviour that we've got for our MCAUSE field, which boils down to this code in `ibex_cs_registers.sv`:
```systemverilog
      // mcause: exception cause
      CSR_MCAUSE: csr_rdata_int = {mcause_q.irq_ext | mcause_q.irq_int,
                                   mcause_q.irq_int ? {26{1'b1}} : 26'b0,
                                   mcause_q.lower_cause[4:0]};
```

Here, we ostensibly have a 31-bit exception code field, but the top bits are all constrained to be equal.

@GregAC: What do you think we should do? A lame response might be to just use 6 bits and represent internal interrupts starting at 32, without the bit extension. Or we could split the field and just set the top bit? Or we could do some work to generalise the DV code to understand this sort of thing: for example, we could say that the field is 6 bits but the other 25 bits are some sort of mirror. But that means some DV work and (eek) getting a patch into riscv-dv.

My preference is probably to just set the top bit, so it ends up looking like this:
```systemverilog
      // mcause: exception cause
      CSR_MCAUSE: csr_rdata_int = {mcause_q.irq_ext | mcause_q.irq_int,
                                   mcause_q.irq_int,
                                   25'b0,
                                   mcause_q.lower_cause[4:0]};
```

> estimate 4
> remaining 2022-07-22 3
> remaining 2022-07-27 1

GregAC:
With the new WARL functionality for the RISCV-DV CSR test generator we
can bring back these CSRs into the test.

Fixes #1663

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
@@ -142,47 +142,52 @@
 # fields need to be constrained such that their value after every operation is within the allowed
 # range of values - too much complexity for the current format
 # MSTATUS
-#- csr: mstatus
-#  description: >
-#    Controls hart's current operating state
-#  address: 0x300
-#  privilege_mode: M
-#  rv32:
-#    - field_name: tw
-#      description: >
-#        Timeout Wait (WFI from U-mode will trap to M-mode)
-#      type: WARL
-#      reset_val: 0
-#      msb: 21
-#      lsb: 21
-#    - field_name: mprv
-#      description: >
-#        Modify Privilege (Loads and stores use MPP for privilege checking)
-#      type: WARL
-#      reset_val: 0
-#      msb: 17
-#      lsb: 17
-#    - field_name: mpp
-#      desription : >
-#        Previous privilege mode
-#      type: R
-#      reset_val: 0
-#      msb: 12
-#      lsb: 11
-#    - field_name: mpie
-#      description: >
-#        Previous value of interrupt-enable bit
-#      type: WARL
-#      reset_val: 1
-#      msb: 7
-#      lsb: 7
-#    - field_name: mie
-#      description: >
-#        M-mode interrupt enable
-#      type: WARL
-#      reset_val: 0
-#      msb: 3
-#      lsb: 3
+- csr: mstatus
+  description: >
+    Controls hart's current operating state
+  address: 0x300
+  privilege_mode: M
+  rv32:
+    - field_name: tw
+      description: >
+        Timeout Wait (WFI from U-mode will trap to M-mode)
+      type: WARL
+      reset_val: 0
+      msb: 21
+      lsb: 21
+    - field_name: mprv
+      description: >
+        Modify Privilege (Loads and stores use MPP for privilege checking)
+      type: WARL
+      reset_val: 0
+      msb: 17
+      lsb: 17
+    - field_name: mpp
+      desription : >
+        Previous privilege mode
+      type: WARL
+      reset_val: 0
+      msb: 12
+      lsb: 11
+      warl_legalize: |
+        if val_in == 0b11 or val_in == 0b00:
+          val_out = val_in
+        else:
+          val_out = 0b00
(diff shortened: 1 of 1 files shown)


dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
```diff
@@ -142,47 +142,52 @@
...
+      lsb: 17
+    - field_name: mpp
+      desription : >
+        Previous privilege mode
+      type: WARL
+      reset_val: 0
+      msb: 12
+      lsb: 11
+      warl_legalize: |
+        if val_in == 0b11 or val_in == 0b00:
```
marnovandermaas, line 173: This seems correct to me, since MPP is WARL and based on the following line in the spec:
> If the machine provides only U and M modes, then only a single hardware storage bit is required to represent either 00 or 11 in MPP.

dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
```diff
@@ -288,25 +293,34 @@
...
+  description: >
+    Indicates trap cause
+  address: 0x342
+  privilege_mode: M
+  rv32:
+    - field_name: Interrupt info
+      description: >
+        Indicates if trap caused by interrupt and if that interrupt was internal
+        or external
+      type: WARL
```
marnovandermaas, line 307: In the privileged spec section 3.1.15 it says that the exception code is from 30 to 0 and:
> The Exception Code is a WLRL field, so is only guaranteed to hold supported exception codes.

Where does this interrupt info come from? The interrupt field of mcause is only specified as being 1 bit wide in the spec.

GregAC: This is due to the Ibex internal interrupts (a concept that's particular to Ibex, it's not a RISC-V architecture concept): https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html#internal-interrupts

I think the WARL behaviour here is fine with regards to the spec as WLRL and WARL and more or less the same from the hardware's point of view:

> Implementations can return arbitrary bit patterns on the read of a WLRL field when the last write was of an illegal value, but the value returned should deterministically depend on the illegal written value and the value of the field prior to the write.

> Implementations can return any legal value on the read of a WARL field when the last write was of an illegal value, but the legal value returned should deterministically depend on the illegal written value and the architectural state of the hart

The only real difference is it's allowable to return an illegal value for a `WLRL` field when an illegal value is written.

The distinction matters more to software (where for WARL it can probe by writing arbitrary values to discover the legal ones for `WLRL` is should only ever write values it knows to be legal).

marnovandermaas: Ok, looks good to me based on that table.

marnovandermaas (approved):
LGTM, thanks!

PR #1780, opened 2022-08-18
[rtl] Cannot add M mode executable PMP regions when MML = 1

### LINKED ISSUE #1740: [rtl,epmp] Prevent addition of executable regions as required by the specification
The Smepmp specification states:

> Adding a rule with executable privileges that either is M-mode-only or a locked Shared-Region is not possible and such pmpcfg writes are ignored, leaving pmpcfg unchanged. This restriction can be temporarily lifted e.g. during the boot process, by setting mseccfg.RLB.****

This was overlooked during the original epmp implementation and must be added.

> estimate 2
> remaining 2022-18-08 1

GregAC:
A rule that allows M mode execution (either M mode only or shared M/U
mode) cannot be added when MML is set, unless RLB is also set.

Fixes #1740

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -124,6 +124,22 @@ module ibex_cs_registers #(
 
   import ibex_pkg::*;
 
+  // Is a PMP config a locked one that allows M-mode execution when MSECCFG.MML is set (either
+  // M mode alone or shared M/U mode execution)?
+  function automatic logic is_mml_m_exec_cfg(ibex_pkg::pmp_cfg_t pmp_cfg);
+    logic unused_cfg;
+    unused_cfg = ^{pmp_cfg.mode};
+
+    if (pmp_cfg.lock) begin
+      case ({pmp_cfg.read, pmp_cfg.write, pmp_cfg.exec})
+        3'b001, 3'b010, 3'b011, 3'b101: return 1'b1;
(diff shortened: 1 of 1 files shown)


rtl/ibex_cs_registers.sv
```diff
@@ -124,6 +124,22 @@ module ibex_cs_registers #(
...
 
+  // Is a PMP config a locked one that allows M-mode execution when MSECCFG.MML is set (either
+  // M mode alone or shared M/U mode execution)?
+  function automatic logic is_mml_m_exec_cfg(ibex_pkg::pmp_cfg_t pmp_cfg);
+    logic unused_cfg;
+    unused_cfg = ^{pmp_cfg.mode};
+
+    if (pmp_cfg.lock) begin
+      case ({pmp_cfg.read, pmp_cfg.write, pmp_cfg.exec})
+        3'b001, 3'b010, 3'b011, 3'b101: return 1'b1;
```
marnovandermaas, line 135: I've verified these values and they look correct to me based on Table 2.1 in: https://raw.githubusercontent.com/riscv/riscv-tee/main/Smepmp/Smepmp.pdf

marnovandermaas (approved):

marnovandermaas (approved):
LGTM, assuming checks pass once CI is fixed.

ctopal (approved):

PR #1781, opened 2022-08-18
[ci] Fix co-sim install

GregAC:
Due to a failure of mirror syncing the co-sim package cannot be
installed. This downloads it directly from the master server and
installs it manually.

### DIFF
--- a/ci/install-build-deps.sh
+++ b/ci/install-build-deps.sh
@@ -33,9 +33,9 @@ case "$ID-$VERSION_ID" in
     $SUDO_CMD apt-get update
 
     # Make spike-cosim repository available
-    curl -Ls https://download.opensuse.org/repositories/home:gac_lowrisc/xUbuntu_18.04/Release.key | $SUDO_CMD apt-key add -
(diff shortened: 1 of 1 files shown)


GregAC:
Not the best solution but it'll do for now. Perhaps we should move off OBS for the cosim package anyway. It does make pushing up new versions a bit of a pain.

Maybe we could setup some hosting on a new lowrisc subdomain, using a GCP bucket? Or even github (which hosts the main site).

GregAC:
Incidently the master server seems to do a redirect when you're not using a browser, so this gets redirected to the same bad mirror. However now there's no integrity check it at least installs it. We just don't get the updated version.

GregAC:
After a couple minutes more thought setting up some CI to automatically build the tip of the ibex-cosim branch when we push something new to it, packaging it up and writing it to a GCP bucket should be pretty easy. I'll investigate today. Still worth merging this though to unblock CI.

ctopal (approved):

marnovandermaas (approved):
LGTM assuming you've tested this.

GregAC:
> LGTM assuming you've tested this.

Yes, see also the CI passing as a demonstration.

PR #1782, opened 2022-08-18
[doc] Made values of mcause in doc 32 bits

marnovandermaas:


GregAC:
Skipping the CI as it's not relevant

PR #1786, opened 2022-08-18
[dv] Use lowRISC IP dir from imports instead of re-deriving it

marnovandermaas:
We'll need this for integrating into OpenTitan DV: https://github.com/lowRISC/opentitan/issues/11794


hcallahan-lowrisc (approved):
LGTM. I also just pushed another commit to this PR which would have caused problems when vendoring into OT. Feel free to merge if that makes sense to you! Thanks ;)

marnovandermaas (commented):
LGTM

PR #1788, opened 2022-08-19
Remove the final cosim flags from core_ibex uvm environment

### LINKED ISSUE #1566: Rip out all the non-cosim stuff from dv/uvm/core-ibex scripts
Depends on #1565 but, once we've got working "cosim tooling", we should get rid of the old way of doing things.

This doesn't include the work to rip out the old sequences for e.g. exceptions: we'll track that separately.

> estimate 2

hcallahan-lowrisc:
We are running with cosim by default now, and no longer support COSIM=0. Hence
this option and all downstream conditional paths are no longer required.

~~This also removes the +disable_cosim flag from all tests, as we are working towards all tests running under cosimulation anyway~~. For the bitmanipulation and csr test, any failures are now no-longer due to cosimulation and should be tracked separately. For bitmanipulation, it appears the tests for the `opentitan` configuration are all passing.

`+disable_cosim=1` has it's behaviour changed to de-escelate cosim mismatches from fatal-errors to info messages. This can be useful for debugging, and remains in place as the default for riscv_csr_test as that test has known failures.

Closes #1566

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
+++ b/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
@@ -136,7 +136,11 @@ class ibex_cosim_scoreboard extends uvm_scoreboard;
                             rvfi_instr.trap)) begin
         // cosim instruction step doesn't match rvfi captured instruction, report a fatal error
         // with the details
-        `uvm_fatal(`gfn, get_cosim_error_str())
+        if (cfg.relax_cosim_check) begin
+          `uvm_info(`gfn, get_cosim_error_str(), UVM_LOW)
+        end else begin
+          `uvm_fatal(`gfn, get_cosim_error_str())
+        end
       end
     end
   endtask: run_cosim_rvfi
--- a/dv/uvm/core_ibex/env/core_ibex_env.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_env.sv
@@ -10,9 +10,7 @@ class core_ibex_env extends uvm_env;
   ibex_mem_intf_response_agent   data_if_response_agent;
   ibex_mem_intf_response_agent   instr_if_response_agent;
   irq_request_agent              irq_agent;
-`ifdef INC_IBEX_COSIM
   ibex_cosim_agent               cosim_agent;
-`endif
   core_ibex_vseqr                vseqr;
   core_ibex_env_cfg              cfg;
   scrambling_key_agent           scrambling_key_agent_h;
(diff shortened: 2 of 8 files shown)


dv/uvm/core_ibex/env/core_ibex_env.sv
```diff
@@ -30,15 +28,7 @@ class core_ibex_env extends uvm_env;
     instr_if_response_agent = ibex_mem_intf_response_agent::type_id::
                            create("instr_if_response_agent", this);
     irq_agent = irq_request_agent::type_id::create("irq_agent", this);
-
-`ifdef INC_IBEX_COSIM
-    if (!cfg.disable_cosim) begin
```
GregAC, line 35: I'd like to leave this config knob in. It may prove useful for tests where we're having issues with the co-sim but we have other ways to check for pass/fail or want to use the test for some investigation/bring-up stuff (so want to examine waves and log and don't care about a cosim mismatch).

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -514,8 +514,6 @@
   iterations: 5
   no_iss: 1
   rtl_test: core_ibex_csr_test
-  sim_opts: >
-    +disable_cosim=1
```
GregAC, line 518: Leave it disabled on this test for now, it does self check after all.

We do want to enable it eventually for the CSR test but I suspect there's still a couple of CSR related mismatches and I'd prefer to have this test working (using it's self checking)).

dv/uvm/core_ibex/env/core_ibex_env.sv
hcallahan-lowrisc: I could change this config knob to instead de-escalate the cosim mismatch fatal error to an info or warning instead? That way the logs would contain all the mismatches. It would be nice to always construct the cosim model even if we don't use it's output, and I doubt the performance cost would be significant.

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
```diff
@@ -136,7 +136,11 @@ class ibex_cosim_scoreboard extends uvm_scoreboard;
                             rvfi_instr.trap)) begin
         // cosim instruction step doesn't match rvfi captured instruction, report a fatal error
         // with the details
-        `uvm_fatal(`gfn, get_cosim_error_str())
+        if (cfg.relax_cosim_check) begin
```
hcallahan-lowrisc, line 139: @GregAC e.g. like this

dv/uvm/core_ibex/env/core_ibex_env.sv
GregAC: Good idea, let's go for that

GregAC (approved):
LGTM thanks @hcallahan-lowrisc

PR #1792, opened 2022-08-22
[ci] Download cosim from lowRISC GCP bucket

GregAC:


GregAC:
Now anytime you push to the ibex_cosim branch of our riscv-isa-sim fork Azure will build spike for you and upload it to a GCP bucket. This should make cosim changes a little less painful, we might be able to use these pre-built versions for private CI too.

ci/install-build-deps.sh
```diff
@@ -61,12 +61,12 @@ case "$ID-$VERSION_ID" in
...
         xz-utils
-       # "ibex-cosim-$IBEX_COSIM_VERSION" \
 
-    # TODO: Remove this hack, workaround due to OBS mirrors not syncing
-    # correctly
-    wget https://download.opensuse.org/repositories/home:/gac_lowrisc/xUbuntu_18.04/amd64/ibex-cosim-0.4_0.4-1_amd64.deb
-    $SUDO_CMD dpkg -i ibex-cosim-0.4_0.4-1_amd64.deb
+    wget https://storage.googleapis.com/ibex-cosim-builds/ibex-cosim-$IBEX_COSIM_VERSION.tar.gz
+    $SUDO_CMD mkdir -p /tools/riscv-isa-sim
+    $SUDO_CMD chmod 777 /tools/riscv-isa-sim
```
hcallahan-lowrisc, line 67:
```suggestion
    $SUDO_CMD chmod 755 /tools/riscv-isa-sim
```

hcallahan-lowrisc (approved):
Looks good. Just a minor nit.

ci/install-build-deps.sh
GregAC: Think I'll leave this as is, we're using 777 elsewhere in the file and I prefer to things to remain consistent. As it's all being executed in a container than 777 mode shouldn't be an issue.

GregAC:
Thanks for the review @hcallahan-lowrisc

PR #1794, opened 2022-08-24
[testplan] Align milestone/stage terminology 

msfschaffner:
This vendors OT into Ibex and alignes the milestone -> stage terminology.


msfschaffner:
This will have to be vendored back into OT once merged.

tjaychen (approved):

PR #1805, opened 2022-09-07
Move to v1.12 Privileged specification

GregAC:


GregAC:
CI breakage is due to using an older cosim version. I've pushed the latest version to the ibex_cosim branch in our spike fork. I'm also going to update private CI so it pulls the binaries from the GCP bucket rather than using the toolnas.

hcallahan-lowrisc:
Do we have a coding-standard about mentioning the spec directly within code-comments? I notice you have remove some references to the spec here.

GregAC:
> Do we have a coding-standard about mentioning the spec directly within code-comments? I notice you have removed some references to the spec here.

We don't but maybe should, see the comment on https://github.com/lowRISC/ibex/pull/1805/commits/f22ca738001af4083026b1f608958ed73f9f889b for more details

dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
```diff
@@ -17,6 +17,9 @@ class ibex_asm_program_gen extends riscv_asm_program_gen;
       MVENDORID,
       MARCHID,
       MHARTID,
+      MCONFIGPTR,
+      MENVCFG,
```
marnovandermaas, line 21: What about `MENCFGH`?

dv/uvm/core_ibex/riscv_dv_extension/riscv_core_setting.tpl.sv
```diff
@@ -120,7 +120,10 @@ const privileged_reg_t implemented_csr[] = {
     MIMPID,           // Implementation ID
     MARCHID,          // Architecture ID
     MHARTID,          // Hardware thread ID
-    MSTATUS,          // Machine status
+    MCONFIGPTR,       // Machine configuration pointer
+    MENVCFG,          // Machine environment configuration (lower 32 bits)
```
marnovandermaas, line 124: What about `MENVCFGH`?

rtl/ibex_controller.sv
```diff
@@ -231,7 +231,7 @@ module ibex_controller #(
   // Is there an instruction in ID or WB that has yet to complete?
   assign id_wb_pending = instr_valid_i | ~ready_wb_i;
 
-  // Exception/fault prioritisation is taken from Table 3.7 of Priviledged Spec v1.11
+  // Logic to determine which exception takes priority where multiple are possible.
```
marnovandermaas, line 234: Table 3.7 still contains the priority in version 1.12

dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
GregAC: I'm going to add that to the CSR waiver, much like MSECCFGH it's not implemented correctly in spike. I could sort it out but I don't think it's worth the time right now.

dv/uvm/core_ibex/riscv_dv_extension/riscv_core_setting.tpl.sv
GregAC: I'm going to add that to the CSR waiver, much like MSECCFGH it's not implemented correctly in spike. I could sort it out but I don't think it's worth the time right now.

rtl/ibex_controller.sv
GregAC: I'd prefer to leave these references out of comments, as explained in the commit message here: https://github.com/lowRISC/ibex/pull/1805/commits/f22ca738001af4083026b1f608958ed73f9f889b I think it's fairly brittle and it's a bit odd we've got a couple of these kinds of comments but there's plenty more places you could give specific spec references for the things implemented.

Tagging everything with specific spec references is one way we could go but it bit-rots far too easily. I prefer more general comments that aren't specific spec pointers. Those interested can always find it for themselves.

marnovandermaas: Ok, that makes sense.

dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
marnovandermaas: Ok, as long as it is tracked in the waiver, I think it is fine to leave it out here. Otherwise I would have suggested adding a comment on this.

dv/uvm/core_ibex/riscv_dv_extension/riscv_core_setting.tpl.sv
marnovandermaas: Ok thanks.

marnovandermaas (approved):
LGTM

GregAC:
CI has passed, see here: https://dev.azure.com/lowrisc/lowrisc-private/_build/results?buildId=93290&view=results

Done as a manual run because changes were needed on the private CI side, so successful run doesn't end up getting reflected here.

Overriding CI failure here so I can merge.

PR #1807, opened 2022-09-09
Scr key valid csr

GregAC:
Before this is merged the cosim needs an update, see this PR here: https://github.com/lowRISC/riscv-isa-sim/pull/14

When that is merged we can update the cosim version used here allowing CI to pass.


ctopal (approved):
LGTM overall, thanks!

A question: Do you think if it makes sense in the future to have all state variable updating methods into a single wrapper? Considering they all get called in every rvfi_item transaction in the scoreboard I think it would make things less crowded in the DPI call side without changing the functionality, not sure about the disadvantages of it though. I am talking about something like 

```c++
riscv_cosim_set_state(Cosim *cosim, const svBitVecVal *mip,  svBit nmi, svBit debug_req){
  assert(cosim);

  cosim->set_mip(mip[0]);
  cosim->set_nmi(nmi);
  cosim->set_debug_req(debug_req);
}
```

GregAC:
> A question: Do you think if it makes sense in the future to have all state variable updating methods into a single wrapper

Yes definitely, I will open an issue for this.

Ideally we would wrap up all of the state that we feed into the cosim along with the instruction PC/reg write info into some kind of structure or packet so we don't need lots of different rvfi IO and DPI calls.

Issue #1761, opened 2022-08-11
[dv] Add instructions to enable/disable random instruction generation to existing tests


GregAC:
Ensure CSR writes that will enable and disable the random instruction generator are emitted for some tests. This is mostly dealt with via on-going CSR work. The task here is just to use the new functionality to stimulate the random instruction generator. Some extra work may be required to ensure the randomly generated instructions don't get fed into co-sim checking (but we do ensure they have no effect on architectural state).

> estimate 2

GregAC:
#1808 gives us the basis of this (random `CPUCTRLSTS` writes will enable/disable random instruction generation).

Currently the dummy instructions are simply not fed through the RVFI interface so the cosim is unaware of their existence. We should add some extra checking to ensure they don't write to the register file (e.g. some assertion that states that if writeback writes to the RF then the first RVFI stage must be valid, as for dummy instructions we prevent it from entering the RVFI stages).

andreaskurth:
According to @GregAC, what basically remains to be done here is adding the assertions that check dummy instructions don't cause register writes. #1853 deals with this.

andreaskurth:
Closing as complete with #1853 merged.

Issue #1760, opened 2022-08-11
[dv] Verify double fault behaviour


GregAC:
Ibex has a double fault detector. A checker should be created to ensure this functions correctly and double faults should be stimulated in some tests.

Various work has been happening in this area, as a result this issue is less effort than originally estimated.
old estimate: 4
> estimate 2
> remaining 2022-11-8 1

GregAC:
Added an estimate to this. The cosim side of things is handled by https://github.com/lowRISC/ibex/pull/1808 what remains is

- Produce a double fault in some test (either a new dedicated one or add them to some existing test that already generates exceptions). A read of `cpuctrlsts` that will see the `double_fault_seen` bit being set should be included. The ideal test would clear this bit, recover from the double fault and continue (some of time, other times some loop of faulting that times out would be good).
- A checker for the `double_fault_seen_o` output. This could be done as an assertion along with some tracking logic.

johngt:
Taking @marnovandermaas off this task assignment for now.

GregAC:
@GregAC to provide more detailed guidance on checking here

GregAC:
Reassigning to @GregAC I should be able to address this rapidly freeing @hcallahan-lowrisc for other tasks

GregAC:
PR https://github.com/lowRISC/ibex/pull/1916/files to add the double fault checker

GregAC:
Closed by https://github.com/lowRISC/ibex/pull/1916

Issue #1654, opened 2022-05-12
[dv] Add instructions to enable/disable data independent timing to RISC-V DV generated code


GregAC:
Our DV environment never enables data independent timing or random instruction generation. These are controlled the CPUCTRL CSR, We should emit some instructions that toggle it.

Whilst strictly security related features they impacts the basic functionality of the pipeline so we should have some basic testing of them in V2. Consideration on any specific scenarios and coverpoints for this feature can be dealt with in V2S.

> estimate 4
> remaining 2022-09-09 2
> remaining 2022-10-25 1

GregAC:
https://github.com/lowRISC/ibex/pull/1808 implements the basis of this, namely random writes to the `CPUCTRLSTS` CSR along with the cosim changes needed to allow this to pass regressions.

That PR only enables `CPUCTRLSTS` writes in the random instruction test. We should expand this to more tests (ideally all of them but there may be some places we can't do it for whatever reason). To do so add `0x7c0` (the CSR address is `CPUCTRLSTS` to `default_include_csr_write` in `dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv`. `+gen_all_csrs_by_default=1` will also need setting for every test we want a `CPUCTRLSTS` write in (look at how riscv_rand_instr_test is setup in the testlist yaml file).

GregAC:
https://github.com/lowRISC/ibex/pull/1882/files fixes up the CPUCTRLSTS writes so they occur in `riscv_rand_instr_test`. This enables and disables data independent timing during that test.

PR #1808, opened 2022-09-09
Cpuctrlsts write

GregAC:
This draft PR gives us the basis of https://github.com/lowRISC/ibex/issues/1654 and https://github.com/lowRISC/ibex/issues/1761. We should look to expand the number of tests we're writing cpuctrlsts so the things it controls get randomly changed under more scenarios.

This also gives us part of: https://github.com/lowRISC/ibex/issues/1760 as it implements the double fault behaviour for cosim.

Also note it's based on https://github.com/lowRISC/ibex/pull/1807 so that needs merging first

### DIFF
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -233,6 +233,9 @@ bool SpikeCosim::step(uint32_t write_reg, uint32_t write_reg_data,
       if (!check_sync_trap(write_reg, pc, initial_spike_pc)) {
         return false;
       }
+
+      handle_cpuctrl_exception_entry();
+
       // This is all the checking possible when consider a
       // synchronously-trapping instruction that never retired.
       return true;
@@ -242,9 +245,13 @@ bool SpikeCosim::step(uint32_t write_reg, uint32_t write_reg_data,
   // We reached a retired instruction, so check spike and the dut behaved
   // consistently.
 
-  if (!sync_trap && pc_is_mret(pc) && nmi_mode) {
-    // Do handling for recoverable NMI
-    leave_nmi_mode();
+  if (!sync_trap && pc_is_mret(pc)) {
+    change_cpuctrlsts_sync_exc_seen(false);
+
+    if (nmi_mode) {
+      // Do handling for recoverable NMI
+      leave_nmi_mode();
+    }
   }
 
   if (pending_iside_error) {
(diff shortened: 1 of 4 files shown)


rtl/ibex_core.sv
```diff
@@ -1279,7 +1279,7 @@ module ibex_core import ibex_pkg::*; #(
     assign rvfi_trap_id = id_stage_i.controller_i.id_exception_o;
     assign rvfi_trap_wb = id_stage_i.controller_i.exc_req_lsu;
     // WB is instantly done in the tracking pipeline when a trap is progress through the pipeline
```
marnovandermaas, line 1281: Nit: update this comment to reflect the change.

dv/cosim/spike_cosim.cc
```diff
@@ -440,6 +447,34 @@ void SpikeCosim::leave_nmi_mode() {
...
+  if (old_sync_exc_seen) {
+    set_cpuctrlsts_double_fault_seen();
+  }
+}
+
+bool SpikeCosim::change_cpuctrlsts_sync_exc_seen(bool flag) {
+  bool old_flag = false;
+  uint32_t cpuctrlsts = processor->get_csr(CSR_CPUCTRLSTS);
+
+  if (cpuctrlsts & 0x40) {
```
marnovandermaas, line 462: Nit: put a comment here that the double fault seen flag is bit 6.

dv/cosim/spike_cosim.cc
```diff
@@ -440,6 +447,34 @@ void SpikeCosim::leave_nmi_mode() {
...
+
+  cpuctrlsts = (cpuctrlsts & 0x1bf) | (flag ? 0x40 : 0);
+  processor->put_csr(CSR_CPUCTRLSTS, cpuctrlsts);
+
+  return old_flag;
+}
+
+void SpikeCosim::set_cpuctrlsts_double_fault_seen() {
+  uint32_t cpuctrlsts = processor->get_csr(CSR_CPUCTRLSTS);
+  cpuctrlsts = (cpuctrlsts & 0x17f) | 0x80;
```
marnovandermaas, line 474: Nit: put a comment here that double fault seen flag is bit 7.

marnovandermaas: Also, is the mask can be `0x1ff`

GregAC: It can, but I prefer to leave it as 0x17f, makes it more explicit we're ignoring the value of double_fault_seen here

rtl/ibex_core.sv
GregAC: I think the comment still stands, this change is correcting a bug in the operation.

Issue #1732, opened 2022-07-28
[dv,cosim] Implement internal interrupts in co-sim


GregAC:
Add support to cosim for internal interrupts (documented at https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html#internal-interrupts)

> estimate 8
> remaining 2022-10-06 4
> remaining 2022-10-14 2
> remaining 2022-11-01 1

ctopal:
>remaining 2022-10-14 2

GregAC:
@GregAC to add write suppression on ECC failures on load to co-sim environment

GregAC:
I have a set of commits on top of Canberk's work in https://github.com/lowRISC/ibex/pull/1811 but no PR yet. Once that is merged this will be complete.

GregAC:
Completed by https://github.com/lowRISC/ibex/pull/1907

PR #1811, opened 2022-09-20
[dv] Integrity error generation in memory busses

### LINKED ISSUE #1731: [dv] Add ecc error generation to memory agent
We need to stimulate ECC errors. The incoming bus data has integrity bits. The driver should provide incorrect versions of these at configurable random internals in a similar manner to how bus errors work.

Also update co-simulation to support internal interrupts from ECC errors.

> estimate 8
> remaining 2022-09-20 3
> remaining 2022-10-14 1

ctopal:
First commit changes the old `memory_error_seq` so that it uses the new sequence library. 
Second commit improves memory interface to have an integrity error response on demand, also changes the `memory_error_seq` to inject integrity errors as well as bus errors.

Not sure but we are seeing an load integrity related internal interrupt which then triggers `alert_major_bus_o`. So, I turned off the assertion that checks no alert is triggering in the testbench.

Note that this is part one of the two part task of having internal interrupts generated and modelled in cosim environment.

Resolves #1731 
Opens up the way for #1732


ctopal:
This does not take into account the type of memory request. That means we won't differentiate between seeing an integrity error on the memory interface while doing a store or a load (either way if rdata integrity bits are faulty, we're seeing an internal NMI -is this okay? WDYT @GregAC ). I believe this might not be ideal and we should only send an integrity error while we are doing a read. Should I change it?

GregAC:
> That means we won't differentiate between seeing an integrity error on the memory interface while doing a store or a load (either way if rdata integrity bits are faulty, we're seeing an internal NMI -is this okay? WDYT @GregAC ). I believe this might not be ideal and we should only send an integrity error while we are doing a read. Should I change it?

With the current Ibex RTL integrity errors on `rdata` for writes are ignored. However we're planning on going back to checking it (see https://github.com/lowRISC/opentitan/issues/14611) so the current behaviour is fine.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1266,6 +1266,7 @@ class core_ibex_mem_error_test extends core_ibex_directed_test;
     memory_error_seq_h = memory_error_seq::type_id::create("memory_error_seq_h", this);
     `uvm_info(`gfn, "Running core_ibex_mem_error_test", UVM_LOW)
     memory_error_seq_h.vseq = vseq;
+    `DV_ASSERT_CTRL_REQ("NoAlertsTriggered", 1'b0)
```
GregAC, line 1269: I think this is a little too blunt. We wouldn't want alerts to occur on bus errors for example.

Perhaps randomly choose whether or not a test will generate integrity errors at all and then only disable this assertion if we'll be generating integrity errors?

GregAC (commented):
This is looking good @ctopal just a few tweaks til it's ready to go.

GregAC:
Oh and we do need a check that an integrity error triggers an alert. This can be done as an assert for the data side. Instruction side requires a little more thought (a fetch may not actually get executed so the integrity error doesn't get consumed and doesn't trigger an alert). Though strictly that's a V2S task. So for this PR at least we don't need to worry about it, but we must not forget about it (best to create an issue to track).

ctopal:
While simulating with this version I found a minor mismatch regarding two back to back integrity errors (with load instructions). RTL side does not take the faulty data from memory into the registers, Spike does.

I might need help about how to tell Spike to not go into a fault but also not process the retired instruction.

ctopal:
Test pass rate for this memory test is now around 60%. We hit internal NMI coverpoints though. I suspect almost all failures are related with the issue that I mentioned in the previous message.

GregAC:
As discussed in the Ibex DV Sync I think we should avoid generating integrity errors in the `riscv_mem_error_test` and introduce a new `riscv_mem_error_intg_test` that does have integrity error. Just add a config option for `core_ibex_mem_error_test` that enables generating integrity errors and a new entry to the testlist that turns that on.

With that done this can pass CI and we can get it merged. If the new test pass rate is 60% that should do for our V2 targets and we can fix up remaining issues in V3.

GregAC (approved):

hcallahan-lowrisc:
I think it might be cleaner to split out the cosim/NMI changes into a separate PR? Those changes are quite significant on their own. And could you add some more detail to the PR comment, the messages for 7222d4a012ba1b28b09cb5bd4d0f39cd5f70e9ac and 66f3e2dfc4416ba8d831f4200d525955f6fd518b are a good start but I think some more context would be really useful in the future.

ctopal:
Thanks @hcallahan-lowrisc for having a look, I updated the commit messages for almost all of them and dropped that one unnecessary commit about catching external RVFI signals -that one should belong to another PR I agree. Latest changes should make the mem_test pass in CI (as can be seen in Ibex CI + Private CI).

Might worth just another quick look @GregAC since now it includes cosim integration too.

dv/cosim/cosim.h
```diff
@@ -95,6 +95,13 @@ class Cosim {
   // When an NMI is due to be taken that will occur at the next call of `step`.
   virtual void set_nmi(bool nmi) = 0;
 
+  // Set the state of the internal NMI (non-maskable interrupt) line.
+  // Behaviour wise this is almost as same as external NMI case explained at
+  // set_nmi method. Main difference to consider is that this interrupt is
+  // synchronous because it comes from the integrity checking logic inside the
```
GregAC, line 101: The internal interrupt is effectively synchronous as currently implemented but it should be consider asynchronous (in particular we don't make any guarantees about when it'll occur relative to a memory access). So I think just drop the final statement here or tweak the language so it doesn't call it 'synchronous'

dv/cosim/spike_cosim.cc
```diff
@@ -200,7 +200,10 @@ bool SpikeCosim::step(uint32_t write_reg, uint32_t write_reg_data,
 
   if (processor->get_state()->last_inst_pc == PC_INVALID) {
     if (!(processor->get_state()->mcause->read() & 0x80000000) ||
-        processor->get_state()->debug_mode) { // (Async-Traps are disabled in debug mode)
+        (processor->get_state()->mcause->read() ==
```
GregAC, line 203: Internal NMIs should be handled as an async trap here, had you encountered difficulties using internal NMI with the sync trap here?

ctopal: I tested removing it now, it didn't changed the pass rate. Possibly it was some other problem that lead me to believe this was the correct behaviour. I'll change it to behave just as another NMI now.

ctopal:
I think this PR increases our test time for mem_error_test (after changing it with the new sequence we probably are injecting way more memory errors) and tips it over 30 mins. What should we do about it?

GregAC:
> I think this PR increases our test time for mem_error_test (after changing it with the new sequence we probably are injecting way more memory errors) and tips it over 30 mins. What should we do about it?

Let's try dropping the frequency of memory errors and see what that does to the run time.

ctopal:
@GregAC can you have a look at https://github.com/lowRISC/ibex/pull/1811/commits/66d8d0fc0b087fa48f74cf8e0d343b582e2ad8e9 to see if this makes sense?

GregAC:
> @GregAC can you have a look at [66d8d0f](https://github.com/lowRISC/ibex/commit/66d8d0fc0b087fa48f74cf8e0d343b582e2ad8e9) to see if this makes sense?

It does make sense, though having reviewed our discussion I think we want to leave the RTL behaviour as is. This means we need to instead alter cosim to not do the register writes on ECC failures. I will take a look at this.

GregAC:
> > @GregAC can you have a look at [66d8d0f](https://github.com/lowRISC/ibex/commit/66d8d0fc0b087fa48f74cf8e0d343b582e2ad8e9) to see if this makes sense?
> 
> It does make sense, though having reviewed our discussion I think we want to leave the RTL behaviour as is. This means we need to instead alter cosim to not do the register writes on ECC failures. I will take a look at this.

I've worked out how I'm going to do this, just need to implement it. I'm going to use a method that doesn't require spike changes (as otherwise these would be quite invasive to suppress a register write on a load).

GregAC:
I've added some stuff and reworked commits here to give https://github.com/lowRISC/ibex/pull/1811 which supersedes this PR and should get the memory integrity test running and passing.

PR #1812, opened 2022-09-21
[rtl] Switch FF RF to use unpacked arrays

GregAC:
Previously it had two packed dimensions. In general we prefer the use of unpacked arrays for this kind of usage. In particular this had an impact on trace viewing in GTKWave, the two dimensions were flattened into a single large bus which made determining individual register values tricky.

I did some git archeology and it looks like it has been this way since Ibex was created from Zero-riscy. I can't see any reason it needs to be two packed dimensions. @msfschaffner adding you as a reviewer to comment in case there's something I'm missing.

### DIFF
--- a/rtl/ibex_register_file_ff.sv
+++ b/rtl/ibex_register_file_ff.sv
@@ -45,9 +45,8 @@ module ibex_register_file_ff #(
   localparam int unsigned ADDR_WIDTH = RV32E ? 4 : 5;
   localparam int unsigned NUM_WORDS  = 2**ADDR_WIDTH;
 
-  logic [NUM_WORDS-1:0][DataWidth-1:0] rf_reg;
-  logic [NUM_WORDS-1:1][DataWidth-1:0] rf_reg_q;
-  logic [NUM_WORDS-1:0]                we_a_dec;
+  logic [DataWidth-1:0] rf_reg   [NUM_WORDS];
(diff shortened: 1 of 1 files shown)


marnovandermaas (commented):
The only downside to this approach is that the index in `rf_reg` and `rf_reg_q` are no longer the same, but the benefit you mention of looking at waveforms might outweigh this downside.

msfschaffner (approved):
This array dimension switch seems fine to me.
Regarding `rf_req_q`, is there a reason why we can't just tie this off in a similar way as in OTBN?
https://github.com/lowRISC/opentitan/blob/5b95d019b95423b14e590a285475abf4d99c50f6/hw/ip/otbn/rtl/otbn_rf_base_ff.sv#L44-L52

The extra indexing operation (`i - 1`) seems a bit unnecessary and may be confusing when looking at array indices in the waves...

GregAC:
> Regarding `rf_req_q`, is there a reason why we can't just tie this off in a similar way as in OTBN? https://github.com/lowRISC/opentitan/blob/master/hw/ip/otbn/rtl/otbn_rf_base_ff.sv#L44-L52
> 
> The extra indexing operation (`i - 1`) seems a bit unnecessary and may be confusing when looking at array indices in the waves...

Good suggestion, I have made that change.

PR #1813, opened 2022-09-21
[rtl] Don't cache instructions in debug mode

### LINKED ISSUE #1472: [bug, dbg] Don't cache instruction whilst in debug mode
The debug module used in conjunction with Ibex implements its functionality by having Ibex execute code that dynamically changes depending upon the commands issued to it. When the icache is present caching the memory region used for this will break the debug module.

A simple fix is to disable instruction caching whilst Ibex is in debug mode.

> estimate 2

GregAC:
RISC-V debug modules may utilise dynamically changing code. Don't cache any instructions in debug mode to correctly support this.

Fixes #1472

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -69,6 +69,7 @@ module ibex_cs_registers #(
 
   // debug
   input  logic                 debug_mode_i,
+  input  logic                 debug_mode_entering_i,
   input  ibex_pkg::dbg_cause_e debug_cause_i,
   input  logic                 debug_csr_save_i,
   output logic [31:0]          csr_depc_o,
@@ -1650,7 +1651,8 @@ module ibex_cs_registers #(
   assign cpuctrlsts_part_wdata.double_fault_seen = cpuctrlsts_part_wdata_raw.double_fault_seen;
   assign cpuctrlsts_part_wdata.sync_exc_seen     = cpuctrlsts_part_wdata_raw.sync_exc_seen;
 
-  assign icache_enable_o = cpuctrlsts_part_q.icache_enable;
+  assign icache_enable_o =
+    cpuctrlsts_part_q.icache_enable & ~(debug_mode_i | debug_mode_entering_i);
 
   ibex_csr #(
     .Width     ($bits(cpu_ctrl_sts_part_t)),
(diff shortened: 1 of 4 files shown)


marnovandermaas (commented):
I see that `debug_mode_entering_o` is set to 1, but is never set to 0. Does that mean that debug mode cannot be disabled?
As far as I can tell this is also the only difference between the value of `debug_mode_d` and `debug_mode_entering_o`.

marnovandermaas:
> I see that `debug_mode_entering_o` is set to 1, but is never set to 0. Does that mean that debug mode cannot be disabled? As far as I can tell this is also the only difference between the value of `debug_mode_d` and `debug_mode_entering_o`.

@ctopal explained that `debug_mode_entering_o` is set to zero in an `always_comb` block so when it is set to 1 it is only high for one cycle.

rtl/ibex_cs_registers.sv
```diff
@@ -1618,7 +1619,7 @@ module ibex_cs_registers #(
   assign cpuctrl_wdata.double_fault_seen = cpuctrl_wdata_raw.double_fault_seen;
   assign cpuctrl_wdata.sync_exc_seen     = cpuctrl_wdata_raw.sync_exc_seen;
 
-  assign icache_enable_o = cpuctrl_q.icache_enable;
+  assign icache_enable_o = cpuctrl_q.icache_enable & (~debug_mode_i | debug_mode_entering_i);
```
marnovandermaas, line 1622: I don't really see the need for `debug_mode_entering_i` here. Based on the waveforms @ctopal  showed me this is the normal waveform:
`.........debug_mode_i = __--`
`debug_mode_entering_i = _-__`
`......icache_enable_o        = --__`

In that scenario, the value of `debug_mode_entering_i` does not affect `icache_enable_o`. Am I overlooking something?

ctopal: I think we can also use `enter_debug_mode` if we want to disable it earlier, right?

GregAC: Ah, this line is incorrect it should be

```systemverilog
assign icache_enable_o = cpuctrl_q.icache_enable & ~(debug_mode_i | debug_mode_entering_i);
```

The `debug_mode_entering_i` exists to cover the first cycle of debug mode such that the first instruction of debug mode is always fetched with `icache_enable_o` set low. Without it there's a risk we may be requesting the fetch of the first instruction of debug mode with `icache_enable_o` asserted (as we can request the fetch the same cycle we decide to enter debug mode, so `debug_mode_i` is only asserted the cycle following that).

marnovandermaas (approved):
LGTM

PR #1815, opened 2022-09-26
[dv] Fix timeout issues

GregAC:
core_ibex_directed_test has a 'disable fork' that was killing processes that were running sequences. Another part of the testbench waits for those sequences to finish. When this 'disable fork' happens too early the sequences are killed before they finish so the testbench never terminated and times out. Instead ensure the sequences have finished before doing the 'disable fork'.


marnovandermaas (approved):
From just looking at the code, LGTM.

PR #1816, opened 2022-09-27
[rtl] Change how we record debug causes

### LINKED ISSUE #1772: Cosim mismatch when reading DCSR.cause
## Observed Behavior
Test failures with the following error : 
```
UVM_FATAL ibex_cosim_scoreboard.sv(138) @ 4049807: Cosim mismatch Register write data mismatch to x22 DUT: 40000100 expected: 400000c0
```

This fails upon reading from `DCSR.cause`
In this case : 

image
(IBEX)

image
(SPIKE)

Observing the waveform below, we can see that `debug_req_i` is de-asserted shortly after `debug_mode_i` goes low due to leaving debug mode from a dret. We do re-enter debug mode, but the cause can be seen to be recorded as `DBG_CAUSE_STEP`. In spike, we read back `DBG_CAUSE_HALTREQ` at the same instruction, hence the mismatch.

## Expected Behavior

Ibex should record `DCSR.cause` as DBG_CAUSE_HALTREQ.

I think this might be due to the [ibex_controller.sv ](https://github.com/lowRISC/ibex/blob/2f8dfa9dfe2d938228118ed8ab6a543804d0c9ae/rtl/ibex_controller.sv#L662)state machine sampling the `debug_req_i` signal to determine the correct `cause` one cycle later than the decision to enter debug_mode.

```
cd dv/uvm/core_ibex/
make TEST=riscv_debug_basic_test ITERATIONS=1 WAVES=1 SEED=9892 -j12 -ik
```
image

> estimate 8
> remaining 2022-09-27 4
> remaining 2022-10-12 1

ctopal:
This commit changes when we catch the debug causes. Since `debug_cause_o` only gets saved when `csr_save_cause_o` is high, it would work if we change the cause with a mux that is connected to the input signals.

Resolves #1772

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -114,6 +114,7 @@ module ibex_controller #(
 
   logic nmi_mode_q, nmi_mode_d;
   logic debug_mode_q, debug_mode_d;
+  dbg_cause_e debug_cause_d, debug_cause_q;
   logic load_err_q, load_err_d;
   logic store_err_q, store_err_d;
   logic exc_req_q, exc_req_d;
@@ -417,6 +418,22 @@ module ibex_controller #(
 
   assign unused_irq_timer = irqs_i.irq_timer;
 
+  assign debug_cause_d = trigger_match_i   ? DBG_CAUSE_TRIGGER :
+                         ebreak_into_debug ? DBG_CAUSE_EBREAK  :
+                         debug_req_i       ? DBG_CAUSE_HALTREQ :
+                         do_single_step_d  ? DBG_CAUSE_STEP    :
+                                             DBG_CAUSE_NONE ;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      debug_cause_q <= DBG_CAUSE_NONE;
+    end else begin
+      debug_cause_q <= debug_cause_d;
+    end
+  end
+
+  assign debug_cause_o = debug_cause_q;
+
   /////////////////////
   // Core controller //
   /////////////////////
@@ -453,7 +470,6 @@ module ibex_controller #(
     flush_id               = 1'b0;
 
     debug_csr_save_o       = 1'b0;
-    debug_cause_o          = DBG_CAUSE_EBREAK;
     debug_mode_d           = debug_mode_q;
     nmi_mode_d             = nmi_mode_q;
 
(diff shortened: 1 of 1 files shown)


rtl/ibex_controller.sv
```diff
@@ -417,6 +418,22 @@ module ibex_controller #(
 
   assign unused_irq_timer = irqs_i.irq_timer;
 
+  assign debug_cause_d = trigger_match_i   ? DBG_CAUSE_TRIGGER :
+                         ebreak_into_debug ? DBG_CAUSE_EBREAK  :
```
hcallahan-lowrisc, line 422: @ctopal I think we agreed that `ebreak_into_debug` wasn't the right signal to use here?

GregAC: `ebreak_into_debug` indicates an ebreak should enter debug, not that it's happening right now.

rtl/ibex_controller.sv
```diff
@@ -417,6 +418,22 @@ module ibex_controller #(
...
+                         debug_req_i       ? DBG_CAUSE_HALTREQ :
+                         do_single_step_d  ? DBG_CAUSE_STEP    :
+                                             DBG_CAUSE_NONE ;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      debug_cause_q <= DBG_CAUSE_NONE;
+    end else begin
+      debug_cause_q <= debug_cause_d;
```
GregAC, line 431: We may be better off only altering `debug_cause_q` when we're actually doing a debug entry, could add an `if(enter_debug_mode)` here?

GregAC (commented):
Looks good but we'll need to fix the `ebreak_into_debug` thing @hcallahan-lowrisc pointed out before merging.

rtl/ibex_controller.sv
hcallahan-lowrisc: Resolved in latest commit

rtl/ibex_controller.sv
hcallahan-lowrisc: The signal `enter_debug_mode` does not go high when we go into debug_mode due to an `EBREAK`. So it's not quite suitable here.
So far testing this PR as-is resolves the issue linked (https://github.com/lowRISC/ibex/issues/1772), but I'm going to run a few more regressions locally before merging to see if some other qualifying signal might be needed.

GregAC: Ah ok, could add an explicit enable signal

```systemverilog
    if (!rst_ni) begin
      debug_cause_q <= DBG_CAUSE_NONE;
    end else if (debug_cause_en) begin
      debug_cause_q <= debug_cause_d;
```

Then add logic to drive the enable within the controller FSM

GregAC (approved):
LGTM.

The enable point would be nice, but certainly not essential. Unless it's something you can do quickly just merge this and open an issue to flag it as an improvement.

Issue #1765, opened 2022-08-11
[dv] Investigate issues in mixed interrupts/traps/debug tests


GregAC:
> estimate 16
> remaining 2022-09-28 8

ctopal:
>remaining 2022-28-09 8

johngt:
@GregAC - Think that @ctopal and @hcallahan-lowrisc will need some guidance on the efforts required for this + https://github.com/lowRISC/ibex/issues/1686

GregAC:
This test is seeing some passes, failures are generally intricate co-sim mismatch bugs (but not RTL bugs) which take significant time to resolve so moving this to V3

GregAC:
Closing as this will be picked up with general regression fixing required for V3

PR #1824, opened 2022-09-30
Irq fixes

GregAC:
I had a go at adding interrupts to the single step test. I ran into various failures but I suspect I'm chasing the same bugs as @ctopal and @hcallahan-lowrisc are in https://github.com/lowRISC/ibex/issues/1765 so I won't continue that work until #1765 is resolved.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -401,8 +401,9 @@ module ibex_controller #(
   // Interrupts including NMI are ignored,
   // - while in debug mode,
   // - while in NMI mode (nested NMIs are not supported, NMI has highest priority and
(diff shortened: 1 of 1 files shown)


GregAC:
It's a little concerning we hadn't yet realised that we had a bug where `mstatus.mie` effects interrupts seen in U mode (that field should be ignored when in U mode, only the per interrupt enable bits should be relevant). Still it's not the most critical bug, I've created a separate issue to track the extra testing/coverage we need to test that and other possible behaviours with different states of `mstatus` that we may also be missing https://github.com/lowRISC/ibex/issues/1825

vogelpi (approved):
LGTM, thanks @GregAC !

Issue #1831, opened 2022-10-04
[riscv-formal] Yosys parsing fails


marnovandermaas:
When running `make -C formal/riscv-formal` the Yosys toolchain emits the following error:
```
12. Executing Verilog-2005 frontend: build/ibex_if_stage.v
Parsing Verilog input from `build/ibex_if_stage.v' to AST representation.
build/ibex_if_stage.v:333: ERROR: syntax error, unexpected TOK_STRING, expecting TOK_ID or '#'
make: *** [Makefile:76: ibex.v] Error 1
```

marnovandermaas:
After some internal discussion, we think it is best to remove the riscv-formal folder from the main branch.

PR #1832, opened 2022-10-04
[riscv-formal] Removing unsupported and broken feature

marnovandermaas:
This PR removes the `formal/riscv-formal` directory.

It also removes the `Makefile` and `run.sby.j2` from the `formal/data_ind_timing` and `formal/icache` directories. I've added read me files to indicate that we don't have any build infrastructure for these sets of assertions, but that they will still be useful in the future.

Closes: https://github.com/lowRISC/ibex/issues/1831


vogelpi (commented):
This looks good to me, thanks @marnovandermaas . There are still some references in the tree to `formal` (but not `riscv-formal` obviously). Let's quickly discuss this in today's meeting.

GregAC (approved):

vogelpi (approved):
LGTM, thanks @marnovandermaas .

PR #1834, opened 2022-10-04
Update SCONTEXT address, add MSCONTEXT csr to match riscv_debug 1.0

hcallahan-lowrisc:
Observing the following spec change:

RISC-V Debug Support Version 1.0.0-STABLE
1.2.1.4 New Features from 0.13 to 1.0
8. Move scontext, renaming original to mscontext, and create hcontext.

MSCONTEXT is a backwards-compatible alias to SCONTEXT. [[1]](https://github.com/riscv/riscv-debug-spec/blob/c40847add0c3e3dcafb6ce568c3979412bb6f341/xml/hwbp_registers.xml#L246)
In Ibex, SCONTEXT is a read-only zero register. Hence MSCONTEXT has the same behaviour.

### DIFF
--- a/doc/03_reference/cs_registers.rst
+++ b/doc/03_reference/cs_registers.rst
@@ -46,6 +46,8 @@ Ibex implements all the Control and Status Registers (CSRs) listed in the follow
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x3BF  | ``pmpaddr15``      | WARL   | PMP Address Register                          |
 +---------+--------------------+--------+-----------------------------------------------+
+|  0x5A8  | ``scontext``       | WARL   | Supervisor Context Register                   |
++---------+--------------------+--------+-----------------------------------------------+
 |  0x747  | ``mseccfg``        | WARL   | Machine Security Configuration                |
 +---------+--------------------+--------+-----------------------------------------------+
 |  0x757  | ``mseccfgh``       | WARL   | Upper 32 bits of ``mseccfg``                  |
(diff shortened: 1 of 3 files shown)


GregAC (commented):
Currently we implement debug v0.13 not v1.0 and I'd prefer not to move versions at this point. Though this is a pretty innocuous change so happy to do it if it's needed.

What motivates this change? Is it some Ibex/Spike mismatch issue?

hcallahan-lowrisc:
> What motivates this change? Is it some Ibex/Spike mismatch issue?

Yes both Spike and riscv-dv already have this change merged. I thought it would be easier to bump Ibex up to the latest spec in this case because the change is so minor, but we could alternatively move spike/riscv-dv backwards to 013.

GregAC (approved):

hcallahan-lowrisc:
> Ok let's do this then. Strictly we end up with an incompatible implementation but I think this a minor corner we can get away with it (especially as something expecting v0.13 will still find a CSR it cannot write to and read as 0 where it wants SCONTEXT) 

Great. Thanks for the review.

> We should probably look at what else a proper move to v1.0 entails, but I think minimal change in this area is best.

I agree. We should probably accumulate a backlog of issues tagged as 'Spec-Compliance' to go through when the time is right.

vogelpi (commented):
Thanks for the PR @hcallahan-lowrisc . I also agree on checking what else needs to be done to bump the spec after the next milestone.

PR #1835, opened 2022-10-04
[rtl/dv] Bring back data integrity check on write responses

GregAC:
Previously Ibex signaled a major alert on an integrity error (where incoming read data doesn't match its integrity bits) for both read and write responses. This was removed as the data part of a response to a write is ignored.

This brings it back in a more measured way. This provides a little extra fault injection hardening as an attacker glitching the memory bus will generate an alert on both read and write responses.

See https://github.com/lowRISC/opentitan/issues/14611 for some more background on this.


vogelpi (approved):
This looks good, thanks @GregAC  for the PR and for providing the references.

ctopal (approved):
LGTM

PR #1842, opened 2022-10-07
[dv] Improve objdump and prettify scripts

marnovandermaas:
- Shellcheck objdump script
- Check for RISCV_TOOLCHAIN variable
- Shellcheck prettify script

### DIFF
--- a/dv/uvm/core_ibex/scripts/objdump.sh
+++ b/dv/uvm/core_ibex/scripts/objdump.sh
@@ -1,4 +1,9 @@
+#!/bin/bash
 _GET_OBJS=$(find ./out/run -type f -iregex '.*test\.o')
+if [[ -z "${RISCV_TOOLCHAIN}" ]]; then
+   echo "Please define RISCV_TOOLCHAIN to have access to objdump."
+   exit 1
+fi
 for obj in $_GET_OBJS; do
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc (commented):
Our docs [currently say](https://github.com/lowRISC/ibex/blame/master/doc/03_reference/verification.rst#L125) that you need to have set the RISCV_TOOLCHAIN environment variable, could we just leave that to be the only working option? I feel like having two options is a little confusing. WDYT?

hcallahan-lowrisc:
The shellcheck improvements are very welcome. Thanks.

hcallahan-lowrisc:
Riscv-dv picks up gcc and objcopy from the environment variables [RISCV_GCC](https://github.com/google/riscv-dv/blob/c6acc1897429f5245cc89b2ecee2e3eefdefd18d/run.py#L432) and RISCV_OBJCOPY. Currently our docs say to define those relative to the RISCV_TOOLCHAIN envvar. 
I think for now it would be preferable to have a single-source-of-truth for how to setup the ibex build system. Adding to the PATH works too but I can imagine a situation where one accidentally ends up with two riscv toolchains accessible to the shell and subtle chaos ensues. I think having a single method supported for now makes sense in the short term.

marnovandermaas:
Ok, that makes sense and thanks for the explanation. I have changed it to just throw an error if `RISCV_TOOLCHAIN` does not exist.

hcallahan-lowrisc (approved):

PR #1848, opened 2022-10-12
Add a double_fault detector to core_ibex uvm environment

### LINKED ISSUE #1823: [dv] Implement double fault loop detector
Add a double fault loop detector to the testbench. It should determine when Ibex has entered a loop of double faults and terminate the simulation after some configurable number of iterations (with a default e.g. 100).

By default is should exit with a fail but a configuration option should be supplied to enable exit with a pass (for tests where a double fault loop is the intended behaviour).

> estimate 4
> remaining 2022-10-12 1

hcallahan-lowrisc:
Add a new scoreboard component to the core_ibex uvm environment, which contains a double_fault detector task. This uses the top-level output 'double_fault_seen_o' to count the number of total and consecutive double_faults seen within a test. A helper task allows the base_test to wait upon each of these counters reaching the configured thresholds, and then to end the test.

The default thresholds are 100 for consecutive faults, and 1000 for total faults.
When either threshold is reached, the test will fail by default.

A plusarg '+enable_double_fault_detector=1' enables/disables the checker.
The double_fault detector is ~~disabled~~ enabled by default.

A plusarg '+is_double_fault_detected_fatal=1' controls the result of reaching the thresholds. '0' will make the test end with a passing result when the threshold is reached.

~~This commit enables it for only the 'pmp_full_random_test', as that is a useful test candidate to begin with.~~

Closes #1823.

### DIFF
--- a/dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
@@ -58,7 +58,6 @@ class core_ibex_scoreboard extends uvm_scoreboard;
             double_fault_cnt_consecutive = 0;
           end
 
-
           // Create an event if either counter reaches its threshold value, then reset the counter.
           if (double_fault_cnt_consecutive == cfg.double_fault_threshold_consecutive) begin
             fault_threshold_consecutive_reached.trigger();
@@ -71,8 +70,12 @@ class core_ibex_scoreboard extends uvm_scoreboard;
 
         end
       end
-      // Latch a signal to show that we have seen a double_fault.
-      // The pulse may be receieved sometime before the rvfi_seq_item.
+      // Latch the 'double_fault_seen_o' signal to catch the fault.
+      // The single pulse may be receieved sometime before the rvfi_seq_item
+      // corresponding to the faulting instruction is generated. Hence we
+      // latch that pulse when it is seen, and then reset above when the
+      // seq_item arrives.
+      // https://github.com/lowRISC/ibex/pull/1848#discussion_r995903762
       begin
         forever begin
           @(posedge cfg.ibex_dut_vif.double_fault_seen);
--- a/dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
+++ b/dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
@@ -1,0 +1,108 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class core_ibex_scoreboard extends uvm_scoreboard;
+
+  uvm_tlm_analysis_fifo #(ibex_rvfi_seq_item) rvfi_port;
+  core_ibex_env_cfg                           cfg;
+
+  // Events for Double-Fault detection
+  uvm_event fault_threshold_consecutive_reached, fault_threshold_total_reached;
+
+  `uvm_component_utils(core_ibex_scoreboard)
+  `uvm_component_new
+
+  function void build_phase(uvm_phase phase);
+    super.build_phase(phase);
+    if (!uvm_config_db#(core_ibex_env_cfg)::get(this, "", "cfg", cfg)) begin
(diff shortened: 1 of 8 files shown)


dv/uvm/core_ibex/tests/core_ibex_base_test.sv
```diff
@@ -211,38 +211,51 @@ class core_ibex_base_test extends uvm_test;
...
         end else begin
           `uvm_fatal(`gfn, "Incorrectly formed handshake received at test-control address.")
         end
       end
+      // - End the test if we see too many of the following...
+      //   - double_faults
+      begin
+        if (cfg.enable_double_fault_detector) begin
+          env.scoreboard.dfd_wait_for_pass_events();
+          `uvm_info(`gfn, "Test done due to double_fault detector.", UVM_LOW)
```
marnovandermaas, line 239: Could we also add a letter to the failure mode like with the timeouts?

marnovandermaas (commented):
I will test this PR on my local setup to see if it improves my simulation time.

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -757,7 +757,9 @@
     +instr_cnt=6000
     +pmp_max_offset=00024000
     +pmp_randomize=1
-    +pmp_allow_addr_overlap=1
+    +pmp_allow_illegal_tor=1
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Whoops, I should have dropped that in a rebase! thanks

marnovandermaas (approved):
I've tested this pull request using the PMP full random test and it seems to pass when detecting the double faults and also make sure the tests do not time out. This has been a great help, thanks @hcallahan-lowrisc

dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
```diff
@@ -0,0 +1,108 @@
...
+
+        end
+      end
+      // Latch a signal to show that we have seen a double_fault.
+      // The pulse may be receieved sometime before the rvfi_seq_item.
+      begin
+        forever begin
+          @(posedge cfg.ibex_dut_vif.double_fault_seen);
+          double_fault_pulse_seen = 1'b1;
+          cfg.ibex_clk_vif.wait_clks(1);
```
GregAC, line 80: With this setup the `double_fault_pulse_seen` being set won't necessarily correspond with the `rvfi_instr`coming from the analysis port (i.e. that instruction may not be the one for which the double fault was seen). However I don't think it really matters, as you get at most one `double_fault_seen` pulse per instruction so they're both working at the same rate and all we need to do is count the pulse per instruction rate (and total pulses).

May be worth adding a comment to explain this.

GregAC (approved):
LGTM with a couple of suggestions

dv/uvm/core_ibex/tests/core_ibex_base_test.sv
GregAC: Ahh looks like you covered this in one of your comments:

> Still TODO:
>
> *   Add option to change the test result from a pass to a fail when the detector activates.

marnovandermaas:
I've found a seed where the double fault detection does not work:
`make -j1 --keep-going IBEX_CONFIG=opentitan ITERATIONS=1 TEST=riscv_epmp_mml_read_only_test WAVES=0 COV=0 SEED=26216`

On this commit:
https://github.com/marnovandermaas/ibex/commit/6a309647723ec763ce304d0cc4bfa2cc4cb31d01

dv/uvm/core_ibex/env/core_ibex_scoreboard.sv
hcallahan-lowrisc: Yes this is a good point, I'll add some more comments to explain it further.
I too assumed while creating this that there could not be more than one pulse of `double_fault_seen_o` per retired instruction, and that pulses of the signal could not occur without an rvfi_item being generated shortly after. So hopefully it will be okay. Follow up on this if it causes any unforeseen issues.

Issue #1846, opened 2022-10-12
[dv] Tests failing when accessing uninitialized memory


marnovandermaas:
When doing a load to uninitialized memory, I get the following error:
```
UVM_ERROR ~/ibex/vendor/lowrisc_ip/dv/sv/mem_model/mem_model.sv(35) @ 16688864: reporter [mem] read from uninitialized addr 0x7
```

When doing a fetch to unitialized memory, I get the following error:
```
UVM_ERROR ~/ibex/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv(162) @ 25601766: reporter [ASSERT FAILED] NoAlertsTriggered
```

Now for most tests this is fine, but for tests like random PMP tests you would like to be able to catch these errors. @hcallahan-lowrisc has seen similar issues when testing debug mode. My proposal would be to have a flag that says to pass when it sees these errors, but am happy to discuss alternative approaches.

> Estimate 2
> remaining 2022-10-14 1

GregAC:
We could introduce something that produces data when you read from uninitialized memory? For instance on a d-side read we return some random value (maybe writing that to the memory model though just random is probably fine), then for i-side read return 2x the compressed cannonical unimplemented RISC-V instruction (as if we fetch from a place we haven't put code we want to ensure we don't end up executing something from it).

Though this will cause co-sim issues. On the ISide you would presume we never actually execute the fetches so fine if cosim and uvm environment have a different view on what lives there. For DSide they need to match up. May actually be fairly simple as cosim runs behind Ibex RTL so when we observe the read from Ibex we can produce some random data and write it into the cosim memory model.

GregAC:
@GregAC to triage current regression and disable integrity failures on uninitialized memory where needed.

PR #1850, opened 2022-10-14
Fixup some DV tests

GregAC:
Note that riscv_csr_test is still broken due to https://github.com/lowRISC/ibex/issues/1846 but when that is sorted we should see it pass.


marnovandermaas (approved):
LGTM, thanks

ctopal (approved):

PR #1855, opened 2022-10-14
Change ibex_mem_intf_response_seq to handle uninit memory differently

hcallahan-lowrisc:
Reading uninit DMEM returns a random value.
Reading uninit IMEM returns returns `{2{C.unimp}}`.

[~Integrity errors are no-longer generated.~](https://github.com/lowRISC/ibex/pull/1855#discussion_r995967300)
Integrity errors are now generated by default, but this behaviour can be disabled by settings the plusarg `+enable_bad_intg_on_uninit_access=0`

Goes towards #1846

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
@@ -21,8 +21,7 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
   `uvm_object_new
 
   virtual task body();
-    if(m_mem ==  null)
-      `uvm_fatal(get_full_name(), "Cannot get memory model")
+    if (m_mem == null) `uvm_fatal(get_full_name(), "Cannot get memory model")
     `uvm_info(`gfn, $sformatf("is_dmem_seq: 0x%0x", is_dmem_seq), UVM_LOW)
     forever
     begin
@@ -33,6 +32,8 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
       bit                  data_was_uninitialized = 1'b0;
 
       p_sequencer.addr_ph_port.get(item);
+      aligned_addr = {item.addr[DATA_WIDTH-1:2], 2'b0};
+
       req = ibex_mem_intf_seq_item::type_id::create("req");
       error_synch = 1'b0;
       if (!req.randomize() with {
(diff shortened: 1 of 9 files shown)


dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
```diff
@@ -57,48 +58,30 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
...
+      end else if(item.read_write == WRITE) begin
+        // Update memory_model
+        write(aligned_addr, item.data);
       end
       // Add integrity bits
       {req.intg, req.data} = prim_secded_pkg::prim_secded_inv_39_32_enc(req.data);
 
       // If data_was_uninitialized is true then we want to force bad integrity bits: invert the
       // correct ones, which we know will break things for the codes we use.
-      if (data_was_uninitialized) req.intg = ~req.intg;
```
hcallahan-lowrisc, line 88: @GregAC I've already forgotten exactly what we said about this, is there a need to keep the option to corrupt the integrity here? Maybe behind a plusarg?

hcallahan-lowrisc:
@GregAC I also haven't yet implemented the update to the cosim memory_model for dmem accesses yet. I do see a few cosim failures testing this locally which are caused by this.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
GregAC: We should still corrupt the integrity by default then have a plusarg that disables this.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
```diff
@@ -110,19 +93,55 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
...
+      data = data << 8;
+      data[7:0] = read_byte(addr + i, byte_is_uninit);
+      if (byte_is_uninit) begin
+        did_access_uninit_mem = 1'b1;
+        // If any byte of the access comes back as uninit, bork the whole access.
+        if (is_dmem_seq) begin
+          // DMEM
+          `DV_CHECK_STD_RANDOMIZE_FATAL(data)
+          `uvm_warning(`gfn, $sformatf("Addr is uninit! DMEM seq, returning random data 0x%0h", data))
+          return data;
```
GregAC, line 114: I think we'll need to write the randomly generated bytes to the unimp memory here (along with writing to the cosim memory model).

GregAC (commented):
Think this should do the trick, just need to sort out the memory model updates when we generate the random data and the plusarg configuration. Thanks @hcallahan-lowrisc

GregAC (approved):

Issue #1827, opened 2022-10-03
`core_busy_o` not protected by lockstep


andreaskurth:
## Observed Behavior
The `core_busy_o` output signal of the non-lockstep core is used to gate the core clock, which is used by the regular Ibex core and its lockstep counterpart.
https://github.com/lowRISC/ibex/blob/3943a4eca3917d0a00406abb36f4951304c0b394/rtl/ibex_top.sv#L211-L219
https://github.com/lowRISC/ibex/blob/3943a4eca3917d0a00406abb36f4951304c0b394/rtl/ibex_top.sv#L286-L287
https://github.com/lowRISC/ibex/blob/3943a4eca3917d0a00406abb36f4951304c0b394/rtl/ibex_top.sv#L927-L928

This means that the lockstep core gets gated in the same clock cycle as the regular core, even though the lockstep core is a few cycles delayed. **A low-glitch in the `core_busy_o` output of the non-lockstep core thus does not lead to a comparison mismatch and a major alert because the lockstep core is already clock-gated when it would produce the value of `core_busy_o` that is used in the comparison.**
https://github.com/lowRISC/ibex/blob/3943a4eca3917d0a00406abb36f4951304c0b394/rtl/ibex_lockstep.sv#L453-L454

## Expected Behavior

A low-glitch/mismatch in the `core_busy_o` output signal of the non-lockstep core should lead to a major alert by the lockstep core.

## Steps to reproduce the issue
Run the `chip_sw_ibex_lockstep_glitch` test in OpenTitan (currently in https://github.com/lowRISC/opentitan/pull/15221) and override `port_idx` to the index of `core_busy_o` in `output_ports`.

> estimate 2

vogelpi:
Thanks @andreaskurth for creating this issue. I think this is a good catch!

I've now had a look at the RTL. If the clock is turned off, also the delay registers have no clock anymore. This means the outputs of the last couple of cycles of the core will be compared with the outputs of the lockstep/shadow core after resuming. Functionality-wise this seem okay to me.

From a security perspective, this could however be problematic. As we just discussed in person, by glitching `core_busy_o` of the main core to low, Ibex could be made to sleep instead of handling e.g. an alert or similar. Eventually, the watchdog timer will intervene and reset the whole chip but Andreas and I think it would be good to just protect the `core_busy_o` signal using a multi-bit encoding (similar to `fetch_enable_i`). Does this sound reasonable @GregAC ?

[...]

PR #1865, opened 2022-10-21
[rtl] Protect core_busy_o with a multi-bit encoding

vogelpi:
This PR protects the `core_busy_o` signal using a multi-bit encoding to reduce the chances of an adversary for glitching this signal to low, thereby putting the core to sleep and e.g. not handling an alert.

Without this change, the glitch would only be detected once both the main core and the shadow core wake up again and the comparison of the core_busy_o signals continues.

This resolves lowRISC/Ibex#1827.


rtl/ibex_pkg.sv
```diff
@@ -653,14 +653,14 @@ package ibex_pkg;
   parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonceDefault =
       64'hf79780bc735f3843;
 
-  // Fetch enable. Mult-bit signal used for security hardening. For non-secure implementation all
-  // bits other than the bottom bit are ignored.
-  typedef logic [3:0] fetch_enable_t;
+  // Mult-bit signal used for security hardening. For non-secure implementation all bits other than
+  // the bottom bit are ignored.
+  typedef logic [3:0] ibex_mubi_t;
```
vogelpi, line 658: I thought it would be cleaner to rename this type because we're now also using it for the `core_busy` signal. Still we don't want to use the OT MuBi4 Type here (we've been discussing already a while back that it's probably better to not add this dependency).

tjaychen:
@vogelpi this looks good to me. But why keep any of the alert handling logic on gated clock anyways? While this change reduces the possibilities, if you move that logic to ungated portion, don't you remove the possibility completely? 

I'm not suggesting we change this PR, but just wanted to understand.

vogelpi:
> @vogelpi this looks good to me. But why keep any of the alert handling logic on gated clock anyways? While this change reduces the possibilities, if you move that logic to ungated portion, don't you remove the possibility completely (i'm probably oversimplifying... the downstream cone is probably bigger than I imagine)?
> 
> I'm not suggesting we change this PR, but just wanted to understand.

Thanks for taking a look @tjaychen . IIRC the problem with moving the comparison logic into the ungated region is that after switching off the clock the comparison logic will not get new inputs from the main core, just from the shadow core for a couple of cycles. Similarly, after re-enabling the clock the shadow core will not produce new output for some cycles. Both this would then need to be handled (and hardened) in the comparison logic. Right now this logic is simple: after reset, there is a delay of a couple of cycles until we enable the comparison. After that point, there is no way to disable the comparison again.

rtl/ibex_top.sv
```diff
@@ -189,26 +189,45 @@ module ibex_top import ibex_pkg::*; #(
...
+    prim_flop #(
+      .Width($bits(ibex_mubi_t)),
+      .ResetValue(IbexMuBiOff)
+    ) u_prim_core_busy_flop (
+      .clk_i (clk_i),
+      .rst_ni(rst_ni),
+      .d_i   (core_busy_d),
+      .q_o   (core_busy_q)
+    );
+  assign clock_en = (core_busy_q != IbexMuBiOff) | debug_req_i | irq_pending | irq_nm_i;
```
andreaskurth, line 214:
```suggestion
    assign clock_en = (core_busy_q != IbexMuBiOff) | debug_req_i | irq_pending | irq_nm_i;
```

andreaskurth (approved):
LGTM, thanks @vogelpi! I only have two nits

rtl/ibex_core.sv
```diff
@@ -368,7 +368,31 @@ module ibex_core import ibex_pkg::*; #(
...
+    for (genvar i = 0; i < $bits(ibex_mubi_t); i++) begin : g_core_busy_bits
+      if (IbexMuBiOn[i] == 1'b1) begin : g_pos
+        assign core_busy_o[i] =  |busy_bits_buf[i*NumBusySignals +: NumBusySignals];
+      end else begin : g_neg
+        assign core_busy_o[i] = ~|busy_bits_buf[i*NumBusySignals +: NumBusySignals];
+      end
+    end
+  end else begin : g_core_busy_non_secure
+    // For non secure Ibex, synthesis is allowed to optimize core_busy_o.
+    assign core_busy_o = ctrl_busy | if_busy | lsu_busy ? IbexMuBiOn : IbexMuBiOff;
```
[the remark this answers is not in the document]
vogelpi: Sound good, also I should have used `||` instead of `|`. This is now changed.

rtl/ibex_top.sv
vogelpi: Good catch, fixed!

PR #1871, opened 2022-10-25
Fix bug in passing cosim_agent handle to the data_intf_seq

hcallahan-lowrisc:
It turns out #1855 could have been tested a little better.

The handle was passed before the cosim object was constructed, so when it came to use the handle it caused a null pointer exception. Moved to the end of the `connect_phase`.


GregAC:
I had noticed the same thing, had yet to get round to a PR though, thanks for the fix.

PR #1878, opened 2022-10-27
[cosim] Fixup ebreak behaviour

GregAC:
When DCSR is set such that ebreak will enter debug mode we were getting cosim mismatches. This was because Ibex produces the ebreak on the RVFI interface and spike effectively skips right over it and executes the first instruction of the debug handler immediately. Traps have similar but not identical behaviour so we need a special case in the step function to handle this.

### DIFF
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -168,6 +168,17 @@ bool SpikeCosim::step(uint32_t write_reg, uint32_t write_reg_data,
   // The DUT has just produced an RVFI item
   // (parameters of this func is the data in the RVFI item).
 
+  // First check to see if this is an ebreak that should enter debug mode. These
+  // need specific handling. When spike steps over them it'll immediately step
+  // the next instruction (i.e. the first instruction of the debug handler) too.
+  // In effect it treats it as a transition to debug mode that doesn't retire
+  // any instruction so needs to execute the next instruction to step a single
+  // time. To deal with this if it's a debug ebreak we skip the rest of this
+  // function checking a few invariants on the debug ebreak first.
+  if (pc_is_debug_ebreak(pc)) {
+    return check_debug_ebreak(write_reg, pc, sync_trap);
+  }
+
   uint32_t initial_spike_pc;
   bool pending_sync_exception = false;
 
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc:
This is interesting, I have been using the `DCSR.ebreakm/u` functionality in the [hardware_breakpoints test/PR](https://github.com/lowRISC/ibex/pull/1801/files#diff-4380451fc55aef8a4edf663b80f708e20fe5f0362639f42a3f669f3277dbddc2R93) and have not seen and co simulation mismatches there. It's possible that test only makes use of one of `m/u` modes and not the other, so it could be that I guess. I haven't thought about it too much.

GregAC:
> This is interesting, I have been using the `DCSR.ebreakm/u` functionality in the [hardware_breakpoints test/PR](https://github.com/lowRISC/ibex/pull/1801/files#diff-4380451fc55aef8a4edf663b80f708e20fe5f0362639f42a3f669f3277dbddc2R93) and have not seen any cosimulation mismatches there. It's possible that PR only makes use of one of `m/u` modes and not the other, so it could be that I guess. I haven't thought about it too much.

Yeah this will only come up if you set the relevant DCSR bit to turn the `ebreak` into a break into debug rather than causing a breakpoint exception. I think the ebreakmu is the only test which does that.

hcallahan-lowrisc:
> I think the ebreakmu is the only test which does that.

So I also use that behaviour in the hardware_breakpoints test / PR, which uses ebreaks to get into debug mode to then change the trigger address. Right at the start of the test it sets the DCSR.ebreakm/u bits = 1. So I believe I'm exercising the same behaviour?

GregAC:
> > I think the ebreakmu is the only test which does that.
> 
> So I also use that behaviour in the hardware_breakpoints test / PR, which uses ebreaks to get into debug mode to then change the trigger address. Right at the start of the test it sets the DCSR.ebreakm/u bits = 1. So I believe I'm exercising the same behaviour.

Ah ok.

May also be a behaviour that altered with the recent ebreak change in Spike (https://github.com/lowRISC/riscv-isa-sim/pull/22)

hcallahan-lowrisc (approved):

Issue #1759, opened 2022-08-11
[dv] Verify ICache Integrity and Scrambling


GregAC:
Verify the ICache integrity and scrambling features

For integrity create a testbench component to corrupt icache data and icache tags. Ensure corruptions triggered the required alert. This will require targeting to corrupt data and tags that relate to instructions that are going to get execute. Understanding of scrambling will need to be employed to determine the appropriate memory locations to corrupt.

For scrambling I'm uncertain if we need to do anything further as we're using a existing OT primitive component. @ctopal any thoughts from your OTBN experience? 

old estimate 16, new estimate based upon more detailed analysis by @andreaskurth 
> estimate 4
> remaining 2022-10-28 1

ctopal:
In OTBN, this checking of responses to integrity failures is done as this:

1. Read scrambled data from RAM using `encrypt_sram_addr` and `decrypt_sram_data` from `sram_scrambler_pkg`.
2. Apply 1 or 2 bit flips to generate an integrity error when said data gets read back.
3. Write corrupted data to scrambling memory using `encrypt_sram_addr` and `encrypt_sram_data` from `sram_scrambler_pkg`.

A task will need to include incorporating  `sram_scrambler_pkg` to our core_ibex DV environment (or maybe block level ICache DV environment?) and writing up the sequence that uses the functionality mentioned above.

For verifying scrambling itself, OTBN side maps smoke and calls it a day (not sure if it's good to do that, maybe we need to have a talk about it as well?) https://github.com/lowRISC/opentitan/pull/13335#discussion_r903922037

andreaskurth:
I've looked into the integrity verification, and I think the following approach should work:
- For each data and tag memory, track which memory entries have been written thus contain valid values (using the `req_i`, `write_i`, and `addr_i` signals).
- For each data and tag memory, track which memory entries are being read (using the same three signals). If an entry is being read that has been written before, corrupt one bit in the output data with a certain probability.
- If the corruption happened, check that the required alert is fired.

I'm estimating 4 for that.

GregAC:
Thanks @andreaskurth I will update the estimate with your new one and assign you to the issue.

andreaskurth:
Closing this as completed in #1879.

PR #1879, opened 2022-10-28
[dv] Add test glitching data/tag of icache

andreaskurth:
This PR addresses the integrity part of https://github.com/lowRISC/ibex/issues/1759 by verifying that a glitch on the data or tag response from the instruction cache raises a minor alert by the core (if the data/tag returned from the cache is used by the core).

The **last** commit implements the actual test. The commits before that are prerequisites as follows:
1. [dv] Add assert control for `NoAlertsTriggered`: duplicate from #1873, will be dropped once that PR is merged.
2. [wip] Extend vendored-in `clk_rst_if`: adds two functions to that IF -- if we agree that this is appropriate and necessary, these functions should be upstreamed.
3. [wip] Locally import `DV_CREATE_SIGNAL_PROBE_FUNCTION`: adds a macro recently added in OpenTitan (I wanted to properly vendor it in, but the update failed). If we agree that this macro should be used here, we need to fix the vendor-update problems.
4. [dv] Add icache signals to `core_ibex_dut_probe_if`: should be fairly uncontroversial.
5. [dv] Create signal probe functions in `core_ibex_dut_probe_if`: these functions have benefits over `uvm_hdl_read()`; see the commit message.

Pass rate is currently 100% (I tried 20 seeds) -- that is, alerts are produced and checked as expected. Every test corrupts a used data/tag of the icache, but not in every test the data/tag is valid (even when not corrupted). The test takes this into account when checking for an alert.


hcallahan-lowrisc:
I tried running this locally and all seeds fail with an error I am unfamiliar with.

Running with environment module `xcelium/latest`

```
$ git rev-parse --short HEAD
7d035124

$ make -j12 -ik TEST=riscv_icache_intg_test ITERATIONS=20
```

regr.log extract
```

riscv_icache_intg_test.13876
----------------------------
binary:          test.bin
rtl_log:         rtl_sim.log
rtl_trace:       trace_core_00000000.log
iss_cosim_trace: spike_cosim_trace_core_00000000.log

[FAILED]: error seen in 'rtl_sim.log'
---------------*LOG-EXTRACT*----------------
    72: /home/harry/projects/ibex/rtl/ibex_top.sv, 180:
    73: UVM_INFO /home/harry/projects/ibex/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv(163) @ 7410: reporter [core_ibex_tb_top.unmblk1] Disabling assertions: core_ibex_tb_top.NoAlertsTriggered
    74: ERROR:          VPI       VISNOR
    75: Attempting to access core_ibex_tb_top.dut.u_ibex_top.u_ibex_core.if_stage_i.gen_icache.icache_i.lookup_valid_ic1 by vpi_get_value() routine which does not have read access.
    76: /home/harry/projects/ibex/rtl/ibex_icache.sv, 106:
[E] 77: UVM_FATAL /home/harry/projects/ibex/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv(248) @ 7430: reporter [uvm_test_top] Check failed (!$isunknown(lookup_valid))
    78: 
    79: --- RISC-V UVM TEST FAILED ---
    80: 
    81: ~
--------------------------------------------

```

Something about different version of tools/libraries maybe?

andreaskurth:
Thanks for pasting the error log, @hcallahan-lowrisc. That's the problem I meant with "the simulator complains that it does not have access to the icache data/tag to glitch it". One can work around it by enabling the capturing of waves (`WAVES=1`). A proper solution could involve passing arguments/flags enabling access to signals that are not captured in waves or disabling the optimization of such signals.

GregAC (approved):
LGTM, thanks @andreaskurth 

Agreed with the changes to the DV library from OT. Obviously we need to get those applied in OT and vendored across here, along with the vendoring issue you found fixed up before we merge this rather than merging in this form.

It's unfortunate the force doesn't work unless we have waves enabled. I think we'll be able to at least partially mitigate this by doing a more specific probe that just adds the necersarry signals to the trace.

> A PR to update the private-CI job to WAVES=1 should be merged with this.

Please hold off on doing this. A general WAVES=1 will slow down the regression massively as well as generating a huge amount of data. For now I don't mind if this test just fails in the nightlies.

I'll have a think about the easiest way to work around this without too major an impact.

ctopal:
@hcallahan-lowrisc suggested to look at how OT repository was handling this. Looks like we need a `-access rw` added `rtl_simulation.yaml`. It shouldn't change the whole simulation time that much, considering this won't actually log the signals but only set the accesses. When I tried with this, it seemed to work. Now trying to see if we'd see any slowdowns from it but considering dvsim uses  `-access rw` for `WAVES=0` and  `-access rwc` for `WAVES=1` I think we can just change it.

ctopal (approved):
LGTM, please have a look at the suggested fix above for enabling forcing without `WAVES=1`

andreaskurth:
100% pass rate (with 50 seeds/iterations) now without any "hacks". Thanks for your inputs!

Issue #1756, opened 2022-08-11
[dv] Verify register file integrity checking and write enable glitch detection


GregAC:
Create a testbench component that can corrupt bits and glitch the one hot write enable in the register file.

Write a test that uses this and checks the appropriate alert is signaled.

original estimate 8
> estimate 4
> remaining 2022-10-28 1

andreaskurth:
The write enable is protected by `prim_onehot_check`:
https://github.com/lowRISC/ibex/blob/ed927be38770b3478f34d4eace6edc1d74decaa8/rtl/ibex_register_file_ff.sv#L70-L81
and its error signal is wired up to the major internal alert of the core:
https://github.com/lowRISC/ibex/blob/ed927be38770b3478f34d4eace6edc1d74decaa8/rtl/ibex_top.sv#L429
https://github.com/lowRISC/ibex/blob/ed927be38770b3478f34d4eace6edc1d74decaa8/rtl/ibex_top.sv#L1035-L1037
and this is [covered by assertions in OpenTitan](https://github.com/lowRISC/opentitan/blob/64ea3bd268b0ad00e771aa65d735a3a89a7febcb/hw/ip/rv_core_ibex/rtl/rv_core_ibex.sv#L886-L889), which allows us to check this counter measure in FPV. Thus, we don't need to check it in simulation.

I'm updating the estimate to 4 since this is therefore reduced to writing a test that glitches the data of the RF and checks for the resulting integrity error.

PR #1881, opened 2022-10-28
[dv] Add test glitching register file read ports

andreaskurth:
This PR addresses the integrity checking part of https://github.com/lowRISC/ibex/issues/1756 by verifying that a glitch on the data read from the register file raises a major alert by the core (if the data read from the register file is used by the core).

The **last** commit implements the actual test. The commits before that are prerequisites as follows:
1. duplicate from #1873, will be dropped once that PR is merged.
2. duplicate from #1879, will be dropped once that PR is merged.
3. [dv] Add RF signals to `core_ibex_dut_probe_if` and
4. [dv] Create signal probe functions in `core_ibex_dut_probe_if`: same approach as in #1879.

Pass rate is currently 100% (I tried 20 seeds) -- that is, alerts are produced and checked as expected -- however the DV environment misinterprets passing tests as failures (probably due to the call to `dv_test_status_pkg::dv_test_status(1)` at the end of a successful test). Hints on how to fix this would be appreciated.


hcallahan-lowrisc:
This test needs `WAVES=1` too.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -10,6 +10,113 @@ class core_ibex_csr_test extends core_ibex_base_test;
...
+      // Release force after one clock cycle.
+      clk_vif.wait_n_clks(1);
+      release_force(port_name);
+
+      // Complete test if alert has been correctly triggered.
+      if (exp_alert) break;
+    end
+
+    // Stop test at this point because cosim will mismatch.
+    dv_test_status_pkg::dv_test_status(1);
```
hcallahan-lowrisc, line 114:
```suggestion
    rs = uvm_report_server::get_server();
    rs.report_summarize();
```
This will allow the test to be detected as passed. (The Ibex testbench only matches on a [single string](https://github.com/lowRISC/ibex/blob/0c0626ebbf6ded92fe84dbb8cdb558383c99adf2/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py#L221) to detect a passing test. This could be improved)

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -10,6 +10,113 @@ class core_ibex_csr_test extends core_ibex_base_test;
 
 endclass
 
+// Test that corrupts data read from the register file and checks that an appropriate alert occurs.
+class core_ibex_rf_intg_test extends core_ibex_base_test;
+
+  `uvm_component_utils(core_ibex_rf_intg_test)
+  `uvm_component_new
+
```
hcallahan-lowrisc, line 18:
```suggestion

  uvm_report_server rs;
  
```

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
andreaskurth: Cool, thanks @hcallahan-lowrisc!

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -10,6 +10,116 @@ class core_ibex_csr_test extends core_ibex_base_test;
...
+      // Release force after one clock cycle.
+      clk_vif.wait_n_clks(1);
+      release_force(port_name);
+
+      // Complete test if alert has been correctly triggered.
+      if (exp_alert) break;
+    end
+
+    // Stop test at this point because cosim will mismatch.
```
GregAC, line 115: I think we should add some functionality to our base test that can force an immediate test stop with a pass. Given you've used this is another of the V2S tests as well and it's producing false failures here better to have a more general method that better integrates with the normal test flow. I can take a look at doing this. Happy to leave this as is to merge things to keep V2S work moving forward in the meantime.

GregAC (approved):
LGTM, to get the false failure sorted I think you just need to output 'RISC-V UVM TEST PASSED' before you terminate. Happy for that to just be hacked in for now. As I said in my other comment I'll take a look at providing a general method in the base test to do this more cleanly.

This is the line from the script that determines pass/fail which checks the pass condition: https://github.com/lowRISC/ibex/blob/63be01b6082d335f04d98056760f029faf60a424/dv/uvm/core_ibex/riscv_dv_extension/ibex_log_to_trace_csv.py#L221

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
andreaskurth: With [Harry's suggestion](https://github.com/lowRISC/ibex/pull/1881#discussion_r1009203099), it's no longer producing false failures. Some code could be generalized to reduce duplication, but I agree that this is a lower prio enhancement

andreaskurth:
Waiting for #1897 and #1898 to land before rebasing this

andreaskurth:
100% pass rate now without any "hacks". Thanks for all your feedback!

PR #1884, opened 2022-10-29
[rtl] Fix dummy instructions

GregAC:
Previously there was a single dummy_instr_id_o signal from ibex_core which the register file used to determine if it could write to the zero register (which reads as zero always for real instructions). However a write occurs in the writeback stage so this signal was not asserted correctly.

This adds a dummy_instr_wb_o signal to control the write to zero register. dummy_instr_id_o remains as it's still employed for register reads for dummy instructions.


vogelpi (approved):
This looks good except for the typo/lint error, thanks @GregAC !

andreaskurth (approved):

PR #1886, opened 2022-10-31
[pmp] Add directed instr streams to increase PMP coverage

ctopal:
The main idea is to see some writes to `mseccfg` CSR while operating on the `pmp_full_random test` to enhance our stimulus generation and see coverage increases around sticky bit features + some no region match crosses.

Also now we enable randomising `mstatus.mprv` bit in `pmp_full_random_test` with second commit which should improve the coverage on `mprv_effect_cross`

Third commit is about increasing coverage in pmpcfg0 region, enables us to see NA4 region getting configured for instruction channel with randomised execute access.

Note that these changes reduces the pass rate while increasing the coverage.


hcallahan-lowrisc (approved):
LGTM

GregAC (approved):

PR #1918, opened 2022-11-09
Fix capture info for ibex cosim in case of multiple interrupts

Saad525:
When there is an interrupt followed by an nmi interrupt or a debug request (which if detected before the first instruction of trap handler was executed), the already captured previous interrupt info needs to be updated with nmi_interrupt or debug_request.

As RTL seems to follow this priority: debug_request << nmi_interrupt << other interrupts (to confirm from documentation), so Spike needs to be fed with captured info for the upcoming nmi_interrupt or debug_request.


GregAC:
This fix is needed though more is required on top of it to fix DV issues associated with it. We'll get this merged and I'll do a follow-up PR with some other changes.

GregAC (approved):

PR #1922, opened 2022-11-11
[dv] Fix iside error notification to cosim

GregAC:
When a writeback exception occurs when the instruction in ID/EX has seen an instruction fetch error we need to ensure that error doesn't get notified to cosim. This requires watching for a writeback exception and removing the latest iside error from the queue if needed.

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
+++ b/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
@@ -17,9 +17,11 @@ class ibex_cosim_scoreboard extends uvm_scoreboard;
   uvm_tlm_analysis_fifo #(ibex_ifetch_seq_item)     ifetch_port;
   uvm_tlm_analysis_fifo #(ibex_ifetch_pmp_seq_item) ifetch_pmp_port;
 
-  virtual core_ibex_instr_monitor_if              instr_vif;
+  virtual core_ibex_instr_monitor_if instr_vif;
+  virtual core_ibex_dut_probe_if     dut_vif;
 
   uvm_event reset_e;
+  uvm_event check_inserted_iside_error_e;
 
   bit failed_iside_accesses [bit[31:0]];
   bit iside_pmp_failure     [bit[31:0]];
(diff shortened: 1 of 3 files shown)


GregAC:
No errors seen with 50 iterations of riscv_mem_error_test and this fix

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
```diff
@@ -261,12 +283,36 @@ class ibex_cosim_scoreboard extends uvm_scoreboard;
...
   endtask: run_cosim_imem_errors;
 
+  task run_cosim_prune_imem_errors();
+    // Errors are added to the iside error queue the first cycle the instruction that sees the error
+    // is in the ID stage. Cycles following this the writeback stage may cause an exception flushing
+    // the ID stage so the iside error never occurs. When this happens we need to pop the new iside
+    // error off the queue.
+    forever begin
+      // Wait until the `run_cosim_imem_errors` task notifies us it's added a error to the queue
+      check_inserted_iside_error_e.wait_ptrigger();
```
ctopal, line 299: Could you explain why this is needed? I think as long as `(!instr_vif.instr_cb.instr_new_id && dut_vif.dut_cb.wb_exception)` holds false this task will do nothing, is it for increasing performance?

ctopal (approved):
LGTM with one question for clarification, thanks!

dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
GregAC: It's for process synchronization, after the event wait is done we know we've just added something to `iside_error_queue` that we need to check. Without it (e.g. some kind of direct wait to observe an error) this might get scheduled before the thing that pushes the error to the queue causing issues.

Though I guess as we have the wait for the clock edge we probably don't need this synchronization. Still I prefer this method, it's more obviously correct. The `run_cosim_imem_errors` is the main process and it tells this to wakeup when it has something to do, rather than `run_cosim_prune_imem_errors` having to infer when it has something to do.

PR #1923, opened 2022-11-11
Icache fcov

GregAC:
This may be a tricky one to review as I'm the only one familiar with the inner workings of the icache (I think?).

I should also add a coverplan to the icache DV documentation but wanted to get the rest of it out as a PR ASAP

### DIFF
--- a/dv/uvm/icache/dv/fcov/ibex_icache_fcov_if.sv
+++ b/dv/uvm/icache/dv/fcov/ibex_icache_fcov_if.sv
@@ -1,0 +1,272 @@
+interface ibex_icache_fcov_if import ibex_pkg::*; #(
+  parameter int NUM_FB = 4
+) (
+  input                   clk_i,
+  input                   rst_ni,
+  input                   req_i,
+  input                   branch_i,
+  input                   icache_enable_i,
+  input                   icache_inval_i,
+  input                   lookup_throttle,
+  input                   lookup_req_ic0,
+  input                   tag_req_ic0,
+  input                   data_req_ic0,
+  input                   fill_req_ic0,
+  input                   inval_write_req,
+  input                   ecc_write_req,
+  input                   skid_valid_q,
+  input                   valid_o,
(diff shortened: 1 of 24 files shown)


dv/uvm/icache/dv/fcov/ibex_icache_fcov_if.sv
```diff
@@ -0,0 +1,269 @@
...
+  `include "dv_fcov_macros.svh"
+  import uvm_pkg::*;
+
+  typedef enum {
+    FBIdle,
+    FBReqing,
+    FBReqingAndFilling,
+    FBFilling,
+    FBAllocating,
+    FBAwaitingOutput
```
ctopal, line 64: nit: I suggest using `FillBuf` instead of `FB`

dv/uvm/icache/dv/fcov/ibex_icache_fcov_if.sv
```diff
@@ -0,0 +1,269 @@
...
+      if (i_other_fb != i_fb) begin
+        assign addr_matches[i_other_fb] = fill_busy_q[i_fb] & fill_busy_q[i_other_fb] &
+          (fill_addr_q[i_fb] == fill_addr_q[i_other_fb]);
+      end else begin
+        assign addr_matches[i_other_fb] = 1'b0;
+      end
+    end
+
+
+    covergroup icache_fb_cg @(posedge clk_i);
```
ctopal, line 249: nit: Suggest to change the name into `icache_fillbuf_cg`

ctopal (approved):
LGTM implementation wise but I believe we also need to have at least a basic documentation to point out to the specific coverpoints. Because it is hard to review thoroughly without that. (Also needed to sign-it off I think)

GregAC:
@ctopal I've added a coverage plan, you can view it here: https://github.com/lowRISC/ibex/blob/feccfb09a4212e260246aeb7faccd98bd1f917cf/dv/uvm/icache/doc/ibex_icache_dv_plan.md PTAL

PR #1928, opened 2022-11-13
[dv] Fixup coverage collection to match OT dvsim flow

GregAC:


GregAC:
This helps improve architectural coverage as the bad immediates meant certain coverpoint bins for these instructions weren't getting recorded as hits when they should have been

marnovandermaas (approved):
LGTM, but good for someone else to have a look too.

GregAC:
@hcallahan-lowrisc could you take a quick look?

dv/uvm/core_ibex/Makefile
```diff
@@ -41,6 +41,9 @@ SIGNATURE_ADDR      := 8ffffffc
 ### Ibex top level parameters ###
 IBEX_CONFIG         := opentitan
 
+# Path to DUT used for coverage reports
+DUT_COV_RTL_PATH		:= "ibex_top"
```
hcallahan-lowrisc, line 45: One tiny nit, could you swap the tabs for spaces in the Makefile changes please.

dv/uvm/core_ibex/Makefile
```diff
@@ -41,6 +41,9 @@ SIGNATURE_ADDR      := 8ffffffc
 ### Ibex top level parameters ###
 IBEX_CONFIG         := opentitan
 
+# Path to DUT used for coverage reports
```
hcallahan-lowrisc, line 44: Is this value something you would want to override on the command line? If not, it might just be cleaner to hardcode it in `metadata.py`.

hcallahan-lowrisc (approved):
LGTM, with one question and a tiny nit.

dv/uvm/core_ibex/Makefile
GregAC: Likely not but I'd prefer not to hardcode it into the script as it's specific to the testbench structure where the scripts are a more general flow that could be used with multiple testbench setups. So in absence of a better place to put such things I'd like to leave it here.

PR #1935, opened 2022-11-15
V2s directed tests

GregAC:


examples/sw/simple_system/dit_test/count_ones_zeros.S
```diff
@@ -0,0 +1,53 @@
...
+  j count_ones_zeros.loop_end
+count_ones_zeros.loop_end:
+  srli a0, a0, 1
+  addi t4, t4, 1
+  bne t4, t0, count_ones_zeros.loop_start
+  csrr t6, mcycle
+  sub a0, t6, t5
+  jr ra
+
+# Divides a0 by every bit in a0 one by one, shifting a0 left one bit every
```
andreaskurth, line 34:
```suggestion
# Divides a0 by every bit in a0 one by one, shifting a0 right one bit every
```

examples/sw/simple_system/dit_test/dit_test.c
```diff
@@ -0,0 +1,158 @@
...
+void set_icache(uint32_t on) {
+  if (on) {
+    __asm__ volatile("csrsi 0x7c0, 0x1");
+  } else {
+    __asm__ volatile("csrci 0x7c0, 0x1");
+  }
+}
+
+// Run NUM_ITERATIONS of count_ones_zeros noting the cycle time from each.
+// Return how many times one iteration differeed in cycle time from the next.
```
andreaskurth, line 45:
```suggestion
// Return how many times one iteration differed in cycle time from the next.
```

examples/sw/simple_system/dit_test/dit_test.c
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    last_cycles = cycles;
+  }
+
+  return mismatches;
+}
+
+// Run NUM_ITERATIONS of div_ones_zeros noting the cycle time from each.
+// Return how many times one iteration differeed in cycle time from the next.
```
andreaskurth, line 66:
```suggestion
// Return how many times one iteration differed in cycle time from the next.
```

examples/sw/simple_system/dit_test/dit_test.c
```diff
@@ -0,0 +1,158 @@
...
+    uint32_t cycles = div_ones_zeros(val);
+    if (cycles != last_cycles) {
+      ++mismatches;
+    }
+
+    last_cycles = cycles;
+  }
+
+  return mismatches;
+}
```
andreaskurth, line 84: Low prio suggestion: I think this code could be deduplicated by letting the test function take a function pointer

examples/sw/simple_system/dit_test/dit_test.c
```diff
@@ -0,0 +1,158 @@
...
+int main(void) {
+  uint32_t mismatches;
+  pcount_enable(1);
+
+  set_dit(0);
+  mismatches = test_b_run_time();
+  if (mismatches < (NUM_ITERATIONS / 4)) {
+    puts(
+        "FAILURE: Too few cycle mismatches seen for branch test when DIT "
+        "disabled: ");
```
andreaskurth, line 95:
```suggestion
        "disabled: 0x");
```
Suggest to prepend base to avoid misinterpretation (also before the following `puthex`es and in `dummy_instr_test.c`)

examples/sw/simple_system/dit_test/dit_test.c
```diff
@@ -0,0 +1,158 @@
...
+
+  set_icache(1);
+  // Warm up ICache
+  test_div_run_time();
+
+  // Run test with ICache enabled
+  mismatches = test_div_run_time();
+  if (mismatches != 0) {
+    puts(
+        "FAILURE: Cycle mismatches seen for divide test when DIT nd ICache "
```
andreaskurth, line 150:
```suggestion
        "FAILURE: Cycle mismatches seen for divide test when DIT and ICache "
```

examples/sw/simple_system/dummy_instr_test/busy_work.S
```diff
@@ -0,0 +1,37 @@
...
+  sub t4, t4, t3
+  and t4, t4, t1
+
+  sw t4, 0(a4)
+
+  addi a0, a0, -1
+  addi a1, a1, 1
+  addi a2, a2, 1
+  addi a3, a3, 1
+  addi a4, a4, 1
```
andreaskurth, line 31:
```suggestion
  addi a1, a1, 4
  addi a2, a2, 4
  addi a3, a3, 4
  addi a4, a4, 4
```
Each array element has 4 bytes, right?

andreaskurth (commented):
Essentially LGTM, thanks @GregAC!

I think a few details should be fixed, though

examples/sw/simple_system/dit_test/dit_test.c
GregAC: I preferred to keep it as simple as possible even where that results in duplication.

examples/sw/simple_system/dummy_instr_test/busy_work.S
GregAC: Whoops! Well spotted.

Doesn't actually matter for the purposes of the test but better to do things properly.

GregAC:
Thanks for the review @andreaskurth I have made your suggested corrections. I've also added a comment block at the top of each file to explain the test in more detail.

andreaskurth (approved):
Great, thanks!

msfschaffner (approved):

PR #1945, opened 2022-11-16
Update Complaince Testing framework for ibex

Abdulwadoodd:
This PR replaces the 0.x version of Compliance test suites with 3.x version which uses RISCOF framework.


GregAC:
Hi @Abdulwadoodd thanks for looking into this, moving to the latest compliance suite framework is something we'd like to do.

Though along with it I'd been planning to retire the compliance suite specific riscv-compliance system (https://github.com/lowRISC/ibex/tree/master/dv/riscv_compliance) in favour of using simple system with co-simulation (https://github.com/lowRISC/ibex/tree/master/dv/verilator/simple_system_cosim). I'd also like to be able to run the compliance suite under our DV framework.

So I won't merge this PR as is but will leave it open as it's got much of the work needed to move over to RISCOF we just need the switch to simple system too. Someone at lowRISC will eventually get around to this but it's not the highest point on our priority list.

In the meantime feel free to take a look at doing this yourself if you want to.

PR #1947, opened 2022-11-17
Pmp region exec fcov

GregAC:


dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -431,6 +439,16 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
           csr_pmp_cfg[i_region].lock &&
           cs_registers_i.g_pmp_registers.g_pmp_csrs[i_region].u_pmp_addr_csr.wr_en_i &&
           csr_pmp_mseccfg.rlb)
+
+        pmp_wr_exec_region : cross pmp_region_priv_bits_wr, csr_pmp_mseccfg.rlb iff
+            (fcov_csr_write &&
+             cs_registers_i.csr_we_int &&
+             cs_registers_i.csr_addr == (CSR_OFF_PMP_CFG + (i_region[11:0] >> 2)) &&
+             !pmp_region_priv_bits[2]) {
```
marnovandermaas, line 447: You're checking whether the X bit of the current config is 0. In non-MML mode this means the current configuration is non-executable. This does not work when MML is set to 0 however. Is that a problem?

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -431,6 +439,16 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+          ignore_bins no_exec_bins = binsof(pmp_region_priv_bits_wr) intersect {NONE, R, W, WR, L,
+            LR, LW, LWR, MML_NONE, MML_RU, MML_WRM_RU, MML_WRU, MML_WRM_WRU, MML_L, MML_RM, MML_WRM,
+            MML_RM_RU};
```
marnovandermaas, line 450: Put a comment here that these are all the values for `pmp_prv_bits_e` that are not executable.

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
```diff
@@ -237,3 +237,73 @@ class ibex_cross_pmp_region_mem_access_stream extends riscv_directed_instr_strea
...
+    if (cfg.init_privileged_mode != MACHINE_MODE) begin
+      // Cannot write to pmpcfgX CSRs in U Mode so skip inserting this sequence
+      return;
+    end
+
+    if (!std::randomize(pmp_region) with {
+          pmp_region > 1;
+          pmp_region < cfg.pmp_cfg.pmp_num_regions;
+          !cfg.pmp_cfg.pmp_cfg[pmp_region].x;
+          !cfg.pmp_cfg.pmp_cfg[pmp_region].l;
```
marnovandermaas, line 267: `MML_XM_XU` is executable but excluded by this rule?

marnovandermaas: Also, I think `MML_RM_RU` passes this test even though it is not executable?

marnovandermaas (commented):
Nice work, I just had some questions.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
GregAC: Ah good point, I was being a bit lazy with just checking the 'X' bit without going back over the ePMP spec properly.

The aim of this is capture any attempt to add an M-mode executable region using a config that doesn't currently allow M-mode execution. So with this check it'll also capture an attempt to modify a locked shared code region too, probably best I exclude that.

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
GregAC: This is seeking to find PMP regions that don't currently allow execution from M mode, so `MML_XM_XU` being excluded is fine, as is `MML_RM_RU` being included.

However it's still flawed as it wouldn't give you a region that allows execution from U mode where just setting the lock bit turns it into execution from M mode.

I think I'll just change it to search for non-locked regions. Locked non-executable regions aren't hugely interesting for this test anyway as it's specifically trying to stimulate the slightly trickier scenario of something adding an executable region and if it starts off locked then it's not modifiable anyway so no need for the more complex 'are we added an M mode executable region' checking anyway.

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
```diff
@@ -237,3 +237,75 @@ class ibex_cross_pmp_region_mem_access_stream extends riscv_directed_instr_strea
...
+    begin
+      `uvm_info(`gfn,
+        {"WARNING: Cannot choose random PMP region with no lock bit, skipping making PMP region ",
+         "executable"}, UVM_LOW)
+      return;
+    end
+
+    // Choose a new config which is executable when MML is set
+    // 4 config bits are {L, X, W, R}
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(new_cfg, new_cfg inside {4'b0100, 4'b0101, 4'b0111,
```
GregAC, line 277: Changed how we choose a new config to write to be more sophisticated as well, we know randomly choose amongst all possible configs that give you an M or U mode executable region with MML.

GregAC (commented):
It may be worth expanding this cross to include the config we started with as well, though to properly hit it all we need a more sophisticated test. I feel this is better as V3 work (where we're reworking PMP test generation anyway).

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
```diff
@@ -237,3 +237,75 @@ class ibex_cross_pmp_region_mem_access_stream extends riscv_directed_instr_strea
...
+      4'b1100, 4'b1010, 4'b1110, 4'b1101};)
+
+    pmpcfg_num = pmp_region / 4;
+    pmpcfg_bits = {new_cfg[3], 4'b0, new_cfg[2:0]} << ((pmp_region % 4) * 8);
+
+    initialize_instr_list(3);
+
+    // Use CSRRS to set the L and X bits of the region we are targetting
+    li_instr = riscv_pseudo_instr::type_id::create("li_instr");
```
GregAC, line 286: Realised the CSRRS won't work correctly with my new set any random config setup. Will rewrite so it does.

GregAC:
@marnovandermaas I've rejigged this, the cross now looks at current configuration vs configuration we're trying to write but excludes all non MML behaviour.

The direct instruction sequence now doesn't aim to find specific PMP configurations it just randomly chooses any and tries to make it a (randomly chosen) executable config.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -431,6 +439,33 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+        // Only sample this cross when we attempt to write to a PMP config that doesn't currently
+        // allow execution with MML enabled
+        pmp_wr_exec_region :
+          cross pmp_region_priv_bits, pmp_region_priv_bits_wr, csr_pmp_mseccfg.rlb iff
+            (fcov_csr_write &&
+             csr_pmp_mseccfg.mml &&
+             cs_registers_i.csr_we_int &&
+             cs_registers_i.csr_addr == (CSR_OFF_PMP_CFG + (i_region[11:0] >> 2)) &&
+             ((!pmp_region_priv_bits[2] && pmp_region_priv_bits != MML_XM_XU) ||
+              pmp_region_priv_bits inside {MML_WRM_WRU, MML_RM_RU})) {
```
marnovandermaas, line 452: Isn't this the same as the `from_exec_bins` ignore bin below?

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
```diff
@@ -237,3 +237,97 @@ class ibex_cross_pmp_region_mem_access_stream extends riscv_directed_instr_strea
...
+      return;
+    end
+
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(pmp_region,
+      pmp_region > 1; pmp_region < cfg.pmp_cfg.pmp_num_regions;)
+
+    // Choose a new config which is executable when MML is set
+    // 4 config bits are {L, X, W, R}
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(new_cfg, new_cfg inside {4'b0100, 4'b0101, 4'b0111,
+      4'b1100, 4'b1010, 4'b1110, 4'b1101};)
```
marnovandermaas, line 270: Just an observation: Same order as in the Smepmp spec, but different from how they're defined in `core_ibex_pmp_fcov_if.sv`.

marnovandermaas (approved):
LGTM, just a few minor comments.

dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
GregAC: Yes but I don't think the tool can translate what passes the `iff` into excluded bins so they need to be added as ignore or illegal anyway.

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
GregAC: Both `core_ibex_pmp_fcov_if` and this (and the RISC-V Privileged spec) use the LXWR order. whilst the Smepmp spec use LRWX (Section 2.1 of https://github.com/riscv/riscv-tee/blob/main/Smepmp/Smepmp.pdf)

PR #1951, opened 2022-11-17
Cov waivers

GregAC:


dv/uvm/core_ibex/waivers/coverage_waivers_xlm.tcl
```diff
@@ -0,0 +1,25 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Exclude standard primitives that are verified elsewhere
+exclude -type "prim_secded_*"
+exclude -type "prim_lfsr"
+exclude -type "prim_flop"
+exclude -type "prim_prince"
+exclude -type "prim_subst_perm"
+exclude -type "prim_ram_1p*"
```
msfschaffner, line 11: @weicaiyang can you take a look at these whether they look good?
I believe we have only have testbenches for `prim_lfsr` and `prim_prince`, but not for the other primitives - see here: https://reports.opentitan.org/hw/top_earlgrey/dv/summary/latest/report.html

msfschaffner: @GregAC do we need to exclude all of them? we typically only exclude the ones that are pre-verifed, and collect coverage on the others...

dv/uvm/core_ibex/waivers/coverage_waivers_xlm.tcl
```diff
@@ -0,0 +1,25 @@
...
+# Exclude code coverage from aux code used for gathering functional coverage
+exclude -type "core_ibex_fcov_if" -metrics code:statement:fsm:assertion
+exclude -type "core_ibex_pmp_fcov_if" -metrics code:statement:fsm:assertion
+
+# RVFI signals not present in real design so not relevant to coverage closure
+exclude -type "ibex_top" -toggle "rvfi*"
+# hart_id_i and boot_addr_i are intended to be hard-wired inputs that do not
+# toggle
+exclude -type "ibex_top" -toggle "hart_id_i"
+exclude -type "ibex_top" -toggle "boot_addr_i"
```
sriyerg, line 18: I feel like these can be covered in V3 tests and should not be excluded since they are design ports. as opposed to parameters.

dv/uvm/core_ibex/waivers/coverage_waivers_xlm.tcl
weicaiyang: Agree with @msfschaffner. We can only exclude pre-verified IPs, otherwise, we need to have a good reason on the exclusion.

For the fair amount of uncovered, is it due to unreachable code? We may consider running UNR to take them out.

GregAC: Thankfully it appears we still hit our V2 coverage goals when we only exclude `prim_prince` and `prim_lfsr`. So I'll leave it like that for now. We can get a UNR flow setup for Ibex post V2.

msfschaffner: Ok this sounds great, thanks for checking @GregAC!

msfschaffner (approved):

msfschaffner (approved):

dv/uvm/core_ibex/waivers/coverage_waivers_xlm.tcl
GregAC: The issue is they probably should be parameters and they're intended to be totally static. As it happens I think you can alter them at will and get some sensible behaviour but they are meant to be wired to a constant.

At this stage I don't want to go altering Ibex's parameter set so I think they should remain design ports but I think it's reasonable not to test toggling, especially for the OT use case where they are wired to a constant in the earlgrey top-level.

PR #1962, opened 2022-12-07
Directed test tooling

hcallahan-lowrisc:
~~Draft PR, probably requires some tidying, nits need picking etc.~~
Now ready for review.

This adds functionality to specify directed tests using a seperate `core_ibex/directed_tests/directed_testlist.yaml` file. Both tests from this file and from the existing riscvdv `testlist.yaml` are added to the regression, and can be selected either/or/both on the command line using the existing TEST= functionality, e.g. `TEST=riscv_loop_test,empty`

I have worked on-top of a commit from @Saad525 that gave me a basic directed test to work with, called `empty`. New directed tests can be added by following the format specified in the new yaml file. This format largely re-uses the keys from the riscvdv file, but adds a few different ones to specify compilation options for building the test.
The directed test sources are ingested, then compiled by creating a gcc command within `compile_test.py::get_directed_compile_cmds()`. This function is a simple text-substitution to assemble the command with correct paths for our directory structure. Currently it only works when provided with assembly sources (eg. asm.S), but this shouldn't be too difficult to extend to tests written in C.

The main changes to achieve the above was to :
1) modify the metadata.py to ingest two yaml files, then store the full testlist with each type of test marked.
2) modify the makefile to have different compile_test targets for the two different types of tests (riscvdv/directed). This is needed because the layout of src files is different between the tests, so the current static-pattern rules no longer work carte blanche.  (I also slightly refactored the makefile by splitting into smaller components.) The riscvdv tests continue to be compiled by requesting the compilation command from the riscv-dv python scripts.

NB.
One small workflow change as this stands, is that if you are running a goal that is not `all` (ie. make/run/check tests, gen regr.log), you now need to specify it using the GOAL arg. e.g. `make TEST=empty ITERATIONS=1 GOAL=compile_test`, which will stop after compiling the 'empty' directed test. See `scripts/wrapper.mk` for the list of phony goals.


dv/uvm/core_ibex/scripts/riscvdv.mk
```diff
@@ -0,0 +1,80 @@
+###############################################################################
```
Saad525, line 1: @hcallahan-lowrisc do you think a short description to be added at the beginning?

dv/uvm/core_ibex/directed_tests/access_pmp_overlap/access_pmp_overlap.S
```diff
@@ -0,0 +1,32 @@
+#include "riscv_test.h"
```
GregAC, line 7: Missing lowRISC copyright header

dv/uvm/core_ibex/directed_tests/encoding.h
```diff
@@ -0,0 +1,2876 @@
+/* See LICENSE for license details. */
```
GregAC, line 1: Where did this come from? It looks derived from either https://github.com/riscv-non-isa/riscv-arch-test/blob/main/riscv-test-suite/env/encoding.h, this earlier version of it looks closer: https://github.com/riscv/riscv-test-env/blob/0666378f353599d01fc48562b431b1dd049faab5/encoding.h but still not identical.

As this isn't a file authored by us we should at least ensure there's a proper header providing attribution (the current comment refers to a 'LICENSE' file but the top-level LICENSE does not apply to this). Preferably we'd vendor this in rather than adding it directly here. As I think we're vendoring in riscv-arch-test anyway to support the architectural test suite maybe we should just use the `encoding.h` from there?

dv/uvm/core_ibex/directed_tests/riscv_test.h
```diff
@@ -0,0 +1,274 @@
+// See LICENSE for license details.
```
GregAC, line 1: As with `encoding.h` I think this has come from some other RISCV repo? The same comments apply in that this should really be vendored in rather than included directly here.

dv/uvm/core_ibex/directed_tests/test_macros.h
```diff
@@ -0,0 +1,754 @@
+// See LICENSE for license details.
```
GregAC, line 1: As with `encoding.h` I think this has come from some other RISCV repo? The same comments apply in that this should really be vendored in rather than included directly here.

dv/uvm/core_ibex/Makefile
```diff
@@ -1,12 +1,15 @@
...
+###############################################################################
 
 .SUFFIXES:
+export
 
 # Explicitly ask for the bash shell
 SHELL := bash
 
+# Build the 'all' target by default, override with e.g. GOAL=rtl_tb_compile
+GOAL ?= all
```
GregAC, line 15: Is this correct? We have no 'all' target in this makefile but we do have a 'run'

dv/uvm/core_ibex/scripts/ibex_sim.mk
```diff
@@ -0,0 +1,99 @@
...
+  | $(BUILD-DIR)
+	@echo Building RTL testbench
+	$(verb)env PYTHONPATH=$(PYTHONPATH) \
+	scripts/compile_tb.py \
+	  --dir-metadata $(METADATA-DIR)
+	$(call dump-vars,$(tb-compile-vars-path),comp,$(rtl-tb-compile-var-deps))
+	@touch $@
+
+###############################################################################
+# Run ibex RTL simulation with randomly-generated program and uvm stimulus
```
GregAC, line 49: If I've understood this correctly this step will also run the directed tests? If so the comment needs updating.

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+##########################################################
+
+# This file largely copies the formatting of the testlist.yaml
+# used by riscv-dv, but only specifies directed tests.
+
+
+##########################################################
```
GregAC, line 14: Shouldn't the 'access_pmp_overlap.S' test be in here?

GregAC (changes requested):
My apologies for taking so long to get around to reviewing this. The flow is looking good though I've set request changes as we need to sort out vendoring in the encoding.h/riscv_test.h/test_macros.h files and as is this PR doesn't have the proper copyright headers on them anyway so we shouldn't merge it into the repo.

As a note for the future when doing a major refactoring like we have here it would be appreciated if it could be broken down into multiple commits, it makes reviewing it a lot easier!

dv/uvm/core_ibex/directed_tests/access_pmp_overlap/access_pmp_overlap.S
Saad525: thanks for pointing out. Added it.

dv/uvm/core_ibex/directed_tests/encoding.h
Saad525: It was from a submodule of riscv-tests ([riscv-test-env](https://github.com/riscv/riscv-test-env/tree/master)). Now, I added a separate file ibex_macros.h for implementation specific stuff while using the encoding.h from the vendored repo without any modification.

dv/uvm/core_ibex/directed_tests/riscv_test.h
Saad525: Added a patch file for it since some modifications was required for test start sequence of instructions, test ending mechanism for Ibex.

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
Saad525: thanks for pointing out. Added it here and modified it with a handler to cleanly fail in case of unexpected exception.

dv/uvm/core_ibex/Makefile
Saad525: In run target, its calling,
```
@$(MAKE) --file wrapper.mk --environment-overrides --no-print-directory $(GOAL)
```
and wrapper.mk has 'all' target, does it make sense or i am missing something? @hcallahan-lowrisc

hcallahan-lowrisc: Yes that's correct, it just passes the GOAL variable onto the wrapper.mk file, which does have an 'all'target

Saad525:
LGTM!

Saad525 (approved):

GregAC (approved):
Thanks for the all work on this @Saad525 and @hcallahan-lowrisc

Angie20182 (approved):

PR #1967, opened 2023-01-06
[rtl] Explicitly specify var type for ibex_core inputs for FI testing

vogelpi:
Without the `var` keyword, some simulators decide to model these inputs as wires. As a results, glitches inserted inside ibex_core can propagate back up to the `ibex_top` hierarchy and from there into `ibex_lockstep` and the shadow core. As a result, both cores get glitched in simulation.


vogelpi:
FYI, without this we get a large amount of erroneous lockstep glitch test failures with Xcelium.

GregAC (changes requested):
Good catch! Though I am hesitant to approve this without understanding  if there's any impact on synthesis results. Given it's altering the semantics of the IO it has the potential to do weird things in synth.

If we do switch to this style I think we need to update the style guide with a recommendation and also consider where else in OT we may want similar.

andreaskurth:
> Without the `var` keyword, some simulators decide to model these inputs as wires.

Those simulators are correct -- in fact, I think *all* simulators should do this -- if I understood the SystemVerilog spec (Section 23.2.2.3 of IEEE 1800-2012) correctly: "An implicit data type declaration implies a net unless the `var` keyword is used."

> As a results, glitches inserted inside ibex_core can propagate back up to the `ibex_top` hierarchy and from there into `ibex_lockstep` and the shadow core. As a result, both cores get glitched in simulation.

On the IC, the `prim_buf` we use to buffer the input signals should prevent this, right? If so, can we address this problem in the simulation model of `prim_buf`?

vogelpi:
The `prim_buf` cells behave as expected (glitches not propagating back) but they are only used on the shadow core inputs and not on the main core inputs. As a result, the issue occurs only when glitching the main core inputs.

We could also insert `prim_buf` cells on the main core inputs but this will certainly have an area impact. From a FI vulnerability point of view, there is little difference though. It's probably best to quickly discuss this tomorrow in the HW sync-up.

vogelpi:
> > Without the `var` keyword, some simulators decide to model these inputs as wires.
> 
> Those simulators are correct -- in fact, I think _all_ simulators should do this -- if I understood the SystemVerilog spec (Section 23.2.2.3 of IEEE 1800-2012) correctly: "An implicit data type declaration implies a net unless the `var` keyword is used."

After some more reading and our discussion from Tuesday, I believe the meaning of this section of the specification is that if for a port declaration you don't write `logic` or `wire` (which is the data type), a net is implied. However, in the Ibex codebase we generally specify the data type of ports as `logic`. So the data type isn't implicit.

I couldn't find in section of the standard explicitly stating that `logic` without net type implies `var` but Section 6.5 "Nets and variables" contains the following example:
```sv
wire w = vara & varb;      // net with a continuous assignment

logic v = consta & constb; // variable with initialization

logic vw;                  // no initial assignment
assign vw = vara & varb;   // continuous assignment to a variable
```
This tells me that if we don't specify the net type for a logic, `var` should be used. So, the tool simulating the back propagation is doing the wrong thing here. From an offline discussion with @rasmus-madsen (thanks again) I know that this is a performance optimization of that tool to collapse ports into a single object to reduce the number of assignments. Explicitly specifying `var` prevents this collapsing but from a language standard viewpoint it shouldn't be necessary.

@johannheyszl also brought up that instead of inserting `prim_buf` cells or the `var` keyword for these input ports, we could change the FI test to work as follows:
- shadow core inputs are forced as before (there is no back propagation)
- when forcing the main core inputs, also the shadow core inputs are forced but to the opposite value to defeat the back propagation done by some tools.

I think this is a great idea as it means we don't have to make big RTL changes. So this PR could actually be closed as we most likely don't need it. However, we will further discuss this matter in an FI meeting tonight and I will update this thread based on the outcome of this discussion.

vogelpi:
We just discussed this in the SCA/FI meeting and it was agreed that both the `var` and the dual forcing approach proposed by @johannheyszl would be fine.

From a modelling perspective we concluded the following:
- We actually want that the entire net is changing its value upon a force. The effect of this is similar to both Laser FI on the net driver and EM FI on the net itself.
- Instead of forcing core inputs to test the lockstep countermeasure, we should actually be forcing internal core signals. However, the problem with this is that it's infeasible to always correctly model how the forcing of any internal signal impacts core behavior and ultimately whether this should be detected. Forcing core inputs is a way to make the test more feasible. 
- We have to remain aware that for some inputs there may exist single points of failure, e.g. some inputs without integrity protection, some control signals without spurious enable detection etc. There are always such single points of failures. Where and how many there are depends on the surrounding modules, backend etc. and is out of scope for the lockstep countermeasure.

I'll give the dual forcing approach a try and will then feed back again.

vogelpi:
I've tried out the dual force approach and it works. The PR is here lowRISC/OpenTitan#17102.

andreaskurth:
Closing this in favor of [lowRISC/opentitan#17102](https://github.com/lowRISC/opentitan/pull/17102), which has now been merged.

PR #1968, opened 2023-01-09
[rtl] Improve FI hardening around data_rvalid_i

vogelpi:
Previously, it was possible to glitch data_rvalid_i at the interconnect level and if the data integrity bits happened to be valid, Ibex would write the current data_rdata_i into the register file even if it wasn't doing a load. Since the glitch is inserted at the interconnect level, both the main and the shadow core are affected equally.

This commit changes the WB stage to only forward the LSU write enable, which is generated from data_rvalid_i, when Ibex is actually waiting for an interconnect response for a load instruction. This substantially narrows down the window for attacks at the interconnect level.

### DIFF
--- a/rtl/ibex_wb_stage.sv
+++ b/rtl/ibex_wb_stage.sv
@@ -166,6 +166,9 @@ module ibex_wb_stage #(
     // that returns too late to be used on the forwarding path.
     assign rf_wdata_fwd_wb_o = rf_wdata_wb_q;
 
+    // For FI hardening, only forward LSU write enable if we're actually waiting for it.
+    assign rf_wdata_wb_mux_we[1] = outstanding_load_wb_o & rf_we_lsu_i;
+
     if (DummyInstructions) begin : g_dummy_instr_wb
       logic dummy_instr_wb_q;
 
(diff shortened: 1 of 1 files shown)


andreaskurth (approved):
Very good catch!

Do we also need to harden the Ibex configuration without WB stage against this? (Probably not for OT, but we could create an issue in the Ibex repo to track this?)

vogelpi:
> Very good catch!
> 
> Do we also need to harden the Ibex configuration without WB stage against this? (Probably not for OT, but we could create an issue in the Ibex repo to track this?)

Thanks @andreaskurth . I think I wouldn't harden the configuration without WB, as this config is less relevant and the hardening there would be more involved. The nice thing about hardening the WB configuration is that the required flop is already there. So it's just about adding a single AND gate.

andreaskurth:
> I think I wouldn't harden the configuration without WB, as this config is less relevant and the hardening there would be more involved. The nice thing about hardening the WB configuration is that the required flop is already there. So it's just about adding a single AND gate.

Thanks for the explanation. I agree that hardening the config without WB is not necessary at the moment, but I would suggest to either track this as an issue or document that some hardening requires the WB stage to be instantiated. Right now, the [Ibex docs](https://ibex-core.readthedocs.io/en/latest/03_reference/security.html#) state "Ibex implements a set of extra features (when the `SecureIbex` parameter is set) to support security-critical applications", which is logically correct but does not mention that there are other parameters that influence the implemented security features.

vogelpi:
This sounds reasonable @andreaskurth . I've created an issue to track this here #1970 and will now merge this PR.

PR #1969, opened 2023-01-16
[doc] Update block diagram

ctopal:


vogelpi (commented):
Thanks for doing this @ctopal . This LGTM but others probably also want to have a say.

Also, I remember @andreaskurth mentioning that something (don't remember what exactly) could  improved in this new diagram.

andreaskurth (commented):
LGTM overall, thanks @ctopal!

I just have two comments:
1. The block at the bottom of the Instruction Fetch stage is supposed to be the _Compressed Decoder_, right? It does not have a label (at least not in my browser -- is it an SVG problem?).
2. The line around _Writeback_ should not be dotted IMO, since only the Writeback register is optional but the mux from the Decode & Execute stage or the LSU is not.

GregAC:
+1 to @andreaskurth's comments

In addition the label on the Controller seems to have suffered the same fate as the compressed decoder.

From prior feedback on my original email thread where I did the B&W version of this @andreaskurth pointed out the decoder shouldn't be nested in the controller, instead they should sit side by side (my mistake I had decided the decoder lived in the controller but they're both inside the ID stage).

andreaskurth (approved):
LGTM, thanks!

PR #1971, opened 2023-01-17
[python-requirements] Pin setuptools version to < 66.0.0

vogelpi:
Starting with setuptools version 66.0.0, legacy package version names such as 0.23ubuntu1 are no longer supported. Since some of our Python dependencies use this format, we pin the setuptools version to the last version before this change. This unblocks CI and gives us time to upgrade/rebase our dependencies.


vogelpi:
For details, see also lowRISC/OpenTitan#17098 and lowRISC/OpenTitan#17099.

ctopal (approved):

jprendes (commented):
LGTM

PR #1975, opened 2023-01-24
[rtl/dv] Add assertions for icache scramble keys

andreaskurth:
This commit adds two assertions in `ibex_top` to ensure that the scramble key is correctly applied to the icache scrambled memory primitives.  Those assertions previously existed in the module that instantiated Ibex in OpenTitan, but the reference into the generate loops was problematic for some EDA tools; see lowRISC/opentitan#17155.

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -620,6 +620,36 @@ module ibex_top import ibex_pkg::*; #(
           .cfg_i       (ram_cfg_i)
         );
 
+        `ifdef INC_ASSERT
+          // Sample scramble key whenever it is valid for use in the assertions below.  This may be
+          // redundant with the sampling performed in the actual design, but that is okay because
+          // the assertions exist to check the correct functioning of the design.
+          logic [SCRAMBLE_KEY_W-1:0] sampled_scramble_key;
+          always_ff @(posedge clk_i, negedge rst_ni) begin
+            if (!rst_ni) begin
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks @andreaskurth for taking care of this, the changes LGTM but the smoketest for the opentitan config which uses these changes is currently failing with Xcelium. This needs to be debugged first.

andreaskurth:
Thanks @vogelpi. I had a quick look, and for 3 out of 4 programs that are part of the smoke test, the assertion fails. Those failures did not occur in the OT tests, but Ibex is operated in more corner conditions in this DV environment. If changing the assertions to `[0:$]`, they don't fail -- but maybe a reset at the end of the test kills the assertions? I plan to debug this with waves as soon as I can allocate some time for that.

andreaskurth:
I found the problem: The assertions previously used the input scramble key (`scramble_key_i`) even though they tolerated a delay after which the input scramble key was not necessarily valid anymore (i.e., `scramble_key_valid_i` could go low and `scramble_key_i` could take any value).  This mistake has been corrected by sampling the input scramble key for the assertions when it is valid and using the sampled value in the comparison of the assertions.

vogelpi:
Thanks for debugging and fixing this @andreaskurth . It makes sense and your changes LGTM.

GregAC (approved):

PR #1982, opened 2023-02-07
Override misa and mtvec CSRs in cosim

Saad525:
Override mtvec and misa CSRs to match the Ibex behavior. Inconsistency with Spike was due misa and mtvec having following behaviour in Ibex:
* misa is hardwired
* mtvec has a mask such that it is 256 byte aligned and always vectored

These issues were caught when misa and mtvec are written and read back in the test.

### DIFF
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -690,7 +690,30 @@ void SpikeCosim::fixup_csr(int csr_num, uint32_t csr_val) {
       if (any_interrupt && int_interrupt) {
         new_val |= 0x7fffffe0;
       }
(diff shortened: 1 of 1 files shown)


GregAC (changes requested):
Nice catch @Saad525 though I think we should use an alternative fix. If you take a look in `spike_cosim.cc` you'll find the `fixup_csr` function. This gets called when a CSR gets written specifically to fix up mismatches like this. You can see it being used for mstatus and mcause already.

Saad525:
@GregAC thanks for pointing to fixup_csr function. I have now used that function to update mtvec/misa.

GregAC (approved):

PR #1989, opened 2023-02-22
Directed test for PMP.LXWR transitions when mseccfg.mml is 1

Saad525:
Added a directed test for covering all scenarios of mml=1 with rlb=0/1 with all combinations of PMP.LXWR and accessing that configured PMP region with RW accesses.


GregAC (approved):

GregAC:
CI issue is the known MISA mismatch problem, so I'll just bypass the CI checks and merge

PR #1992, opened 2023-03-01
[rtl] Fix MISA X bit for balanced bitmanip config

GregAC:
All RV32B configs include non-ratified sub-extensions so the 'X' bit MISA must be set for all of them.

This will fix the failure in private CI


marnovandermaas (approved):
Thanks, LGTM!

PR #1995, opened 2023-03-06
Avoid explicit module names references to signals

flaviens:
Hi there!

Some tools modify module names during their flow, or allow to do it optionally, such as the `-p` flag of [Morty](https://github.com/pulp-platform/morty).
However, there is one display occurrence which refers to explicit module names, which breaks compatibility with such passes.
Arguably this is a problem of the tool, but:
* Such absolute references are rare across designs and appear only here in Ibex afaik.
* It looks much easier to fix in Ibex than in the tools.

Please feel free to edit the fix in place!
Flavien

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -174,8 +174,8 @@ module ibex_controller #(
   always_ff @(negedge clk_i) begin
     // print warning in case of decoding errors
     if ((ctrl_fsm_cs == DECODE) && instr_valid_i && !instr_fetch_err_i && illegal_insn_d) begin
-      $display("%t: Illegal instruction (hart %0x) at PC 0x%h: 0x%h", $time, ibex_core.hart_id_i,
-               ibex_id_stage.pc_id_i, ibex_id_stage.instr_rdata_i);
+      $display("%t: Illegal instruction (hart %0x) at PC 0x%h: 0x%h", $time, u_ibex_core.hart_id_i,
+               pc_id_i, id_stage_i.instr_rdata_i);
     end
   end
   // synopsys translate_on


GregAC (changes requested):
I am not a fan of adding IO to modules where that IO isn't actually needed by the logic. I think there's a couple of ways to address this that can work with morty without needing to add the extra IO:

1. Use instance names to reference the signals rather than module names: 
```systemverilog
      $display("%t: Illegal instruction (hart %0x) at PC 0x%h: 0x%h", $time, u_ibex_core.hart_id_i,
               id_stage_i.pc_id_i, id_stage_i.instr_rdata_i);
```
2. Move this display into `ibex_id_stage.sv`. We can't recreate it precisely (as we don't have visibility of the controller state) but should be able to do something that more or less behaves the same.

GregAC:
Probably worth adding something to the lowRISC style guide about this too, state that where hierarchical references are used instance names rather than module names should be used to avoid issues where people wish to rename modules.

flaviens:
Hi @GregAC, thank you for your quick review! I very much agree with avoiding useless wires.
I did not know this was possible with the instance names! :slightly_smiling_face:
I also removed the instance name before the pc id because the pc id is readily available in the module.

One could say that we are just moving the problem from the module names to the instance names, but indeed it seems that changing the instance names would be more exotic. Since you are more experienced, I let you judge of that.

Thanks!
Flavien

GregAC (approved):
> One could say that we are just moving the problem from the module names to the instance names, but indeed it seems that changing the instance names would be more exotic

Indeed we are but I think, as you say, wanting to changing instance names is significantly rarer. You need module renaming if you get a module naming conflict (e.g. perhaps you've got a modified Ibex core in the same system as another Ibex) but similar issues don't arise with instancing naming.

GregAC:
Thanks for the fix @flaviens

PR #1999, opened 2023-03-14
[doc] Remove misleading sentence above block diagram

GregAC:
The block diagram now represents both the 2 and 3 stage Ibex configurations.


marnovandermaas (approved):
Thanks for this change!

PR #2000, opened 2023-03-17
[sw/simple_system] Declare `pcount_enable` function as inline

andreaskurth:
The `pcount_enable` function can be used to selectively enable performance counters only for a specific code segment.  For example:
```c
pcount_enable(0);
pcount_reset();
pcount_enable(1);
/* code to be measured */
pcount_enable(0);
```

The `pcount_enable` function consists of a single CSR instruction, so the overhead and thus the impact on the measurement is potentially low. When the function is called, however, many instructions have to be executed in addition to the single CSR instruction, which influences measurements.

This commit moves the `pcount_enable` function to the `simple_system_common.h` header file and declares it as `inline` (and `static`, to prevent link-time collisions when the header file is included in multiple compilation units).  This helps the compiler inline the function even without LTO.

### DIFF
--- a/examples/sw/simple_system/common/simple_system_common.c
+++ b/examples/sw/simple_system/common/simple_system_common.c
@@ -102,18 +102,6 @@ void pcount_reset() {
       "csrw mhpmcounter31h, x0\n");
 }
 
-void pcount_enable(int enable) {
-  // Note cycle is disabled with everything else
-  unsigned int inhibit_val = enable ? 0x0 : 0xFFFFFFFF;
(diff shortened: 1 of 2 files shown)


andreaskurth:
CI checks fail with a Python `TypeError`. I don't think that's related to this PR :thinking:

GregAC:
> CI checks fail with a Python `TypeError`. I don't think that's related to this PR thinking

Yes I'm trying to diagnose and fix that one, looks like a failure in the typeguard code itself rather than typeguard finding a typing error in the DV python scripts. Frustratingly only occurs in CI and I cannot reproduce it locally.

GregAC (approved):

PR #2007, opened 2023-03-27
[rtl] Add missing `include to ibex_if_stage

GregAC:
The include is needed for a FCOV related macro. Lack of this include can cause issues in some simulators.


marnovandermaas (approved):
Looks fine to me, it's already included in the ID and WB stage as well.

marnovandermaas:
Let's wait for CI to finish before merging.

PR #2013, opened 2023-04-10
[util] Ignore Verilator stderr in version check

wallento:
With funny CI environments, the locale might not be set properly, making perl complain. The version check will fail then. I believe it is sufficient to check for the tool version, stderr can be ignored safely, if I am not missing something.


rswarbrick (approved):
This looks sensible to me.

The change means it will now allow a system where running `verilator --version` now spits out some message to `stderr` but still returns an ok exit code. That seems reasonable!

Note that this file actually comes from OpenTitan and was "manually vendored" by me. Someone (possibly me!) needs to make the corresponding change there if we merge this.

Issue #1919, opened 2022-11-09
Mismatch of dcsr.prv when interrupt followed by a debug request


Saad525:
## Observed Behavior

When there is an interrupt from U-mode followed by debug request, Ibex fetches the first instruction of the handler and when the debug request comes in, it takes it and jumps to corresponding address. Privilege mode already switched to M-mode and after debug request, when dcsr is read, dcsr.prv is 'b11.

![wave](https://user-images.githubusercontent.com/55890346/200887006-145b9916-c07a-4703-81ec-59141cd0a49f.png)

## Expected Behavior

Since we are feeding Spike with debug request or nmi interrupt (when multiple interrupts happen back to back before previous interrupt handler's first instruction execution), it doesn't have the transition from U-mode to M-mode, so we see dcsr.prv = 'b00 from Spike.

It seems that changes might be needed in Ibex cosim.

## Steps to reproduce the issue

```
cd dv/uvm/core_ibex
make --keep-going IBEX_CONFIG=opentitan SIMULATOR=xlm ISS=spike ITERATIONS=1 SEED=23054 TEST=riscv_assorted_traps_interrupts_debug_test WAVES=1
```

## My Environment

**Operating system:**

Ubuntu Linux 20.04

**Version of the Ibex source code:**

ba527fed80ccd3d98f47f01a712dbb60f5d53860 (Draft PR)

/cc @GregAC @hcallahan-lowrisc @ctopal 

> estimate 8

GregAC:
estimate range: 4 - 8

PR #2014, opened 2023-04-14
Cosim irq debug fix

GregAC:
This helps fix some failures in the assorted_traps_interrupts_debug_test though hasn't really effected the overall pass rate as other failures lurk behind.

Longer-term I think we need a better fix here (will open a separate issue to discuss) but felt this would do for now. It's primary purpose was to see if any RTL bugs surfaced after fixing these DV issues and there's higher priorities than continuing to chase the remaining problems. I'll open issues for the current failures seen in  assorted_traps_interrupts_debug_test explaining what the issue is and why it doesn't indicate RTL issues.


dv/cosim/spike_cosim.cc
```diff
@@ -585,11 +633,15 @@ void SpikeCosim::set_nmi(bool nmi) {
...
+
+    early_interrupt_handle();
   }
 }
 
 void SpikeCosim::set_nmi_int(bool nmi_int) {
-  if (nmi_int && !nmi_mode && !processor->get_state()->debug_mode) {
+  if (nmi_int && !nmi_mode && !processor->get_state()->debug_mode &&
+      processor->halt_request != processor_t::HR_REGULAR) {
+    bool old_nmi_int = processor->get_state()->nmi_int;
```
marnovandermaas, line 644: This value doesn't seem to be used. Do we need this line?

rtl/ibex_core.sv
marnovandermaas, line 1: I'm a bit wary that there is an RTL change here, we should think carefully before merging this.

marnovandermaas (commented):
A couple initial comments and CI seems to be throwing a lint error:
```
ERROR: %Warning-PINMISSING: ../src/lowrisc_ibex_ibex_top_0.1/rtl/ibex_lockstep.sv:348:5: Cell has missing pin: 'rvfi_ext_irq_valid'
```

dv/cosim/spike_cosim.cc
GregAC: Ah whoops we don't, failed to delete it after making some changes.

rtl/ibex_core.sv
GregAC: This change only effects the RVFI (which is used by the cosim to get information on retired instructions) and doesn't touch the actual core logic.

GregAC:
This issue also provides some context to these fixes: https://github.com/lowRISC/ibex/issues/2015

GregAC:
The problem this aims to fix was highlighted by this issue: https://github.com/lowRISC/ibex/issues/1919 there's some further info there that may help with understanding.

rtl/ibex_core.sv
marnovandermaas: Ah yes, you're right. All this code requires `RVFI` to be defined.

dv/cosim/spike_cosim.cc
```diff
@@ -571,11 +571,59 @@ void SpikeCosim::initial_proc_setup(uint32_t start_pc, uint32_t start_mtvec,
 }
 
 void SpikeCosim::set_mip(uint32_t mip) {
+  uint32_t old_mip;
+  uint32_t new_mip;
+
+  new_mip = mip;
+  old_mip = processor->get_state()->mip->read();
```
andreaskurth, line 578: Since this is C++ (and not ANSI C or (System)Verilog), combining declaration and value assignment is probably preferable:
```cpp
uint32_t new_mip = mip;
uint32_t old_mip = processor->get_state()->mip->read();
```
(suggestion also applies to other C++ variables added in this PR)

andreaskurth (approved):
LGTM overall, thanks @GregAC!

I have a few smaller points, but none of them is critical I think

dv/cosim/spike_cosim.cc
```diff
@@ -571,11 +571,59 @@ void SpikeCosim::initial_proc_setup(uint32_t start_pc, uint32_t start_mtvec,
...
+
   processor->get_state()->mip->write_with_mask(0xffffffff, mip);
+
+  if (processor->get_state()->debug_mode ||
+      (processor->halt_request == processor_t::HR_REGULAR) ||
+      (!get_field(processor->get_csr(CSR_MSTATUS), MSTATUS_MIE) &&
+       processor->get_state()->prv == PRV_M)) {
+    // Return now if new MIP won't trigger an interrupt handler either because
+    // we're in or heading to debug more or interrupts are disabled.
```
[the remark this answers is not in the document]
GregAC: Yes just  typo more->mode

GregAC:
Thanks for the reviews @andreaskurth and @marnovandermaas

PR #2016, opened 2023-04-18
[dv] Double fault detector should sample with clocking block

GregAC:


marnovandermaas (approved):
Yes, this matches what is done for other signals in the rest of the Ibex DV setup (mostly under the name `dut_vif.dut_cb`).

PR #2020, opened 2023-04-21
Remove TODOs

GregAC:
- rvfi_trap now correctly handled for writeback
 - issue created to track coverpoint for pmpcfg reserved bits writes.
 - flush pipe on debug CSR writes is reasonable

For the debug CSR write it may be a pipeline flush isn't required but it's safer to do so and doesn't present any real performance concerns so we may as well leave it in.


GregAC:
Issue here track extra coverage: https://github.com/lowRISC/ibex/issues/2019

PR #2028, opened 2023-05-10
[dv] Add asserts to check alerts for memory integrity failures

GregAC:


marnovandermaas (approved):
LGTM, this PR checks that when an instruction or data is passed to the Ibex core that causes a SECDED error that it also raises a major bus alert within the next 5 clock cycles.

marnovandermaas:
We should wait to merge this until CI is fixed right?

PR #2034, opened 2023-05-11
Drop a double entry in rtl_simulation.yaml

rswarbrick:
We ended up with the Unicode fix twice because of two colliding PR merges. We only need one copy of the -CFLAGS argument, and VCS generates a rather strange message about an unknown argument if you use it twice. Fortunately, it's easy to fix once you've worked out how to get the system to print out what it's doing.


hcallahan-lowrisc (approved):
LGTM!

PR #2037, opened 2023-05-17
Tweaks to Ibex to get core_ibex UVM to build with VCS

rswarbrick:
We're not quite there yet, but it's much closer*! I've put sensible comments on all the commits in the PR, but they are mostly to do with cleaving more carefully to the IEEE SystemVerilog spec and not assuming things that happen to be true for a different simulator.

(*) With the current code, running something like `make IBEX_CONFIG=opentitan SIMULATOR=vcs ISS=spike ITERATIONS=1 SEED=1 TEST=riscv_arithmetic_basic_test WAVES=0 COV=0` gets as far as actually generating some binaries. Next: getting sensible trace logs!

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
+++ b/dv/uvm/core_ibex/common/ibex_cosim_agent/ibex_cosim_scoreboard.sv
@@ -6,7 +6,6 @@
 `include "cosim_dpi.svh"
 
 class ibex_cosim_scoreboard extends uvm_scoreboard;
-  import ibex_pkg::*;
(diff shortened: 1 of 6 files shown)


rswarbrick:
Incidentally, I'm pleased to say that we were closer than I thought. I just saw a successful test run of `riscv_arithmetic_basic_test` (with seed 1). Phew!

hcallahan-lowrisc (approved):
All seems sensible to me. Thanks for getting this going again!

GregAC (approved):

GregAC:
Thanks for this @rswarbrick had been meaning to fix the VCS build for a while!

Issue #1419, opened 2021-08-05
Add support for mtvec direct mode


tomroberts-lowrisc:
Some SW use cases would prefer to have support for "direct" interrupt mode rather than "vectored" mode in the mtvec csr (support for FreeRTOS and for boot ROMs e.g. in OpenTitan).

The rtl change is trivial but the impact on DV needs to be investigated.

medexs:
Hi, what is the status on the direct mode support for `mtvec`? I am working on porting [NuttX](https://github.com/apache/nuttx) to Ibex and the support for the direct mode would be much appreciated as handling of all exceptions/interrupts in NuttX is done in only one routine.

PR #2044, opened 2023-05-31
[rtl] Add support for direct mode in mtvec

leesum1:
The Ibex RISC-V core currently lacks support for the direct mode in the mtvec (#1419). This commit make mtvec support both vectored and direct mode.

To ensure compatibility with existing test programs, the initial value of mtvec.MODE is set to vectored mode.For both vectored mode and direct mode, mtvec.BASE must be 256-byte aligned.

## Test
Pass test in rt-thread(direct mode)
Pass `examples/sw/simple_system/hello_test` (vectored mode)


medexs:
Hi @leesum1 , @GregAC  , what is the current status on the direct mode support for `mtvec`? From the checks above I can see that almost all of them but three passed successfully. When could be the support finished?

 I am working on porting [NuttX](https://github.com/apache/nuttx) to Ibex and the support for the direct mode would be much appreciated as handling of all exceptions/interrupts in NuttX is done in only one routine.

leesum1:
> Hi @leesum1 , @GregAC , what is the current status on the direct mode support for `mtvec`? From the checks above I can see that almost all of them but three passed successfully. When could be the support finished?
> 
> I am working on porting [NuttX](https://github.com/apache/nuttx) to Ibex and the support for the direct mode would be much appreciated as handling of all exceptions/interrupts in NuttX is done in only one routine.

NO DV NOW，I don't know how to add the DV  properly

PR #2052, opened 2023-06-22
Pmp fcov2

GregAC:
The large diff in `directed_testlist.yaml` appears to be because when I run the python script to generate it it's re-ordered some of the tests, but I don't think it's dropped any.


vogelpi (approved):
LGTM, thanks @GregAC !

The changes look sensible to me but I am no PMP expert. If also @hcallahan-lowrisc could have a look this would be great (can happen post merge to not block this PR and other work).

GregAC:
@marnovandermaas could also do a review, left him off the original list as he's on holiday but can do a post merge review when he returns.

PR #2062, opened 2023-07-18
Remove srecord and use objcopy instead

ctopal-rl:
Now that `objcopy` have support for `--verilog-data-width`, we don't need `srecord` anymore. To be fair, the current setup already supports pointing just to an `.elf` file but it's still nice to have the `.vmem`.


ctopal-rl:
CI fails because the version of riscv-compliance-suit that the CI uses also has srecord. I think that's not the case with its current version but moving to it would be bigger task. I'll put back the srecord requirement but edit it to say it's for the riscv-compliance test suit. Rest is up to you 😄

rswarbrick:
I just checked, and the `--verilog-data-width` option got added to binutils in version 2.33 (back in 2019). (Thanks for the heads-up! It's cool!)

I'm reasonably sure that this is ancient enough that we can just depend on it, but I couldn't figure out what base versions we use in our documentation. @GregAC, any concerns, or should we just merge?

jwnrt:
> I couldn't figure out what base versions we use in our documentation. @GregAC, any concerns, or should we just merge?

I believe [here](https://ibex-core.readthedocs.io/en/latest/03_reference/verification.html#prerequisites-environment-setup) we recommend either using the [pre-built lowRISC toolchain](https://github.com/lowRISC/lowrisc-toolchains) ([binutils v2.35](https://github.com/lowRISC/lowrisc-toolchains/blob/master/lowrisc-toolchain-gcc-rv32imcb.config#L61)) or building the [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain) from source ([binutils ~v2.40](https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=b2bc62b)).

Note that [this fix](https://sourceware.org/bugzilla/show_bug.cgi?id=25202) landed in binutils v2.40 for an issue with `--verilog-data-width` which I believe [caused us problems](https://github.com/lowRISC/opentitan/issues/1107#issuecomment-616448211) in the OpenTitan repository.

I'm not sure whether that issue applies here or whether 2.40 fixes it, but maybe we should update the version our toolchain ships.

PR #2074, opened 2023-08-15
Minor cleanups in get_fcov.py

rswarbrick:
Avoid importing some stuff that we're not using. Break a long line. No functional change.

### DIFF
--- a/dv/uvm/core_ibex/scripts/get_fcov.py
+++ b/dv/uvm/core_ibex/scripts/get_fcov.py
@@ -8,9 +8,8 @@
 import sys
 import argparse
 import pathlib3x as pathlib
-import shutil
 
-from metadata import RegressionMetadata, LockedMetadata
+from metadata import LockedMetadata
 from scripts_lib import run_one
 import riscvdv_interface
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
Hmm, I think @hcallahan-lowrisc remembers that these were actually needed for some library to work properly. We should probably fix things or at least add a documentation comment, but I'm not exactly sure what was going on! So I'm closing this for now: we can always tidy it up later.

hcallahan-lowrisc:
I just tried this now to see if I remembered it right. If you remove the import, you get the following traceback:
(I removed it in compile_tb.py, but the outcome would be the same)
```
  File "dv/uvm/core_ibex/scripts/compile_tb.py", line 110, in <module>
    sys.exit(_main())
  File "dv/uvm/core_ibex/scripts/compile_tb.py", line 51, in _main
    with LockedMetadata(args.dir_metadata, __file__) as md:
  File "dv/uvm/core_ibex/scripts/metadata.py", line 496, in __enter__
    self.md = pickle.load(self.handle)
AttributeError: Can't get attribute 'RegressionMetadata' on <module '__main__' from 'dv/uvm/core_ibex/scripts/compile_tb.py'>
```
This seems to be something strange to do with re-loading the pickled file. At a glance, I'm out of ideas!

PR #2075, opened 2023-08-15
Tweak ibex_tracer.sv to build with recent versions of Verilator

rswarbrick:
These currently fail, complaining about accesses to `decoded_str` and `data_accessed`. Verilator doesn't trace through far enough to see that these all come from the same `always_comb` process and can't race with each other.

This PR switches things so that we update those two variables with a non-blocking assignment on a clock negedge, then use them on the following posedge. It satisfies Verilator, and should be equivalent in any sensible system.

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -81,9 +81,6 @@ module ibex_tracer (
 
   import ibex_tracer_pkg::*;
 
-  int          file_handle;
-  string       file_name;
-
   int unsigned cycle;
   string       decoded_str;
   logic        insn_is_compressed;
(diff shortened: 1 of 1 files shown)


rswarbrick:
Darn. This ain't gonna work. The cunning trick was to use a non-blocking assignment in `always_ff` blocks to "thread the needle" and find something that was acceptable to several Verilator versions and Verible too.

But it turns out that VCS doesn't allow non-blocking assignments to dynamic types (here: string). That's quite reasonable of VCS, but it scuppers my grand plan! I'll try to come up with another approach, but I'll drop this PR for now.

PR #2076, opened 2023-08-16
Remove (empty) Verible waiver file

rswarbrick:
The recent versions of Verible that I've tried die when they are given an empty waiver file. The error message is "Fatal error: Broken waiver config handle". I can't see a way to add a comment ("// No waiver" or similar), so I think the best way to keep everything alive is to delete the empty file.

(This is motivated by trying to get a separate change that I was working on to pass CI. The first step was to reproduce the CI lint check that was failing for it!)

### DIFF
--- a/ibex_core.core
+++ b/ibex_core.core
@@ -42,10 +42,6 @@ filesets:
     files:
       - lint/verilator_waiver.vlt: {file_type: vlt}
 
-  files_lint_verible:
-    files:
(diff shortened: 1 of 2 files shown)


GregAC (approved):

dmcardle:
Thanks @rswarbrick! I just ran into this issue in lowRISC/opentitan#19426.

PR #2078, opened 2023-08-25
[doc] Fix background in block diagram

marnovandermaas:
The background was originally a gradient background on the right together with a solid background on the left. This caused some distortion on closer inspection. This commit changes it to have one background that is a gradient from left to right.

The background of the Icache block was a not-well-fitted path that was causing the background to seep through. This commit updates that background to more tightly align with the lines and letters.


marnovandermaas:
You can see that the original background is not correct when viewing it on a dark background:
image

hcallahan-lowrisc (approved):
LGTM!

Issue #2071, opened 2023-08-12
make[1]: *** [out/metadata/tb.compile.stamp] Error 2


cllll402:
I am a beginner in ibex, and when I followed the github steps to test something wrong, can you tell me what I missed or did wrong?
The error message is as follows:

autumn@localhost ~/ibex/dv/uvm/core_ibex$ make --keep-going IBEX_CONFIG=opentitan SIMULATOR=vcs ISS=spike ITERATIONS=1 SEED=1 TEST=riscv_arithmetic_basic_test WAVES=0 COV=0
Building RTL testbench
23:54:56 (snpslmd) OUT: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823296] 
23:55:19 (snpslmd) IN: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823296] 
23:55:19 (snpslmd) OUT: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823319] 
23:55:21 (snpslmd) IN: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823319] 
make[1]: *** [out/metadata/tb.compile.stamp] Error 2
Generating core configuration file
Building randomized test generator
23:55:24 (snpslmd) OUT: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823324] 
23:55:41 (snpslmd) IN: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823324] 
23:55:41 (snpslmd) OUT: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823341] 
23:55:43 (snpslmd) IN: "VCSCompiler_Net" autumn@localhost.localdomain  [snps_checkout_1691823341] 
Running randomized test generator to create assembly file /home/autumn/ibex/dv/uvm/core_ibex/out/run/tests/riscv_arithmetic_basic_test.1/test.S
23:55:44 (snpslmd) OUT: "VCSRuntime_Net" autumn@localhost.localdomain  [snps_checkout_1691823344] 
23:55:51 (snpslmd) IN: "VCSRuntime_Net" autumn@localhost.localdomain  [snps_checkout_1691823344] 
Compiling riscvdv test assembly to create binary at /home/autumn/ibex/dv/uvm/core_ibex/out/run/tests/riscv_arithmetic_basic_test.1/test.bin
[...]

PR #2082, opened 2023-08-30
Tweak ibex_cmd.py to fail more cleanly

rswarbrick:
This shouldn't change the behaviour when it works. On a failure, we now print out a bit more about what's going on.

When asked to do something impossible now, I think the output is a bit clearer. For example, if you try to run riscv_bitmanip_full_test with an OpenTitan configuration (which doesn't have full bitmanip), the warning message is now:

    WARNING:ibex_cmd:Rejecting test: riscv_bitmanip_full_test. It specifies rtl_params of ['ibex_pkg::RV32BFull'], which doesn't contain the expected 'ibex_pkg::RV32BOTEarlGrey'.

(The following stuff that appears is a bit messy, but at least the first line is now clearer!)

Triggered by issue #2071.

### DIFF
--- a/dv/uvm/core_ibex/scripts/ibex_cmd.py
+++ b/dv/uvm/core_ibex/scripts/ibex_cmd.py
@@ -150,13 +150,18 @@ def filter_tests_by_config(cfg: ibex_config.Config,
 
     for test in test_list:
         if "rtl_params" not in test:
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Great! Thanks

PR #2086, opened 2023-09-13
Misc fixes

GregAC:
A collection of small fixes/changes produced whilst I was running coremark to get the latest Ibex performance numbers across different configurations.

### DIFF
--- a/examples/sw/benchmarks/README.md
+++ b/examples/sw/benchmarks/README.md
@@ -11,8 +11,12 @@ All of these benchmarks run on Simple System. A verilator simulation suitable
 for running them can be built with:
 
 ```
-fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
+fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system `./util/ibex_config.py maxperf fusesoc_opts`
 ```
+
+This will build a simulation of Ibex in the 'maxperf' configuration.
+It is one of several pre-defined ibex configurations, others can be used.
+These are specified in the `ibex_configs.yaml` file.
 
 See examples/simple_system/README.md for full details.
 
--- a/examples/sw/simple_system/common/simple_system_common.h
+++ b/examples/sw/simple_system/common/simple_system_common.h
@@ -95,4 +95,21 @@ void timer_disable(void);
  */
 uint64_t get_elapsed_time(void);
 
+/**
+ * Enables/disables the instruction cache. This has no effect on Ibex
+ * configurations that do not have an instruction cache and in particular is
+ * safe to execute on those configurations.
+ *
+ * @param enable if non-zero enables, otherwise disables
+ */
+static inline void icache_enable(int enable) {
+  if (enable) {
+    // Set icache enable bit in CPUCTRLSTS
+    asm volatile("csrs 0x7c0, 1");
+  } else {
+    // Clear icache enable bit in CPUCTRLSTS
+    asm volatile("csrc 0x7c0, 1");
+  }
+}
+
 #endif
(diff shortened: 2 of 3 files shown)


examples/sw/simple_system/common/simple_system_common.h
```diff
@@ -95,4 +95,19 @@ void timer_disable(void);
...
+ * configurations that do not have an instruction cache and in particular is
+ * safe to execute on those configurations.
+ *
+ * @param enable if non-zero enables, otherwise disables
+ */
+static inline void icache_enable(int enable) {
+  if (enable) {
+    asm volatile("csrs 0x7c0, 1");
+  } else {
+    asm volatile("csrc 0x7c0, 1");
```
hcallahan-lowrisc, line 109:
```suggestion
    asm volatile("csrc 0x7c0, 0");
```

hcallahan-lowrisc (approved):

examples/sw/benchmarks/README.md
```diff
@@ -11,9 +11,12 @@ All of these benchmarks run on Simple System. A verilator simulation suitable
 for running them can be built with:
 
 ```
-fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
+fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system `./util/ibex_config.py [ibex_config] fusesoc_opts`
```
rswarbrick, line 14: Hmm, maybe it's cleaner to give an example config here? (That way a reader can copy-paste from the doc to their terminal!)

rswarbrick (approved):
Looks good to me, with a slight style question about a README.

marnovandermaas (commented):
Looks fine to me, except the change that Harry requested.

examples/sw/simple_system/common/simple_system_common.h
GregAC: It should be 1 in both cases, the first instruction is csr**s** (set a CSR bit) and the second is csr**c** (clear a CSR bit).

GregAC:
My apologies for the rather delayed response! Thanks for reviewing @hcallahan-lowrisc @rswarbrick @marnovandermaas would you mind taking another look?

rswarbrick:
I still have my question about the example config, but this looks reasonable otherwise.

examples/sw/benchmarks/README.md
GregAC: So I switched this to specify 'maxperf' as the config so it can be copy/pasted. I didn't want to include explicit flags as the idea is you should just use the ibex_config.py script to generate them for you based upon a config name.

Some may wish to specify manually but I'd consider that a more advanced usage and these instructions are targetted for people just trying to get Ibex doing a few things, such as reproduce benchmark results against supported configs.

GregAC: @rswarbrick does that address your question?

rswarbrick: Oh, I see. Yep, that sounds reasonable to me.

rswarbrick: Aargh! Sorry, I forgot! I'd suggest picking a config here, and writing something below about how to change it. It seems importantant (to me) that a reader can just copy/paste *something* into the terminal.

examples/sw/simple_system/common/simple_system_common.h
marnovandermaas: Very easy to miss this one letter difference. Would it make sense to put this in a comment?

marnovandermaas (approved):
Thanks, looks good to me.

examples/sw/benchmarks/README.md
GregAC: The example command line can be directly copy/pasted and used.

Though the wording below maybe didn't make this obvious, I've updated it to make it explicitly (hopefully!).

examples/sw/simple_system/common/simple_system_common.h
GregAC: Done

examples/sw/benchmarks/README.md
rswarbrick: Duh, sorry. I was looking at the thing from GitHub at the top of this discussion, which is what I originally commented on. What you've got looks good to me.

PR #2092, opened 2023-10-03
Dv fixes2

GregAC:
These are needed to fix the nightly regression


GregAC:
@mundaym I can kick off a regression once this is merged.

Issue #2101, opened 2023-11-17
Strange performance counters after following Simple System guide


noas1019:
Hi, I closely followed the guide in https://github.com/lowRISC/ibex/blob/master/examples/simple_system/README.md to try and simulate Ibex to run some tests, but I get strange results from the performance counters no matter the test used (everything but cycles and instructions retired is 0).  I've attached the results of hello_test and dit_test respectively, can someone point me to where the simulation is going wrong?
![hello_test](https://github.com/lowRISC/ibex/assets/55861223/e48b192e-be0e-4c35-942d-23935ebb297d)
![dit_test](https://github.com/lowRISC/ibex/assets/55861223/4c828e0e-6582-41be-b562-15138a68126e)

## My Environment

I'm running Ibex in Visual Studio Code using an Ubuntu wsl terminal.  All the pre-requisites in the simple_system guide were installed (along with RISC-V toolchain from https://github.com/lowRISC/lowrisc-toolchains/releases/tag/20230811-1), and the commands to build the simulation + software and run them are straight from the guide.

**EDA tool and version:**

N/A

**Operating system:**

Ubuntu 22.04.2 LTS

**Version of the Ibex source code:**

7aec1404f1d406ff9d3e68f9844387fc9f44c9ac

rswarbrick:
Hmm, this is a bug in our simple system guide. Kind of! The reason we're getting zeroes here is that the default configuration that you're picking has no performance counters. Unsurprisingly, this means that the counter numbers that are reported at the end of the simulation are all zero.

Fixes that we need to do:
1. Update the simple system documentation to use `ibex_config.py`. Something like this should work: `fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system $(util/ibex_config.py opentitan fusesoc_opts)`
2. Update the simple system wrapper C++ code not to report numbers that it doesn't know.

PR #2102, opened 2023-11-20
[dv] Fix performance counter printing in simple system

rswarbrick:
This comes in two commits, but the first is a silly typo fix. The commit message for the second is:

> [dv] Fix performance counter printing in simple system
> 
> This will avoid printing out a load of spurious zeros if the Ibex
> config doesn't enable the corresponding counter.

Partial fix for #2101.

### DIFF
--- a/dv/verilator/pcount/cpp/ibex_pcounts.cc
+++ b/dv/verilator/pcount/cpp/ibex_pcounts.cc
@@ -33,7 +33,7 @@ const std::vector<std::string> ibex_counter_names = {
     "Divide Wait"};
 
 std::string ibex_pcount_string(bool csv) {
-  char seperator = csv ? ',' : ':';
+  char separator = csv ? ',' : ':';
   std::string::size_type longest_name_length;
 
   if (!csv) {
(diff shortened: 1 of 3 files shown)


dv/verilator/pcount/cpp/ibex_pcounts.cc
```diff
@@ -32,24 +33,46 @@ const std::vector<std::string> ibex_counter_names = {
     "Multiply Wait",
     "Divide Wait"};
 
+static bool has_hpm_counter(int index) {
+  // The "cycles" and "instructions retired" counters are special and always
+  // exist.
+  if (index == 0 || index == 2)
+    return true;
+
+  // The "NONE" counter is fake. That never exists.
```
nbdd0121, line 42:
```suggestion
  // The "NONE" counter is fake is a placeholder for the "MTIME" CSR. That never exists.
```

rswarbrick: Ah, that makes sense. I couldn't figure this out: how can we tell that index 1 corresponds to the MTIME CSR?

nbdd0121: It's in the order of MCYCLE, MTIME, MINSTRET, MHPMCOUNTER3, ...

rswarbrick: I'm clearly being slow here. Where does that list of CSRs come from? Is it a standard RISC-V thing? My web searches find lots of Ibex...

rswarbrick: Looking at e.g. the RISC-V privileged spec, I see MCYCLE, (a gap here), MINSTRET, MHPMCOUNTER3, ... in table 2.5.

Is there a spec that explains that the gap is MTIME?

nbdd0121: MTIME was once an actual CSR but later removed from the spec since people want it to be MMIO instead of actual CSR. But you can guess that this counter is reserved for MTIME by looking at the list of user-mode CSRs, which is CYCLE, TIME, INSTRET, HPMCOUNTER3, ...

rswarbrick: Ahah! Thanks for the explanation. I'll tweak this PR to match.

nbdd0121 (approved):

PR #2103, opened 2023-11-20
[vendor] Update lowrisc_ip from OpenTitan

rswarbrick:
Pull the latest version of lowrisc_ip from the current head of the OpenTitan repository.

One handy result (the reason for re-vendoring!) is that prim_prince.sv has been rewritten to avoid triggering errors with recent versions of Verilator. See https://github.com/lowRISC/opentitan/pull/19785 for an explanation.

### DIFF
--- a/dv/uvm/common_project_cfg.hjson
+++ b/dv/uvm/common_project_cfg.hjson
@@ -5,6 +5,7 @@
   project:          ibex
 
   // These keys are expected by dvsim.py, so we have to set them to something.
+  book:              bogus.book.domain
   doc_server:        bogus.doc.server
   results_server:    bogus.results.server
   results_html_name: report.html
--- a/vendor/lowrisc_ip.lock.hjson
+++ b/vendor/lowrisc_ip.lock.hjson
@@ -9,6 +9,6 @@
   upstream:
   {
     url: https://github.com/lowRISC/opentitan
-    rev: 0deeaa99e5760ee4f5c0a08e5fc1670509d22744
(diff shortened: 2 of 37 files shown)


dv/uvm/common_project_cfg.hjson
```diff
@@ -5,6 +5,7 @@
   project:          ibex
 
   // These keys are expected by dvsim.py, so we have to set them to something.
+  book:              bogus.book.domain
```
nbdd0121, line 8: I tried to revendor locally and it doesn't give me this line

rswarbrick: Oh! Now I remember! I had to do this manually. The problem is that the latest version of dvsim (which we're pulling from OpenTitan) now depends on that key in a file that it imports (and smushes into Python datastructures in a rather horrible way).

The dependency comes from [OpenTitan commit ff2314b16b919aff2d1149947ed6e2432f2a27b7](https://github.com/lowRISC/opentitan/commit/ff2314b16b919aff2d1149947ed6e2432f2a27b7) from back in July.

Can you see a cleaner way to do this update?

nbdd0121: I think the solution is fine, but you'd want to mention this in the commit message.

rswarbrick: Silly me: there's a much easier way! I've just split this into two commits. The first adds this line to `common_project_cfg.hjson` and the second is the vendoring commit. Hopefully this will make everything a bit more obvious.

nbdd0121 (approved):

PR #2104, opened 2023-11-20
[rtl] Avoid name collision in ibex_pmp.sv

rswarbrick:
Recent versions of Verilator complain about the code that was there because the csr_pmp_cfg argument clashes with a name in ibex_core.sv.

What's more, they mean different things! In ibex_core.sv, it was the PMP configuration for the entire core. In the functions, it's the PMP configuration for a single region. This patch adds a "region_" prefix to the names, which fixes both the Verilator warning and my confusion!


marnovandermaas (approved):
Thanks, I'm happy with this change.

PR #2105, opened 2023-11-20
[doc] Describe counters more explicitly in simple_system README

rswarbrick:
This should avoid some of the confusion that caused issue #2101.

### DIFF
--- a/examples/simple_system/README.md
+++ b/examples/simple_system/README.md
@@ -27,12 +27,19 @@ run stand-alone binaries. It contains:
 
 ## Building Simulation
 
-The Simple System simulator binary can be built via FuseSoC. From the Ibex
-repository root run:
+The Simple System simulator binary can be built via FuseSoC. This can be built
+with different configurations of Ibex, specified by parameters. To build the
+"small" configuration, run the following command from the Ibex repository root.
+
 
 ```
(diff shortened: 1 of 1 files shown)


examples/simple_system/README.md
```diff
@@ -27,13 +27,19 @@ run stand-alone binaries. It contains:
...
+"small" configuration, run the following command from the Ibex repository root.
+
 
 ```
-fusesoc --cores-root=. run --target=sim --setup --build lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
+fusesoc --cores-root=. run --target=sim --setup --build \
+        lowrisc:ibex:ibex_simple_system $(util/ibex_config.py small fusesoc_opts)
 ```
 
+For a larger design, which implements more performance counters, replace `small`
```
nbdd0121, line 40: It feels weird to specifically mention performance counter since there are way more things (and more important things) that are changed by switching configuration. Perhaps just say "For a larger design, replace ..."?

rswarbrick: Good point: this was rather unclear. What I meant was "In order to get a design with more performance counters (which will be larger) ...". I've turned it around, and the result should hopefully make a bit more sense.

nbdd0121 (approved):

PR #2109, opened 2023-12-07
Add GitHub actions version of the CI

nbdd0121:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+          fusesoc --version
+          verilator --version
+          riscv32-unknown-elf-gcc --version
+          verible-verilog-lint --version
+
+      # Verible format is experimental so only run on default config for now,
+      # will eventually become part of the per-config CI
+      - name: Format all source code with Verible format (experimental)
+        run: |
+          set +e
```
jwnrt, line 57: Does GitHub's runner set `set -e` where Azure's doesn't?

jwnrt (approved):
Thanks @nbdd0121, this LGTM

I checked that the action matches `ibex-rtl-ci-steps.yml` and that the workflow matches `azure-pipelines.yml`. The scripts all look the same, except obviously with the Azure-specific variables ported as well as the error message format converted.

.github/workflows/ci.yml
nbdd0121: Yes, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

GregAC (approved):
Looks good thanks for doing this @nbdd0121 

We don't need to do this for this PR but I think it'd be reasonable to move to Ubuntu 22 for the Ibex github actions. We'll need to build a verilator binary for it but that's straightforward. Everything else should just work.

We should also look at moving to a newer verilator version (again will require building new binaries)

nbdd0121:
Upgrading Ubuntu and verilator version sounds great. I'll add that to my TODO list.

jwnrt:
Discussed with @nbdd0121, I would suggest disabling the Azure pipeline in the web UI and leaving the configuration in this repo for at least one real run so that we can re-enable it if there's an issue.

jwnrt:
I have disabled the pipeline.

I notice that the GitHub Actions pipeline didn't run during the merge queue, do we need to add it to some list for that, or will it run on the next merge @nbdd0121?

Also, if it runs in the merge queue, should we disable the `on: push: ["master"]` run which I think will be redundant?

PR #2117, opened 2023-12-22
[rtl] Fix FI vulnerability in RF

nasahlpa:
As described in [#20715](https://github.com/lowRISC/opentitan/issues/20715), a single fault-induced bit-flip inside the register file could change which of the register file value is provided to Ibex.

This PR fixes this issue by (i) encoding raddr_a/b to one-hot encoded signals, (ii) checking these signals for faults, and (iii) using an one-hot encoded MUX to select which register file value is forwarded to rdata_a/b.

Area increases by ~1% (Yosys + Nangate45 synthesis).

I conducted a formal fault injection verification at the Yosys netlist to ensure that the issue really is fixed.


rtl/ibex_register_file_ff.sv
```diff
@@ -39,7 +40,7 @@ module ibex_register_file_ff #(
   input  logic [DataWidth-1:0] wdata_a_i,
   input  logic                 we_a_i,
 
-  // This indicates whether spurious WE are detected.
+  // This indicates whether spurious WE or one-hot encoded raddr are detected.
```
vogelpi, line 43: Nit: one-hot encoded -> non-one-hot encoded

rtl/ibex_register_file_ff.sv
```diff
@@ -130,8 +133,101 @@ module ibex_register_file_ff #(
...
+      .AddrWidth(ADDR_WIDTH),
+      .OneHotWidth(NUM_WORDS),
+      .AddrCheck(1),
+      .EnableCheck(0)
+    ) u_prim_onehot_check_raddr_a (
+      .clk_i,
+      .rst_ni,
+      .oh_i   (raddr_onehot_a_buf),
+      .addr_i (raddr_a_i),
+      .en_i   (1'b0),
```
vogelpi, line 183: Maybe want to add a comment that the `en_i` input is unused as the `EnableCheck` parameter is disabled?

rtl/ibex_register_file_ff.sv
```diff
@@ -130,8 +133,101 @@ module ibex_register_file_ff #(
     assign rf_reg[0] = WordZeroVal;
   end
 
-  assign rdata_a_o = rf_reg[raddr_a_i];
-  assign rdata_b_o = rf_reg[raddr_b_i];
+  if (RdataMuxCheck) begin : gen_rdata_mux_check
+    // Encode raddr_a/b into one-hot encoded signals.
```
vogelpi, line 137: IIUC, it is sufficient to do the one-hot encoding in here (not covered by the lockstep) because the same gates driving the `raddr_a/b_i` signals used here also drive the wires ultimately going into the lockstep comparison. I.e., if someone managed to fault `raddr_a_i` here, this would still be caught by the lockstep countermeasure, right?

If yes, this would be perfect because it means we don't have to touch the lockstep interfacing and checking logic.

rtl/ibex_register_file_ff.sv
```diff
@@ -130,8 +133,101 @@ module ibex_register_file_ff #(
...
+    );
+
+    prim_buf #(
+      .Width(NUM_WORDS)
+    ) u_prim_buf_raddr_b (
+      .in_i (raddr_onehot_b),
+      .out_o(raddr_onehot_b_buf)
+    );
+
+    // Check the one-hot encoded signals for glitches.
```
vogelpi, line 172: Please also add a `// SEC_CM: DATA_REG_SW.GLITCH_DETECT` comment above this line. We use these labels inside OpenTitan to reference / track countermeasures. For additional information, please refer to the comportability spec: https://opentitan.org/book/doc/contributing/hw/comportability/index.html?highlight=comportability#security-countermeasures

nasahlpa: done.

rtl/ibex_register_file_ff.sv
nasahlpa: done.

rtl/ibex_register_file_ff.sv
nasahlpa: Yes, exactly.

In `ibex_top.sv` the `rf_raddr_a/b` signals are generated by the Ibex core:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/rtl/ibex_top.sv#L336-L337
forwarded to the RF:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/rtl/ibex_top.sv#L433-L435
and buffered for the lockstep:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/rtl/ibex_top.sv#L862-L863

Inside the lockstep, again the `rf_raddr_a/b` signals are generated by the shadow Ibex core:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/rtl/ibex_lockstep.sv#L374-L375

And finally the comparison happens. Here, the buffered `rf_raddr_a/b` signals from the main core are then compared to the signals generated by the shadow core.

If a bit-flip inside the RF manipulates the `rf_raddr_a/b` signals, these faulty signals will get buffered for the lockstep:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/rtl/ibex_top.sv#L862-L863

And the fault is detected when comparing the main vs shadow `rf_raddr_a/b` signals.

rtl/ibex_register_file_ff.sv
nasahlpa: Thanks for the comment. I actually had to set `en_i = 1'b1` as `EnableCheck `needs to be on when having `AddrCheck=1`.
The reason for this is the following condition:
https://github.com/lowRISC/ibex/blob/38da151a251de5b13afeedf9c5947707f87c5992/vendor/lowrisc_ip/ip/prim/rtl/prim_onehot_check.sv#L54
I've added comments.

nasahlpa:
Thanks for your valuable feedback, @vogelpi! I've addressed all of your comments and CI is also passing.

doc/03_reference/security.rst
```diff
@@ -90,6 +90,13 @@ When Ibex is configured with the SecureIbex parameter, the write enable signal i
 This can be useful to detect fault injection attacks.
 No attempt is made to correct detected errors, but an internal major alert is signaled for the system to take action.
 
+Register file read addresses glitch detection
+-------------------------------------------
+
+When Ibex is configured with the SecureIbex parameter, the read addresses provided to the register file are converted to one-hot encoded signals and these signals are checked.
```
GregAC, line 96: may want to expand slightly on what the checks are.

rtl/ibex_register_file_ff.sv
```diff
@@ -130,8 +133,110 @@ module ibex_register_file_ff #(
...
+    // SEC_CM: DATA_REG_SW.GLITCH_DETECT
+    // Check the one-hot encoded signals for glitches.
+    prim_onehot_check #(
+      .AddrWidth(ADDR_WIDTH),
+      .OneHotWidth(NUM_WORDS),
+      .AddrCheck(1),
+      // When AddrCheck=1 also EnableCheck needs to be 1.
+      .EnableCheck(1),
+      // Avoid that raddr=0 triggers an error.
+      .StrictCheck(0)
```
GregAC, line 181: Under what scenario does this happen? When the incoming address is 0 we'll get a one-hot encoded signal with the bottom bit set and the onehot encoders are always enabled.

GregAC (approved):
LGTM to me, nice catch @nasahlpa couple of minor comments then should be good to go.

rtl/ibex_register_file_ff.sv
nasahlpa: Ah yeah, you are right! I've removed `StrictCheck=0`.

nasahlpa:
Full DV regression run is now completed:
`89.82% PASS 1271 PASSED, 144 FAILED`

msfschaffner (approved):
Thanks @nasahlpa, this is a nice security improvement!

nasahlpa:
Thanks to all of you for the review! Merging it now.

rtl/ibex_register_file_ff.sv
vogelpi: Thanks for adding the comments.

rtl/ibex_register_file_ff.sv
vogelpi: Thanks for doing the analysis and writing it up here. LGTM!

vogelpi (approved):
LGTM, thanks for addressing the remaining comments :+1:

PR #2118, opened 2023-12-22
[ci] introduce GitHub actions based private CI

nbdd0121:
The private CI works by using workflow dispatch to trigger a CI on another repository.

Pending status checks are created before calling the dispatch to indicate their queued status before they are picked up in private CI, and also serves to block merge group from success.

### DIFF
--- a/.github/workflows/private-ci.yml
+++ b/.github/workflows/private-ci.yml
@@ -1,0 +1,45 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Ibex Private CI
+
+on:
+  push:
+    tags:
+      - "*"
+  merge_group:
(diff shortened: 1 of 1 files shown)


jwnrt:
Given this only runs on `pull_request_target`, would you mind triggering a run on your fork or similar to show it working?

jwnrt (approved):
This LGTM, but I worry slightly about the names of checks here becoming out of sync with the private repo. I think it would be difficult to factor those names out, though.

Doesn't the private CI create pending checks when it starts? Are these pending checks meant to fill the gap between those starting?

Could we instead have a job in this workflow that waits for the private CI to have started and we can block the merge queue on that?

nbdd0121:
The pending status will only be created when jobs are picked up, so if we have a busy CI pool then we'd be waiting for quite a while, wasting compute.

jwnrt:
The new approach looks a lot cleaner, thanks a lot for working on that

PR #2121, opened 2024-01-03
[doc] Update cosim version

nasahlpa:
Not sure if this is helpful, but:

Currently, the verification documentation recommends to using ibex-cosim-v0.3. However, this version of spike causes some compilation issues for `spike_cosim `on my side:
`class "processor_t" has no member "set_mhpm_counter_num`
when calling `make` in `dv/uvm/core_ibex/`.
This issue occurs when using the latest pre-build lowrisc toolchain "20231205-1" or "20230427-1".

When using ibex-cosim-v0.5 DV works.

### DIFF
--- a/doc/03_reference/verification.rst
+++ b/doc/03_reference/verification.rst
@@ -111,8 +111,7 @@ In order to run the co-simulation flow, you'll need:
   + Some custom CSRs
   + Custom NMI behavior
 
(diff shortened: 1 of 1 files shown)


GregAC:
Ultimately I think we should drop the version tagging here, perhaps instead just have a file that refers to a commit hash of the cosim that's compatible with the Ibex commit you got the hash from. However this is a good fix for now. Thanks @nasahlpa

GregAC (approved):

PR #2123, opened 2024-01-08
Add NOTICE file

mundaym:
Add a NOTICE file containing some information about the project.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,10 @@
+The Ibex Project
+Copyright 2024 lowRISC contributors.
+
+This product includes hardware and/or software developed as part of the
+Ibex(R) (https://github.com/lowRISC/ibex) and OpenTitan(R) projects.
+
+Ibex was originally developed by the PULP team at ETH Zurich and University of
+Bologna under the name zero-riscy. Ibex verification, performance enhancement
+and security hardening have been supported by the OpenTitan project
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):

NOTICE
```diff
@@ -0,0 +1,10 @@
+The Ibex Project
+Copyright 2024 lowRISC contributors.
+
+This product includes hardware and/or software developed as part of the
+Ibex(R) (https://github.com/lowRISC/ibex) and OpenTitan(R) projects.
```
nbdd0121, line 5: Do we want (R) or ®?

NOTICE
```diff
@@ -0,0 +1,10 @@
...
+
+Ibex was originally developed by the PULP team at ETH Zurich and University of
+Bologna under the name zero-riscy. Ibex verification, performance enhancement
+and security hardening have been supported by the OpenTitan project
```
luismarques, line 9: regarding "performance enhancement[s]", I guess we also added features? (and might add more in the future)

NOTICE
mundaym: Unfortunately ® is rejected by our CI system at the moment as it is not ASCII.

NOTICE
mundaym: Addressed in the OT repo review.

PR #2129, opened 2024-01-17
[rtl] Harden lockstep enable against FI

nasahlpa:
Currently, the dual-core lockstep FI mitigation is enabled/disabled using a single bit. For transient bit-flips, this is not problematic, as one bit-flip into this signal and one bit into the Ibex is required to threaten the security of the system.

However, a permanent stuck-at-0 fault could disable the lockstep completely by targeting this signal. Then, only a single, additional fault (transient or permanent) is required.

This PR enhances the FI resilience of the Ibex lockstep by encoding this single bit into a `ibex_mubi_t` signal, i.e., a 4-bit multi-bit signal. Note that the MUXes:
https://github.com/lowRISC/ibex/blob/2fbf04636b329bdb643501ab0ef3ee7d6b3303a9/rtl/ibex_lockstep.sv#L138
and
https://github.com/lowRISC/ibex/blob/2fbf04636b329bdb643501ab0ef3ee7d6b3303a9/rtl/ibex_lockstep.sv#L468
are not redundant. However, as the goal is to protect the previously 1-bit select signal against stuck-at-0 faults, increasing the bit-width to 4-bit significantly increases security against this threat.

I opted for going with a `ibex_mubi_t` instead of a `mubi4/8` as it seems that only `ibex_mubi_t` is frequently used in the core.

Tracked in #[20779](https://github.com/lowRISC/opentitan/issues/20779).

Full DV regression looks good.

### DIFF
--- a/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
+++ b/dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
@@ -331,6 +331,13 @@ module core_ibex_tb_top;
     run_test();
   end
 
+  // Manually set unused_assert_connected = 1 to disable the AssertConnected_A assertion for
+  // prim_count in case lockstep (set by SecureIbex) is enabled. If not disabled, DV fails.
+  if (SecureIbex) begin : gen_disable_count_check
+    assign dut.u_ibex_top.gen_lockstep.u_ibex_lockstep.u_rst_shadow_cnt.
+          unused_assert_connected = 1;
+  end
+
   // Disable the assertion for onhot check in case WrenCheck (set by SecureIbex) is enabled.
   if (SecureIbex) begin : gen_disable_onehot_check
     assign dut.u_ibex_top.gen_regfile_ff.register_file_i.gen_wren_check.u_prim_onehot_check.
--- a/rtl/ibex_lockstep.sv
+++ b/rtl/ibex_lockstep.sv
@@ -120,33 +120,50 @@ module ibex_lockstep import ibex_pkg::*; #(
   // - The reset of the shadow core is synchronously released.
   // The comparison is started in the following clock cycle.
 
-  logic [LockstepOffsetW-1:0] rst_shadow_cnt_d, rst_shadow_cnt_q, rst_shadow_cnt_incr;
-  // Internally generated resets cause IMPERFECTSCH warnings
-  /* verilator lint_off IMPERFECTSCH */
-  logic                       rst_shadow_set_d, rst_shadow_set_q;
-  logic                       rst_shadow_n, enable_cmp_q;
-  /* verilator lint_on IMPERFECTSCH */
+  logic [LockstepOffsetW-1:0] rst_shadow_cnt;
+  logic                       rst_shadow_cnt_err;
+  ibex_mubi_t                 rst_shadow_set_d, rst_shadow_set_q;
+  logic                       rst_shadow_n, rst_shadow_set_single_bit;
+  ibex_mubi_t                 enable_cmp_d, enable_cmp_q;
 
-  assign rst_shadow_cnt_incr = rst_shadow_cnt_q + 1'b1;
+  // This counter primitive starts counting to LockstepOffset after a system
+  // reset. The counter value saturates at LockstepOffset.
+  prim_count #(
+    .Width      (LockstepOffsetW        ),
+    .ResetValue (LockstepOffsetW'(1'b0) )
+  ) u_rst_shadow_cnt (
+    .clk_i      (clk_i                  ),
+    .rst_ni     (rst_ni                 ),
+    .clr_i      (1'b0                   ),
+    .set_i      (1'b0                   ),
+    .set_cnt_i  ('0                     ),
+    .incr_en_i  (1'b1                   ),
+    .decr_en_i  (1'b0                   ),
+    .step_i     (LockstepOffsetW'(1'b1) ),
+    .cnt_o      (rst_shadow_cnt         ),
+    .cnt_next_o (                       ),
+    .err_o      (rst_shadow_cnt_err     )
+  );
 
-  assign rst_shadow_set_d = (rst_shadow_cnt_q == LockstepOffsetW'(LockstepOffset - 1));
-  assign rst_shadow_cnt_d = rst_shadow_set_d ? rst_shadow_cnt_q : rst_shadow_cnt_incr;
+  // When the LockstepOffset counter value is reached, activate the lockstep
+  // comparison. We do not explicitly check whether rst_shadow_set_q forms a valid
+  // multibit signal as this value is implicitly checked by the enable_cmp
+  // comparison below.
+  assign rst_shadow_set_d =
+    (rst_shadow_cnt >= LockstepOffsetW'(LockstepOffset - 1)) ? IbexMuBiOn : IbexMuBiOff;
 
-  always_ff @(posedge clk_i or negedge rst_ni) begin
-    if (!rst_ni) begin
-      rst_shadow_cnt_q <= '0;
-      enable_cmp_q     <= '0;
-    end else begin
-      rst_shadow_cnt_q <= rst_shadow_cnt_d;
-      enable_cmp_q     <= rst_shadow_set_q;
-    end
-  end
+  // Enable lockstep comparison.
+  assign enable_cmp_d = rst_shadow_set_q;
+
+  // This assignment is needed in order to avoid "Warning-IMPERFECTSCH" messages.
+  // TODO: Remove when updating Verilator #2134.
+  assign rst_shadow_set_single_bit = rst_shadow_set_q[0];
 
   // The primitives below are used to place size-only constraints in order to prevent
   // synthesis optimizations and preserve anchor points for constraining backend tools.
   prim_flop #(
-    .Width(1),
-    .ResetValue(1'b0)
+    .Width(IbexMuBiWidth),
+    .ResetValue(IbexMuBiOff)
   ) u_prim_rst_shadow_set_flop (
     .clk_i (clk_i),
     .rst_ni(rst_ni),
@@ -154,10 +171,20 @@ module ibex_lockstep import ibex_pkg::*; #(
     .q_o   (rst_shadow_set_q)
   );
 
+  prim_flop #(
+    .Width(IbexMuBiWidth),
+    .ResetValue(IbexMuBiOff)
+  ) u_prim_enable_cmp_flop (
+    .clk_i (clk_i),
+    .rst_ni(rst_ni),
+    .d_i   (enable_cmp_d),
+    .q_o   (enable_cmp_q)
+  );
+
   prim_clock_mux2 #(
     .NoFpgaBufG(1'b1)
   ) u_prim_rst_shadow_n_mux2 (
-    .clk0_i(rst_shadow_set_q),
+    .clk0_i(rst_shadow_set_single_bit),
     .clk1_i(scan_rst_ni),
     .sel_i (test_en_i),
     .clk_o (rst_shadow_n)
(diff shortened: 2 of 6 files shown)


rtl/ibex_lockstep.sv
```diff
@@ -458,7 +464,8 @@ module ibex_lockstep import ibex_pkg::*; #(
 
   logic outputs_mismatch;
 
-  assign outputs_mismatch       = enable_cmp_q & (shadow_outputs_q != core_outputs_q[0]);
+  assign outputs_mismatch =
+    (enable_cmp_q == IbexMuBiOn) & (shadow_outputs_q != core_outputs_q[0]);
```
andreaskurth, line 468: Wouldn't single-bit FI on `enable_cmp_q` still be sufficient to evade this comparison? Put differently, shouldn't we use `enable_cmp_q != IbexMuBiOff` as condition here?

rtl/ibex_lockstep.sv
```diff
@@ -157,7 +163,7 @@ module ibex_lockstep import ibex_pkg::*; #(
   prim_clock_mux2 #(
     .NoFpgaBufG(1'b1)
   ) u_prim_rst_shadow_n_mux2 (
-    .clk0_i(rst_shadow_set_q),
+    .clk0_i(rst_shadow_set_single_bit),
```
nasahlpa, line 166: This is needed, because when doing this:
```suggestion
    .clk0_i(rst_shadow_set_q[0]),
```
I get the `IMPERFECTSCH` linting error.

rtl/ibex_lockstep.sv
```diff
@@ -123,19 +123,25 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  assign rst_shadow_set_single_bit = rst_shadow_set_q[0];
 
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
       rst_shadow_cnt_q <= '0;
-      enable_cmp_q     <= '0;
+      enable_cmp_q     <= IbexMuBiOff;
     end else begin
       rst_shadow_cnt_q <= rst_shadow_cnt_d;
       enable_cmp_q     <= rst_shadow_set_q;
```
andreaskurth, line 147: `enable_cmp_q` may also benefit from being implemented with a `prim_flop` to avoid optimization, right?

rtl/ibex_lockstep.sv
```diff
@@ -123,19 +123,25 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  logic                       rst_shadow_n;
+  ibex_mubi_t                 enable_cmp_q;
   /* verilator lint_on IMPERFECTSCH */
 
   assign rst_shadow_cnt_incr = rst_shadow_cnt_q + 1'b1;
 
-  assign rst_shadow_set_d = (rst_shadow_cnt_q == LockstepOffsetW'(LockstepOffset - 1));
-  assign rst_shadow_cnt_d = rst_shadow_set_d ? rst_shadow_cnt_q : rst_shadow_cnt_incr;
+  assign rst_shadow_set_d =
+    (rst_shadow_cnt_q == LockstepOffsetW'(LockstepOffset - 1)) ? IbexMuBiOn : IbexMuBiOff;
```
andreaskurth, line 136: Here a MuBi is computed from a strict bit-by-bit equality, so wouldn't a single-bit FI on `rst_shadow_cnt_q` be sufficient to control the MuBi?

I think this could be mitigated by using a hardened counter for `rst_shadow_cnt`

andreaskurth (commented):
Thanks for catching this and working on a fix, @nasahlpa! I agree we should make changes to harden lockstep enable against FI, but I'm not sure the PR as-is suffices; please see my comments

rtl/ibex_lockstep.sv
nasahlpa: Yes, you are right. I've changed this.

rtl/ibex_lockstep.sv
nasahlpa: I've implemented `enable_cmp` with a `prim_flop` to avoid optimization. Thanks!

rtl/ibex_lockstep.sv
rswarbrick: Does the IMPERFECTSCH warning cause CI to fail? If not, I'd be inclined to use the `rst_shadow_set_q[0]` form. I notice that the warning was removed from Verilator with version 5 (https://verilator.org/guide/latest/warnings.html).

nasahlpa: Yes, CI fails because of this IMPERFECTSCH warning.

I would also prefer using `rst_shadow_set_q[0]`, but I couldn't find a way of getting rid of this warning in CI.

rtl/ibex_lockstep.sv
nasahlpa: I also recognized that the counter here could be a weak point. However, does using `prim_count` actually mitigate this problem? I was thinking of replacing the counter with the following:
```Verilog
  prim_count #(
    .Width(LockstepOffsetW)
  ) u_rst_shadow_cnt (
    .clk_i(clk_i),
    .rst_ni(rst_ni),
    .clr_i('0),
    .set_i('0), Done
    .set_cnt_i('0), Done
    .incr_en_i(rst_shadow_set_d == IbexMuBiOn),
    .decr_en_i(1'b0), Done
    .step_i(LockstepOffsetW'(1'b1)), Done
    .cnt_o(rst_shadow_cnt_q),
    .cnt_next_o(), Done
    .err_o(cnt_err)
  );
```
However, if a stuck-at-0 fault inside the `prim_count` permanently sets the `incr_en` signal to 0, we would have the same issue again?

rtl/ibex_lockstep.sv
rswarbrick: A solution: upgrade our Verilator dependency :-)

I *think* we can probably just do that. Would you mind filing an issue to remind us? ("Upgrade Verilator, then remove hack I had to do in ibex_lockstep.sv" or something similar)

nasahlpa: Good idea, tracked in issue #2131.

rtl/ibex_lockstep.sv
andreaskurth: `prim_count` saturates and doesn't overflow, so how about setting `incr_en_i` as well as `step_i` to 1, and then use `>=` for comparing `cnt_o`?

rtl/ibex_pkg.sv
```diff
@@ -655,7 +655,9 @@ package ibex_pkg;
 
   // Mult-bit signal used for security hardening. For non-secure implementation all bits other than
   // the bottom bit are ignored.
-  typedef logic [3:0] ibex_mubi_t;
+  parameter int IbexMuBiWidth = 4;
+
```
vogelpi, line 659: Nit: maybe remove the empty line here.

rtl/ibex_lockstep.sv
vogelpi: I agree with @andreaskurth .

I am not sure if we should also check whether `rst_shadow_set_q` contains a valid value and raise an alert if not. We do this in other places as well. However, if `rst_shadow_set_q` was invalid, also `enable_cmp_q` would become invalid eventually and the comparison was started: whenever it's not off, the comparison is enabled. So it's probably fine. We may want to document this though to prevent someone from changing this accidentally.

vogelpi (commented):
Thanks for the PR @nasahlpa , this looks good. Adding a prim_count sounds like a good idea to me.

rtl/ibex_lockstep.sv
```diff
@@ -120,44 +120,74 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+  // This signal is needed in order to avoid "Warning-IMPERFECTSCH" messages.
+  logic                       rst_shadow_set_single_bit;
+  logic                       rst_shadow_n;
+  ibex_mubi_t                 enable_cmp_d, enable_cmp_q;
   /* verilator lint_on IMPERFECTSCH */
 
-  assign rst_shadow_cnt_incr = rst_shadow_cnt_q + 1'b1;
+  // This counter primitive starts counting to LockstepOffset after a system
+  // reset. The counter value saturates at LockstepOffset.
+  prim_count #(
```
nasahlpa, line 136: @rswarbrick I again get strange linting errors. When using the `prim_count` module (which is not yet used by Ibex), I get a Verilator linting error inside the `prim_count`:
```
ERROR: %Warning-WIDTHCONCAT: ../src/lowrisc_prim_count_0/rtl/prim_count.sv:53:85: Unsized numbers/parameters not allowed in concatenations.
                                                                         : ... In instance ibex_top_tracing.u_ibex_top.gen_lockstep.u_ibex_lockstep.u_rst_shadow_cnt
   53 |   localparam logic [NumCnt-1:0][Width-1:0] ResetValues = {{Width{1'b1}} - ResetValue,  
      |                                                                                     ^
                      ... Use "/* verilator lint_off WIDTHCONCAT */" and lint_on around source to disable this message.
```

rswarbrick: Oof, that's rather weird. Does the linting error come out with recent versions of Verilator? If not, I guess I still think that jumping to more recent tooling is the right approach! [*]

Looking at the code, I don't think the tool warning makes any sense. The only concatenation that's happening is the replication that makes `{Width{1'b1}}`. Maybe the complaint is that the number of copies to use doesn't itself have a size? It's pretty clear this is a tooling bug!

[*] As a quick sanity test, I tried it locally! Once I comment out the instantiation of `prim_flop`, I can run `verilator --lint-only prim_count.sv` with no problems. Local version: `Verilator 5.020 2024-01-01 rev (Debian 5.020-1)`.

nasahlpa: Thanks for testing this! I just tested it with the Verilator version that is also used in OpenTitan (4.210). However, linting still fails. This is a bit strange as then linting in the OpenTitan repo should also fail? Or are there other tools involved?

rtl/ibex_lockstep.sv
```diff
@@ -120,44 +120,75 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+    .Width      (LockstepOffsetW        ),
+    .ResetValue (LockstepOffsetW'(1'b0) )
```
nasahlpa, line 138: @rswarbrick explicitly defining the `ResetValue` helped to get rid of the linting error mentioned above.

nasahlpa:
> Thanks for the PR @nasahlpa , this looks good. Adding a prim_count sounds like a good idea to me.

Thanks for all the helpful suggestions! I've updated the code and this PR is ready for a second round of reviews :-)

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -331,6 +331,12 @@ module core_ibex_tb_top;
     run_test();
   end
 
+  // Disable the assertion for prim_count in case lockstep (set by SecureIbex) is enabled.
```
rswarbrick, line 334: Hmm, this "assert connected" flag is an ugly feature of the design of the prim. Yuck!

But can we get it wired up without hacking things by using something like the `ASSERT_ERROR_TRIGGER_ALERT` macro? I suspect not, because that macro seems to have an opinion about how alerts should be signaled (true in OpenTitan; not true for Ibex, I think).

If that doesn't work, could we add a bit of explanation in this comment about what we're working around?

rtl/ibex_lockstep.sv
```diff
@@ -120,44 +120,71 @@ module ibex_lockstep import ibex_pkg::*; #(
...
+    .decr_en_i  (1'b0                   ),
+    .step_i     (LockstepOffsetW'(1'b1) ),
+    .cnt_o      (rst_shadow_cnt         ),
+    .cnt_next_o (                       ),
+    .err_o      (rst_shadow_cnt_err     )
+  );
 
-  assign rst_shadow_cnt_incr = rst_shadow_cnt_q + 1'b1;
+  // When the LockstepOffset counter value is reached, activate the lockstep
+  // comparison. We do not explicitly check whether rst_shadow_set forms a valid
```
rswarbrick, line 149: Nit: Does this need to say "rst_shadow_set_q" here?

rswarbrick (approved):
A couple of last nitty questions, but this looks really good to me.

vogelpi (approved):
LGTM, thanks @nasahlpa !

Issue #2132, opened 2024-01-18
docs/readthedocs.org:ibex-core CI step keeps failing


vogelpi:
The docs/readthedocs.org:ibex-core CI steps rebuilding the documentation seems to keep failing. I've seen it for multiple PRs now. The error message we get is always:
```
 python -m sphinx -T -E -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
Running Sphinx v4.5.0
/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/setuptools_scm/git.py:163: UserWarning: "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/checkouts/2130" is shallow and may cause errors
  warnings.warn(f'"{wd.path}" is shallow and may cause errors')
loading translations [en]... done

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/sphinx/registry.py", line 438, in load_extension
    metadata = setup(app)
               ^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/sphinxcontrib/applehelp/__init__.py", line 230, in setup
    app.require_sphinx('5.0')
  File "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/sphinx/application.py", line 393, in require_sphinx
    raise VersionRequirementError(version)
sphinx.errors.VersionRequirementError: 5.0

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/sphinx/cmd/build.py", line 272, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/ibex-core/envs/2130/lib/python3.11/site-packages/sphinx/application.py", line 219, in __init__
    self.setup_extension(extension)
[...]

PR #2137, opened 2024-01-31
[doc] Require sphinx version >= 7.0

GregAC:
The previous ~= 4.2 was failing build on readthedocs

This will (hopefully) fix https://github.com/lowRISC/ibex/issues/2132

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,5 +1,5 @@
 setuptools_scm
-sphinx~=4.2
+sphinx>=7.0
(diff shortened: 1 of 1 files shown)


nbdd0121:
The code says >=5.0 but the commit message says >=7.0?

GregAC:
Ah whoops, initially set it to 5.0 then realised 7.0 was the latest version, trialled that locally and it worked but forgot to update the commit!

nbdd0121 (approved):

PR #2138, opened 2024-01-31
Fix stale merge commit issue in private CI

nbdd0121:
This is a very rare condition where GitHub triggers the workflow a little bit too fast before the asynchronous mergeability check has completed (only an issue for pull_request_target because pull_request would block on mergeability check completion). To address this issue I added a polling loop to do a few retries.

An example run is here: https://github.com/nbdd0121/ibex/actions/runs/7728003457/job/21067811451 (it fails in a later step due to lack of secrets, but that's expected).

### DIFF
--- a/.github/workflows/private-ci.yml
+++ b/.github/workflows/private-ci.yml
@@ -23,6 +23,35 @@ jobs:
     name: Trigger Private CI
     runs-on: ubuntu-latest
     steps:
+      # Find a merge commit. We cannot use merge_commit_sha from context directly because
+      # mergeability check is asynchronous to pull_request_target trigger..
+      - name: Find the merge commit
+        id: merge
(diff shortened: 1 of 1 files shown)


jwnrt:
How would you feel about bundling all of this into a GitHub action so it's encapsulated away from the workflow YAML?

nbdd0121:
Do you mean create another lowRISC repo for this?

jwnrt:
If they need another repo then possibly yeah, we can use it for OpenTitan too in that case. Alternatively I think some actions you can put in an `actions/` directory in this repo.

GregAC (approved):
Approving and merging to fix CI issues. We can refactor to address the suggestions from @jwnrt in a follow-up

PR #2145, opened 2024-02-29
Lowrisc ip update

GregAC:
Vendors in the latest IP from opentitan.

prim_count has changed so a small tweak to `ibex_lockstep` was required. There are no functional differences.

### DIFF
--- a/rtl/ibex_lockstep.sv
+++ b/rtl/ibex_lockstep.sv
@@ -132,17 +132,18 @@ module ibex_lockstep import ibex_pkg::*; #(
     .Width      (LockstepOffsetW        ),
     .ResetValue (LockstepOffsetW'(1'b0) )
   ) u_rst_shadow_cnt (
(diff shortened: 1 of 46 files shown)


msfschaffner:
Does it maybe make sense to wait until the CIP lib changes have been merged here https://github.com/lowRISC/opentitan/pull/21226? Just because all of that stuff gets vendored.

GregAC:
> Does it maybe make sense to wait until the CIP lib changes have been merged here [lowRISC/opentitan#21226](https://github.com/lowRISC/opentitan/pull/21226)? Just because all of that stuff gets vendored.

I don't think it matters, Ibex doesn't use those parts of the CIP lib (the icache testbench uses some of it, but it doesn't care about interrupts).

marnovandermaas (approved):
Had a quick look through the vendored in IP and I think the only thing is the update to `prim_count`.

PR #2147, opened 2024-03-01
[pmp] Use top-level straps for PMP reset values

Razer6:
By using top-level straps for the PMP reset configuration its easier to implement different reset configurations if there are multiple Ibex cores in the system.


rswarbrick (commented):
This looks like a sensible change to me.

@GregAC can you see any problem with this? If not, I'm happy to take it.

GregAC:
I had intentionally avoid doing this with parameters when I first wrote it to avoid adding yet more parameters to Ibex, though given we've already got loads perhaps it doesn't matter.

Ultimately this is probably what we have to do but I'll have a think about alternative methods.

GregAC:
@Razer6 we intend to take this, I just need to do a couple of checks to ensure there's no issues for non PMP configs or breakages in other things we don't check in CI I hope to get this done over the next couple of days.

Razer6:
@GregAC No hurry on that one. We have changed that in our environment and are not dependent on vendoring in a new version of ibex at the moment.

Razer6:
@GregAC @rswarbrick Could you take another look?

GregAC:
@Razer6 yes will do

doc/03_reference/pmp.rst
```diff
@@ -50,5 +50,5 @@ Custom Reset Values
 
 By default all PMP CSRs (include ``mseccfg``) are reset to 0.
 Some applications may want other reset values.
-Default reset values are defined in :file:`ibex_pmp_reset_default.svh`.
-An implementation can either modify this file or define ``IBEX_CUSTOM_PMP_RESET_VALUES`` and place a copy of :file:`ibex_pmp_result_default.svh` in a new file, :file:`ibex_pmp_reset.svh`, changing the values as required and adding the new file to the include path of whatever build flow is being used.
+Default reset values are defined in :file:`ibex_pkg.svh`.
```
GregAC, line 53: It's just `ibex_pkg.sv` (no h)

rtl/ibex_lockstep.sv
```diff
@@ -9,36 +9,39 @@
...
+  parameter bit                     ICacheECC         = 1'b0,
+  parameter int unsigned            BusSizeECC        = BUS_SIZE,
+  parameter int unsigned            TagSizeECC        = IC_TAG_SIZE,
+  parameter int unsigned            LineSizeECC       = IC_LINE_SIZE,
+  parameter bit                     BranchPredictor   = 1'b0,
+  parameter bit                     DbgTriggerEn      = 1'b0,
+  parameter int unsigned            DbgHwBreakNum     = 1,
+  parameter bit                     ResetAll          = 1'b0,
   parameter lfsr_seed_t  RndCnstLfsrSeed   = RndCnstLfsrSeedDefault,
   parameter lfsr_perm_t  RndCnstLfsrPerm   = RndCnstLfsrPermDefault,
```
GregAC, line 36: Alignment needs adjusting here also

GregAC (approved):
This looks good to me thanks @Razer6 just a couple of small nits to address.

Razer6:
Thanks @GregAC for the review. I just updated the changes for the nits.

GregAC:
Thanks @Razer6

PR #2151, opened 2024-03-21
[icache,dv] Fix types in ibex_icache_sim_cfg.hjson

rswarbrick:
I actually made this change in OpenTitan (editing the vendored Ibex code by accident). Oops! The easiest path to getting everything in sync again is probably to manually make the identical change here and then re-vendor the result into OpenTitan.

For reviewers who also happen to be on the OT side: this was a silly mistake from me, and it seems that a bug in the CI code in OpenTitan means we can't currently see the mistake. But I've just fixed the CI bug... which means CI checks in OpenTitan are currently failing, for silly reasons. This is the simplest path I can see to getting everything green again.

### DIFF
--- a/dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson
+++ b/dv/uvm/icache/dv/ibex_icache_sim_cfg.hjson
@@ -29,12 +29,11 @@
 
   sim_tops: ["ibex_icache_fcov_bind"]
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
I take it back! There's an easier way, that I can do on the Ibex side for now. :-)

PR #2157, opened 2024-03-27
[simple_system] Bump C++ version in core files

### LINKED ISSUE #2153: Installation Issue: FuseSoC build doesn't work
Ubuntu 22.04
Verilator version: 5.022 2024-02-24 rev v5.020-157-g2b4852048
gcc version: 12.3.0
g++ version: 12.3.0
clang version: 14.0.0

After the installation of all the prerequisites, I'm stuck.
Idk how to solve this, why does it seems that I haven't a C++14 compiler?

Screenshot from 2024-03-25 10-32-09
ERROR: Failed to build lowrisc:ibex:ibex_simple_system:0 : '['make', '-j', '3']'

rswarbrick:
This sets the C++ standard that's being used for a compiler that's building a Verilator simulation. Recent versions of Verilator (since 5.020) require the compiler to be in C++14 mode, so passing -std=c++11 breaks the build with them.

Looking at history, -std=c++14 has been supported since GCC 6.1 (released in 2016), so I don't think this argument is going to cause any tooling problems.

Fixes #2153.

### DIFF
--- a/dv/riscv_compliance/ibex_riscv_compliance.core
+++ b/dv/riscv_compliance/ibex_riscv_compliance.core
@@ -161,6 +161,6 @@ targets:
           - '--trace-structs'
           - '--trace-params'
           - '--trace-max-array 1024'
-          - '-CFLAGS "-std=c++11 -Wall -DVM_TRACE_FMT_FST -DTOPLEVEL_NAME=ibex_riscv_compliance -g"'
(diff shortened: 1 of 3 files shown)


GregAC (approved):
I think the CI failure is some transient issue, so have rerun the job

nbdd0121:
GitHub's API seems to be quite laggy today, even 30 seconds of retry doesn't seem long enough.

PR #2164, opened 2024-05-02
[rtl] Expose ICacheScrNumPrinceRoundsHalf parameter

vogelpi:
This parameter allows integrators controlling the number of PRINCE half rounds in the scrambled ICache SRAM primitives, e.g., to balance timing impact and security guarantees.

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -13,28 +13,29 @@
  * Top level module of the ibex RISC-V core
  */
 module ibex_top import ibex_pkg::*; #(
-  parameter bit          PMPEnable        = 1'b0,
-  parameter int unsigned PMPGranularity   = 0,
-  parameter int unsigned PMPNumRegions    = 4,
-  parameter int unsigned MHPMCounterNum   = 0,
-  parameter int unsigned MHPMCounterWidth = 40,
-  parameter bit          RV32E            = 1'b0,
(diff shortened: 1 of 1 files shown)


vogelpi:
FYI: This is needed for OpenTitan where the ICache scrambling is acceptable to be slightly weaker compared to the scrambling of the main SRAM.

rswarbrick (commented):
I think this is changing behaviour because the currently checked-in version of the prim has a default of 2 (in the Ibex repository).

Is the change intentional? If so, would you mind doing it as two commits? The first could be a no-op and pass the default value, which would mean all the line noise from the reformatting was for the no-op. Then the second commit would be a 1-line 2->3 change.

vogelpi:
Would you prefer me switching to 2 here?

rswarbrick:
I think I'm suggesting switching it to 2 here to keep things unchanged locally. Then presumably `rv_core_ibex` can be instantiated with the parameter set to 3 in OpenTitan, right?

rswarbrick:
Oh! Sorry! Now I understand (having read *both* messages...) I guess I'd suggest leaving things so that we have the values from the prim that is currently vendored into Ibex. If not, I'd definitely split the change out to a separate commit in the same PR.

vogelpi:
Alright, I've switched it to 2 in this PR.

rswarbrick (approved):

Issue #2144, opened 2024-02-29
Security vulnerability: Ibex leaks data when multiplication is aborted


KatCe:
## Observed Behavior

A MULH instruction that is aborted by an unsolicited memory error response leaks information about its input values to a subsequent MUL instruction. The subsequent MUL instruction stores a value dependent on MULH's operands into its result register, thus produces a wrong result and leaking information.

I believe that the reason is that the multiplication's module's internal state machine is not reset in case of an abort. 

## Expected Behavior

Ibex should not leak data in case of external interface errors.

## Steps to reproduce the issue

The testbench in https://github.com/KatCe/ibex/tree/bug_mulh_reg_leakage/dv/bug_mulh_reg_leakage demonstrates the problem.
This testbench uses the default configuration. We have also observed the bug in secure configuration with the parameters commented in the testbench.

## My Environment

**EDA tool and version:**

Questa Sim-64, Version 2022.3_1 linux_x86_64 Aug 12 2022

**Operating system:**

Ubuntu 20.04.6 LTS
**Version of the Ibex source code:**

5a8a1a99930afbb9efe107ff2c80d482239d7a5f
The exact list of files run by the test bench can be found in https://github.com/KatCe/ibex/blob/bug_mulh_reg_leakage/dv/bug_mulh_reg_leakage/flist.f

GregAC:
Thank you for the report @KatCe. We are looking at this internally and will report back once we've examined the issue in detail.

PR #2166, opened 2024-05-16
[rtl] Guard against false memory responses for secure configurations

GregAC:
With this change all memory responses are only acted on if Ibex is expecting them for all secure configurations. Previously an error response that was injected onto the bus would trigger an exception that shouldn't occur (in particular breaking the functioning of the multiply state machine). In addition for configurations without the writeback stage an injected load data response could trigger an incorrect write to the register file.

This is only applied to the secure configurations, non-secure configurations assume correct adherence to the bus protocol meaning a response will only be seen if a request is outstanding.

Fixes https://github.com/lowRISC/ibex/issues/2144

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -211,10 +211,13 @@ module ibex_core import ibex_pkg::*; #(
   exc_cause_t  exc_cause;                      // Exception cause
 
   logic        instr_intg_err;
-  logic        lsu_load_err;
-  logic        lsu_store_err;
+  logic        lsu_load_err, lsu_load_err_raw;
+  logic        lsu_store_err, lsu_store_err_raw;
   logic        lsu_load_resp_intg_err;
   logic        lsu_store_resp_intg_err;
+
+  logic        expecting_load_resp_id;
+  logic        expecting_store_resp_id;
 
   // LSU signals
   logic        lsu_addr_incr_req;
@@ -290,6 +293,7 @@ module ibex_core import ibex_pkg::*; #(
   logic [1:0]  lsu_type;
   logic        lsu_sign_ext;
   logic        lsu_req;
+  logic        lsu_rdata_valid;
   logic [31:0] lsu_wdata;
   logic        lsu_req_done;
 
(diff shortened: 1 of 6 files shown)


marnovandermaas (commented):
I think this change looks reasonable to me, although it would be good to get more people to approve before merging.

rtl/ibex_ex_block.sv
```diff
@@ -196,4 +196,22 @@ module ibex_ex_block #(
...
+
+  if (RV32M == RV32MSlow) begin : gen_multdiv_sva_idle_slow
+    assign sva_multdiv_fsm_idle = gen_multdiv_slow.multdiv_i.sva_fsm_idle;
+  end else if (RV32M == RV32MFast || RV32M == RV32MSingleCycle) begin : gen_multdiv_sva_idle_fast
+    assign sva_multdiv_fsm_idle = gen_multdiv_fast.multdiv_i.sva_fsm_idle;
+  end else begin : gen_multdiv_sva_idle_none
+    assign sva_multdiv_fsm_idle = 1'b1;
+  end
+
+  // Mark the sva_multduv_fsm_idle as unused to avoid lint issues
```
vogelpi, line 212: Mintor typo: sva_multduv_fsm_idle -> sva_multdiv_fsm_idle

vogelpi (approved):
Very nice work, many thanks @GregAC !

rtl/ibex_core.sv
```diff
@@ -1810,6 +1839,9 @@ module ibex_core import ibex_pkg::*; #(
   // Certain parameter combinations are not supported
   `ASSERT_INIT(IllegalParamSecure, !(SecureIbex && (RV32M == RV32MNone)))
 
+  // If the ID stage signals its ready the mult/div FSMs must be idle in the following cycle
+  `ASSERT(MultDivFSMIdleOnIdReady, id_in_ready |=> ex_block_i.sva_multdiv_fsm_idle)
```
KatCe, line 1843: ex_block_i.sva_multdiv_fsm_idle is not accessible if INC_ASSERT is not defined

PR #2170, opened 2024-05-28
[rtl] Readback mode and iCache MuBi encoding update

nasahlpa:
This PR updates the codebase as well as `vendor/lowrisc_ip` to factor in the SRAM readback mode (lowRISC/opentitan#23212) and the iCache MuBi encoding (lowRISC/opentitan#23292) changes.


vogelpi (approved):
LGTM, verbilelint fails because the vendor commit touches so many files (there was patch affecting most FuseSoC core files in OpenTitan + the patch changing the copyright headers since we last updated the prims).

Given how big the vendor change is, I think it's also fine to keep the other changes related to the vendor update in a separate commit. What is your view @GregAC ?

GregAC (approved):

GregAC:
> LGTM, verbilelint fails because the vendor commit touches so many files (there was patch affecting most FuseSoC core files in OpenTitan + the patch changing the copyright headers since we last updated the prims).
> 
> Given how big the vendor change is, I think it's also fine to keep the other changes related to the vendor update in a separate commit. What is your view @GregAC ?

Given the priority here I'm happy to just merge and bypass the verible lint failure. I don't think it'll run in the merge queue because it's a PR specific check so merge should go ahead fine. Should it result in some CI failure we can fix it up with a late rPR.

Issue #2173, opened 2024-06-07
[dv] Implement fault injection tests for `prim_ram_1p_scr` MUBI values


GregAC:
Implement new test which inject faults into various MUBI values used in the `prim_ram_1p_scr` module. Proposed test should:

- Run a random program using the same RISC-DV test and config as `riscv_rand_instr_test`
- After a random interval choose a random `prim_ram_1p_scr` instance from the following (within `ibex_top`).
  - `gen_rams.gen_rams_inner.gen_scramble_rams.tag_bank`
  - `gen_rams.gen_rams_inner.gen_scramble_rams.data_bank`
- Chose a random MUBI rather within that instance from the following:
  - `read_en`
  - `write_en_d`/`write_en_q`
  - `addr_match`
  - `rvalid_q`
  - `req_d`/`req_q`- Within `prim_ram_1p_adv`
  - `write_d`/`write_q` - Within `prim_ram_1p_adv`
  - `rvalid_d`/`rvalid_q` - Within `prim_ram_1p_adv`
  - `rvalid_sram_d`/`rvalid_sram_q` - Within `prim_ram_1p_adv`
 - Check an alert has occurred within a bounded number of cycles (bound TBD)

PR #2182, opened 2024-07-01
[dv] Add riscv_rf_ctrl_intg_test

GregAC:
This tests new hardening added to the register file around read and write control signals.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -199,6 +199,78 @@ class core_ibex_rf_intg_test extends core_ibex_base_test;
 
 endclass
 
+class core_ibex_rf_ctrl_intg_test extends core_ibex_base_test;
+  `uvm_component_utils(core_ibex_rf_ctrl_intg_test)
+  `uvm_component_new
+
+  uvm_report_server rs;
+
+  virtual task send_stimulus();
+    int          rnd_delay;
+    int unsigned bit_idx;
+    logic [31:0] orig_val, glitch_val;
+    logic        alert_major_internal;
+    string       glitch_path, alert_major_internal_path;
+    string       ctrl_signals[];
+    int unsigned ctrl_signal_idx;
+    string       top_path = "core_ibex_tb_top.dut.u_ibex_top";
+    string       ibex_rf_path = {top_path, ".gen_regfile_ff.register_file_i"};
+
+    ctrl_signals = {
+      "we_a_dec",
+      "gen_rdata_mux_check.raddr_onehot_a",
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -199,6 +199,81 @@ class core_ibex_rf_intg_test extends core_ibex_base_test;
...
+    ctrl_signals = {
+      "we_a_dec",
+      "gen_rdata_mux_check.raddr_onehot_a",
+      "gen_rdata_mux_check.raddr_onehot_b"
+    };
+
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(ctrl_signal_idx, ctrl_signal_idx < ctrl_signals.size();)
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(rnd_delay, rnd_delay > 1000; rnd_delay < 10_000;)
+
+    ctrl_signal = ctrl_signals[ctrl_signal_idx];
```
rswarbrick, line 229: Nit: I'd probably inline this into the computation of `glitch_path` and drop a variable.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -199,6 +199,81 @@ class core_ibex_rf_intg_test extends core_ibex_base_test;
...
+    `uvm_info(`gfn, $sformatf("Forcing %s to value 'h%0x", glitch_path, glitch_val), UVM_LOW)
+    `DV_CHECK_FATAL(uvm_hdl_force(glitch_path, glitch_val));
+
+    // Leave glitch applied for one clock cycle.
+    clk_vif.wait_n_clks(1);
+
+    // Check that the alert matches our expectation.
+    alert_major_internal_path = $sformatf("%s.alert_major_internal_o", top_path);
+    `DV_CHECK_FATAL(uvm_hdl_read(alert_major_internal_path, alert_major_internal))
+    `DV_CHECK_EQ_FATAL(alert_major_internal, 1'b1, "Major alert did not fire!")
```
rswarbrick, line 260: Nit: For this sort of thing, I'd probably use something like `DV_CHECK_FATAL(alert_major_internal, "blah")` because otherwise the reader needs to figure out how the first two arguments work. Honestly, I'm rather unconvinced about `DV_CHECK_EQ` in general, but the default print message is slightly useful, espeically if the actual and expected values are variable and/or multi-bit.

rswarbrick (approved):
A couple of nitty style comments, but this looks good. I particularly like the "early stop" at the bottom: I hadn't seen that trick before.

nasahlpa (approved):
Thanks Greg for working on this, LGTM.

Will use this as a base when working on issue #2173.

PR #2185, opened 2024-07-02
[dv] Add spurious responses to memory agent

GregAC:
A spurious response is one that isn't associated with any on-going request. With this new feature the memory agent can generate them randomly when the interface is idle (i.e. there are no outstanding requests).

### DIFF
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
@@ -10,18 +10,28 @@ class ibex_mem_intf_monitor extends uvm_monitor;
 
   protected virtual ibex_mem_intf vif;
 
+  // The monitor tick event fires every clock cycle once any writes to
+  // outstanding_access_port and addr_ph_ports have occurred.
+  event monitor_tick;
+
   mailbox #(ibex_mem_intf_seq_item)          collect_response_queue;
   uvm_analysis_port#(ibex_mem_intf_seq_item) item_collected_port;
   uvm_analysis_port#(ibex_mem_intf_seq_item) addr_ph_port;
+  // The number of outstanding accesses is written to this port every clock cycle
+  uvm_analysis_port#(int)                    outstanding_accesses_port;
 
   `uvm_component_utils(ibex_mem_intf_monitor)
   `uvm_component_new
+
+  int outstanding_accesses = 0;
 
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
     item_collected_port = new("item_collected_port", this);
     addr_ph_port = new("addr_ph_port_monitor", this);
     collect_response_queue = new();
+    outstanding_accesses_port = new("outstanding_accesses_port", this);
+
     if(!uvm_config_db#(virtual ibex_mem_intf)::get(this, "", "vif", vif)) begin
        `uvm_fatal("NOVIF",{"virtual interface must be set for: ",get_full_name(),".vif"});
     end
@@ -52,9 +62,23 @@ class ibex_mem_intf_monitor extends uvm_monitor;
 
   virtual protected task collect_address_phase();
     ibex_mem_intf_seq_item trans_collected;
+
+
     forever begin
+      @(vif.monitor_cb);
+
       trans_collected = ibex_mem_intf_seq_item::type_id::create("trans_collected");
-      while(!(vif.monitor_cb.request && vif.monitor_cb.grant)) @(vif.monitor_cb);
+
+      while (!(vif.monitor_cb.request && vif.monitor_cb.grant)) begin
+        if (vif.monitor_cb.rvalid && !vif.monitor_cb.spurious_response) begin
+          outstanding_accesses--;
+        end
+        outstanding_accesses_port.write(outstanding_accesses);
+
+        -> monitor_tick;
+        @(vif.monitor_cb);
+      end
+
       trans_collected.addr                       = vif.monitor_cb.addr;
       trans_collected.be                         = vif.monitor_cb.be;
       trans_collected.misaligned_first           = vif.monitor_cb.misaligned_first;
@@ -63,7 +87,7 @@ class ibex_mem_intf_monitor extends uvm_monitor;
       trans_collected.m_mode_access              = vif.monitor_cb.m_mode_access;
       `uvm_info(get_full_name(), $sformatf("Detect request with address: %0x",
                 trans_collected.addr), UVM_HIGH)
-      if(vif.monitor_cb.we) begin
+      if (vif.monitor_cb.we) begin
         trans_collected.read_write = WRITE;
         trans_collected.data = vif.monitor_cb.wdata;
         trans_collected.intg = vif.monitor_cb.wintg;
@@ -73,7 +97,14 @@ class ibex_mem_intf_monitor extends uvm_monitor;
       addr_ph_port.write(trans_collected);
       `uvm_info(get_full_name(),"Send through addr_ph_port", UVM_HIGH)
       collect_response_queue.put(trans_collected);
-      @(vif.monitor_cb);
+
+      outstanding_accesses++;
+      if (vif.monitor_cb.rvalid && !vif.monitor_cb.spurious_response) begin
+        outstanding_accesses--;
+      end
+      outstanding_accesses_port.write(outstanding_accesses);
+
+      -> monitor_tick;
     end
   endtask : collect_address_phase
 
@@ -83,7 +114,7 @@ class ibex_mem_intf_monitor extends uvm_monitor;
       collect_response_queue.get(trans_collected);
       do
         @(vif.monitor_cb);
-      while(vif.monitor_cb.rvalid === 0);
+      while(vif.monitor_cb.rvalid === 0 || vif.monitor_cb.spurious_response === 1);
 
       if (trans_collected.read_write == READ) begin
         trans_collected.data = vif.monitor_cb.rdata;
--- a/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
+++ b/dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
@@ -18,10 +18,19 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
   bit                    error_synch = 1'b1;
   bit                    is_dmem_seq = 1'b0;
   bit                    suppress_error_on_exc = 1'b0;
+  bit                    enable_spurious_response = 1'b0;
+
 
   `uvm_object_utils(ibex_mem_intf_response_seq)
   `uvm_declare_p_sequencer(ibex_mem_intf_response_sequencer)
   `uvm_object_new
+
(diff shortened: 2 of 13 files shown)


GregAC:
I still need to do a full regression with this version but regressions with earlier prototypes were working.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -10,18 +10,28 @@ class ibex_mem_intf_monitor extends uvm_monitor;
 
   protected virtual ibex_mem_intf vif;
 
+  // The monitor tick event fires every clock cycle once any writes to
+  // outstanding_access_port and addr_ph_ports have occurred.
+  event monitor_tick;
```
rswarbrick, line 15: This is reasonable, but it's a pity to need it! I think we're just using this in `ibex_mem_intf_response_seq`. Can't we just run that sequence on the negedge? I think it might be simpler.

If that doesn't work, I think the comment still needs tweaking. I think the event comes before the write to `addr_ph_port` on line 95.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -10,18 +10,28 @@ class ibex_mem_intf_monitor extends uvm_monitor;
...
+
   mailbox #(ibex_mem_intf_seq_item)          collect_response_queue;
   uvm_analysis_port#(ibex_mem_intf_seq_item) item_collected_port;
   uvm_analysis_port#(ibex_mem_intf_seq_item) addr_ph_port;
+  // The number of outstanding accesses is written to this port every clock cycle
+  uvm_analysis_port#(int)                    outstanding_accesses_port;
```
rswarbrick, line 21: I can see where this port gets attached to `outstanding_accesses_imp` in the sequencer. But I can't see any other use of it. How do items go further?

More generally, a stream of numbers seems slightly surprising here. Is there a reason that we don't expose it as a "status"?

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -52,9 +62,23 @@ class ibex_mem_intf_monitor extends uvm_monitor;
 
   virtual protected task collect_address_phase();
     ibex_mem_intf_seq_item trans_collected;
+
+    @(vif.monitor_cb);
```
rswarbrick, line 66: Can you add a comment here to explain why we have to delay a cycle? (Or it might be worth moving the clock edge in the forever loop to the start?)

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -52,9 +62,23 @@ class ibex_mem_intf_monitor extends uvm_monitor;
...
+
     forever begin
       trans_collected = ibex_mem_intf_seq_item::type_id::create("trans_collected");
-      while(!(vif.monitor_cb.request && vif.monitor_cb.grant)) @(vif.monitor_cb);
+
+      while(!(vif.monitor_cb.request && vif.monitor_cb.grant)) begin
```
rswarbrick, line 71: I know this was here already, but maybe worth adding a space after "while" here.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
```diff
@@ -52,9 +62,23 @@ class ibex_mem_intf_monitor extends uvm_monitor;
...
+
+        if (vif.monitor_cb.rvalid && !vif.monitor_cb.spurious_response) begin
```
rswarbrick, line 73: The `!spurious_response` bit seems slightly surprising here. Should the monitor really care about whether the response was the one we were waiting for? I guess the problem is the `outstanding_accesses` counter, but maybe that really belongs in the driver or the sequence rather than the monitor, since those components know whether they have anything "in flight".

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
```diff
@@ -41,12 +53,48 @@ class ibex_mem_intf_response_seq extends uvm_sequence #(ibex_mem_intf_seq_item);
...
+            `DV_CHECK_RANDOMIZE_WITH_FATAL(req, rvalid_delay == 0;)
+
+            req.spurious_response = 1'b1;
+            {req.intg, req.data} = prim_secded_pkg::prim_secded_inv_39_32_enc(req.data);
+
+            `uvm_info(`gfn, $sformatf("Generated spurious response:\n%0s", req.sprint()), UVM_HIGH)
+            start_item(req);
+            finish_item(req);
+
+            `DV_CHECK_MEMBER_RANDOMIZE_FATAL(spurious_response_delay_cycles)
```
rswarbrick, line 84: I think it might be simpler to do this before starting the `while ` loop and then run it as `repeat (spurious_response_delay_cycles`, which I think makes the code a bit easier. Untested, but I think you end up with something like:
```systemverilog
      if (enable_spurious_response) begin
        bit real_request = 1'b0;

        forever begin
          // When spurious responses are enabled, we insert a spurious response after a randomised
          // delay.
          int unsigned delay_cycles;
          `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(
            delay_cycles,
            delay_cycles inside [p_sequencer.cfg.spurious_response_delay_min:
                                 p_sequencer.cfg.spurious_response_delay_max];)

          repeat (delay_cycles) begin

            @p_sequencer.monitor_tick;

            if (p_sequencer.addr_ph_port.try_get(item)) begin
              // If we have a new request, proceed as normal.
              real_request = 1'b1;
              break;
            end
          end

          if (p_sequencer.outstanding_accesses == 0 && !real_request) begin
            // If it's time to generate a new spurious response and there are no outstanding
            // responses (we must only generate a spurious response when the interface is idle) send
            // one to the driver.
            req = ibex_mem_intf_seq_item::type_id::create("req");

            `DV_CHECK_RANDOMIZE_WITH_FATAL(req, rvalid_delay == 0;)

            req.spurious_response = 1'b1;
            {req.intg, req.data} = prim_secded_pkg::prim_secded_inv_39_32_enc(req.data);

            `uvm_info(`gfn, $sformatf("Generated spurious response:\n%0s", req.sprint()), UVM_HIGH)
            start_item(req);
            finish_item(req);
          end
        end
      end else begin
        // Without spurious responses just wait for the monitor to report a new request
        p_sequencer.addr_ph_port.get(item);
      end
```

Note that we can also get rid of the `spurious_response_delay_cycles` class variable and tighten the scope a bit.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
GregAC: I'm not a huge fan of running things on the negedge. It can make the waves confusing to read sometimes (results in signals being a particular value for only have a cycle in same cases and I think it's better if the testbench stimulus looks to behave the same way every other signal inthe design does) and I worry it's an easy way out around race conditions in the testbench which means you don't actually think about them properly.

In this specific case we're not doing stimulus on negedge elsewhere and it's good to keep things consistent. Plus switching to negedge might come with some other issues I don't have the time to debug right now.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
GregAC: The items go to the `write` function of `ibex_mem_intf_response_sequencer`. It's a little awkward e.g. being able to call it `write_outstanding_accesses` instead would be preferable but I don't think there's a way to alter the name of the function the imp calls. We could do a seperate UVM component that purely exists to implement the `write` function and hold the number of outstanding accesses.

As for why it exists in the first place it's to keep consistency with other abstractions. The sequencer is connected to the monitor but it's done in a loosely coupled fashion, we've got a couple of analysis ports that get connected up and an event that's connected across but in particular the sequencer doesn't need a direct pointer to the monitor nor does it need to know anything about the monitor type. Without this we'd need a direct pointer to the monitor and have to require it contains an 'outstanding_accesses' field.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
GregAC: Once I've remembered why I did this, sure :) Could be putting it into the forever loop is actually the better thing to do.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
GregAC: This is partially down to laziness on my part but pre this change spurious responses didn't exist and downstream users of the monitor (i.e. those things connecting to the `item_collected_port`) were expecting everything they receive to represent an actual transaction that occurred (in particular the co-sim scoreboard). If we output spurious responses here we'll need all downstream users of the monitor to filter out spurious responses. This felt like a cleaner design.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
rswarbrick: Ok, cool, makes sense.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
rswarbrick: Ahah, that makes sense. Thanks for explaining.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_seq_lib.sv
GregAC: The issue with that setup is the spurious responses can be effectively prevented from ever occurring (or at least more a lot rarer) where we've got a steady stream of real requests as we abort out spurious response countdown any time we see a real request and will restart it again after that's been handled.

With the code in the PR we only restart the countdown once a spurious response has been sent out which avoid this issue.

rswarbrick: Ah, I see. Thanks for explaining: I'm convinced :-)

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_monitor.sv
rswarbrick: Ok, that feels reasonable. Maybe an example of where "the principled solution" is also the wrong one :-)

rswarbrick (approved):
Thanks for the responses. This looks sensible to me.

Issue #2193, opened 2024-07-18
Pipeline not being flushed on PMP register clears


mndstrmr:
In [this commit](https://github.com/lowRISC/ibex/commit/46404cf86f60f72fb84a2bc693d2ab08011bdb37) the pipeline was made to flush on changes to PMP registers, changes made via `CSR_OP_CLEAR` however are missed.

This is likely because prior to this commit only mstatus and mie were listed, and clearing is safe for those.

This meant that instructions being fetched were passing PMP checks which they should not have been.

GregAC:
Thanks for finding this @mndstrmr I'll be putting a fix together shortly.

In the scenario you describe any instruction that's in the IF stage when the clear is executed has already has its PMP check done under the old pre-clear state. Ultimately whilst clearly an issue that must be fixed I think the practical impact of this is limited.

First it is only the instruction in the IF stage that immediately follows the clear that 'sees' the old PMP state (before clear) and hence could execute when it shouldn't. It's also notable that only the execute part of the PMP permissions is effected as the read/write is checked when we attempt to being the relevant memory transaction and will always see the latest version of the PMP state.

Second there are limited scenarios under which an csrrc can cause a PMP state change that is immediately relevant. You must alter the configuration of a region that you are about to execute (i.e. a region that covers the instruction after your clear) and any change to the PMP CSRs must be done in M mode so you need to alter M mode relevant state. This means either the region is locked (so cannot be altered unless msseccfg.RLB is set) or is one of the epmp shared data regions (which don't provide execute permissions so don't matter for this issue). This means you have to have msseccfg.RLB set to produce a code sequence that can be effected by this issue. The MMWP and MML bits in msseccfg cannot be cleared once set so cannot be affected by this issue (which only relates to CSR bit clears). 

For any software running on a version of Ibex effected by this issue it's recommended the CSR clears are never used to alter PMP state to avoid this scenario entirely.

korran:
Is there anything users can do to avoid the pipeline flushing on every PMP register update? I'd like to update pmpcfg0-3 and pmpaddr0-15 in a trap handler, and I don't see why it needs to take more than 42 cycles total. Ideally the pipeline flush would be deferred until the first relevant memory access after updating these CSRs.

PR #2195, opened 2024-07-24
Flush pipeline when PMP CSRs are cleared

marnovandermaas:
Closes: https://github.com/lowRISC/ibex/issues/2193

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -522,7 +522,9 @@ module ibex_id_stage #(
     // - When modifying any PMP CSR, PMP check of the next instruction might get invalidated.
     //   Hence, a pipeline flush is needed to instantiate another PMP check with the updated CSRs.
     // - When modifying debug CSRs.
-    if (csr_op_en_o == 1'b1 && (csr_op_o == CSR_OP_WRITE || csr_op_o == CSR_OP_SET)) begin
+    if (csr_op_en_o == 1'b1 && (csr_op_o == CSR_OP_WRITE ||
+                                csr_op_o == CSR_OP_SET   ||
+                                csr_op_o == CSR_OP_CLEAR)) begin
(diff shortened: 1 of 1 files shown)


nbdd0121 (approved):

mndstrmr:
This is my suggested solution, but I think it's wrong: Doesn't this prevent the `else` case ever being entered? Indeed was the `else` case ever correct? It would previously not be entered e.g. on a `WRITE` to a debug register, which it clearly should be.

I think the `if` and `else` should be merged, and the inner `if` statements merged into their parents, something like the below:

```sv
assign csr_pipe_flush = 
    csr_op_en_o == 1'b1 && csr_op_o != CSR_OP_READ && (
    	csr_num_e'(instr_rdata_i[31:20]) == CSR_MSTATUS   ||
     	csr_num_e'(instr_rdata_i[31:20]) == CSR_MIE       ||
      	csr_num_e'(instr_rdata_i[31:20]) == CSR_MSECCFG   ||
      	// To catch all PMPCFG/PMPADDR registers, get the shared top most 7 bits.
      	instr_rdata_i[31:25] == 7'h1D 		  	  ||
    	csr_num_e'(instr_rdata_i[31:20]) == CSR_DCSR      ||
	csr_num_e'(instr_rdata_i[31:20]) == CSR_DPC       ||
	csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH0 ||
	csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH1
    );
```

GregAC (changes requested):
I wanted to discuss different ways to approach this so please do not merge this for now.

GregAC:
I also don't want to make any RTL changes to Ibex until I've sorted the OT earlgrey release branch

marnovandermaas:
Closing this PR because it probably needs a different solution.

PR #2204, opened 2024-08-13
Fix icache regression failure on VCS

nbdd0121:
It appears that VCS require expression after `iff` to be wrapped inside parenthesis otherwise it will complain about syntax error.

This should fix the weekly VCS regression, which seems to be failing [forever](https://github.com/lowRISC/lowrisc-private-ci/actions/workflows/ibex-weekly-vcs.yml) due to this.


hcallahan-lowrisc (approved):
LGTM, thanks!

PR #2206, opened 2024-08-23
[rtl] Remove ECC related data_rdata_i -> instr_X_o feedthroughs

GregAC:
Prior to this commit an ECC failure on the incoming data memory response factored directly into the outputs for the instruction memory interfaces. This existed due to a desire to take an NMI on an ECC failure as soon as possible but causes timing issues so it has been altered.

Now rather than directly raise the NMI the same cycle the assertion of 'irq_nm_int' is delayed by a cycle which breaks the feedthrough path.


vogelpi (approved):
Thanks @GregAC , this looks good to me!

PR #2214, opened 2024-09-17
[rtl] Flush pipe on all CSR modifications

### LINKED ISSUE #2193: Pipeline not being flushed on PMP register clears

GregAC:
This fixes #2193, an issue that meant bit clears in PMP related CSRs didn't immediately apply to an instruction already in the fetch stage due to a lack of a pipeline flush.

With this change the pipeline will flush in that scenario, fixing the issue. It now flushes the pipeline on all CSR modifications as this makes the pipeline more resliant against similar issues in the future (where the list of CSRs to flush on should have been updated but wasn't).

### DIFF
--- a/rtl/ibex_id_stage.sv
+++ b/rtl/ibex_id_stage.sv
@@ -509,36 +509,13 @@ module ibex_id_stage #(
     .branch_in_dec_o(branch_in_dec)
   );
 
-  /////////////////////////////////
-  // CSR-related pipeline flushes //
-  /////////////////////////////////
-  always_comb begin : csr_pipeline_flushes
-    csr_pipe_flush = 1'b0;
-
-    // A pipeline flush is needed to let the controller react after modifying certain CSRs:
-    // - When enabling interrupts, pending IRQs become visible to the controller only during
-    //   the next cycle. If during that cycle the core disables interrupts again, it does not
-    //   see any pending IRQs and consequently does not start to handle interrupts.
-    // - When modifying any PMP CSR, PMP check of the next instruction might get invalidated.
-    //   Hence, a pipeline flush is needed to instantiate another PMP check with the updated CSRs.
-    // - When modifying debug CSRs.
-    if (csr_op_en_o == 1'b1 && (csr_op_o == CSR_OP_WRITE || csr_op_o == CSR_OP_SET)) begin
-      if (csr_num_e'(instr_rdata_i[31:20]) == CSR_MSTATUS ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_MIE     ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_MSECCFG ||
-          // To catch all PMPCFG/PMPADDR registers, get the shared top most 7 bits.
-          instr_rdata_i[31:25] == 7'h1D) begin
-        csr_pipe_flush = 1'b1;
-      end
-    end else if (csr_op_en_o == 1'b1 && csr_op_o != CSR_OP_READ) begin
-      if (csr_num_e'(instr_rdata_i[31:20]) == CSR_DCSR      ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_DPC       ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH0 ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH1) begin
-        csr_pipe_flush = 1'b1;
-      end
-    end
-  end
+  // Flush pipe on any CSR modification. Some CSR modifications alter how instructions execute (e.g.
+  // the PMP CSRs) so this ensures all instructions always see the latest architectural state when
+  // entering the fetch stage. This causes some needless flushes but performance impact is
+  // limited. We have a single fetch stage to flush not many stages of a deep pipeline and CSR
+  // instructions are in general rare and not part of performance critical parts of the code.
+  assign csr_pipe_flush =
+    (csr_op_en_o == 1) && (csr_op_o inside {CSR_OP_WRITE, CSR_OP_SET, CSR_OP_CLEAR});
 
   ////////////////
   // Controller //


GregAC:
Still need to run a full regression on this as well a specific directed test to ensure the particular bug has indeed been solved.

marnovandermaas (commented):
Looks like the `riscv_multiple_interrupt_test` is failing on this commit.

rtl/ibex_id_stage.sv
```diff
@@ -509,36 +509,13 @@ module ibex_id_stage #(
...
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH0 ||
-          csr_num_e'(instr_rdata_i[31:20]) == CSR_DSCRATCH1) begin
-        csr_pipe_flush = 1'b1;
-      end
-    end
-  end
+  // Flush pipe on any CSR modification. Some CSR modifications alter how instructions execute (e.g.
+  // the PMP CSRs) so this ensures all instructions always see the latest architectural state when
+  // entering the fetch stage. This causes some needless flushes but performance impact is
+  // limited. We have a single fetch stage to flush not many stages of a deep pipeline and CSR
```
rswarbrick, line 515: Grammar question: Do you mean something like "We only have a single fetch stage and the pipeline is not deep. Also, CSR instructions are normally rare and aren't part of performance critical parts of the code."

GregAC:
> Looks like the `riscv_multiple_interrupt_test` is failing on this commit.

I think that's a known failure, though may be we're getting more of them now as it's to do with assertions around pipeline flushes and we're doing more pipeline flushes. Could be we should use a less flaky test in our smoke suite.

rtl/ibex_id_stage.sv
GregAC: Yes I do, I shall rewrite it to make it clearer

rtl/ibex_id_stage.sv
```diff
@@ -509,36 +509,13 @@ module ibex_id_stage #(
...
+  // instructions are in general rare and not part of performance critical parts of the code.
```
nbdd0121, line 516: Note that it's a common pattern to do a `csrrw sp, mscratch, sp` to switch to a dedicated stack for exception handling.

For exception handling it also performs quite a few other CSR reads, so I wouldn't call CSR instructions rare and not part of performance critical code.

GregAC: These are good points.

Though do note we don't flush on reads only modifications. Whilst strictly all CSR instructions do perform modifications `CSRRS` and `CSRRC` with an `x0` `rs1` (the register that specifies which bits to set/clear) and the immediate versions with a 0 immediate do not alter the CSRs and will have `csr_op_o` in this RTL set to 'CSR_OP_READ` so won't trigger the flush.

Do we expect any CSR writes other than to `mscratch` during normal exception handling code?

I'm wonder if we should operate a white-list system here where modifications to specific CSRs do not trigger a flush but everything else does. `mscratch` is a good candidate for that list, anything else?

nbdd0121: `mepc` would also be updated in case the exception needs to skip over the current instruction. `mstatus` may be updated but that actually require flushes.

GregAC: Just realised use of `ecall` will require `mepc` modification (otherwise you get an infinite loop of `ecall`). Performance degradation from a pipe flush is modest (a single cycle in cases where you hit in icache or have single cycle access to instruction memory I believe) and there'd only be one `mepc` write per `ecall` so I don't think it'd be a huge issue but as above it seems a safe one to whitelist so we can do that.

nbdd0121: For exception handling I think it's not uncommon to save mepc & all registers to memory, call some routine, and unconditionally restore mepc. Also, for things like ecall, it's also needed to unconditionally increment by 4. But you're right they're not touched for interrupt handling, which matters for than synchronous exceptions.

marnovandermaas:
Last force push is a rebase.

GregAC:
The prior CI failure here was a bad assertion. I've recommend we just remove that assertion and related ones as they've been a lot of trouble to get working and provide little utility (https://github.com/lowRISC/ibex/pull/2255).

The latest version of this PR introduces the whitelist described above where we don't flush on certain CSRs being written. It also adds a new commit that directly reads the CSR address from the relevant instruction bits rather than via the operand mux. This helps timing in some cases and was useful to use in `ibex_id_stage` to implement the whitelist system (rather than reading directly from the raw instruction as was done previously).

@marnovandermaas @vogelpi PTAL

GregAC:
I've run a full regression with limited iterations (10 max for each test) which didn't show any issues, doing a full regression now.

Razer6:
I just wanted to comment on this PR to see its state. Glad there is some progess.

vogelpi (approved):
Thanks @GregAC , this change makes sense to me. I like the idea of moving to a whitelist system for when to avoid flushes. If it turns out there are performance issues for certain CSRs, we can take a close look an reason specifically for those. Identifying and fixing issues with CSR modifications that erroneously don't lead to flushes is more painful.

marnovandermaas (approved):
Thanks Greg, this looks good to me.

GregAC:
Latest version just fixes a small lint issue (explicit cast to `csr_num_e`) which isn't an issue for our CI verilator version but does appear under verilator v5.

I've manually confirmed this fixes the specific PMP issue by observing waves of a suitable test program

@marnovandermaas the formal flow should be run on this as well to double check all is good. Can do that once this is merged into master.

marnovandermaas:
Sounds good, the formal flow has a counter-example which I still need to resolve even before this PR goes in. That work is captured in this PR: https://github.com/lowRISC/ibex/pull/2245

PR #2221, opened 2024-11-08
Block diagram update

marnovandermaas:
My hackday project, which I think improves the block diagram formatting without doing any major re-organization.

### DIFF
--- a/doc/03_reference/images/blockdiagram.svg
+++ b/doc/03_reference/images/blockdiagram.svg
@@ -7,7 +7,7 @@
    viewBox="0 0 278.31393 171.93503"
    version="1.1"
    id="svg8"
-   inkscape:version="1.2.2 (732a01da63, 2022-12-09, custom)"
+   inkscape:version="1.4 (1:1.4+202410161351+e7c3feb100)"
    sodipodi:docname="blockdiagram.svg"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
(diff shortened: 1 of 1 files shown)


elliotb-lowrisc (changes requested):
This looks nice in Inkscape on my local computer, but has problems when rendered on GitHub: https://github.com/lowRISC/ibex/tree/c06a6b5c2efa093e0e5a686e3c208b7a6b37cebf

Note the loss of white background behind the "Instruction Memory Interface" box, the "Instruction Fetch" text is not shown as Exo 2 and is overrunning the edge of its containing box, and the tiny text below the Prefetch Buffer is not rendered at all.

elliotb-lowrisc (changes requested):
Ah, the CSS approach I suggested for the font doesn't seem to work on GitHub after all.

marnovandermaas:
Ok, I've done the boring thing and moved back to liberation sans.

elliotb-lowrisc (changes requested):
1. debug_req_i is unreadable on a dark background (e.g. dark-mode GitHub). Suggest increasing size back to what it was and adding either a white outline or background shape.
2. Text below Prefetch Buffer not rendering on GitHub. Suggest looking for layer issues.

Nits:
1. I liked the `sans-serif` "Ibex Core" title better.
2. "PMP Check" text is rather large (in both places).

elliotb-lowrisc (approved):
Looking good

hcallahan-lowrisc (approved):

PR #2228, opened 2024-11-29
[rtl] Fix non-DSP reset in ibex_counter

nasahlpa:
When targeting Xilinx FPGAs, we utilize a DSP for counters with a width of less than 49-bit. In this case, a sync. reset is needed. However, currently, there is a bug in the RTL where also a sync. reset is used for the non-DSP counters on the FPGA.

### DIFF
--- a/rtl/ibex_counter.sv
+++ b/rtl/ibex_counter.sv
@@ -51,26 +51,37 @@ module ibex_counter #(
   end
 
 `ifdef FPGA_XILINX
-  // Set DSP pragma for supported xilinx FPGAs
-  localparam int DspPragma = CounterWidth < 49 ? "yes" : "no";
-  (* use_dsp = DspPragma *) logic [CounterWidth-1:0] counter_q;
-
-  // DSP output register requires synchronous reset.
-  `define COUNTER_FLOP_RST posedge clk_i
+  // On Xilinx FPGAs, 48-bit DSPs are available that can be used for the
+  // counter. Hence, use Xilinx specific flop implementation. The datatype for
+  // UseDsp is on purpose int as with string Xilinx throws an error for the
+  // use_dsp pragma.
+  localparam int UseDsp = CounterWidth < 49 ? "yes" : "no";
+  (* use_dsp = UseDsp *) logic [CounterWidth-1:0] counter_q;
 `else
+  localparam int UseDsp = "no";
   logic [CounterWidth-1:0] counter_q;
-
-  `define COUNTER_FLOP_RST posedge clk_i or negedge rst_ni
 `endif
 
-  // Counter flop
-  always_ff @(`COUNTER_FLOP_RST) begin
-    if (!rst_ni) begin
-      counter_q <= '0;
-    end else begin
-      counter_q <= counter_d;
+  if (UseDsp == "yes") begin : g_cnt_dsp
+    // Use sync. reset for DSP.
+    always_ff @(posedge clk_i) begin
+      if (!rst_ni) begin
+        counter_q <= '0;
+      end else begin
+        counter_q <= counter_d;
+      end
+    end
+  end else begin : g_cnt_no_dsp
+    // Use async. reset for flop.
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        counter_q <= '0;
+      end else begin
+        counter_q <= counter_d;
+      end
     end
   end
+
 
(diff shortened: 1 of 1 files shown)


nasahlpa:
Apologies, I should have re-generated the bitstream after addressing this review [comment](https://github.com/lowRISC/ibex/pull/2226#discussion_r1861219466). I assumed this is purely cosmetically but it actually led to a Vivado synthesis failure :/

However, Vivado still gives me the following warning:
`WARNING: [Synth 8-2306] macro COUNTER_FLOP_RST redefined [11464cc7017f312f2c274e572db74bf2/sandbox/linux-sandbox/1586/execroot/_main/bazel-out/k8-fastbuild-ST-9c1027dad682/bin/hw/bitstream/vivado/build.fpga_cw340/src/lowrisc_ibex_ibex_core_0.1/rtl/ibex_counter.sv:63]`

Nevertheless, Vivado still seems to create a DSP for a counter with width less than 48-bits, .e.g.: the [mcounters_variable_i](https://github.com/lowRISC/ibex/blob/54985d21b055967c39d0a77b45ae0d573b55b0f7/rtl/ibex_cs_registers.sv#L1380) counter:
Screenshot from 2024-11-29 20-10-58
And for the 64-bit counters no DSP is used, e.g., the [mcycle](https://github.com/lowRISC/ibex/blob/54985d21b055967c39d0a77b45ae0d573b55b0f7/rtl/ibex_cs_registers.sv#L1328) counter:
Screenshot from 2024-11-29 20-11-14

If this warning still is a problem, any ideas @vogelpi or @GregAC how I can rewrite the `COUNTER_FLOP_RST` macro?

vogelpi (changes requested):
Thanks for fixing this @nasahlpa .

To me it looks like this PR adds a lot more code and makes things much less readable and more complex. It negatively impacts maintenance. I think what we should do:
1. Revert the previous PR which introduced the issue.
2. Do a PR that maybe doesn't address the comment you mentioned but that implements the change without making it much more complex.

nasahlpa:
As discussed offline:

- I've reverted the previous commit in PR #2229 
- Created two new modules for a Xilinx specific and a general flop implementation that is used in the counters. However, I've decided to not create a dedicated prim for this as we only would use this prim here in Ibex but not in the OpenTitan repository. However, if you think @vogelpi  it would make sense to have this as a dedicated prim I can move it.

rtl/ibex_counter.sv
```diff
@@ -51,26 +51,39 @@ module ibex_counter #(
...
-  // DSP output register requires synchronous reset.
-  `define COUNTER_FLOP_RST posedge clk_i
+  // On Xilinx FPGAs, 48-bit DSPs are available that can be used for the
+  // counter. Hence, use Xilinx specific flop implementation. The datatype for
+  // UseDsp is on purpose int as with string Xilinx throws an error for the
+  // use_dsp pragma.
+  localparam int UseDsp = CounterWidth < 49 ? "yes" : "no";
+  (* use_dsp = UseDsp *) logic [CounterWidth-1:0] counter_q;
 `else
+  localparam string UseDsp = "no";
```
vogelpi, line 61: I think it would be better to use an int type here as well for consistency. Thanks for adding the comment on Vivado requiring `int` instead of `string`.

rtl/ibex_counter.sv
```diff
@@ -51,26 +51,39 @@ module ibex_counter #(
...
-  `define COUNTER_FLOP_RST posedge clk_i or negedge rst_ni
 `endif
 
-  // Counter flop
-  always_ff @(`COUNTER_FLOP_RST) begin
-    if (!rst_ni) begin
-      counter_q <= '0;
-    end else begin
-      counter_q <= counter_d;
+  generate
```
vogelpi, line 65: Please don't use generate regions, some tools don't support this and this is why our style guide doesn't permit it: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#generate-constructs

vogelpi (approved):
Thanks @nasahlpa , this is pretty elegant :-)

rtl/ibex_counter.sv
nasahlpa: Fixed, thanks.

rtl/ibex_counter.sv
nasahlpa: Fixed, thanks.

vogelpi:
Thanks Pascal for addressing the remaining comments, I think we can merge this today.

PR #2229, opened 2024-12-02
Revert "[rtl] Fix counter reset value on FPGA"

nasahlpa:
Reverts lowRISC/ibex#2226 because the scope for counter_q is incorrect resulting in synthesis failure for Xilinx targets.


vogelpi (approved):
LGTM

PR #2232, opened 2024-12-13
[rtl,pmp] Allow all accesses to Debug Module in debug mode

andreaskurth:
The RISC-V Debug Specification ([current release 1.0.0-rc4](https://github.com/riscv/riscv-debug-spec/releases/tag/1.0.0-rc4)) in Section A.2 states that the PMP must not disallow accesses to addresses of the Debug Module when the hart is in debug mode, regardless of how the PMP is configured.  This commit changes the PMP accordingly.


rtl/ibex_pmp.sv
```diff
@@ -226,14 +232,23 @@ module ibex_pmp #(
                              pmp_req_addr_i[c][PMPGranularity+2-1:0]};
     end
 
+    // Determine whether the core is in debug mode and the access is to an address in the range of
+    // the Debug Module.  According to Section A.2 of the RISC-V Debug Specification, the PMP must
```
rswarbrick, line 236: Nit: Double space before "According" and I don't think the rest of this file has standardised on that style.

rtl/ibex_pmp.sv
```diff
@@ -226,14 +232,23 @@ module ibex_pmp #(
...
+    // not disallow fetches, loads, or stores in the address range associated with the Debug Module
+    // when the hart is in debug mode.
+    assign debug_mode_allowed_access[c] = debug_mode_i &
+                                         ((pmp_req_addr_i[c][31:0] & ~DmAddrMask) == DmBaseAddr);
```
rswarbrick, line 240: Nit: I think this line needs another space of indentation?

rtl/ibex_pmp.sv
```diff
@@ -226,14 +232,23 @@ module ibex_pmp #(
...
     // denied by figuring out the matching region and its permission check.
     assign pmp_req_err_o[c] = access_fault_check(csr_pmp_mseccfg_i.mmwp,
                                                  csr_pmp_mseccfg_i.mml,
                                                  pmp_req_type_i[c],
                                                  region_match_all[c],
                                                  priv_mode_i[c],
-                                                 region_perm_check[c]);
+                                                 region_perm_check[c]) &
+                              // No error if the access is allowed as Debug Module access.
+                              ~debug_mode_allowed_access[c];
```
rswarbrick, line 251: Nit: The indentation looks a bit odd to me here. Shouldn't it be the same as `region_perM_check[c]`?

rswarbrick (commented):
A couple of silly nitty comments, but the logic looks sensible to me.

Razer6 (approved):
Appart from the nits, this looks good to me.

rtl/ibex_pmp.sv
GregAC: That is one of the arguments to `access_fault_check` this new line is ANDed with the result of that function so I think the indentation is reasonable. However you could put `~debug_mode_allowed_access[c]` at the top rather than the bottom which may make it more readable.

GregAC (commented):
Thanks @andreaskurth this looks good to me, though one thing gives me pause. The current debug mode only effects ID/EX stage behaviour, this introduces something new in the micro-architecture where debug mode also effects IF stage behaviour. We only have one debug mode flag so cannot support a execution where something in ID/EX is running in debug mode and something in IF is not or vice-versa (contrast to the privilege level which has separate ID/EX and IF signals).

From reviewing `ibex_controller` RTL it looks like on all debug mode entry/exit we will flush the pipeline (as we do a branch) so I don't think this will be an issue however we should write some kind of assertion that checks we never get a debug mode transition without such a flush.

I do want to run a full regression on this before it gets merged. Assuming that looks ok I'll approve.

We can add the assertion referred to above in a follow up PR (e.g. as part of the DV work).

rtl/ibex_pmp.sv
andreaskurth: Thanks, implemented

andreaskurth:
> From reviewing `ibex_controller` RTL it looks like on all debug mode entry/exit we will flush the pipeline (as we do a branch) so I don't think this will be an issue however we should write some kind of assertion that checks we never get a debug mode transition without such a flush.

Thanks for the suggestion, I added an assertion. Could you please review it?

> I do want to run a full regression on this before it gets merged. Assuming that looks ok I'll approve.

Ready for running a full regression from my view

rtl/ibex_controller.sv
```diff
@@ -997,6 +997,11 @@ module ibex_controller #(
     // If there's a pending exception req that doesn't need a PC set we must not see one
     `ASSERT(IbexNoPCSetOnSpecialReqIfNotExpected,
       exception_req_pending && !expect_exception_pc_set |-> ~pc_set_o)
+
+    // If entering or exiting debug mode, the pipeline must be flushed. This is because Ibex
+    // currently does not support some of the pipeline stages being in debug mode; either all or
+    // none of the pipeline stages must be in debug mode.
+    `ASSERT(IbexPipelineFlushOnChangingDebugMode, debug_mode_d != debug_mode_q |-> flush_id_o)
```
GregAC, line 1004: Unfortunately it's not this simple to write the assertion. For one thing `flush_id` only effects the ID/EX stage. It sets `instr_valid_clear_i` which will clear out the current instruction in ID/EX but doesn't prevent a fetched instruction from immediately coming into ID/EX the following cycle.

In particular consider some scenario where we have an instruction in the fetch stage, and it is allowed to execute in debug mode under the PMP rule mode bypass, but wouldn't be allowed to execute outside of debug mode. The PMP check is done whilst it's in the fetch stage. So should we exit debug mode and flush ID/EX (but not IF) then that instruction will proceed into the ID/EX stage and get execute where it should instead cause a PMP exception.

Ideally we'd have an assertion that just generically checks whatever is in IF isn't making it into ID/EX this cycle but that could be fiddly to write. Switch this to `flush_id_o & pc_set_o` could do the trick. There may be reasonable behaviours that don't meet these requirements however by inspection all of our current debug entry/exit does meet this. So seems reasonable to mandate it for now and if there's some future micro-architectural tweak that breaks the assertion (whilst still executing correctly) we can think about it again then.

GregAC:
Regression run was good, though I should try again with the suggested assertion tweak before we merge.

andreaskurth:
> Regression run was good, though I should try again with the suggested assertion tweak before we merge.

Great, thanks @GregAC! Could you please try again with the updated assertion?

GregAC:
> > Regression run was good, though I should try again with the suggested assertion tweak before we merge.
> 
> Great, thanks @GregAC! Could you please try again with the updated assertion?

Regression looks good, thanks @andreaskurth

GregAC (approved):

andreaskurth:
Perfect, thank you! Merging

PR #2243, opened 2025-01-09
[ibex_tracer] Use static variables in always/final blocks

Razer6:
Without, this causes warnings in xcelium:

```
Local static variable with initializer requires 'static' keyword.
```

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -740,7 +740,7 @@ module ibex_tracer (
       // as a blocking assignment to xx. They then complain about the mixture with that an the
       // non-blocking assignment we use when opening the file. The bug is fixed with recent versions
       // of Verilator, but this hack is probably worth it for now.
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
Thanks for looking at this. I understand why it might be needed for the `final` block, but is it also needed for the `always` block? I'm a little surprised if this isn't considered an automatic variable?

Razer6:
> Thanks for looking at this. I understand why it might be needed for the `final` block, but is it also needed for the `always` block? I'm a little surprised if this isn't considered an automatic variable?

Xcelium warns exactly for these 3 locations.

rswarbrick:
Huh, you're completely right! This is in the 1800 spec, section 6.21. And, thinking about it, it's kind of reasonable (and Xcelium is a bit silly to warn about this, I think).

Anyway, this looks good to me.

rswarbrick (approved):

Issue #2230, opened 2024-12-06
[regfile_fpga] oh_raddr_*_err signals unassigned if RdataMuxCheck=0


glaserf:
## Observed Behavior
The onehot-error signals in relation to the read addresses (`oh_raddr_a_err`, `oh_raddr_b_err`) in [rtl/ibex_register_file_fpga.sv](../blob/master/rtl/ibex_register_file_fpga.sv) are unassigned for non-secure Ibex configurations (which is for example used for opentitan FPGA top-levels IIRC):
https://github.com/lowRISC/ibex/blob/667fd20d2ede51caececccbcbda3652074424ce2/rtl/ibex_register_file_fpga.sv#L152-L158

This leads to the regfile `err_o` signal becoming `x`
https://github.com/lowRISC/ibex/blob/667fd20d2ede51caececccbcbda3652074424ce2/rtl/ibex_register_file_fpga.sv#L55
which triggers assertions (at least for us) when simulating with the FPGA regfile. It however seems to not cause issues during synthesis, at least with Vivado.

I am also unsure about the comments that refer to `rdata_*_o` as asynchronous:
https://github.com/lowRISC/ibex/blob/667fd20d2ede51caececccbcbda3652074424ce2/rtl/ibex_register_file_fpga.sv#L153-L154

## Expected Behavior
Same as for the other regfile variants; signals in question being tied to `1'b0` for `RdataMuxCheck = 0`.

rswarbrick:
Oh! Looks like it's fixed! Thanks for adding a note.

About the asynchronous comments, they date back to the original version of the file (from 2020). I think the author probably meant to point out that `raddr_a_i` feeds through to `rdata_a_o` without any register stage, so isn't synchronous to some clock. But it's rather confusingly phrased.

I've just checked, and the behaviour is the same for e.g. `ibex_register_file_ff`. I'll file a PR to drop the comment for the FPGA register file so that things are a bit less confusing.

PR #2250, opened 2025-01-23
[ibex_register_file_fpga] Drop two confusing comments

rswarbrick:
These were noticed by someone responding to issue #2230. I think the author's original logic was to point out that there's a path from e.g. raddr_a_i to rdata_a_o which doesn't depend on any clock, so is "asynchronous".

But that's the same in the other modes and also for the other register file implementations, which don't have analogous comments.

Drop these ones.


nasahlpa (approved):
Ah yeah, makes sense. Thanks!

PR #2254, opened 2025-02-14
[ibex,tracer] Uniquify trace log with module hierarchy

Razer6:
In case of having multiple Ibexes, this caues a conflict in openening the same filename from different instances. Add the unique module hierarchy into the filename.

### DIFF
--- a/rtl/ibex_tracer.sv
+++ b/rtl/ibex_tracer.sv
@@ -753,7 +753,7 @@ module ibex_tracer (
       if (fh == 32'h0) begin
         static string file_name_base = "trace_core";
         void'($value$plusargs("ibex_tracer_file_base=%s", file_name_base));
-        $sformat(file_name, "%s_%h.log", file_name_base, hart_id_i);
+        $sformat(file_name, "%s_%h_%m.log", file_name_base, hart_id_i);
 
         $display("%m: Writing execution trace to %s", file_name);
         fh = $fopen(file_name, "w");


nasahlpa (commented):
Do you know if the trace log is automatically parsed by some other tool where we also should adapt the path after your change? If not, LGTM.

Razer6:
Not that I am aware of.

Razer6:
@GregAC or anyone with access, can you look into the failing CI?

Razer6:
> Before proceeding with that, I have a question for @Razer6: Is the hart ID, which gets appended to the file name, not sufficient for separating the trace log files? That is, do you have multiple harts with the same ID?

We have multiple Ibexes in the system. These are independent systems, thus the RV spec specifies all of them having hart id 0. Hart ID only becomes different, if let's say, Ibex has multiple harts within itself.

andreaskurth:
> We have multiple Ibexes in the system. These are independent systems, thus the RV spec specifies all of them having hart id 0. Hart ID only becomes different, if let's say, Ibex has multiple harts within itself.

RISC-V just specifies that "at least one hart must have a hart ID of zero" and that "hart IDs must be unique within the execution environment." So this change would address the case when there are actually multiple disjoint execution environments in the same RTL simulation.

rswarbrick:
I believe this PR is stale. Please rebase and re-open if you intend to work on it again.

PR #2255, opened 2025-02-14
[rtl] Remove low utility assertions

GregAC:
This removes several assertions from `ibex_controller`. They aimed to ensure that controller behaviour was correct on exception behaviour (e.g. ensuring that a pending interrupt will actually trigger an interrupt). However they've proved to be flaky and hard to maintain with multiple edge cases needing to be accounted for.

The co-simulation checking in functional verification will catch the same issues these assertions catch. The assertions (when working correctly) would cause a failure directly when the bug happens which makes debugging easier. However they've added significant effort in regression triage due to their many false failures so it's not worth the maintenance burden.

Within formal they don't really add any value now we have the full end-to-end formal flow.

I decided this was a sensible course of action when looking at the CI failure in https://github.com/lowRISC/ibex/pull/2214 (which was down to an incorrect failure in one of the removed assertions).


marnovandermaas (approved):
I think removing these assertions is reasonable.

Issue #2233, opened 2024-12-13
[dv] Verify access to Debug Module in debug mode even if disallowed by PMP


andreaskurth:
Verify the feature implemented in #2232.

PR #2262, opened 2025-03-28
[dv] Plan test for DM accesses in debug mode

andreaskurth:
This adds a test item (commented out as the test isn't implemented yet) for #2233.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
@@ -278,6 +278,20 @@
     compare_final_value_only: 1
     verbose: 1
 
+# TODO(#2233): Implement the following test (also note the priorities in the issue).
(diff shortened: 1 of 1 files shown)


Razer6:
Should we also perform negative testing here? I.e, try to access the DM when not in debug mode?

andreaskurth:
@Razer6:
> Should we also perform negative testing here? I.e, try to access the DM when not in debug mode?

Right, so when debug mode is not enabled, PMP checks apply as usual. Verifying PMP is outside the scope of this test, but a minimal negative check could make sense. How about:
> When debug mode is not enabled, accesses to the Debug Module address space are governed by PMP settings. Verifying PMP is the focus of other tests. This test verifies a simple case: when debug mode is disabled and the PMP does not allow accesses to the Debug Module address space, a random access to that address space fails.

Razer6 (approved):

PR #2266, opened 2025-04-19
[rtl] Add common security features of regfiles to separate module

### LINKED ISSUE #2217: [rtl] Refactor register files to bring common logic together
We have 3 register file implementations, one flip-flop based, one latch based and one targetting FPGA implementation. They all duplicate various bits of security hardening logic. The register files should be refactored so the security hardening logic is in a common module as much as is possible and keeping the per-implementation logic to a minimum.

aklsh:
Resolves #2217


aklsh:
whoops sorry, marked it as ready by mistake.

aklsh:
can someone with access to the private ci let me know what the fails are?

rtl/ibex_register_file_common.sv
```diff
@@ -11,7 +11,7 @@
 module ibex_register_file_common #(
     parameter bit          FPGA          = 0,
     parameter int unsigned AddrWidth     = 5,
-    parameter int unsigned NumWords      = 5,
+    parameter int unsigned NumWords      = 2 ** AddrWidth,
```
marnovandermaas, line 15: Can you include this change in the previous commit?

rtl/ibex_register_file_ff.sv
marnovandermaas, line 1: Please avoid making unnecessary white space changes.

marnovandermaas (commented):
Thanks for your interest in the Ibex repository. In its current state this PR is a bit hard to review. First, please avoid making lots of white space changes. And if you do want to make a whole load of white space changes please separate them from the commits that have functional changes. Would you mind adding a comment in the common register file to say what functionality you've pulled out. And it would also be nice if you could put a short description in the pull request itself to describe what you've done and how what you've done still has the same functionality as before.

aklsh:
I've tried resolving all the whitespace changes I had introduced unintentionally @marnovandermaas 

> And it would also be nice if you could put a short description in the pull request itself to describe what you've done and how what you've done still has the same functionality as before.

#2217 was opened to bring together common logic present in each of the three regfile implementations into one common module which can then be instantiated in each of the three implementations.

This PR mainly combines the security hardening logic that checks for onehot encoding in various signals for glitches in addresses and write enable(s). It tries to maintain the same logic as before wrt parameterization of these checks.

marnovandermaas (changes requested):
Thanks for the clean up and the explanation. At the moment this does not build. An extract from CI:
```
xmvlog: *E,ILLMIX (/home/runner/_work/lowrisc-private-ci/lowrisc-private-ci/rtl/ibex_register_file_common.sv,79|4): ordered and named parameter list mixed syntax [7.2.3(AMSLRM)].
    ) u_prim_onehot_check (
    |
xmvlog: *E,ILLPRI (/home/runner/_work/lowrisc-private-ci/lowrisc-private-ci/rtl/ibex_register_file_common.sv,79|4): illegal expression primary [4.2(IEEE)].
package dv_utils_pkg;
```
Have you tested this PR yourself?

PR #2267, opened 2025-04-22
[ibex] Pass mvendorid and mimpid as top-level params

Razer6:
Instead of using default values from a package, create a top-level parameter to define these and pass them down. This allows integrators to specify them on a per-instance basis.

### DIFF
--- a/rtl/ibex_top.sv
+++ b/rtl/ibex_top.sv
@@ -43,7 +43,16 @@ module ibex_top import ibex_pkg::*; #(
   parameter int unsigned            DmExceptionAddr              = 32'h1A110808,
   // Default seed and nonce for scrambling
   parameter logic [SCRAMBLE_KEY_W-1:0]   RndCnstIbexKey          = RndCnstIbexKeyDefault,
-  parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonce        = RndCnstIbexNonceDefault
+  parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonce        = RndCnstIbexNonceDefault,
+  // mvendorid: encoding of manufacturer/provider
+  // 0 indicates this field is not implemented. Ibex implementors may wish to set their
+  // own JEDEC ID here.
+  parameter logic [31:0]            CsrMvendorId                 = 32'b0,
+  // mimpid: encoding of processor implementation version
+  // 0 indicates this field is not implemented. Ibex implementors may wish to indicate an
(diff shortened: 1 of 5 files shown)


davidschrammel (approved):
LGTM

SamuelRiedel:
This looks good to me, thanks.

I looked into the CI failure and can't see how it relates to your changes, so I assume it’s a quirk of our current (older) Verilator version. I created a PR addressing this failure here: https://github.com/lowRISC/ibex/pull/2270. If that gets merged, CI should be fixed and this PR should be good to go, IMO. If we decide not to implement the fix, we could disable the linter on the specific line where Verilator is currently complaining.

rtl/ibex_top.sv
```diff
@@ -43,7 +43,16 @@ module ibex_top import ibex_pkg::*; #(
   parameter int unsigned            DmExceptionAddr              = 32'h1A110808,
   // Default seed and nonce for scrambling
   parameter logic [SCRAMBLE_KEY_W-1:0]   RndCnstIbexKey          = RndCnstIbexKeyDefault,
-  parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonce        = RndCnstIbexNonceDefault
+  parameter logic [SCRAMBLE_NONCE_W-1:0] RndCnstIbexNonce        = RndCnstIbexNonceDefault,
+  // mvendorid: encoding of manufacturer/provider
+  // No JEDEC ID has been allocated to lowRISC so the value is 0 to indicate the field is not
+  // implemented
```
andreaskurth, line 49: This is no longer true, as lowRISC now has a JEDEC ID (0xEF in bank 13). We could default to that value, but that would be a separate change IMO. For this PR, I suggest we just replace this with
```suggestion
// 0 indicates this field is not implemented. Ibex implementors may wish to set their own JEDEC ID here.
```

rtl/ibex_top.sv
```diff
@@ -43,7 +43,16 @@ module ibex_top import ibex_pkg::*; #(
...
+  parameter logic [31:0]            CsrMvendorId                 = 32'b0,
+  // mimpid: encoding of processor implementation version
+  // 0 indicates this field is not implemeted. Ibex implementors may wish to indicate an RTL/netlist
```
andreaskurth, line 52:
```suggestion
  // 0 indicates this field is not implemented. Ibex implementors may wish to indicate an RTL/netlist
```

andreaskurth (commented):
Two small comments, otherwise LGTM

andreaskurth (approved):

PR #2269, opened 2025-04-23
[rtl] Minor tweak to decoder to avoid dead code

### LINKED ISSUE #2268: [bug][ibex_decoder] Report a bug in Ibex
Hi all!

I am using a static analysis to found a various bugs in open source RTL designs.

It was founded that in `ibex_decoder.sv` RTL file dead code is exists [here](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_decoder.sv#L439).

This assignment under `else` statement is never executed due to [this](https://github.com/lowRISC/ibex/blob/master/rtl/ibex_decoder.sv#L406) external `else` branch with `!instr[26]` condition.

rswarbrick:
The "else" part of the if/else check here wasn't possible because the surrounding else branch (starting at line 406) is already in the case where instr[26] is zero.

Fixes #2268.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -433,11 +433,10 @@ module ibex_decoder #(
                   end
                 end
                 5'b0_0001: begin
-                  if (instr[26] == 1'b0) begin                                        // unshfl
-                    illegal_insn = (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) ? 1'b0 : 1'b1;
-                  end else begin
-                    illegal_insn = 1'b1;
-                  end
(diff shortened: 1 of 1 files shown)


marnovandermaas (commented):
This makes sense to me. Is it worth plopping an assertion that instr[26] must be zero if the code gets here?

rswarbrick:
I don't *think* so, since it's visible "syntactically". And we don't really have procedural assertions in our RTL code, I don't think. And I don't think I know a clever way to fake it for this sort of thing?

andreaskurth (approved):
LGTM, thx @rswarbrick

PR #1250, opened 2021-01-25
[verilator] Remove clock gating waiver and add extra warning

rswarbrick:
The `-Wwarn-IMPERFECTSCH` flag is suggested in the Verilator docs (to
help spot if we get a clock gating hint wrong). The waiver is no
longer needed because we now add the relevant hint in the
RTL (vendored in `prim_generic_clock_gating.sv` from OpenTitan).

Closes #212.

PR #2270, opened 2025-04-24
Disable Verilator's IMPERFECTSCH warnings

SamuelRiedel:
Our current version of Verilator issues "IMPERFECTSCH" warnings, which sometimes appear spurious. According to the [manual](https://verilator.org/guide/latest/warnings.html#cmdoption-arg-IMPERFECTSCH) this warning is disabled by default, but we explicitly enabled it a while back in https://github.com/lowRISC/ibex/pull/1250 and created some RTL workarounds to avoid it in https://github.com/lowRISC/ibex/pull/2129.

Now, the same warning has reappeared at the same point due to the changes in https://github.com/lowRISC/ibex/pull/2267. However, this PR does not touch the code that leads to the warning, which is why I propose removing the warning altogether now and updating Verilator in the near future. Alternatively, we could disable it for the specific lines where it appears in https://github.com/lowRISC/ibex/pull/2267 as a hotfix.

This PR disables the warning in Verilator and reverts the workarounds introduced in https://github.com/lowRISC/ibex/pull/2129.

For reference, this is the warning triggered by https://github.com/lowRISC/ibex/pull/2267:
```sh
INFO: verilator -f lowrisc_ibex_ibex_simple_system_cosim_0.vc --trace --trace-fst --trace-structs --trace-params --trace-max-array 1024 -CFLAGS "-std=c++14 -Wall -DVL_USER_STOP -DVM_TRACE_FMT_FST -DTOPLEVEL_NAME=ibex_simple_system -g `pkg-config --cflags riscv-riscv riscv-disasm riscv-fdt`" -LDFLAGS "-pthread -lutil -lelf `pkg-config --libs riscv-riscv riscv-disasm riscv-fdt`" -Wall -Wwarn-IMPERFECTSCH --unroll-count 72

ERROR: %Warning-IMPERFECTSCH: ../src/lowrisc_ibex_ibex_top_0.1/rtl/ibex_lockstep.sv:131:45: Imperfect scheduling of variable: 'ibex_simple_system.u_top.u_ibex_top.gen_lockstep.u_ibex_lockstep.rst_shadow_set_single_bit'
                                                                                   : ... In instance ibex_simple_system
  131 |   logic                       rst_shadow_n, rst_shadow_set_single_bit;
      |                                             ^~~~~~~~~~~~~~~~~~~~~~~~~
                       ... For warning description see https://verilator.org/warn/IMPERFECTSCH?v=4.210
                       ... Use "/* verilator lint_off IMPERFECTSCH */" and lint_on around source to disable this message.
```

### DIFF
--- a/dv/verilator/simple_system_cosim/ibex_simple_system_cosim.core
+++ b/dv/verilator/simple_system_cosim/ibex_simple_system_cosim.core
@@ -182,7 +182,6 @@ targets:
           - '-CFLAGS "-std=c++14 -Wall -DVL_USER_STOP -DVM_TRACE_FMT_FST -DTOPLEVEL_NAME=ibex_simple_system -g `pkg-config --cflags riscv-riscv riscv-disasm riscv-fdt`"'
           - '-LDFLAGS "-pthread -lutil -lelf `pkg-config --libs riscv-riscv riscv-disasm riscv-fdt`"'
           - "-Wall"
-          - "-Wwarn-IMPERFECTSCH"
           # RAM primitives wider than 64bit (required for ECC) fail to build in
           # Verilator without increasing the unroll count (see Verilator#1266)
           - "--unroll-count 72"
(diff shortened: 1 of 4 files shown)


marnovandermaas (commented):
I think this makes sense. To be honest I hadn't realised that this warning was no longer used in version 5.

andreaskurth (approved):
Thanks @SamuelRiedel, this LGTM

nasahlpa (approved):
LGTM, makes sense. Thanks for also removing the signal that only was introduced to suppress the linting issue.

PR #2272, opened 2025-05-08
[dv/formal] Performance improvement for divide PC

marnovandermaas:
The MType_{Div,DivU,Rem,RemU}_PC properties were proving very slow before this. They were proving with Hp 45, which was very slow. This change groups them together so that the engines can help each other out, the N engine helps with proving these properties much faster than with Hp alone. In the long-term it is better to capture a relavant assumption in the proof files instead of putting this in the TCL script.


hcallahan-lowrisc (approved):

marnovandermaas:
Just ran the whole proof on a big Hetzner machine and it proved in less than one hour and twenty minutes.

PR #2275, opened 2025-05-14
[simple_system] Use correct RAM base address and size

alees24:
Correct the base address and size of the RAM memory model within the simulator environment. Previously the error in the RAM size was masking the incorrect base address (zero) and it simulated as intended but could fail if the base address and/or RAM size were modified.


marnovandermaas (commented):
Looks reasonable to me, just waiting for CI to run after fixing clang lint.

marnovandermaas (approved):

PR #2276, opened 2025-05-16
[dv/formal] Lint cleanup

marnovandermaas:
While going through the formal flow I resolved a few warnings that Jasper complains about. This PR doesn't resolve everything, but makes a start.

Besides a few simple fixes it mainly does a rearranging of the protocol/mem.sv, peek/mem.sv and peek/alt_lsu.sv files.

### DIFF
--- a/dv/formal/check/top.sv
+++ b/dv/formal/check/top.sv
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /*
-This is the top module. Everything else in check/* will be included into this file.
+This is the top module. Everything else in the "check" directory is included in this file.
 This module contains the ibex instance, the specification instance and the assertions
 (included via the built psgen file), most of the non assertion code is setting up
 infrastructure for those assertions.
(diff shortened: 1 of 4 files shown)


dv/formal/check/top.sv
```diff
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /*
-This is the top module. Everything else in check/* will be included into this file.
+This is the top module. Everything else in the "check" directory is be included into this file.
```
rswarbrick, line 8: Nit: "is be" ??

rswarbrick (approved):
This looks good to me (modulo the grammar nit!)

I notice that there's an unchanged file with a check annotation. Would you mind pushing a parallel PR that adds the newline needed?

mndstrmr (approved):
This looks great! It looks much cleaner. I had no idea I had so much dead stuff.

dv/formal/check/top.sv
marnovandermaas: Thanks for the careful review!

PR #2277, opened 2025-05-17
Environment Setup and Makefile Target Errors

WajahatRiaz:
I cloned the repository and followed the steps outlined in the README.md. To verify functionality, I ran the following Makefile targets. However, I encountered errors in both cases. Screenshots of the corresponding error logs are attached below.
 
**Target/Issue No. 1: make build-simple-system**

When I run `make build-simple-system` I get the following error:

image
 
**Target/Issue No. 2: make build-csr-test**

When I run `make build-csr-test` I get the following error: 

image

**Environment Details**
- FuseSoC version: 0.1
- Verilator version: 5.035 (development version: v5.034-27-g258becd17)
- RISC-V Toolchain version: 13.2.0
- Operating System: Ubuntu 24.10 (Codename: oracular)

Please let me know if any additional setup steps are required or if the errors are known issues with the current branch/version. I'm happy to help troubleshoot further.

### DIFF
--- a/dv/cs_registers/Makefile
+++ b/dv/cs_registers/Makefile
@@ -23,11 +23,21 @@ LDFLAGS  = -shared
 CC       = $(CXX)
 
 # Add svdpi include
-TOOLDIR = $(subst bin/$(TOOL),,$(shell which $(TOOL)))
 ifeq ($(TOOL),vcs)
+  # For VCS, add include path
+  TOOLDIR = $(subst bin/$(TOOL),,$(shell which $(TOOL)))
   INCLUDES += -I$(TOOLDIR)include
 else
-  INCLUDES += -I$(TOOLDIR)share/verilator/include/vltstd
+  # For Verilator, use verilator command to find root directory
+  ifeq ($(shell which verilator),)
+    $(error "Verilator not found in PATH. Please install Verilator or add it to your PATH")
+  endif
+  VERILATOR_ROOT = $(shell verilator --getenv VERILATOR_ROOT 2>/dev/null)
+  ifeq ($(VERILATOR_ROOT),)
+    $(error "Could not determine VERILATOR_ROOT. Please make sure Verilator is properly installed")
+  endif
+  # Add both main include directory and vltstd directory where svdpi.h is located
+  INCLUDES += -I$(VERILATOR_ROOT)/include -I$(VERILATOR_ROOT)/include/vltstd
 endif
 
 .PHONY: all clean
--- a/dv/cs_registers/Makefile
+++ b/dv/cs_registers/Makefile
@@ -23,21 +23,41 @@ LDFLAGS  = -shared
 CC       = $(CXX)
 
 # Add svdpi include
-ifeq ($(TOOL),vcs)
-  # For VCS, add include path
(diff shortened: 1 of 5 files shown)


rswarbrick:
Something seems rather odd about the first commit. I think that GCC defaults to `fexceptions` (see [here](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html)). Is there something interesting about your environment that turns this off?

WajahatRiaz:
Well I built verilator from their GitHub repository. Would that be causing this issue?

WajahatRiaz:
@rswarbrick can you test things out using the environment details I provided?

examples/sw/simple_system/common/common.mk
```diff
@@ -19,9 +19,9 @@ SRCS = $(COMMON_SRCS) $(PROGRAM_C) $(EXTRA_SRCS)
 C_SRCS = $(filter %.c, $(SRCS))
 ASM_SRCS = $(filter %.S, $(SRCS))
 
-CC = riscv32-unknown-elf-gcc
+CC = /home/wajahat/RISCV/riscv64-unknown-elf-toolchain/riscv/bin/riscv64-unknown-elf-gcc
```
marnovandermaas: You shouldn't have to add an absolute path here if you add your toolchain to your path. Also you should use riscv32 instead of riscv64. Please use the lowRISC toolchain: https://github.com/lowRISC/lowrisc-toolchains/releases

marnovandermaas (commented):
Interesting that you need to add these include directories relative to verilator root. I appreciate you putting the effort into getting the Verilator root directly from Verilator.

examples/sw/simple_system/common/common.mk
marnovandermaas: I see you update this in a later commit but maybe drop the commit where you add it in the first place.

WajahatRiaz: I see, I will squash my commits.

WajahatRiaz: > You shouldn't have to add an absolute path here if you add your toolchain to your path. Also you should use riscv32 instead of riscv64. Please use the lowRISC toolchain: https://github.com/lowRISC/lowrisc-toolchains/releases

Let me test this out using the toolchain releases by lowRISC.

WajahatRiaz:
@marnovandermaas @rswarbrick Here you go, added appropriate changes. Please review.

dv/cs_registers/Makefile
```diff
@@ -35,9 +35,13 @@ else
...
     $(TOOLDIR)share/verilator/include/vltstd \
     /usr/share/verilator/include \
+    /usr/share/verilator/include/vltstd \
     /usr/local/share/verilator/include \
+    /usr/local/share/verilator/include/vltstd \
     /opt/verilator/share/verilator/include \
-    $(HOME)/verilator/include
+    /opt/verilator/share/verilator/include/vltstd \
+    $(HOME)/verilator/include \
+    $(HOME)/verilator/include/vltstd
```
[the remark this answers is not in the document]
marnovandermaas: If that's the problem I'm on board with adding the "vlstd" path to the Verilator root, but I don't think it's a good idea to add all the possible locations where Verilator could be installed.

PR #2279, opened 2025-05-21
[fcov] Add fcov for the interaction of pmp with debug module accesses

hcallahan-lowrisc:
Accesses to the debug module in debug mode should never be denied by the PMP unit. This commit implements fcov to confirm we have stimulated this particular behaviour in relevant related states.

Illegal bins are used for incorrect behaviour (e.g. denied access in debug mode). Other behaviours such as debug module accesses outside of debug mode are left as ignore_bins for now. This is not explicitly disallowed by the specification, and our implementation does not have any opinion about its validity, but external debug modules opine that it should not be allowed. We could possibly expand the stimulus in the future to test this condition, but it is low priority.

This PR does not include stimulus to hit the new coverage bins. That will come in another request.

Goes towards #2233

### DIFF
--- a/dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
+++ b/dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
@@ -10,7 +10,10 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
     // 0 = No restriction, 1 = 8 byte, 2 = 16 byte, 3 = 32 byte, etc.
     parameter int unsigned PMPGranularity = 0,
     // Number of implemented regions
-    parameter int unsigned PMPNumRegions  = 4
+    parameter int unsigned PMPNumRegions  = 4,
+    // Debug Module Address Space
+    parameter int unsigned DmBaseAddr     = 32'h1A110000,
+    parameter int unsigned DmAddrMask     = 32'h00000FFF
 ) (
   input clk_i,
   input rst_ni,
@@ -19,6 +22,8 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   input logic[33:0]          csr_pmp_addr    [PMPNumRegions],
   input logic                pmp_req_err     [3],
   input pmp_mseccfg_t        csr_pmp_mseccfg,
+
+  input logic debug_mode,
 
   input logic data_req_out,
 
@@ -97,6 +102,8 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
   end
 
   if (PMPEnable) begin : g_pmp_cgs
+    logic [PMPNumChan-1:0] fcov_access_attempted_into_dm;
+
     logic [PMPNumRegions-1:0] pmp_iside_match;
     logic [PMPNumRegions-1:0] pmp_iside2_match;
     logic [PMPNumRegions-1:0] pmp_dside_match;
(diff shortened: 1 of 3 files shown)


dv/uvm/core_ibex/fcov/core_ibex_pmp_fcov_if.sv
```diff
@@ -709,6 +722,108 @@ interface core_ibex_pmp_fcov_if import ibex_pkg::*; #(
...
+      cp_ls_pmp_exception: coverpoint load_store_unit_i.fcov_ls_pmp_exception;
+
+      // Cross access attempts with the result of the access check
+
+      dm_fetch_access_iside_cross: cross
+        dm_fetch_iside_debug_mode_cp,
+        pmp_iside_req_err,
+        pmp_iside_access_fault_check
+      {
+        // Fetches should never fail the access check in debug mode
```
rswarbrick, line 767: Maybe leave a note about why we are certain we'll see an error if this happens for some reason.

rtl/ibex_core.sv
```diff
@@ -1944,6 +1944,20 @@ module ibex_core import ibex_pkg::*; #(
           g_pmp.pmp_i.region_match_all[PMP_D][i_region];
       end
     end
+
+    // Create a signal 'fcov_access_attempted_into_dm' for each PMP channel, which becomes true at
+    // the point when a memory access into the debug_module address space via that channel is
+    // accesses-checked by it's PMP unit. If true we know that the current instruction at least
```
rswarbrick, line 1950: Nit: I think you want "its" here.

rtl/ibex_core.sv
```diff
@@ -1944,6 +1944,20 @@ module ibex_core import ibex_pkg::*; #(
...
+    // attempted to make an access, and we can then cross this signal with the result of the pmp
+    // check and the current debug_mode state to capture all appropriate coverage.
+    logic [PMPNumChan-1:0] access_check_into_dm;
```
rswarbrick, line 1953: Can we bind this in with an interface to avoid touching the RTL? (I think you'll want an "upwards hierarchical reference")

rtl/ibex_pmp.sv
```diff
@@ -241,14 +242,14 @@ module ibex_pmp #(
...
-    // No error is raised if the access is allowed as Debug Module access (first term).
-    assign pmp_req_err_o[c] = ~debug_mode_allowed_access[c] &
-                              access_fault_check(csr_pmp_mseccfg_i.mmwp,
-                                                 csr_pmp_mseccfg_i.mml,
-                                                 pmp_req_type_i[c],
-                                                 region_match_all[c],
-                                                 priv_mode_i[c],
-                                                 region_perm_check[c]);
+    // No error is raised if the access is allowed via the Debug Module in Debug Mode exception.
+    assign access_fault_check_res[c] = access_fault_check(csr_pmp_mseccfg_i.mmwp,
```
rswarbrick, line 246: This makes sense to me (making a big load of combinatorial logic less atomic): maybe it's worth splitting the change into a separate commit?

andreaskurth: Yes please (to splitting this RTL change into a separate commit) :-)

rtl/ibex_core.sv
hcallahan-lowrisc: I moved the new helper logic into the existing pmp_fcov_if. So no change needed to the existing ibex_core rtl.
By this logic though, there is a quantity of existing code in `rtl/` that uses `DV_FCOV_SIGNAL` and `DV_FCOV_SIGNAL_GEN_IF` that should get the boot as well in a future cleanup PR.

rswarbrick: I agree: sounds sensible to me.

rswarbrick (approved):
Thanks for this: it looks nice and clean to me!

PR #2282, opened 2025-05-28
[ibex] Pass mvendorid and mimpid also to lockstep core

Razer6:
Otherwise an alert is raised if a core has a different configuration than zero. There, the primary core would return the non-zero value but the lockstep core zero. This difference causes an alert.

This PR fixes this bug and passes the mvendirid and mimpid parameters also to the lockstep core so that they are in sync


andreaskurth:
There are currently problems with the Ubuntu package repository, causing merge queue checks to fail. Let's retry later today or early tomorrow

PR #2283, opened 2025-05-30
[ci] Use HTTPS for package repostories

andreaskurth:


ci/install-build-deps.sh
```diff
@@ -25,6 +25,9 @@ if [ -z "$GITHUB_ACTIONS" ]; then
   GITHUB_PATH=/dev/null
 fi
 
+# Use non-default mirror for Ubuntu packages, because the default mirror currently have problems.
+$SUDO_CMD perl -pi -e 's!http://(archive|security).ubuntu.com!http://europe-west2.gce.archive.ubuntu.com!g' /etc/apt/sources.list
```
nbdd0121, line 29: Perhaps use sed instead perl?

andreaskurth: Done

vogelpi (approved):

PR #2284, opened 2025-06-02
MemInitFile parameter has string type.

marnovandermaas:
Adding string types where they are expected for memory initialisation files.


Razer6 (changes requested):
`string` should not be used in synthesizable code. There are many PD tools that cannot deal with the explicit `string` type. Even Ascentlint flags this as a listing error `SYN_TYPE`.

There were attempts also make similar changes in OpenTitan such as in https://github.com/lowRISC/opentitan/pull/26584

PR #2285, opened 2025-06-09
[syn] Add keep_hierarchy constraints to prim_generic

nasahlpa:
Some FI countermeasures in Ibex use redundancy to detect fault attacks (e.g., the onehot encoding & checker in the RF). As synthesis tools are great in detecting redundant logic and reducing it, this commit puts a keep_hierarchy synthesis constraint on prim_generic* modules. This is exactely the purpose of the prim_generic_buf, prim_generic_flop, and prim_generic_and2 modules - having synthesis barriers to avoid optimizations.


syn/tcl/yosys_run_synth.tcl
```diff
@@ -30,6 +30,11 @@ yosys "chparam -set RV32M $lr_synth_ibex_multiplier $lr_synth_top_module"
 
 yosys "chparam -set RegFile $lr_synth_ibex_regfile $lr_synth_top_module"
 
+# Place keep_hierarchy contraints on relevant modules to prevent aggressive
+# synthesis optimzations across the boundaries of these modules.
+yosys "hierarchy -check -top $lr_synth_top_module"
+yosys "setattr -mod -set keep_hierarchy 1 *prim_generic*"
```
nasahlpa: @vogelpi do we want to make this explicit by having:
```suggestion
yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_and2*"
yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_buf*"
yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_clock_mux2*"
yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_flop*"
```
instead? Only those four `prim_generics` are used in the `SecureIbex` configuration. 

The advantage of using `*prim_generic*` is that we don't accidentally forget a constraint.

marnovandermaas (commented):
Thanks! Out of curiosity how much does the synthesis area go up by after this change?

syn/tcl/yosys_run_synth.tcl
nasahlpa:
```suggestion
if { $lr_synth_ibex_secure_ibex } {
  yosys "hierarchy -check -top $lr_synth_top_module"
  yosys "setattr -mod -set keep_hierarchy 1 *prim_generic*"
}
```

We also only could use these constraints when we are using the `SecureIbex` configuration. However, currently, the synthesis flow does not support the `SecureIbex` config as some prims are missing and the `lr_synth_ibex_secure_ibex` does not yet exist.

nasahlpa:
> Thanks! Out of curiosity how much does the synthesis area go up by after this change?

From 26.68kGE to 52.98kGE.

As suggested in [this](https://github.com/lowRISC/ibex/pull/2285#discussion_r2135403297) comment, it could make sense to only use this synthesis constraint in the `SecureIbex` config.

syn/tcl/yosys_run_synth.tcl
nasahlpa: Done, implemented.

vogelpi: I think this would be better. The clock mux should be constrained independent of the parameter btw.

vogelpi: Great!

nasahlpa: Done.

You mean
```tcl
yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_clock_mux2*"
```
should also be set for the non-`SecureIbex` version? `clock_mux2` is not used in this Ibex configuration so constraining does not change anything.

vogelpi:
> > Thanks! Out of curiosity how much does the synthesis area go up by after this change?
> 
> From 26.68kGE to 52.98kGE.
> 
> As suggested in [this](https://github.com/lowRISC/ibex/pull/2285#discussion_r2135403297) comment, it could make sense to only use this synthesis constraint in the `SecureIbex` config.

Which configuration did you synthesize and did you enable flattening?

syn/tcl/yosys_run_synth.tcl
```diff
@@ -30,6 +30,17 @@ yosys "chparam -set RV32M $lr_synth_ibex_multiplier $lr_synth_top_module"
...
+  yosys "chparam -set SecureIbex 1 $lr_synth_top_module"
+  # Place keep_hierarchy contraints on relevant modules to prevent aggressive
+  # synthesis optimzations across the boundaries of these modules.
+  yosys "hierarchy -check -top $lr_synth_top_module"
+  yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_and2*"
+  yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_buf*"
+  yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_clock_mux2*"
+  yosys "setattr -mod -set keep_hierarchy 1 *prim_generic_flop*"
+}
+
```
vogelpi, line 43: I think you should again remove these constraints before the very final flattening step, no? Can you compare with the AES setup please?

vogelpi (commented):
Thanks for your work on this, this is important to study Yosys netlists.

syn/tcl/yosys_run_synth.tcl
vogelpi: Yes exactly

syn/tcl/yosys_run_synth.tcl
nasahlpa: Thanks for spotting this, done!

syn/tcl/yosys_run_synth.tcl
vogelpi: Sorry, I misunderstood this. We've now discussed in person: the `clock_mux2` is only used in the SecureIbex configuration. It's thus fine to only apply the constraint for this configuration.

syn/syn_yosys.sh
```diff
@@ -47,9 +56,15 @@ for file in "${LR_DEP_SOURCES[@]}"; do
...
         --define=SYNTHESIS --define=YOSYS \
+        ../vendor/lowrisc_ip/ip/prim/rtl/prim_count_pkg.sv \
+        ../vendor/lowrisc_ip/ip/prim/rtl/prim_cipher_pkg.sv \
         -I../vendor/lowrisc_ip/ip/prim/rtl \
         "$file" \
         > "$LR_SYNTH_OUT_DIR"/generated/"${module}".v
+    # Make sure auto-generated primitives are resolved to generic primitives
+    # where available.
+    sed -i 's/prim_and2/prim_generic_and2/g' "$LR_SYNTH_OUT_DIR"/generated/"${module}".v
+    sed -i 's/prim_flop/prim_generic_flop/g' "$LR_SYNTH_OUT_DIR"/generated/"${module}".v
```
vogelpi: Would it be easier to just have all 4 sed commands here and below? I think this would reduce the risk of spending much time when debugging an issue when a particalur sed command is added in the wrong place.

nasahlpa: Done, I have placed the four sed commands here and below.

vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #2286, opened 2025-06-19
Fix `dsim` support

akashlevy:
This PR fixes up Ibex testbench support for `dsim` simulator. Several things had to be fixed:

- Simulator compatibility: `dsim` requires that enum values must use enum constants
    - Use `ZERO` enum value instead of `'0` and `0` for instr enums
    - Protect Ibex-specific CSRs with  \`ifndef DSIM to ensure these are not included in `implemented_csr[]`. `dsim` requires that all array values be enumerated in `privileged_reg_t`, which they are currently not
- Simulator compatibility: `dsim` does not appear to respect "re-exports"
    - Replace all re-exported `ibex_mem_intf_agent_pkg::ADDR_WIDTH` and  `ibex_mem_intf_agent_pkg::DATA_WIDTH` with `ibex_mem_intf_pkg::ADDR_WIDTH` and  `ibex_mem_intf_pkg::DATA_WIDTH`
- Script fix: Add `dsim` case for `reloc_word` in [dv/uvm/core_ibex/scripts/run_instr_gen.py](https://github.com/lowRISC/ibex/compare/master...Silimate:ibex:dsim_fixes?expand=1#diff-5415dd86d435fd7155622795acb1fb1c440b6e2e595c0c1a857b891fdc6d9e65)
- Script fix: Remove extra quote in [dv/uvm/core_ibex/yaml/rtl_simulation.yaml](https://github.com/lowRISC/ibex/compare/master...Silimate:ibex:dsim_fixes?expand=1#diff-01ac65450dedc1df3421dee3e0874335b0bb15d6b8db2035aef4b0d77af903cd)
- Script fix: Replace `dsim_compile_opts` with `dsim_opts` in [util/ibex_config.py](https://github.com/lowRISC/ibex/compare/master...Silimate:ibex:dsim_fixes?expand=1#diff-0a9f6797b07226a176327f28cfe58b039f1eb20d18e7bdee723ec94668f3be01)

### DIFF
--- a/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
+++ b/dv/uvm/core_ibex/yaml/rtl_simulation.yaml
@@ -143,7 +143,7 @@
           +acc+rwb
           -f <core_ibex>/ibex_dv.f
           -l <tb_build_log>
-          -suppress EnumMustBePositive"
+          -suppress EnumMustBePositive
   sim:
     cmd:
       - >-
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
@@ -79,7 +79,7 @@ class ibex_rand_mseccfg_stream extends riscv_directed_instr_stream;
     csrrw_instr = riscv_instr::get_instr(CSRRWI);
     csrrw_instr.atomic = 1'b0;
     csrrw_instr.csr = MSECCFG;
-    csrrw_instr.rd = '0;
+    csrrw_instr.rd = ZERO;
     // Randomize between 3'b000 and 3'b111 to hit every combination of RLB/MMWP/MML bits.
     csrrw_instr.imm_str = $sformatf("0x%0x", $urandom_range(7,0));
     instr_list = {csrrw_instr};
@@ -159,7 +159,7 @@ class ibex_rand_cpuctrlsts_stream extends riscv_directed_instr_stream;
     instrs[3] = riscv_instr::get_instr(CSRRW);
     instrs[3].atomic = 1'b0;
     instrs[3].csr = 12'h7c0;
-    instrs[3].rd = '0;
+    instrs[3].rd = ZERO;
     instrs[3].rs1 = cfg.gpr[0];
 
     instr_list = instrs;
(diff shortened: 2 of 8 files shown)


hcallahan-lowrisc:
Hey Akash, thanks for the PR getting dsim support going again! This is looking pretty good. Please mark as ready for review when you're happy with the state of things. If there is still some minor breakages, it would be better to get the bulk of things merged and then follow up with some smaller PRs, just to keep this changeset from getting too large.

akashlevy:
Hey @hcallahan-lowrisc, this is ready for review. Most of the dsim tests pass, so I think this can be reviewed and merged as a starting point as I root cause the remaining failures.

dv/uvm/core_ibex/yaml/rtl_simulation.yaml
```diff
@@ -142,11 +142,21 @@
...
           -f <core_ibex>/ibex_dv_defines.f
           -f <core_ibex>/ibex_dv.f
           -l <tb_build_log>
-          -suppress EnumMustBePositive"
+          -suppress EnumMustBePositive
+          <cmp_opts>
+          <cosim_opts>
+      # dsim does not link image.so with ISS_LDFLAGS and ISS_LIBS, so symbols are missing
+      # so we need to re-link image.so with the linker flags and libraries
+      - gcc -shared -Bsymbolic -o <tb_dir>/image.so @<tb_dir>/obj/objfiles <ISS_LDFLAGS> <ISS_LIBS> -lstdc++
```
hcallahan-lowrisc, line 155: Ah that's unfortunate it needs to be done as a seperate step. But this seems pretty clean. Thanks

akashlevy:
Could we potentially just squash this to one commit?

dv/uvm/core_ibex/yaml/rtl_simulation.yaml
akashlevy: Yes definitely unfortunate. I tried everything to get `dsim` to link it, but it doesn't seem possible

akashlevy:
Also I have a nice Dockerfile setup for dsim+ibex if you'd like to speed up the process of testing out the dsim changes. Happy to share that if helpful

akashlevy:
Continuing this in #2297, which is ready for review.

PR #2287, opened 2025-06-21
Change '/bin/bash' shebangs to '/usr/bin/env bash'

hcallahan-lowrisc:
This improves portability across different unix-like operating systems by using bash from the PATH, instead of bash from a hardcoded location.


SamuelRiedel (approved):
Good point. LGTM.

PR #2290, opened 2025-06-26
Fix various typos

elliotb-lowrisc:
Attempt at neutral correction of what I hope are universal seen as spelling and capitalisation errors in some documentation and code comments.

Please flag any accidental changes to something that was already correctly spelt. My spell-checker is set to British, not American English, so there's a chance I accidentally "fixed" a spelling I didn't recognise as being the American variant.


elliotb-lowrisc:
Revert "practiced" -> "practised" change that turned out to be American spelling

dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
```diff
@@ -112,7 +112,7 @@ class ibex_asm_program_gen extends riscv_asm_program_gen;
   endfunction
 
   // Re-define gen_test_done() to override the base-class with an empty implementation.
-  // Then, our own overrided gen_program() can append new test_done code.
+  // Then, our own override gen_program() can append new test_done code.
```
AlexJones0: I don't think "override" is correct here. "override" is a verb or noun, of which neither really make sense here, since we are using it as an adjective. "overrided" is not a word but could be considered semantically as either "overridden" or "overriding". I would suggest "overriding" is the correct choice here based on context.

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
```diff
@@ -51,7 +51,7 @@ class ibex_breakpoint_stream extends riscv_directed_instr_stream;
     la_instr.rd                = cfg.gpr[1];
 
     // Create the ebreak insn which will cause us to enter debug mode, and run the
-    // special code in the debugrom.
+    // special code in the debug ROM.
```
AlexJones0, line 54: This change is fine, but based on the context of 352f83fc747c65120f2df2d8a9b1f10b8f2eb76c it might be more appropriate as "debug_rom" for consistency?

rtl/ibex_alu.sv
```diff
@@ -863,7 +863,7 @@ module ibex_alu #(
       // where P denotes lower 32 bits of the corresponding CRC polynomial, rev(a) the bit reversal
       // of a, n = 8,16, or 32 for .b, .h, .w -variants. {a, b} denotes bit concatenation.
       //
-      // Using barret reduction, one can show that
+      // Using Barret reduction, one can show that
```
AlexJones0:
```suggestion
      // Using Barrett reduction, one can show that
```

rtl/ibex_if_stage.sv
```diff
@@ -541,7 +541,7 @@ module ibex_if_stage import ibex_pkg::*; #(
     logic        prev_instr_seq_q, prev_instr_seq_d;
 
     // Do not check for sequential increase after a branch, jump, exception, interrupt or debug
-    // request, all of which will set branch_req. Also do not check after reset or for dummys.
+    // request, all of which will set branch_req. Also do not check after reset or for dummies.
```
AlexJones0: Nit: my personal preference would be to pluralize precisely as "dummy instructions", not "dummies".

dv/uvm/core_ibex/riscv_dv_extension/ibex_directed_instr_lib.sv
elliotb-lowrisc: Hmm, I think "debug ROM" is better for comments and documentation referring to the ROM itself and not a `debug_rom` class. This approach can already be found here: https://github.com/lowRISC/ibex/blob/f0d408a546186d455ba6cec1dda84f15cecc3627/dv/uvm/core_ibex/riscv_dv_extension/ibex_debug_triggers_overrides.sv#L113-L121
Is there a particular reason you prefer "debug_rom" here?

AlexJones0: Just for consistency, but I hold no strong opinion either way. I think what you've said sounds fine.

elliotb-lowrisc: Alright, I'll keep the current change, if you have no strong opinion

PR #2299, opened 2025-07-21
Support testbench elaboration with Verific

akashlevy:
This PR adds support for testbench elaboration using [Verific](https://www.verific.com/), which is the backbone of many EDA tools that do static analysis on SystemVerilog.

Verific is a bit more strict than other parsers about DV includes. It also does not support functional coverage binding. I've added the necessary includes and protection around functional coverage binding assigns.

Let me know if any changes are needed!


hcallahan-lowrisc (approved):
Makes sense. Thanks!

PR #2300, opened 2025-07-22
Use +define instead of -define to fix unknown option compiling with vcs

rickliu2000:
Commit 7a66bf9 breaks riscv-dv flow on vcs, since vcs won't take the -define as a macro definition. Using -define instead of +define leads to the following output:

image

And since the marco is not correctly defined, vcs failed to build the sim

image

Fixing this by using +define instead of -define in build options.
Tested on vcs and xlm.

### DIFF
--- a/dv/uvm/core_ibex/scripts/compile_tb.py
+++ b/dv/uvm/core_ibex/scripts/compile_tb.py
@@ -119,9 +119,9 @@ def _main() -> int:
                 # Base address of the debug module. This is passed as a parameter
                 # at compile time as the PMP module must always allow debug mode
                 # accesses to it.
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Thanks, yes you're absolutely right. I thought I tested this under vcs, but clearly not!
For some historical reason, [here](https://github.com/lowRISC/ibex/blob/11ce0407b67f543bc1d29ef3908e0ff896990f98/util/ibex_config.py#L258-L278) we use the -define syntax to pass defines for xcelium only, and all other tools use the plusarg form. Since xcelium clearly also accepts the plusarg form, we should probably standardize on that form only.

PR #2301, opened 2025-07-23
Fix incorrect privilege mode detection causing failed dret riscv-dv tests

rickliu2000:
The testbench of the riscv-dv flow failed to compare the privilege mode of the DUT correctly and failing the riscv_dret_test. The testbench assumes the DUT will always be in machine mode while different random seeds can generate a testcase that would bring the DUT into either user mode or machine mode at the point of comparison. 
Some seeds of the riscv_dret_test that leads to failure: 
VCS:  1574
XLM:  28930

image

image

This commit fixed the comparison and make riscv_dret_test of the riscv-dv pass under either circumstances.
image

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -1125,7 +1125,7 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
     // we are blocking the current instruction until the instruction from WB stage is ready.
     wait (dut_vif.dut_cb.ctrl_fsm_cs == FLUSH);
     clk_vif.wait_clks(2);
-    check_priv_mode(PRIV_LVL_M);
+    check_priv_mode(init_operating_mode);
     wait_ret("mret", 15000);
   endtask
 
@@ -1657,7 +1657,7 @@ class core_ibex_dret_test extends core_ibex_directed_test;
 
   virtual task check_stimulus();
     forever begin
-      wait (dut_vif.dut_cb.dret === 1'b1);
+      @(negedge dut_vif.dut_cb.dret);
       check_illegal_insn("Core did not treat dret like illegal instruction");
     end
   endtask


dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1125,7 +1125,7 @@ class core_ibex_directed_test extends core_ibex_debug_intr_basic_test;
     // we are blocking the current instruction until the instruction from WB stage is ready.
     wait (dut_vif.dut_cb.ctrl_fsm_cs == FLUSH);
     clk_vif.wait_clks(2);
-    check_priv_mode(PRIV_LVL_M);
+    check_priv_mode(init_operating_mode);
```
marnovandermaas, line 1128: When you get an illegal instruction it must trap in M-mode regardless of what the initial operating mode is.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -1657,7 +1657,7 @@ class core_ibex_dret_test extends core_ibex_directed_test;
 
   virtual task check_stimulus();
     forever begin
-      wait (dut_vif.dut_cb.dret === 1'b1);
+      @(negedge dut_vif.dut_cb.dret);
```
marnovandermaas, line 1660: I think you're changing what this test does now. This test sees whether there is a DRET instruction and makes sure it is detected as illegal. Have you configured your Ibex to support DRET?

marnovandermaas (changes requested):
Thanks for making these changes, just marking them as request changes because they are currently not correct and will affect other tests as well.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
rickliu2000: Well, I think the time of checking was on trap exit. If the design here was to test the mode on trap entry, we'll probably have to discuss how to rewrite this module. The wait on line 1126 here can not capture the flush signal on trap entry. In this specific testcase(riscv_dret_test), the previous functions before the wait line have a handshake mechanism and blocks the wait line until handshake finished, so the first flush on trap entry signal will not be captured in this case. That's probably why I think this checking is wrong.

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
rickliu2000: Yes I have the dret enabled, and it is only a legal instruction in debug mode. 
I don't think I am changing the test here. Somehow, the wait line failed to capture here sometime, only negedge can correctly triggers all the time.

PR #2303, opened 2025-07-28
[nix] bring formal environment up-to-date

nbdd0121:
This updates nixpkgs to 25.05 and bring sail to the latest of lowrisc branch (which is needed as the old commit is not compatible with OCaml 5).

I have verified the change with Jasper and it still proves.


mndstrmr (approved):
Awesome, thanks for the cleanup.

PR #2304, opened 2025-08-04
[doc] Fix typo'd alert name

jwnrt:
`alert_major_internal_o` was repeated twice when I think the second alert should be `alert_major_bus_o`.


Razer6 (approved):
Thanks for fixing the docs.

PR #2306, opened 2025-08-12
Bring out debug_mode_o port

rishabhmitra-adi:
This brings out a port called 'debug_mode_o' from the core and at the top, signifying the state of debug mode when IBEX is connected to other peripherals dependent on the debug mode to update their operation like halt, etc. No functional changes to it.

### DIFF
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -113,7 +113,8 @@ module ibex_core import ibex_pkg::*; #(
   output logic                         irq_pending_o,
 
   // Debug Interface
-  input  logic                         debug_req_i,
+  input  logic                         debug_req_i,    // Request from Top to Core to enter debug mode
+  output logic                         debug_mode_o,   // Indicates that the Core has entered debug mode
   output crash_dump_t                  crash_dump_o,
   // SEC_CM: EXCEPTION.CTRL_FLOW.LOCAL_ESC
   // SEC_CM: EXCEPTION.CTRL_FLOW.GLOBAL_ESC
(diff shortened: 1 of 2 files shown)


marnovandermaas (commented):
Would you mind explaining what use-case you have for this? I'm unsure why a peripheral would need to know about whether the Ibex is in debug mode or not.

rtl/ibex_core.sv
```diff
@@ -706,6 +707,9 @@ module ibex_core import ibex_pkg::*; #(
     .instr_id_done_o  (instr_id_done)
   );
 
+  // Debug Mode output
+  assign debug_mode_o = debug_mode;
```
marnovandermaas, line 711: I see that debug mode is already exposed when RVFI is defined. Does that suffice in your use-case?

rtl/ibex_top.sv
```diff
@@ -100,7 +100,8 @@ module ibex_top import ibex_pkg::*; #(
   output logic                         scramble_req_o,
 
   // Debug Interface
-  input  logic                         debug_req_i,
+  input  logic                         debug_req_i,   // Request to Core to enter debug mode
+  output logic                         debug_mode_o,  // Indicates that the core has entered debug mode
```
marnovandermaas, line 104: You would need to add this anywhere where ibex_top is instantiated as well.

marnovandermaas (commented):
Some comments from my end. I still don't fully understand what these timers are that you need to halt. What do the timers represent? Number of cycles passed not in debug mode?

PR #2309, opened 2025-08-20
Introduce an OSS Formal Flow

mndstrmr:
This PR introduces a formal flow for Ibex that does not depend on Jasper, instead using yosys and our fork of the yosys-slang frontend to produce an aiger file, which the rIC3 model checker, under the instruction of conductor.py, can then verify. Just one aiger file is created, and is subsequently reconfigured by the Rust aig-manip tool.

This commit also introduces a new ci job which invokes the OS formal flow. It uses the nix flake, which contains all the necessary dependencies.

See dv/formal/README.md for more details, but in general:

1. We generate the specification module and psgen RTL in roughly the same way as the Jasper flow, with small changes to the psgen output in particular.
2. We use a fork of yosys-slang as the frontend for yosys. It has comprehensive SystemVerilog support, and handles both the specification and Ibex easily. It lacks SVA support by default, hence the need for the fork, though hopefully this will be upstreamed soon. The RTL source list is constructed by fusesoc in the same way as for Jasper.
3. The yosys build script is then specified in build_all.ys. It makes use of three custom passes, which may be found in yosys_formal. See their source files for more explanation of what they do. Yosys takes ~10 minutes to build an aiger file. See aig-manip/src/main.rs for more description of aiger files.
4. In order to prove a property, the rIC3 model checker can be used. However, the aiger file produced by yosys in step 3 contains all properties as assertions. To focus on just one, and to take all prior steps as assumptions the aig-manip tool is used. After this invocation of rIC3 is straightforward for IC3, k-induction or BMC. If a CEX is discovered it may be analysed by the aig-manip tool, which can lift the AIW (aiger witness) into a VCD file, for use in gtkwave etc.
5. When considering the proof as a whole, the conductor.py script is used to discover proof strategies, and then run proofs while maximising memory usage. It caches proofs for each step under strategies/step*.json.

Additonal notes:
1. The actual proof is relatively unchanged, with just a couple of new properties.
2. The proof is fast (it takes 40 minutes or so in CI)
3. smt2 based proofs are also available, e.g. for k-induction, since the build script (build_all.ys) will produce an smt2 file, which can be manipulated by smt2manip.py in the same that aig-manip does for an aiger file.
4. The transformation made to a global clock will confuse clock gating. It *should* be fine, since the clock signal it generates is essentially ignored anyway.


dv/formal/strategies/step0.json
marnovandermaas, line 1: It would be good to mark these as auto-generated, see this example file: https://github.com/lowRISC/sonata-system/blob/main/.gitattributes

dv/formal/spec/spec_api.sv
```diff
@@ -62,8 +57,6 @@ module spec_api #(
     output mstatus_t mstatus_o,
     input logic [31:0] mcause_i,
     output logic [31:0] mcause_o,
-    input logic [63:0] mcycle_i,
-    output logic [63:0] mcycle_o,
```
marnovandermaas, line 66: Wonder why these are removed?

dv/formal/check/top.sv
```diff
@@ -233,7 +241,7 @@ logic ex_has_compressed_instr;
 logic wbexc_post_int_err; // Spec had an internal error
 
 logic [31:0] wbexc_post_wX;
-logic [5:0] wbexc_post_wX_addr;
+logic [4:0] wbexc_post_wX_addr;
```
marnovandermaas, line 245: Nice bug fix!

dv/formal/check/top.sv
```diff
@@ -105,6 +108,13 @@ module top import ibex_pkg::*; #(
   input logic                          scan_rst_ni
 );
 
+// Create our reset signal for symbiosis
```
marnovandermaas: Do we still use Symbiosis?

dv/formal/check/protocol/mem.sv
```diff
@@ -42,14 +42,21 @@ interface mem_assume_t(
     // 2. Grants can only be sent when they are requested
     MemGntOnRequest: assume property (@(posedge clk_i) ~req_o |-> ~gnt_i);
 
+    `ifdef YOSYS
+    // Give yosys stronger bounds for now.
+    // FIXME: Would be nice to relax this, but I'm not sure how long proofs would take.
+    //        It might not be too long based on the alt lsus and whatnot, hard to say.
+    StrictGnt: assume property (@(posedge clk_i) req_o |-> gnt_i);
+    StrictRValid: assume property (@(posedge clk_i) outstanding_reqs_q != 8'b0 |-> rvalid_i);
+    `else
```
marnovandermaas, line 51: Once this is merged we should create an issue to track this.

dv/formal/yosys_formal/compile_flags.txt
```diff
@@ -0,0 +1,18 @@
...
+-DYOSYS_ENABLE_READLINE
+-DYOSYS_ENABLE_PLUGINS
+-DYOSYS_ENABLE_GLOB
+-DYOSYS_ENABLE_ZLIB
+-I/usr/include/tcl8.6
+-DYOSYS_ENABLE_TCL
+-DYOSYS_ENABLE_ABC
+-DYOSYS_ENABLE_COVER
+
+-I/nix/store/xd3azr82hh34wac4bqydbimizrisq4gb-yosys-0.51/share/yosys/include/
```
marnovandermaas: Is this constant?

marnovandermaas: Can we reference flake lock at all?

marnovandermaas (commented):
It would be nice to split the formal fixing changes from the open source flow (as you suggested in person). Here are just a few comments based on an initial read though. Todo for me is read through conductor.py and aig-manip

dv/formal/spec/spec_api.sv
mndstrmr: They shouldn't have ever been there, we don't do anything with mcycle since there's essentially no specification interpretation of it.

dv/formal/strategies/step0.json
mndstrmr: Done, thanks!

dv/formal/yosys_formal/compile_flags.txt
mndstrmr: Decided to just remove the whole thing, I never really meant to commit it in the first place!

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+import json
+import os
+from math import floor
+from datetime import datetime
+import hashlib
+
+# --------------------- Utilities ----------------------
+
+PROOFLOG = os.environ["PROOFLOG"] if "PROOFLOG" in os.environ else "prooflog.txt"
+LOGFILE = os.environ["LOGFILE"] if "LOGFILE" in os.environ else "logfile.txt"
```
marnovandermaas: Maybe make these environment names a bit more unique to avoid conflicts. Like IBEX_PROOF_LOG and IBEX_PROOF_LOG_FILE

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+# Send all printed text to a log file, and prepend the date and time.
+# This is helpful for looking back at how long things took, or how
+# long until we expect something to be done.
+oldprint = print
+def print(*args):
+    timestr = datetime.now().strftime('[%d/%m/%y %H:%M:%S.%f]')
+    oldprint(gray(timestr), *args)
+    if not "NO_LOG" in os.environ:
+        with open(LOGFILE, "a") as f:
+            oldprint(timestr, *[re.sub("\033\\[[0-9;]+m", "", str(x)) for x in args], file=f)
```
marnovandermaas: Not sure if overwriting print is the best way to do this. Maybe we can use a logging module like: https://docs.python.org/3/library/logging.html instead?

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+# Returns the global free memory according to /proc/meminfo, note though that reclaimable
+# memory (e.g. cached files) are not added to this when they could be, leading to the impression
+# of less available memory than there is for our use cases.
+def global_memory_free():
+    with open("/proc/meminfo", "r") as f:
+        c = f.read()
+        for line in c.split("\n"):
+            if line.startswith("MemFree:"):
+                return float(line[8:-2].strip()) / (1024 * 1024)
+    return 0
```
marnovandermaas: As discussed in person it would be great if we could add a parameter that can be passed by the Github action to limit the amount of memory used. Then you can return the minimum of that at proc/meminfo

dv/formal/conductor.py
marnovandermaas, line 1: In general it would be nice to use typed Python. If you have spare time would you mind implementing that? The checker I usually use to check this is mypy: `mypy --strict *.py`

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+def pid_memory_used(pid):
+    with open(f"/proc/{pid}/status", "r") as f:
+        c = f.read()
+        for line in c.split("\n"):
+            if line.startswith("RssAnon:"):
+                return float(line[8:-2].strip()) / (1024 * 1024)
+    return 0
+
+# Finishes when a process exits, and cancelling kills it.
+# Result is (exit code, max mem, dt, status string, stdout?, stderr?)
```
marnovandermaas: Can we encode these into the class type so that we don't have the confusing type signature later in the Process.__init__ function?

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+    MAX_RUNNING = len(os.sched_getaffinity(0)) - 2 # Number of CPUs, with a couple spares
+
+    def __init__(self):
+        self.pending = []
+        self.running = []
+        self.first_start = time.time()
+        self.debug_count = 0
+
+    def append(self, proc):
+        # FIXME: Use a circular buffer
```
marnovandermaas: Do you still need this or can we do without?

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+
+        n = floor((time.time() - self.first_start) / 30.0)
+        if n > self.debug_count:
+            print(gray(f"Running {len(self.running)} processes, with {len(self.pending)} pending. Memory used right now: {self.children_used_mem():.3f}GB"))
+            self.debug_count = n
+
+        asyncio.get_running_loop().call_later(ProcessRunner.POLL_DELAY, lambda: self.poll())
+
+print(f"Process runner will have a maximum of {ProcessRunner.MAX_RUNNING} processes, and currently sees {global_memory_free():.3f}GB free.")
+process_runner = ProcessRunner()
```
marnovandermaas: Can we put these calls into a main function?

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+async def explore(step, props, configs, timeout=0.0, debug_slow=None) -> asyncio.Future[None | tuple[str, tuple[int, float, float, str]]]:
+    specialised = await prepare_aiger(step, props)
+
+    async def explore_one(engine_config, idx, expected_memory):
+        result = await shell(f"{engine_config} {specialised}", timeout=timeout, expected_memory=expected_memory, debug_slow=debug_slow)
+        proof_done(engine_config, specialised, step, props, result)
+        if result[0] in {10, 20}:
+            won_race((engine_config, result))
+
+    return await race([explore_one(config, idx, mem) for idx, (config, mem) in enumerate(configs)])
```
marnovandermaas: Very neat!

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+    return group(chunked_names)
+
+# ---------------------------------- Main ------------------------------
+
+parser = argparse.ArgumentParser(prog="conductor.py", description="Constructs and executes proofs.")
+parser.add_argument("mode", choices=["prove", "explore", "bmc", "info", "logopt"],
+                    help="Proof mode. "
+                    "prove will run existing proofs where they exist, "
+                    "explore will attempt to discover new proofs, "
+                    "bmc will run BMC on each property individually, "
```
marnovandermaas: Maybe spell out that this is bounded model checking.

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+parser.add_argument("--missing", action="store_true", help="Equivalent to -p <each property that is not skipped but has no proof in a step where there are some proofs>")
+parser.add_argument("-s", "--start", type=int, default=0, help="First step to start at. (default: 0)")
+parser.add_argument("--bmc-step", type=int, default=1, help="Step size for BMC. (default: 1)")
+parser.add_argument("--bmc-start", type=int, default=4, help="Start length for BMC. (default: 4)")
+parser.add_argument("--hard", action="store_true", help="In explore mode, try harder to prove properties (1hr timeout, more engines).")
+parser.add_argument("--no-run", action="store_true", help="In explore mode don't run proofs again to check steps.")
+parser.add_argument("--no-kill", action="store_true", help="Don't kill proof processes due to running out of memory.")
+parser.add_argument("--check-complete", action="store_true", help="In prove mode, fail if there are unskipped properties without proofs.")
+parser.add_argument("--max-mem", type=float, default=0, help="Max memory in GB, otherwise use all available memory")
```
marnovandermaas, line 543: You already have a maximum memory, I guess the only thing necessary is to link this up to the Github action.

dv/formal/aig-manip/src/bitvec.rs
marnovandermaas, line 1: This and all other Rust code should have the lowRISC copyright header.

dv/formal/conductor.py
machshev: You could also simplify this using `.get` with a default value i.e.

```python
os.environ.get("PROOFLOG", "prooflog.txt")
os.environ.get("LOGFILE", "logfile.txt")
```

dv/formal/conductor.py
machshev: Yes I'd agree, looks like you just implemented your own logger ;)

If you use the built-in logger then it's easier to set the log level and choose how verbose you want the logs to be as a user. It's not too complicated to setup the logger although there are some common mistakes. So if you wanted something quicker then I'd recommend [logzero](https://logzero.readthedocs.io/en/latest/) (MIT license) which is a wrapper around the built-in logger and has some sensible defaults. It's easy to log to a file as well as the terminal.

dv/formal/conductor.py
machshev: Another option is `pyright` which is deals with some of the corner cases a bit better than `mypy`.

dv/formal/spec/fix_bugs.py
```diff
@@ -0,0 +1,46 @@
...
+    logic [7:0] bits;
+} t_Pmpcfg_ent;
+"""
+
+T = """
+typedef logic [127:0] sail_int;
+"""
+
+with open("build/ibexspec.sv", "r") as f:
+    c = f.read()
```
machshev: Maybe use `pathlib` here? There are a number of lint rules that ask people to migrate. I think the main reason is it handles cross platform paths for you.

```
from pathlib import Path

c = Path("build/ibexspec.sv").read_text()
```

dv/formal/spec/fix_bugs.py
```diff
@@ -0,0 +1,46 @@
...
+        sail_reached_unreachable_loc = 0;
+        main_result = main(insn_bits, mode); \
+""")
+c = c.replace("sail_reached_unreachable = 1;", "if (!sail_reached_unreachable) begin sail_reached_unreachable = 1; sail_reached_unreachable_loc = `__LINE__; end")
+c = c.replace("module sail_ibexspec(", "module sail_ibexspec(\n    output logic sail_reached_unreachable,\n    output logic [31:0] sail_reached_unreachable_loc,")
+c = c.replace("logic sail_reached_unreachable;", "")
+c = c.replace("logic [31:0] sail_reached_unreachable_loc;", "")
+
+with open("build/ibexspec.sv", "w") as f:
+    f.write(c)
```
machshev: This could be something like:

```
Path("build/ibexspec.sv").write_text(c, encoding="utf8")
```

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+    with open(LOGFILE, "a") as f:
+        f.write("\n")
+
+def green(s): return f"\033[32;1m{s}\033[0m"
+def red(s): return f"\033[31;1m{s}\033[0m"
+def white(s): return f"\033[1m{s}\033[0m"
+def orange(s): return f"\033[33;1m{s}\033[0m"
+def gray(s): return f"\033[90m{s}\033[0m"
+
+# Races allow async processes to cleanly run in parrallel, until one declares victory, then
```
machshev: This comment could be a docstring for the class. Which means it becomes visible for LSPs making it easier for people to read your code.

dv/formal/smt2manip.py
```diff
@@ -0,0 +1,61 @@
...
+    m = re.match(r"; yosys-smt2-assert ([0-9]+) ([^ ]+)", line)
+    if m is not None:
+        id = m[1]
+        name = m[2][6:].split("$")[0]
+        assert name.startswith("Step")
+        [step, name] = name.split("_", maxsplit=1)
+        step = int(step[4:])
+
+        if name.endswith("_Cover") or step > args.step or (step == args.step and len(args.props) > 0 and name not in args.props):
+            print(f"Delete {name} ({id})")
```
machshev, line 38: Generally preferable to use `log.info` instead of print for massages you always want to see. Alternatively `log.debug` for messages that should only show in verbose mode.

dv/formal/smt2manip.py
```diff
@@ -0,0 +1,61 @@
...
+        else:
+            print(f"Keep {name} ({id})")
+        mapped.append(line)
+        continue
+
+    m = re.match(r"  \(\|top_a ([0-9]+)\| state\)", line)
+    if m is not None:
+        if m[1] in to_assume:
+            continue
+        elif m[1] in delete:
```
machshev:
```suggestion
        if m[1] in to_assume or m[1] in delete:
            continue
```

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+            oldprint(timestr, *[re.sub("\033\\[[0-9;]+m", "", str(x)) for x in args], file=f)
+if not "NO_LOG" in os.environ:
+    with open(LOGFILE, "a") as f:
+        f.write("\n")
+
+def green(s): return f"\033[32;1m{s}\033[0m"
+def red(s): return f"\033[31;1m{s}\033[0m"
+def white(s): return f"\033[1m{s}\033[0m"
+def orange(s): return f"\033[33;1m{s}\033[0m"
+def gray(s): return f"\033[90m{s}\033[0m"
```
machshev, line 47: For terminal colours I'd use something like: https://pypi.org/project/colorama/
This is BSD licensed and provides cross platform support for terminal colours.

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+        for result in e.exceptions:
+            winner = result.result
+    return winner
+
+# --------------------- Process Management ----------------------
+
+# Returns the global free memory according to /proc/meminfo, note though that reclaimable
+# memory (e.g. cached files) are not added to this when they could be, leading to the impression
+# of less available memory than there is for our use cases.
+def global_memory_free():
```
machshev, line 95: Perhaps consider using the `psutil` library?
https://github.com/giampaolo/psutil?tab=readme-ov-file#memory (DSB-3)

Again while your solution works, if there is any intention of cross platform support it's thinks like things that would break it.

dv/formal/conductor.py
```diff
@@ -0,0 +1,802 @@
...
+    with open("/proc/meminfo", "r") as f:
+        c = f.read()
+        for line in c.split("\n"):
+            if line.startswith("MemFree:"):
+                return float(line[8:-2].strip()) / (1024 * 1024)
+    return 0
+
+# Memory used by a specific PID according to /proc/{pid}/status
+def pid_memory_used(pid):
```
machshev: Again `psutil` will provide this for you in a cross platform way.

machshev (commented):
Nice work! Some nits...

dv/formal/conductor.py
mndstrmr: Since this whole setup is very nix based, and given that nix doesn't have Windows support, are we trying to support it anyway? As far as I know everything else supports the standard ANSI colors.

dv/formal/conductor.py
mndstrmr: I can agree that overwriting print was maybe not the best move!

If instead of overwriting `print` I called it `log`, would everyone be happy? I struggle to see the value of adding an entire third party library to replace 6 lines of code that work. It's not hard to handle verbosity either if that's something we want.

dv/formal/thm/riscv.proof
```diff
@@ -112,15 +112,15 @@ lemma riscv
...
-    in L:(wbexc_is_pres_load_instr) S:(wbexc_is_pres_store_instr)
+    in L:(wbexc_is_load_instr) S:(wbexc_is_store_instr)
       use spec_compliant
           split_bool Err:(wbexc_err)
 
   IRQ: in (wbexc_handling_irq)
     PC: have (pc_match)
     CSR: have (csrs_match)
 
-  Illegal: in (wbexc_illegal & can_check_illegal & wbexc_finishing & ~wbexc_fetch_err & (`ISS_CSR -> wbexc_is_checkable_csr))
```
marnovandermaas, line 123: Why should you remove the "can_check_illegal" here?

.github/workflows/ci-formal.yml
```diff
@@ -0,0 +1,48 @@
...
+  fv:
+    name: Run the Open-Source FV flow
+    runs-on: nixos
+    concurrency:
+      group: oss-fv
+      cancel-in-progress: false # Only do one oss-fv job at a time, since we will use a huge amount of memory
+    steps:
+      - name: Setup env
+        run: |
+          echo "MAX_MEM_GB=200" >> $GITHUB_ENV # Maximum memory available to conductor.py, which is inversely proportional to time. 200GB -> 40 mins
```
marnovandermaas: @nbdd0121 Is it acceptable to use 200GiB?

.github/workflows/ci-formal.yml
```diff
@@ -0,0 +1,48 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# GitHub Actions CI build configuration
+
+name: Ibex OSS Formal CI
+
+on:
+  pull_request:
```
marnovandermaas, line 10: @nbdd0121 are you happy for this to run on pull request?

marnovandermaas (commented):
Left a few comments. I still need to run the flow again before I approve.

dv/formal/thm/riscv.proof
mndstrmr: `can_check_illegal` used to be defined as 1 if all the instructions that could raise illegal instruction exceptions were included in the specification. If they weren't then the property would fail since we would reach states marked as unreachable in the specification, making the property false. Since all those instructions are always included now anyway (that's the purpose of that commit) this is now implicitly always 1, so it can just be removed from the `&`.

.github/workflows/ci-formal.yml
marnovandermaas: @mndstrmr is this half of Pomona's memory?

mndstrmr: I'm not actually sure, I'm under the impression it's a bit less than all of it, though with that concurrency group no two formal flows should run at the same time anyway.

.github/workflows/ci-formal.yml
marnovandermaas: Maybe @jwnrt can comment on this as well?

dv/formal/thm/riscv.proof
marnovandermaas: That makes sense, thanks!

.github/workflows/ci-formal.yml
marnovandermaas: That makes sense. It probably still makes sense to lower this a bit so that it doesn't affect other jobs running on the machine.

dv/formal/conductor.py
marnovandermaas: I'm happy with leaving this as is and we can always change this if needed later.

.github/workflows/ci-formal.yml
jwnrt: I think it's okay given the time it takes and the volume of PRs that Ibex gets. PR CI won't run for external contributors unless approved as well.

dv/formal/conductor.py
marnovandermaas: I'm happy with this. If we ever see the need of using a logging library that shouldn't be a difficult change for us to add later.

.github/workflows/ci-formal.yml
marnovandermaas: Thanks, marking this as resolved since we can easily change this in the future.

jwnrt (approved):
Briefly reviewed the Rust and didn't see any problems, LGTM

.github/workflows/ci-formal.yml
mndstrmr: Updated this too 100GB now

flake.nix
```diff
@@ -38,13 +38,25 @@
...
-      url = "github:lowrisc/sail?ref=lowrisc";
+      url = "github:lowrisc/sail?ref=ast_translate";
       flake = false;
     };
     lowrisc_sail_riscv = {
       url = "github:lowrisc/sail-riscv?ref=81a266b6f65365b34180af7b91708265da653878";
       flake = false;
     };
+    lowrisc_yosys_slang = {
+      url = "https://github.com/mndstrmr/yosys-slang";
```
hcallahan-lowrisc, line 49: It would be good to move this under the lowrisc org at some point.

yosys.nix
```diff
@@ -0,0 +1,187 @@
+{
```
hcallahan-lowrisc: I can't see where this file is used?

dv/formal/Makefile
```diff
@@ -32,33 +40,52 @@ fusesoc:
...
 
-.PHONY: psgen
-psgen:
+# The compiled property structure, and the associated TCL
+build/psgen-jg.sv: $(PSGEN) $(PSGEN_SRCS) Makefile
 	mkdir -p build
-	psgen $(addprefix -path ,$(PSGEN_SRCS)) $(PSGEN_FLAGS) -sv-out build/psgen.sv -tcl-out build/psgen.tcl
+	$(PSGEN) $(addprefix -path ,$(PSGEN_SRCS)) $(PSGEN_FLAGS) -sv-out build/psgen-jg.sv -tcl-out build/psgen.tcl
 
+# Same as above, but generate our own covers (not very well at the minute!), and tweak for consumption by our tools
```
hcallahan-lowrisc: Probably can remove the pessimism now! :)

dv/formal/Makefile
```diff
@@ -32,33 +40,52 @@ fusesoc:
...
+build/psgen-yosys.sv: $(PSGEN) $(PSGEN_SRCS) Makefile
+	mkdir -p build
+	$(PSGEN) $(addprefix -path ,$(PSGEN_SRCS)) $(PSGEN_FLAGS) -sv-out build/psgen-yosys.sv -clocking -covers -cover-assert -step-prefix
+
+# The aiger file for the OSS flow, using the yosys_formal/* plugins. The main yosys script is in build_all.ys
+build/all.aig: build/psgen-yosys.sv build/fusesoc build/ibexspec.sv check build_all.ys $(LOWRISC_YOSYS_SLANG) build/yosys_formal.so
+	yosys -m $(LOWRISC_YOSYS_SLANG) -m build/yosys_formal.so -v 1 -s <(cat ./build_all.ys | sed "s|LOWRISC_SAIL_SRC|$(LOWRISC_SAIL_SRC)|g")
```
hcallahan-lowrisc, line 68: Slightly annoying yosys can't do environment variable substitution, would be nice to not have to do this dance with sed.

hcallahan-lowrisc (approved):
This looks great. I'm super happy to see this get merged. Well done @mndstrmr for all the great work!

One note before merging. Some files seem to be missing copyright/license headers. @marnovandermaas would you mind going through all the formal files and checking?

flake.nix
mndstrmr: Ideally it would in the upstream `yosys-slang`, but since it's not a particularly complete implementation of SVA I have some doubts as to whether or not they would want to accept it.

yosys.nix
mndstrmr: Good point!

.github/workflows/ci-formal.yml
```diff
@@ -0,0 +1,48 @@
...
+    branches:
+      - '*'
+
+jobs:
+  fv:
+    name: Run the Open-Source FV flow
+    runs-on: nixos
+    concurrency:
+      group: oss-fv
+      cancel-in-progress: false # Only do one oss-fv job at a time, since we will use a huge amount of memory
```
jwnrt, line 20: It's a shame that we can't cancel the old job when you force push to a PR... I think with this configuration it will still run the queued job even if you force push something new?

@nbdd0121 any insight here? Maybe we can limit the total concurrency between jobs by adding a special label to exactly one of the `nixos` runners and queuing on that?

mndstrmr: Yeah, I don't cancel the old run since e.g. if two PRs are both opened at the same time then the first would be cancelled. This is annoying if you have just one PR with several updates though, I agree.

However, if you push three times to the same PR, then the first and third run, but the second is cancelled by the third, so that's something.

I think GitHub concurrency groups are just not meant for handling this kind of thing, and ideally this would be handled inside the job maybe? Or the runner could be smarter? Not sure.

dv/formal/Makefile
mndstrmr: It unfortunate. Even using `envsubst` won't work due to this line:

https://github.com/lowRISC/ibex/blob/eb6e0ac7c7fc3cd77a0a99567f244277c405e9c8/dv/formal/build_all.ys#L18

.github/workflows/ci-formal.yml
nbdd0121: I plan to limit each runner to 128G of RAM so two concurrent runs would not OOM the machine. Then we can proceed to use concurrent groups as normal?

flake.nix
nbdd0121: How big are the patches? We might be able to supply a patched version via lowrisx-nix if the patch sizes are not excessive

mndstrmr: They're not particularly big. Max 400 lines maybe?

marnovandermaas: We should open an issue to track this once this PR is merged.

dv/formal/check/protocol/mem.sv
marnovandermaas: https://github.com/lowRISC/ibex/issues/2320

flake.nix
marnovandermaas: https://github.com/lowRISC/ibex/issues/2321

povik: Let me leave a drive-by comment as the maintainer of yosys-slang I'm happy to take a look at the patch. An incomplete implementation does not worry me as long as the user experience is good and the supported subset is implemented correctly.

dv/formal/Makefile
povik: You might be able to turn `build_all.ys` into a Tcl script and look up the environment via `$::env(FOO)`. Yosys has an embedded Tcl interpreter in most builds

PR #2289, opened 2025-06-23
[simple_system] Add CSR to architecture extensions

ekigwana:
Simple system uses CSR's and thus this is required.

Closes #2288

PR #2326, opened 2025-11-03
Add missing zicsr to -march in example Makefile

Timmmm:
Newer GCC toolchains require you to explicitly enable zicsr otherwise you get errors when trying to compile code with CSR access instructions.


SamuelRiedel:
This is a duplicate of https://github.com/lowRISC/ibex/pull/2289

PR #2327, opened 2025-11-03
Remove dependency on srecord

Timmmm:
The linked issue was resolved 6 years ago so it's probably safe to use the new flag.

Draft because I'm not sure how to actually test this. Is it used anywhere with Verilator?

### DIFF
--- a/examples/sw/benchmarks/coremark/ibex/core_portme.mak
+++ b/examples/sw/benchmarks/coremark/ibex/core_portme.mak
@@ -92,8 +92,8 @@ $(OPATH)$(PORT_DIR)/%$(OEXT) : %.s
 port_postbuild:
 	riscv32-unknown-elf-objdump -SD $(OPATH)coremark.elf > $(OPATH)coremark.dis
 	riscv32-unknown-elf-objcopy -O binary $(OPATH)coremark.elf $(OPATH)coremark.bin
-	srec_cat $(OPATH)coremark.bin -binary -offset 0x0000 -byte-swap 4 -o  $(OPATH)coremark.vmem -vmem
-
+	riscv32-unknown-elf-objcopy -O verilog --reverse-bytes=4 --verilog-data-width=4 \
+	    $(OPATH)coremark.elf $(OPATH)coremark.vmem
 
 # FLAG : OPATH
 # Path to the output folder. Default - current folder.
--- a/examples/sw/simple_system/common/common.mk
+++ b/examples/sw/simple_system/common/common.mk
@@ -52,15 +52,11 @@ endif
 %.dis: %.elf
 	$(OBJDUMP) -fhSD $^ > $@
 
-# Note: this target requires the srecord package to be installed.
-# XXX: This could be replaced by objcopy once
-# https://sourceware.org/bugzilla/show_bug.cgi?id=19921
-# is widely available.
-%.vmem: %.bin
-	srec_cat $^ -binary -offset 0x0000 -byte-swap 4 -o $@ -vmem
+%.vmem: %.elf
+	$(OBJCOPY) -O verilog --reverse-bytes=4 --verilog-data-width=4 $< $@
 
 %.bin: %.elf
-	$(OBJCOPY) -O binary $^ $@
+	$(OBJCOPY) -O binary $< $@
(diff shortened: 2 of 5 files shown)


examples/sw/benchmarks/coremark/ibex/core_portme.mak
```diff
@@ -92,8 +92,7 @@ $(OPATH)$(PORT_DIR)/%$(OEXT) : %.s
 port_postbuild:
 	riscv32-unknown-elf-objdump -SD $(OPATH)coremark.elf > $(OPATH)coremark.dis
 	riscv32-unknown-elf-objcopy -O binary $(OPATH)coremark.elf $(OPATH)coremark.bin
-	srec_cat $(OPATH)coremark.bin -binary -offset 0x0000 -byte-swap 4 -o  $(OPATH)coremark.vmem -vmem
-
+	objcopy --input-target=binary --output-target=verilog --reverse-bytes=4 --verilog-data-width=4 $(OPATH)coremark.bin $(OPATH)coremark.vmem
```
elliotb-lowrisc: A couple of tweaks here should make this change fit in better with the existing code. Could you please swap out `objcopy` for `riscv32-unknown-elf-objcopy`, use the short version of flags where available (`-O` and `-I`) to match the lines above, and wrap the line before column 100?

examples/sw/simple_system/common/common.mk
```diff
@@ -52,12 +52,8 @@ endif
 %.dis: %.elf
 	$(OBJDUMP) -fhSD $^ > $@
 
-# Note: this target requires the srecord package to be installed.
-# XXX: This could be replaced by objcopy once
-# https://sourceware.org/bugzilla/show_bug.cgi?id=19921
-# is widely available.
 %.vmem: %.bin
-	srec_cat $^ -binary -offset 0x0000 -byte-swap 4 -o $@ -vmem
+	objcopy --input-target=binary --output-target=verilog --reverse-bytes=4 --verilog-data-width=4 $< $@
```
elliotb-lowrisc: Again, a few tweaks would be a big help for maintaining style. Could you please replace `objcopy` with `$(OBJCOPY)`, and use the short version of flags where available (`-O` and `-I`) to match the recipe below?

examples/sw/benchmarks/coremark/ibex/core_portme.mak
Timmmm: I generally prefer long arguments for scripts, where readability is more important than saving a few keystrokes... I guess `-I` and `-O` aren't too bad.

I also changed it so it directly converts from .elf to .vmem - a bit simpler and more regular.

examples/sw/simple_system/common/common.mk
Timmmm: Updated

SamuelRiedel:
This PR is similar to https://github.com/lowRISC/ibex/pull/2062. I am not the expert on these tools, but it might be useful to check out the discussions that were held there as well. Once this PR gets merged, the other one can be closed.

elliotb-lowrisc:
My apologies @Timmmm, it seems there is more to this dependency than I realised.

From what I can gather following the trail of links from https://github.com/lowRISC/ibex/pull/2062, we will probably have to stick with `srec_cat` for now after all. It seems that the we'd need Binutils 2.40 or later for these arguments to be supported correctly (https://github.com/lowRISC/opentitan/issues/1107#issuecomment-1386758734), but the RISC-V fork of Binutils used in our toolchain seems to stop at 2.38 (https://github.com/riscvarchive/riscv-binutils-gdb) and we are currently using 2.35 (https://github.com/lowRISC/lowrisc-toolchains/blame/7471191b10e2a7fbb84cf8c8ab84f729a76a5612/sw-versions.sh#L15). This may have been confirmed by the CI failing just now with the error "riscv32-unknown-elf-objcopy: cannot reverse bytes: length of section .debug_info must be evenly divisible by 4", although I'm not entirely sure if it isn't some other issue. Until our toolchain is updated, we probably have to park this change.

That said, I wonder if it would be possible to update the toolchain. @jwnrt would likely be able to comment once he's back from holiday.

PR #2328, opened 2025-11-05
[pmp] MSB and LSB of PMP address now params

marnovandermaas:
Instead of hardcoding 33 and 2 all over the code base as the most significant and least significant bit of the PMP address, this change creates appropriate parameters in the Ibex package that can be re-used in the rest of the code base.

This commit also replaces some hard-coded 16 values with PMP_MAX_REGIONS which was already a parameter in the Ibex package.

Finally it makes a slight clarifying change to avoid mixing pmp_mseccfg and pmp_mseccfg_q in the same assignment even though they refer to the same value.

### DIFF
--- a/rtl/ibex_cs_registers.sv
+++ b/rtl/ibex_cs_registers.sv
@@ -1183,7 +1183,7 @@ module ibex_cs_registers import ibex_pkg::*; #(
       // When MSECCFG.MML is set cannot add new regions allowing M mode execution unless MSECCFG.RLB
       // is set
       assign pmp_cfg_wr_suppress[i] = pmp_mseccfg_q.mml                   &
-                                      ~pmp_mseccfg.rlb                    &
+                                      ~pmp_mseccfg_q.rlb                  &
                                       is_mml_m_exec_cfg(pmp_cfg_wdata[i]);
 
       // --------------------------


rtl/ibex_cs_registers.sv
```diff
@@ -9,28 +9,28 @@
...
-  parameter bit                     PMPEnable         = 0,
-  parameter int unsigned            PMPGranularity    = 0,
-  parameter int unsigned            PMPNumRegions     = 4,
-  parameter ibex_pkg::pmp_cfg_t     PMPRstCfg[16]     = ibex_pkg::PmpCfgRst,
-  parameter logic [33:0]            PMPRstAddr[16]    = ibex_pkg::PmpAddrRst,
-  parameter ibex_pkg::pmp_mseccfg_t PMPRstMsecCfg     = ibex_pkg::PmpMseccfgRst,
-  parameter bit                     RV32E             = 0,
-  parameter ibex_pkg::rv32m_e RV32M                   = ibex_pkg::RV32MFast,
-  parameter ibex_pkg::rv32b_e RV32B                   = ibex_pkg::RV32BNone,
+module ibex_cs_registers import ibex_pkg::*; #(
```
SamuelRiedel, line 12: Nit: After moving this here, you can remove the import on line 133.

rtl/ibex_pkg.sv
```diff
@@ -387,6 +387,12 @@ package ibex_pkg;
   // PMP constants
   parameter int unsigned PMP_MAX_REGIONS      = 16;
   parameter int unsigned PMP_CFG_W            = 8;
+  // For RV32 the most significant bit of PMP address refers to the physical
+  // address bit index 33.
+  parameter int unsigned PMP_ADDR_MSB         = 33;
```
SamuelRiedel, line 392: In the very end of the package, we have the definition of `PmpAddrRst`, which also uses the `PMP_ADDR_MSB` for its width. This is the only instance I found where you didn't update the hardcoded value with the new parameter:

rtl/ibex_cs_registers.sv
marnovandermaas: Thanks for the detailed review, I've removed this line.

rtl/ibex_pkg.sv
marnovandermaas: Thanks, I also found two more instances of PMP max regions that I replaced.

PR #2324, opened 2025-10-13
Add Zcb and Zcmp extensions DV

SamuelRiedel:
This PR adds support for the **Zcb** and **Zcmp** code-saving extensions. This support is implemented in a parameterizable way via the `RV32ZC` parameter, which allows choosing none, either, or both extensions. By default, both extensions are enabled because their combined hardware overhead is small, approximately 800 gate equivalents.

The Zcb extension introduces new compressed encodings for common instructions already supported in Ibex. The Zcmp extension introduces single compressed instructions that expand into multiple existing instructions within Ibex. For example, a stack push expands into multiple store instructions and a stack pointer update. Therefore, both extensions are implemented in the IF stage.

Adding these new instructions requires updates to the following components:
- [x] Ibex RTL: Implemented in: https://github.com/lowRISC/ibex/pull/2347
- [x] Documentation: Updated in: https://github.com/lowRISC/ibex/pull/2347
- [x] Verification: Implemented in this PR as well. Some missing parts still (see at the bottom)
  + [x] `riscv-isa-sim` (Spike): Updated Spike to support those new extensions https://github.com/lowRISC/riscv-isa-sim/pull/27
  + [x] `riscv-dv`: Add both extensions to [riscv-dv](https://github.com/SamuelRiedel/riscv-dv/tree/zcbzcmp). We can either try to upstream those or create a fork.
    * [x] Small fix to simplify Zcb/Zcmp extension: https://github.com/chipsalliance/riscv-dv/pull/1024
    * [x] Support for Zcb extension: https://github.com/chipsalliance/riscv-dv/pull/1025
    * [x] Support for the Zcmp's CMMV instructions: https://github.com/chipsalliance/riscv-dv/pull/1026
  + [x] Ibex co-simulation: Updated in this PR.
    * [x] Support for Zcb extension
    * [x] Support for the Zcmp's CMMV instructions
    * [x] Support for Zcmp's CMPP instructions
- [ ] Compiler Support: We need to update the compiler to a version that supports the Zcb and Zcmp instructions. Currently, we are still on an [rather old](https://github.com/lowRISC/ibex/blob/master/ci/vars.env#L10) version. The latest RISC-V GCC and Clang support those instructions already.

Missing Components and Known Issues:
[...]

PR #2346, opened 2025-12-18
[dv] Use `id_done` to accurately track instruction monitor

SamuelRiedel:
This refactors `instr_vif` to use `rvfi_id_done` instead of `instr_new_id` to track when a new instruction appears in the ID stage. This interface and signal are only used to keep track of instruction fetch errors by using the aforementioned valid signal and checking whether the `rvfi_order_id` has advanced. However, `instr_new_id` will be gated if an instruction fetch error occurs after another error, causing us to overlook the `fetch_err` in the verification. With this change, we also no longer need to look at `valid_id` since the `rvfi_id_done` already contains this information.

This will fix failing `riscv_mem_error_test` in the nightly runs and the failing CI in https://github.com/lowRISC/ibex/pull/2324


rswarbrick (approved):
Looks good to me, and thanks for the careful explanation.

PR #2350, opened 2026-01-14
[doc] PMP instruction fetch check clarified

marnovandermaas:
Clarify that instruction fetch bus should not contain devices that experience side-effects due to reads.


nasahlpa (approved):
Thanks for clarifying this!

PR #2354, opened 2026-01-18
[syn] Adapt for Yosys 0.60

Zxis233:
Fix the prim module path and the kGE area calculation to adapt Yosys 0.60

### DIFF
--- a/syn/python/get_kge.py
+++ b/syn/python/get_kge.py
@@ -49,15 +49,17 @@ def get_kge(report_path, weighted_dict):
     ge = 0.0
     for line_idx, line in enumerate(report):
         data = line.split()
-        if not data:
+        if len(data) < 3:
             continue
-        weight = weighted_dict.get(data[0])
+        cell_name = data[2]
+        weight = weighted_dict.get(cell_name)
         if weight is not None:
             try:
-                ge += float(data[1]) * weight
+                count = float(data[0])
+                ge += count * weight
             except (IndexError, ValueError):
                 raise RuntimeError('{}:{} Cell {} matched but was misformatted'
-                                   .format(report_path, line_idx + 1, data[0]))
+                                   .format(report_path, line_idx + 1, cell_name))
     print("Area in kGE = ", round(ge/1000, 2))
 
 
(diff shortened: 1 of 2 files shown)


syn/python/get_kge.py
```diff
@@ -77,4 +79,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
marnovandermaas, line 82: Did you mean to change how this file ends?

syn/python/get_kge.py
marnovandermaas, line 1: Can you explain the changes in this file. Previously the code used data 0 and data 1. Now you are also using data 2 but not data 1.

Zxis233: > Can you explain the changes in this file. Previously the code used data 0 and data 1. Now you are also using data 2 but not data 1.

The area report in Yosys 0.60 contains content similar to the following:
```
        +----------Local Count, excluding submodules.
        |        +-Local Area, excluding submodules.
        |        | 
      372  395.808   AND2_X1
       98   130.34   AND3_X1
```

That is, the real cell name is the third field, while the old code assumed that the  cell name is the first field

syn/python/get_kge.py
Zxis233: > Did you mean to change how this file ends?

This change is unnecessary because, according to the specifications, the last line of the document should be a blank line. Sorry for that.

Zxis233:
> Can you share the command you used to test this?

As the Wiki shows:

```
cd syn && ./syn_yosys.sh
```

syn/python/get_kge.py
marnovandermaas: Thanks for the explanation. Is that a change that was introduced in Yosys?

syn/python/get_kge.py
marnovandermaas: Thank you for reverting this.

syn/python/get_kge.py
Zxis233: > Thanks for the explanation. Is that a change that was introduced in Yosys?

I've checked the commit history and I find it:
[https://github.com/YosysHQ/yosys/commit/a0dde68487da600c569c0b61b684dc845fd5c554
](https://github.com/YosysHQ/yosys/commit/a0dde68487da600c569c0b61b684dc845fd5c554
)

Clearly the area report had been refactored, but it was NOT mentioned in the update log.

Zxis233:
@marnovandermaas

PR #2357, opened 2026-01-19
[rtl] Lint fixes for recent Zc code

andreaskurth:
These were flagged by CI lint checks in https://github.com/lowRISC/opentitan/pull/28997.

### DIFF
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -57,9 +57,9 @@ module ibex_compressed_decoder #(
   function automatic logic [4:0] cm_stack_adj_word(input logic [3:0] rlist,
                                                    input logic [1:0] spimm);
     logic [6:0] tmp;
-    logic [1:0] _unused;
+    logic [1:0] unused_tmp;
     tmp = cm_stack_adj(.rlist(rlist), .spimm(spimm));
-    _unused = tmp[1:0];
+    unused_tmp = tmp[1:0];
     return tmp[6:2];
   endfunction
 
@@ -79,8 +79,12 @@ module ibex_compressed_decoder #(
   function automatic logic [31:0] cm_push_store_reg(input logic [4:0] rlist,
                                                     input logic [4:0] sp_offset);
     logic [11:0] neg_offset;
+    logic signed [11:0] neg_offset_signed;
     logic [31:0] instr;
-    neg_offset = ~{5'b00000, sp_offset, 2'b00} + 12'd1;
+    // Compute two's complement on signed variable, then cast back to unsigned
+    // for the part select operations below.
+    neg_offset_signed = -signed'({5'b00000, sp_offset, 2'b00});
+    neg_offset = unsigned'(neg_offset_signed);
     instr[ 6: 0] /* opcode       */ = OPCODE_STORE;
     instr[11: 7] /* offset[4:0]  */ = neg_offset[4:0];
     instr[14:12] /* width        */ = 3'b010; // 32 bit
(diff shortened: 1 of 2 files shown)


rtl/ibex_core.sv
```diff
@@ -1932,10 +1932,13 @@ module ibex_core import ibex_pkg::*; #(
   end
 
 `else
-  logic unused_instr_new_id, unused_instr_id_done, unused_instr_done_wb;
+  logic unused_instr_new_id, unused_instr_id_done, unused_instr_done_wb
+        unused_instr_expanded_id, unused_instr_gets_expanded_id;
```
nasahlpa:
```suggestion
  logic unused_instr_new_id, unused_instr_id_done, unused_instr_done_wb,
        unused_instr_expanded_id, unused_instr_gets_expanded_id;
```

nasahlpa (approved):
Thanks Andreas, adding a `,` should fix the failing CI.

rtl/ibex_compressed_decoder.sv
```diff
@@ -80,7 +80,7 @@ module ibex_compressed_decoder #(
                                                     input logic [4:0] sp_offset);
     logic [11:0] neg_offset;
     logic [31:0] instr;
-    neg_offset = ~{5'b00000, sp_offset, 2'b00} + 12'd1;
+    neg_offset = -{5'b00000, sp_offset, 2'b00};
```
vogelpi: I am not sure this is okay. I don't think it complies with our style guide, see https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#signed-arithmetic

The term within `{ }` is probably an unsigned logic and then you invert the sign. How it was before seems safer to me, but I am not sure what the lint warnings/errors were, the width seems fully defined for both operands.

rtl/ibex_compressed_decoder.sv
```diff
@@ -108,7 +108,7 @@ module ibex_compressed_decoder #(
     logic [31:0] instr;
     imm[11:7] = '0;
     imm[ 6:0] = cm_stack_adj(.rlist(rlist), .spimm(spimm));
-    if (decr) imm = ~imm + 12'd1;
+    if (decr) imm = -imm;
```
vogelpi: This doesn't look right, you're changing the sign of an unsigned logic, see https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#signed-arithmetic .

rtl/ibex_compressed_decoder.sv
```diff
@@ -57,9 +57,9 @@ module ibex_compressed_decoder #(
   function automatic logic [4:0] cm_stack_adj_word(input logic [3:0] rlist,
                                                    input logic [1:0] spimm);
     logic [6:0] tmp;
-    logic [1:0] _unused;
+    logic [1:0] unused_;
     tmp = cm_stack_adj(.rlist(rlist), .spimm(spimm));
-    _unused = tmp[1:0];
+    unused_ = tmp[1:0];
```
vogelpi: Would it hurt to to just name this `unused_tmp`?

rtl/ibex_compressed_decoder.sv
marnovandermaas: So according to the style guide maybe this would be a good substitution?
```
logic signed [11:0] neg_offset;
logic [31:0] instr;
neg_offset = 12'sh000 - signed'({5'b00000, sp_offset, 2'b00});
```

andreaskurth: Thanks for your feedback. The lint error that gets fixed here is on the explicit calculation of a two's complement, in the form of `~value + 1`. I replaced this with `-value`, as suggested in the linter's manual. I also ran it through simulation to check that the result is correct.

The problem with a signed variable -- that is
```sv
logic signed [11:0] neg_offset;
neg_offset = -signed'({5'b00000, sp_offset, 2'b00});
```
is that the linter then complains about the subsequent part selects (e.g., `neg_offset[4:0]`) of that signed variable.

rtl/ibex_compressed_decoder.sv
andreaskurth: This is an application of the linter's suggestion, and it behaves correctly in simulation.

If I make `imm` signed, the linter flags the subsequent assignment to an unsigned target (`instr[31:20]`) as well as the part selects.

rtl/ibex_compressed_decoder.sv
andreaskurth: Converting back and forth between signed and unsigned should make everyone happy. Not pretty, but let's go with that.

rtl/ibex_core.sv
andreaskurth: Thx, fixed.

rtl/ibex_compressed_decoder.sv
andreaskurth: Thx, done.

PR #2364, opened 2026-02-03
[rtl] Split data and ECC of the register file

nasahlpa:
Until now, a single instance of the register file containing 32-bit data words and 7-bit ECC tags was used. While the input of the register file was driven by the main core, the output of the register file was distributed to the main and the shadow core.

In this PR, we are splitting up the data and ECC parts into two different register file instances:

1. This instance is driven by the main core and only operates on the 32-bit data words. The outputs (32-bit data words) are forwarded to the main and the shadow core.
2. This instance is driven by the shadow core and only operates on the 7-bit ECC words. The 7-bit ECC output is combined with the delayed 32-bit data output of the (1) RF instance. The shadow core uses ECC checkers to check if data and ECC (a) match and (b) are not manipulated using FI.

This helps us to save around 6 kGE of area.

### DIFF
--- a/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
+++ b/dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
@@ -132,12 +132,15 @@ class core_ibex_rf_intg_test extends core_ibex_base_test;
     int    rnd_delay;
     bit    port_idx;
     string port_name;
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks @nasahlpa , I've reviewed the RTL and it's identical to what I've reviewed before, LGTM!

PR #2367, opened 2026-02-04
RAM delay in Simple system

marnovandermaas:
The aim of this pull request is to make it possible to do bench-marking on Simple System where the instruction cache actually makes a difference. The RAM must thus respond with a delay on the instruction side.

### DIFF
--- a/examples/simple_system/rtl/ibex_simple_system.sv
+++ b/examples/simple_system/rtl/ibex_simple_system.sv
@@ -22,6 +22,10 @@
   `define RegFile ibex_pkg::RegFileFF
 `endif
 
+`ifndef INSTR_CYCLE_DELAY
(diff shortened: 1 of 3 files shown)


examples/simple_system/rtl/ibex_simple_system.sv
```diff
@@ -274,6 +274,7 @@ module ibex_simple_system (
   // SRAM block for instruction and data storage
   ram_2p #(
       .Depth(1024*1024/4),
+      .BExtraDelay(5),
```
luismarques: How hard would it be to move this to a symbolic constant that we can override in the command line?

marnovandermaas: I've moved it to a FuseSoC command line.

PR #2374, opened 2026-03-18
[rtl] Flush Zcmp state machine on exception or interrupt and make the committing instructions atomic

SamuelRiedel:
Correctly reset the Zcmp expanded instruction state machine when a trap (exception or interrupt) occurs.

Previously, if a trap interrupts the execution of expanded instructions, the state machine retains its progress. This stale state will cause the processor to incorrectly resume or misinterpret the next Zcmp instruction, leading to an incorrect execution. This bug was reported by "Xingzhi Zhang" (https://github.com/zxz2004626).

I tested this fix manually for now. We still need to implement dedicated regression tests in the future.

### DIFF
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -25,6 +25,7 @@ module ibex_compressed_decoder #(
   output logic [31:0]          instr_o,
   output logic                 is_compressed_o,
   output ibex_pkg::instr_exp_e gets_expanded_o,
+  input  logic                 flush_expanded_i,
   output logic                 illegal_instr_o
 );
   import ibex_pkg::*;
@@ -799,7 +800,7 @@ module ibex_compressed_decoder #(
     if (!rst_ni) begin
       cm_state_q <= CmIdle;
     end else begin
-      cm_state_q <= cm_state_d;
+      cm_state_q <= flush_expanded_i ? CmIdle : cm_state_d;
     end
   end
 
--- a/rtl/ibex_compressed_decoder.sv
+++ b/rtl/ibex_compressed_decoder.sv
@@ -671,6 +671,8 @@ module ibex_compressed_decoder #(
                       instr_o = cm_sp_addi(.rlist(instr_i[7:4]),
                                           .spimm(instr_i[3:2]),
                                           .decr(1'b0));
+                      // Ensure the SP adjustment commits atomically with the subsequent micro-ops.
+                      gets_expanded = INSTR_EXPANDED_COMMIT;
                       if (id_in_ready_i) begin
                         unique case (instr_i[12:8])
                           5'b11100: cm_state_d = CmPopZeroA0; // cm.popretz
@@ -685,6 +687,8 @@ module ibex_compressed_decoder #(
                     end
                     CmPopZeroA0: begin
                       instr_o = cm_zero_a0();
+                      // Ensure the `ret` after is executed atomically with this one.
+                      gets_expanded = INSTR_EXPANDED_COMMIT;
                       if (id_in_ready_i) begin
                         cm_state_d = CmPopRetRa;
                       end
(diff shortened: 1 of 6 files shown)


SamuelRiedel:
@zxz2004626 Can you please verify that those changes fix the bugs also on your side?

zxz2004626:
> @zxz2004626 Can you please verify that those changes fix the bugs also on your side?

Sure, I will test it and report to you later.

zxz2004626:
Hi @SamuelRiedel , I carefully verified these patches. Your patches fix most of the problems. The state machine is capable of successfully resetting itself when encountering interruptions or exceptions. Thank you for your quick response and serious attitude!

However, there still some subtle problems. The spec ([cm.popret](https://docs.riscv.org/reference/isa/unpriv/zc.html#insns-cm_popret) and [cm.popretz](https://docs.riscv.org/reference/isa/unpriv/zc.html#insns-cm_popretz)) statement 
_"The final section of pseudocode executes atomically, and only executes if the section above completes without any exceptions or interrupt"_ 
requires that adjusting the stack register, writing to the a0 register and returning should be executed atomically. The state machine transient sequence is `CmIdle`->`CmPopLoadReg`->`CmPopIncrSp`->`CmPopZeroA0`->`CmPopRetRa`. If the state machine is interrupted and reset after completing the `CmPopIncrSp`, the `sp` will be adjusted two times after the replay finished. I have also built a PoC for this and proved it. So `CmPopIncrSp`->`CmPopZeroA0`->`CmPopRetRa` should be executed atomically.

I have some improvements for this. You could see my [PR](https://github.com/SamuelRiedel/ibex/pull/1) in your fork.

SamuelRiedel:
Thank you for the verification. And you're right, we'll fix that as well.

zxz2004626:
Hi @SamuelRiedel,

I would like to know if the fix for this bug is still being worked on.  Have there been any suggestions to my [PR](https://github.com/SamuelRiedel/ibex/pull/1) ? I am more than willing to assist in advancing this matter.

SamuelRiedel:
Hi @zxz2004626 

Sorry for the delay, this got a bit lost. We implemented a fix that also ensures the mv operations were not interruptable and that the debugging also works correctly. Finally, we also added a quick patch to ensure those instructions are correctly counted in the instruction retired counter. Feel free to try it out. I'll ask for feedback and hopefully we can merge this soon and close this issue.

rtl/ibex_controller.sv
```diff
@@ -410,8 +414,10 @@ module ibex_controller #(
   // - while in NMI mode (nested NMIs are not supported, NMI has highest priority and
   //   cannot be interrupted by regular interrupts),
   // - while single stepping.
+  // - while the atomic committing instructions of a Zcmp sequence.
   assign handle_irq = ~debug_mode_q & ~debug_single_step_i & ~nmi_mode_q &
-      (irq_nm | (irq_pending_i & irq_enabled));
+      (irq_nm | (irq_pending_i & irq_enabled)) &
+      !(instr_gets_expanded_i == INSTR_EXPANDED_COMMIT);
```
gautschimi, line 420: interrupts can be handled during `INSTR_EXPANDED` because the instruction does not yet have an internal state, right?

rtl/ibex_compressed_decoder.sv
gautschimi, line 578: should this not be `cm_rlist_d = cm_rlist_q - 5'd1;`

gautschimi (commented):
looks good to me.

One comment is not really relevant to this PR but maybe you can comment on it.

gautschimi (approved):

rtl/ibex_controller.sv
SamuelRiedel: Yes, exactly. Otherwise, the state is `INSTR_EXPANDED_COMMIT`.

rtl/ibex_compressed_decoder.sv
SamuelRiedel: No, because we directly do the decrement for the next state here.

marnovandermaas (commented):
This is a good change to go in. Should we open an issue to track that we eventually need to test this in the DV environment as well? Should we document how to do the manual testing in the meantime?

SamuelRiedel:
Thank you both for the reviews. @marnovandermaas, yes definitely. I'll create an issue reporting what I did so far and with a proposal on how to implement something in riscv-dv.

zxz2004626:
I think the PR is feasible. Thank you all for your contribution!

PR #2376, opened 2026-04-09
[icache/rtl] Tweak infection

nasahlpa:
Before writing data and tag into the data and tag banks, XOR a tweak on top. When reading back the data and tag from the banks, un-XOR them using the tweak.


vogelpi (approved):
This PR is identical to the one that I've already reviewed in depth previously.

LGTM, thanks @nasahlpa !

PR #2380, opened 2026-04-11
Fix: Ensure safe control signal reset on illegal instruction

Anubhav-30:
This PR improves decoder robustness by ensuring all relevant control
signals are safely reset when an illegal instruction is detected.

Previously, signals like rf_ren_a_o, rf_ren_b_o, icache_inval_o and csr_op_o
could retain stale values, potentially causing unintended hardware behavior.

This change aligns the decoder with safe hardware design practices
and improves reliability.


SamuelRiedel:
Duplicate of https://github.com/lowRISC/ibex/pull/2384

PR #2381, opened 2026-04-11
Add: Basic RAW hazard detection in decoder

Anubhav-30:
This PR introduces a simple RAW (Read After Write) hazard detection
mechanism in the decoder.

The added logic detects when the current instruction reads a register
that is also being written, which can lead to incorrect data usage
in pipelined execution.

Currently, this change only exposes a hazard_detected_o signal and
does not implement stalling or forwarding. This provides a clean
foundation for future pipeline hazard handling.

This improves pipeline awareness and aligns with standard CPU
design practices.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -96,6 +96,7 @@ module ibex_decoder #(
   // jump/branches
   output logic                 jump_in_dec_o,         // jump is being calculated in ALU
   output logic                 branch_in_dec_o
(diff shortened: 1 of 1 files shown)


SamuelRiedel:
The proposed logic only checks whether a single instruction reads and writes to the same register. Ibex already handles RAW hazards. Furthermore, this PR introduces syntax errors. Please ensure all future contributions pass basic tests.

PR #2382, opened 2026-04-11
Add: Pipeline stall request signal for RAW hazard handling in decoder

Anubhav-30:
This PR enhances the Ibex decoder by introducing a stall request signal
(stall_req_o) based on RAW (Read After Write) hazard detection.

When a hazard is detected, stall_req_o is asserted to indicate that the
pipeline should be temporarily stalled to prevent incorrect operand usage
due to data dependencies between instructions.

This change builds upon basic hazard detection and provides a clean,
non-intrusive interface for future pipeline control improvements such as
stalling and forwarding.

Key points:
- Adds stall_req_o signal for pipeline control
- Ensures safer execution under data dependency conditions
- Does not modify existing pipeline behavior
- Maintains compatibility with current design

This contribution improves pipeline reliability and aligns with standard
CPU microarchitecture design practices.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -96,6 +96,7 @@ module ibex_decoder #(
   // jump/branches
   output logic                 jump_in_dec_o,         // jump is being calculated in ALU
   output logic                 branch_in_dec_o
+  output logic                 stall_req_o
 );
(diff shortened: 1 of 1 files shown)


SamuelRiedel:

PR #2383, opened 2026-04-11
decoder: Add hazard indication, stall request and forwarding hint signals

Anubhav-30:
This change extends the decoder with additional signals to improve
pipeline hazard visibility and provide a foundation for future
pipeline control enhancements.

The following signals are introduced:
- hazard_detected_o: Indicates RAW (Read After Write) hazards based on
  register dependencies within the decoder stage
- stall_req_o: Signals when a pipeline stall may be required due to
  detected hazards
- forward_rs1_o / forward_rs2_o: Provide forwarding hints for source
  operands to support future data forwarding mechanisms

The implementation is non-intrusive and does not alter existing
pipeline behavior or timing. All signals are purely informational
and intended for integration with higher-level pipeline control
logic in future work.

This improves observability of data hazards at the decoder level
and aligns with common CPU microarchitecture practices.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -96,6 +96,9 @@ module ibex_decoder #(
   // jump/branches
   output logic                 jump_in_dec_o,         // jump is being calculated in ALU
   output logic                 branch_in_dec_o
+  output logic stall_req_o,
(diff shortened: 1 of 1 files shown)


SamuelRiedel:
Duplicate of #2381

PR #2384, opened 2026-04-12
decoder: Ensure safe handling of illegal instructions by blocking all side effects

Anubhav-30:
This change strengthens the handling of illegal instructions in the decoder
by ensuring that no unintended side effects are triggered when an illegal
instruction is detected.

In addition to disabling write and control signals, this update explicitly:
- disables register read enables (rf_ren_a_o, rf_ren_b_o)
- blocks instruction cache invalidation (icache_inval_o)
- forces CSR operations to a safe read-only mode

This prevents any unintended data access, control flow changes, or CSR
side effects from occurring due to partially decoded illegal instructions.

The change is non-intrusive and aligns with safe CPU design practices,
ensuring that illegal instructions are fully contained within the decoder
stage without affecting downstream pipeline behavior.

### DIFF
--- a/rtl/ibex_decoder.sv
+++ b/rtl/ibex_decoder.sv
@@ -656,16 +656,24 @@ module ibex_decoder #(
     // insufficient privileges), or when accessing non-available registers in RV32E,
     // these cases are not handled here
     if (illegal_insn) begin
-      rf_we           = 1'b0;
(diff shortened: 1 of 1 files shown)


SamuelRiedel:
This PR gates signals that do not actually produce any architectural side effects.

PR #2386, opened 2026-04-12
decoder: Prevent memory side-effects on taken branch and add verification testbench

Anubhav-30:
This PR improves pipeline safety by ensuring that no memory
requests are issued when a branch is taken.

This avoids unintended memory accesses during control flow changes.

A testbench is added to verify that data_req_o and data_we_o
are correctly disabled when branch_taken_i is asserted.

This aligns the decoder behavior with real hardware expectations
and improves pipeline correctness.

### DIFF
--- a/dv/dv/tb_branch_flush.sv
+++ b/dv/dv/tb_branch_flush.sv
@@ -1,0 +1,50 @@
+`timescale 1ns/1ps
+
+module tb_branch_flush;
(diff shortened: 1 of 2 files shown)


SamuelRiedel:
The gated signals are already tied to zero for branches, meaning this PR does not introduce any functional change.

PR #2388, opened 2026-04-13
Add Load-Use Hazard Detection and Stall Mechanism in Decoder

Anubhav-30:
This PR adds load-use hazard detection in the decoder to prevent incorrect execution due to data dependencies.
🔧 Changes
Introduced load_use_hazard signal to detect RAW dependency after load instructions
Added control gating to suppress:
register write (rf_we)
memory access (data_req_o, data_we_o)
branch/jump signals
Ignored x0 register to avoid false hazards
🧪 Verification
Added assertion to ensure no register write during hazard
Included testbench to validate stall behavior
📈 Impact
Improves pipeline correctness and prevents execution errors caused by load-use hazards.

### DIFF
--- a/dv/tb_load_use_hazard.sv
+++ b/dv/tb_load_use_hazard.sv
@@ -1,0 +1,47 @@
+module tb_load_use_hazard;
+
+  logic clk;
+  logic rst_n;
+
+  // Inputs
+  logic [31:0] instr;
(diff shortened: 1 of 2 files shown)


SamuelRiedel:
Duplicate of https://github.com/lowRISC/ibex/pull/2381

PR #2390, opened 2026-04-14
Enforce x0 Register Write Protection in Decoder with Verification Testbench

Anubhav-30:
This PR ensures compliance with RISC-V specification by preventing writes to the x0 register.

- Updated write enable logic to block writes when destination register is x0
- Maintains existing illegal instruction protections
- Added System Verilog testbench to verify behavior

This improves architectural correctness and prevents silent data corruption.


SamuelRiedel:

PR #2395, opened 2026-04-17
Disable register file reads for LUI instruction in decoder

Anubhav-30:
Ensure that LUI instructions do not trigger register file reads, as they do not use rs1 or rs2.

This improves decoder correctness and avoids unnecessary register access. 
Includes a simple testbench for validation.


SamuelRiedel:

Issue #2288, opened 2025-06-23
make run-simple-system failure


ekigwana:
## Observed Behavior
make run-simple-system fails to compile simple program

ibex/examples/sw/simple_system/common/simple_system_common.c:40: Error: unrecognized opcode `csrw minstret,x0', extension `zicsr' required

## Expected Behavior
make run-simple-system compiles simple program

## Steps to reproduce the issue
# Setup virtual environment

python -m venv ~/.ibex-eda/

. ~/.ibex-eda/bin/activate

pip install mako
pip install packaging

# build system

. ~/.ibex-eda/bin/activate
git clone https://github.com/olofk/edalize.git
cd edalize
pip install .
cd ..

git clone https://github.com/lowRISC/fusesoc.git
cd fusesoc
pip install .
cd ..

git clone https://github.com/lowRISC/ibex.git

./util/ibex_config.py small fusesoc_opts

`--RV32E=0 --RV32M=ibex_pkg::RV32MFast --RV32B=ibex_pkg::RV32BNone --RegFile=ibex_pkg::RegFileFF --BranchTargetALU=0 --WritebackStage=0 --ICache=0 --ICacheECC=0 --ICacheScramble=0 --BranchPredictor=0 --DbgTriggerEn=0 --SecureIbex=0 --PMPEnable=0 --PMPGranularity=0 --PMPNumRegions=4 --MHPMCounterNum=0 --MHPMCounterWidth=40`

./util/check_tool_requirements.py
`INFO: Tool verilator present: Sufficiently recent version (found 5.028; needed 4.210)
INFO: Tool edalize present: Sufficiently recent version (found 0.6.2.dev2+g1eedb7f; needed 0.2.0)`

make build-simple-system
make run-simple-system

As a quick aside having the above steps would be helpful in the documentation.

No modifications to any sources.
No parameters set.

## My Environment
Gentoo
riscv32-unknown-elf-gcc: gcc version 15.1.0 (Gentoo 15.1.0 p55)

Using Ibex as a micro controller in an FPGA

**EDA tool and version:**
Xilinx Vivado 2025.1

**Operating system:**
Gentoo

**Version of the Ibex source code:**
git master db07ab174e336d6feb82c96249539badd19f0d6c

no modifications to source

PR #2398, opened 2026-04-24
[dv] Add RV32ZC handling to ISA strings

AnastasiyaChernikova:
Use the RV32ZC configuration parameter when constructing the ISA string passed to the cosimulator. This lets DV runs distinguish between Zca, ZcaZcb, ZcaZcmp and ZcaZcbZcmp configurations.

### DIFF
--- a/dv/uvm/core_ibex/scripts/ibex_cmd.py
+++ b/dv/uvm/core_ibex/scripts/ibex_cmd.py
@@ -91,13 +91,27 @@ def get_config(cfg_name: str) -> Config:
     return parse_config(cfg_name, yaml_path)
 
 
+def _rv32zc_isa_parts(rv32zc: str) -> list[str]:
+    mapping = {
+        'ibex_pkg::RV32Zca': ['zicsr', 'zifencei', 'zca'],
+        'ibex_pkg::RV32ZcaZcb': ['zicsr', 'zifencei', 'zca', 'zcb'],
+        'ibex_pkg::RV32ZcaZcmp': ['zicsr', 'zifencei', 'zca', 'zcmp'],
+        'ibex_pkg::RV32ZcaZcbZcmp': ['zicsr', 'zifencei', 'zca', 'zcb', 'zcmp'],
+    }
+    part = mapping.get(rv32zc)
+    if part is None:
+        raise ValueError(f'Unknown RV32ZC value ({rv32zc}) in config YAML')
+    return part
+
+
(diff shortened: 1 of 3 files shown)


AnastasiyaChernikova:
Hi, @nasahlpa!
To simplify future testing of Ibex, I suggest adding the ability to select different extensions at startup. Please review this commit. Also, unfortunately, I can't view the hidden stages of your pipeline to fix the crash.
Thanks in advance!

SamuelRiedel:
Thanks @AnastasiyaChernikova for looking into this. Your changes look good at first glance to me. However, we won't be able to merge them until we update our compiler version. Some changes are similar to commits in the PR adding DV for the Zc* extensions (https://github.com/lowRISC/ibex/pull/2324), but as noted there, this cannot be merged until we update the compiler. For more context please also check out this comment on why the `Zicsr` and `Zifencei` flags currently lead to CI failing: https://github.com/lowRISC/ibex/issues/2288#issuecomment-3481512040

I am afraid we cannot merge this currently, but I'll leave the PR open and revisit it once we update the compiler.

AnastasiyaChernikova:
@SamuelRiedel, thanks for the explanation and for keeping the PR open.

I understand that this is currently blocked by the toolchain/compiler version used in CI rather than by the patch itself, and I would like to help unblock it if possible.
For reference, I tested this locally with the lowRISC toolchain release lowrisc-toolchain-rv32imcb-x86_64-20250710-1, and with that toolchain the build and tests I ran complete successfully. So I am not currently able to reproduce the CI failure locally.
The main difficulty for me is that I do not have access to the full CI pipelines/logs, so I cannot see which exact stage is failing or what additional constraints need to be satisfied before the toolchain can be updated.
I could contribute to the toolchain migration if that would be useful — for example by preparing a follow-up PR, adjusting the ISA flags, updating the affected DV/simple-system flows, or running additional checks locally. At the moment I am mostly missing context on what the remaining difficulties are and what still needs to be done before these changes can be merged.
It would be great to find a way forward here, because this change allows the DV flow to distinguish the concrete RV32ZC configuration passed to the cosimulator. In particular, this is critical for my current work, so I would really appreciate any guidance on the remaining blockers and on whether there is a useful way for me to contribute to resolving them.

AnastasiyaChernikova:
Hi @SamuelRiedel, sorry for the ping.

Just gently following up on this PR. Is there any update on the toolchain/compiler blocker, or anything I could help with from my side?

This is still relevant for my current work, so I would appreciate any guidance on the next steps.

Thanks!

PR #2403, opened 2026-05-11
Add `mcounteren` register

### LINKED ISSUE #1973: [rtl] Implement cycle/instret/time/hmpcountern CSRs
The cycle/instret/time/hmpcountern CSRs are optionally accessible from U-mode. In Ibex we hardwire `mcounteren` to 0 meaning these counters cannot be accessed from U-mode. However they are still accessible from M-mode. Ibex currently doesn't implement them at all which isn't standards compliant. They should be implemented.

### LINKED ISSUE #2312: [Bug,Test] Zicntr in CoSim not compliant with the standard
## Observed Behavior

I encountered some cosim mismatches around operations on CSR `instret` (user mode delegate of `minstret`), which is not (planned to be) implemented (according to #1973 ).
```
FAILURE: Co-simulation mismatch at time                   36
PC mismatch, DUT retired : 100000 , but the ISS retired: 100008
[36] %Error: ibex_simple_system_cosim_checker.sv:74: Assertion failed in TOP.ibex_simple_system.u_ibex_simple_system_cosim_checker_bind: Co-simulation mismatch seen
Received stop request, shutting down simulation.
```
trace_core_00000000.log:
```
Time    Cycle   PC      Insn    Decoded instruction     Register and memory contents
             30         11      00100004        c0206073        csrrsi  x0,instret,0      x0=0x00000000
=> trap IllegalInsn here
             36         14      00100000        0670    c.addi4spn      x12,x2,780       x12=0x0000030c
```

Image

## Expected Behavior

As the project has not yet planned to impl Zicntr, spike used as reference model should be aligned.
Currently the fork lowrisc/riscv-isa-sim is far behind the upstream. While the upstream has made Zicntr optional regarding the ISA string, there exists a fragment in the fork that always enable Zicntr:
```c
riscv/isa_parser.cc:L37
// enable zicntr and zihpm unconditionally for backward compatibility
  extension_table[EXT_ZICNTR] = false;
```

1. Build lowrisc:ibex:ibex_simple_system_cosim
2. Run Cosim with the provided testcase binary (which may need conversion? bin is not officially supported, I run with my modified simulating c++ code) build/lowrisc_ibex_ibex_simple_system_cosim_0/sim-verilator/Vibex_simple_system -c 10000 --meminit=ram,errors/ba8335cdf5815390859b835ec904d458
3. Mismatch reported

[ibex_zicntr_unimpl.zip](https://github.com/user-attachments/files/21960935/ibex_zicntr_unimpl.zip)

## My Environment

**EDA tool and version:**

Verilator 5.006 2023-01-22 rev (Debian 5.006-3)

**Operating system:**

master (commit 587e9fe47ddf54a387b7976f6f720455f5a0b95d)

SamuelRiedel:
This PR adds the `mcounteren` register, which has so far been tied to zero and implements the U-mode performance counter aliases. The `mcounteren` register can be locked with an external MUBI signal (`mcounteren_writable_i`).

**DV:**
This PR adds two directed tests:
- `mcounteren_test` goes through multiple configurations of the mcounteren register and validates that u-mode can only access the ones enabled.
- `mcounteren_lock_test` verifies that the mcounteren CSR cannot be modified without the `mcounteren_writable_i` signal being set.

**To Dos:**
- [x] For the directed tests and the cosim to pass, we have to update riscv-isa-sim to also support those performance counters: https://github.com/lowRISC/riscv-isa-sim/pull/29
- [x] Currently, u-mode counter reads will lead to illegal instruction exceptions. However, we are not gating the value at the CSR module's interface for simplicity. The signal will not leave the core, but we should double-check that this is no security issue.
- [x] The `dv/uvm/core_ibex/directed_tests/directed_testlist.yaml` changes a lot everytime I regenerate it after even small modifications. This polluted the diff quite a bit. Is this expected?

resolves https://github.com/lowRISC/ibex/issues/1973
fixes https://github.com/lowRISC/ibex/issues/2312


rtl/ibex_cs_registers.sv
```diff
@@ -507,6 +507,35 @@ module ibex_cs_registers import ibex_pkg::*; #(
...
+      CSR_INSTRET,
+      CSR_HPMCOUNTER3,
+      CSR_HPMCOUNTER4,  CSR_HPMCOUNTER5,  CSR_HPMCOUNTER6,  CSR_HPMCOUNTER7,
+      CSR_HPMCOUNTER8,  CSR_HPMCOUNTER9,  CSR_HPMCOUNTER10, CSR_HPMCOUNTER11,
+      CSR_HPMCOUNTER12, CSR_HPMCOUNTER13, CSR_HPMCOUNTER14, CSR_HPMCOUNTER15,
+      CSR_HPMCOUNTER16, CSR_HPMCOUNTER17, CSR_HPMCOUNTER18, CSR_HPMCOUNTER19,
+      CSR_HPMCOUNTER20, CSR_HPMCOUNTER21, CSR_HPMCOUNTER22, CSR_HPMCOUNTER23,
+      CSR_HPMCOUNTER24, CSR_HPMCOUNTER25, CSR_HPMCOUNTER26, CSR_HPMCOUNTER27,
+      CSR_HPMCOUNTER28, CSR_HPMCOUNTER29, CSR_HPMCOUNTER30, CSR_HPMCOUNTER31: begin
+        csr_rdata_int = mhpmcounter[mhpmcounter_idx][31:0];
```
gautschimi, line 524: should we assign this to zero if the condition below does not match?

rtl/ibex_top.sv
```diff
@@ -296,6 +298,11 @@ module ibex_top import ibex_pkg::*; #(
     .out_o(fetch_enable_buf)
   );
 
+  prim_buf #(.Width($bits(ibex_mubi_t))) u_mcounteren_writable_buf (
```
gautschimi, line 301: what is this buffer used for?

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_test.S
```diff
@@ -0,0 +1,224 @@
+# Copyright lowRISC contributors.
```
gautschimi, line 1: This test is good! It verifies that the counters are accessible or not depending on mcounteren.

Maybe we can add a short test that also verifies that the read values are correct.

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
```diff
@@ -60,6 +60,14 @@
   test_srcs: empty/empty.S
```
gautschimi, line 60: did something change in this file (except for the new mcounteren_test)? The diff is big. Did the order of the tests change or was something real added?

gautschimi (commented):
The changes look all good to me! Just a few minor comments. (I did not look at the two WIP commits about formal)

rtl/ibex_cs_registers.sv
SamuelRiedel: I was wondering the same. The values currently never make it to the register file because we trap, but we could also gate it to prevent any potential security concerns.

rtl/ibex_top.sv
SamuelRiedel: I was following the `fetch_enable` example here, which says to buffer security critical signals to prevent synthesis optimisation removing them.

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_test.S
SamuelRiedel: I can add this, that should not be difficult.

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
SamuelRiedel: No, this is what I was also confused about. The order seems to change every time I add a test or even modify the description of the test. This makes it quite hard to check the diff. I was wondering whether someone with more experience with this tool had any insights on this.

SamuelRiedel:
I disabled the check of the mcounteren CSR in the formal verification for now. Another reason for the failure of the formal verification was that we ran out of memory on the runner that has twice as many cores but half as much memory available. I'll address this in a separate PR and rebase once merged.

rtl/ibex_cs_registers.sv
gautschimi: Ok, I thought so. If the write enable to the register file is disabled, I think we are good

rtl/ibex_top.sv
gautschimi: ok I see, lets keep it. doesn't hurt

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
gautschimi: ok do you know who to contact?

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_test.S
SamuelRiedel: This is added now

dv/uvm/core_ibex/directed_tests/directed_testlist.yaml
SamuelRiedel: Not besides the people I asked for a review here. But I investigated this myself and was able to fix the python script here: https://github.com/lowRISC/ibex/commit/42f3a007e960c6fbd57580cb3f5959bbec34c2f7

The big diff should now be gone.

gautschimi: very nice!

gautschimi (approved):
Looks good to me!

I mostly looked at RTL, but the changes in python, C, yaml, etc. look also good to me!

rtl/ibex_cs_registers.sv
```diff
@@ -258,6 +258,11 @@ module ibex_cs_registers import ibex_pkg::*; #(
   logic [MHPMCounterNum+3-1:0] mcountinhibit_d, mcountinhibit_q;
   logic                        mcountinhibit_we;
 
+  // mcounteren: machine counter enable (controls U-mode counter access)
+  logic [31:0]                 mcounteren;
+  logic [MHPMCounterNum+3-1:0] mcounteren_d, mcounteren_q;
```
marnovandermaas: It probably makes sense to have at least a local parameter with the 3 offset since this is shared with the mcountinhibit above.

rtl/ibex_pkg.sv
marnovandermaas, line 1: I double checked this mapping and I think they are right.

dv/formal/check/top.sv
```diff
@@ -103,6 +103,7 @@ module top import ibex_pkg::*; #(
 
   // CPU Control Signals
   input  ibex_mubi_t                                                   fetch_enable_i,
+  input  ibex_mubi_t                                                   mcounteren_writable_i,
```
marnovandermaas, line 106: What's the point of making this an input as opposed to a parameter? Is there a use-case to be able to switch this dynamically during runtime?

ci/vars.env
```diff
@@ -6,7 +6,7 @@
 # Quote values to ensure they are parsed as string (version numbers might
 # end up as float otherwise).
 VERILATOR_VERSION=v4.210
-IBEX_COSIM_VERSION=6d5b660
+IBEX_COSIM_VERSION=aadf648
```
marnovandermaas, line 9: This bump adds ZIHPM extension to Spike but it also pulls in the Zc* extensions. Is there any risk in that inclusion?

marnovandermaas (commented):
Since we're adding an extension do we need to do an update of ISA strings that are passed to compilers?

dv/uvm/core_ibex/directed_tests/gen_testlist.py
```diff
@@ -469,11 +469,11 @@ def _main() -> int:
     add_configs_and_handwritten_directed_tests()
 
     if 'riscv-tests' in test_suite_list or test_suite == 'all':
-        isa_tests = {'rv32mi', 'rv32uc', 'rv32ui', 'rv32um'}
+        isa_tests = ['rv32mi', 'rv32uc', 'rv32um', 'rv32ui']
         append_directed_testlist(isa_tests, '../../../../vendor/riscv-tests/isa/', 'riscv-tests', 1)
 
     if 'riscv-arch-tests' in test_suite_list or test_suite == 'all':
-        arch_tests = {'rv32i_m/B/src', 'rv32i_m/C/src', 'rv32i_m/I/src', 'rv32i_m/M/src', 'rv32i_m/Zifencei/src'}
+        arch_tests = ['rv32i_m/M/src', 'rv32i_m/C/src', 'rv32i_m/Zifencei/src', 'rv32i_m/I/src',  'rv32i_m/B/src']
```
marnovandermaas: Nit extra space here.

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_test.S
marnovandermaas, line 1: Thanks for adding this test, very useful!

dv/cosim/cosim.h
```diff
@@ -128,6 +128,17 @@ class Cosim {
   // A full 64-bit value is provided setting both the mcycle and mcycleh CSRs.
   virtual void set_mcycle(uint64_t mcycle) = 0;
 
+  // Set the value of minstret.
+  //
+  // The co-simulation model doesn't alter the value of minstret itself (other
```
marnovandermaas: How come this is true? Spike should have all the information it needs to calculate minstret right?

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_lock_test.S
```diff
@@ -0,0 +1,62 @@
...
+  # Small delay loop to let the hardware pin force propagate
+  .rept 5
+  nop
+  .endr
+
+  # Try to overwrite mcounteren with 0x0 while locked
+  li   s1, 0x0
+  csrw mcounteren, s1
+  csrr t1, mcounteren
+  bne  t1, s0, fail
```
marnovandermaas, line 39: Should this be s1?

marnovandermaas, line 39: Also shouldn't this one fail if it is equal?

dv/uvm/core_ibex/tests/core_ibex_test_lib.sv
```diff
@@ -2025,3 +2025,44 @@ class core_ibex_assorted_traps_interrupts_debug_test extends core_ibex_directed_
    endtask
 
 endclass
+
+class core_ibex_mcounteren_lock_test extends core_ibex_base_test;
```
marnovandermaas, line 2029: Thanks for adding this test, this is great!

marnovandermaas (commented):
I realised there are some more commits so here are some more comments. Very good work on this and please disregard my previous mention that the docs needed updating.

dv/formal/check/top.sv
SamuelRiedel: Yes. This allows locking the `mcounteren` in a certain configuration. For example, the boot code could enable only the `instret` counter to be accessible to the user mode and then lock it to prevent the software from switching it later and use other counters to profile the core. This option, gives us the most flexibility and security.

ci/vars.env
SamuelRiedel: You're right. But that will be fine, because we anyway already have the Zc* extension in the RTL. Only the DV is not fully merged yet, but already having Spike support those extensions will not hurt.

dv/cosim/cosim.h
SamuelRiedel: You're right. And it does, the Spike counter is just lagging one instruction behind Ibex. It should match though, so I'll investigate a bit more...

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_lock_test.S
SamuelRiedel: No, because in this instance, the register should be locked. So the `csrw mcounteren, s1` must have no effect. So the read should return the previous write value.

SamuelRiedel: Wait, they are both the same lines you commented. So here we either **fail** if:

- The current value is not equal to the previous value (`s0=0x5`) from before we locked it and tried to change it to `s1=0`
- The current value is equal to `s1=0`, because this means that the write was successful despite the locking.

I decided to do the first to also catch behaviour where the overwrite had some sort of effect, like changing just a few bits to zero.

rtl/ibex_cs_registers.sv
SamuelRiedel: I added this

dv/cosim/cosim.h
SamuelRiedel: There was an issue with the counter which is fixed here and included in this PR: https://github.com/lowRISC/ibex/pull/2446

I reverted the changes in the cosim.

SamuelRiedel:
@marnovandermaas thank you for your review. Now that the CI works again, this PR would be ready to merge from my side. I included a check for the `minstret` value in the test to directly verify that instructions are counted correctly. This PR now also includes the commits from https://github.com/lowRISC/ibex/pull/2446 because we rely on them to pass the tests. We can either directly merge this one and close https://github.com/lowRISC/ibex/pull/2446 or merge the other one first and then rebase and merge this one.

dv/uvm/core_ibex/directed_tests/mcounteren_test/mcounteren_lock_test.S
marnovandermaas: Good point.

marnovandermaas: That makes sense, thanks

rtl/ibex_cs_registers.sv
marnovandermaas: Looks good now.

ci/vars.env
marnovandermaas: Ok, thanks for checking this.

dv/cosim/cosim.h
marnovandermaas: Thanks!

marnovandermaas (approved):
These changes look good to me now. Thanks for including the instret fixes as well. And for checking everything in CI.

PR #2156, opened 2024-03-27
Proof-of-concept Nix environments for Ibex development

hcallahan-lowrisc:
> This PR is just a PoC, and will likely never be merged. It contains some unrelated changes that I didn't bother completely tidying up, and some short-term fixes to make things work. Hopefully it is obvious which parts are unrelated.

Add a Nix flake to provide development environments for Ibex simulation and synthesis.
The goal of this is to try and utilize the principled properties of Nix (the tool, the language, etc) to easily bootstrap reproducible development environments. This should work on any x86_64-linux system where nix is installable (macOS support is also possible), as there is no dependency on system libraries at all. 

### Precursors
1. Install Nix (I suggest to use this guide: https://zero-to-nix.com/start/install). If you use another installer, you will need to ensure the flakes and nix-command experimental features are enabled.
2. Clone/checkout this branch
```
git clone git@github.com:hcallahan-lowrisc/ibex.git --branch nix_env_init_poc
cd ibex/
```

N.B. The first time running `nix develop` may take a little while to fetch and build dependencies. This is normal (no system library deps etc.), and the second time everything will be cached.

### Demo 1 (verilator sims)
1. Enter a development shell
```
nix develop # This drops you into a subshell with all the deps on your path
# When you are done with the devshell...
exit
```
2. Run some verilator sims (My makefile changes are incomplete, so ensure you clean with `make reset` when switching between sims)
```
# Run the simple_system under verilator
make run-simple-system-hello
# Run coremark under verilator
make run-coremark
# Run a pmp smoketest
make run-pmp-smoke-test
```
### Demo 2 (synthesis)
1. Enter a different development shell, which contains the synthesis deps
```
nix develop .#syn_shell
```
2. Run a yosys synthesis using the Nangate45 open cell library
```
make run_syn_yosys_nangate45
```

PR #2437, opened 2026-05-14
[tooling]: Add Nix devshells for core_ibex functional DV and example synthesis flow

hcallahan-lowrisc:
Add some nix development shells for reproducibly bootstrapping the FOSS project dependencies for the core_ibex UVM environment and the example synthesis flow.

This is a spiritual successor to #2156, which has remained useful as a PoC and should have been completed and merged long ago. The missing link, and what cause me many headaches in the interim period, was modifying the spike derivation to build using a suitably old toolchain and statically-linking some libraries such that host dependencies are minimized and the symbol version floor is low. This allows the library to be successfully linked and remain ABI compatible with proprietary simulators running on older distro releases using relatively-ancient bundled C/Cpp toolchains (such as gcc93).

To get a devShell with deps for the UVM / cosim DV flow in `dv/uvm/core_ibex/`:
```
nix develop .#eda_shell
# Then assuming you already have the proprietary EDA tool on your path / system...
cd dv/uvm/core_ibex
make SIMULATOR=xlm/vcs ITERATIONS=1 SEED=1 TEST=riscv_arithmetic_basic_test
```

To get a devShell for running the example yosys synthesis flow in `syn`:
```
nix develop .#syn_shell
# All the deps for this flow are FOSS, so you can simply do..\
cd syn/
./syn_yosys.sh
```


nix/pythonEnv/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+    "pandas<3.0.0,>=2.1.4",
+    "PyBoolector==3.2.4.20240823.1",
+]
+
+[tool.uv]
+# Uv2nix refuses to apply the sourcePreference = "wheel" to this package
+# unless given this directive. PyBoolector cannot easily be built from
+# sdist as the extension module requires some patches to be applied.
+# (see <boolector>/pypi/build.sh)
+only-binary = ["PyBoolector"] # Dependency of pyvsc, used by riscv-dv
```
SamuelRiedel: nit: missing newline

rtl/ibex_tracer.sv
```diff
@@ -853,18 +853,22 @@ module ibex_tracer (
...
     end
   end
 
   // log execution
   always @(posedge clk_i) begin
     if (rvfi_valid && trace_log_enable) begin
-      static int fh = file_handle;
+
+      int fh;
+      fh = file_handle;
```
SamuelRiedel, line 867: Why was this change necessary?

SamuelRiedel (approved):
Thanks a lot! I used these environments now for a bit in my own workflow and everything worked, so I am really happy with that. I just have a small question and a small nit.

SamuelRiedel:
Ah, and the formal CI should pass when you rebase on master.

rtl/ibex_tracer.sv
hcallahan-lowrisc: The sv2v tool used by the example synthesis flow does not like the explicit static annotation. It was originally added in https://github.com/lowRISC/ibex/pull/2243, but since the variable is implicitly static we can just separate the declaration and the initial assignment and keep both tools happy.

PR #2440, opened 2026-05-22
[rtl] Remove redundant halt_if gating in DECODE state

adsurg:
In the DECODE state of ibex_controller's FSM, halt_if is set in two places when (enter_debug_mode || handle_irq) is true:

  1. An outer conditional gated by (stall || id_wb_pending).
  2. An inner conditional guarded by (!stall && !special_req && !id_wb_pending) that always sets halt_if = 1 when entering DBG_TAKEN_IF or IRQ_TAKEN.

The outer gating is logically redundant — the inner already sets halt_if = 1 whenever (debug || irq) holds and the FSM transitions through the non-special_req path. In the case the outer is meant to cover (special_req high), the value of halt_if is observationally irrelevant: special_req drives retain_id = 1 (controller.sv:577-581), and id_in_ready_o is computed as

  assign id_in_ready_o = ~stall & ~halt_if & ~retain_id;

so retain_id forces id_in_ready_o low independent of halt_if. halt_if has no other consumers (verified: not a module output).

The redundant gating introduced a long combinational arc on the longest topological path:

  special_req → !special_req → AND with (stall||id_wb_pending) and
  (enter_debug_mode||handle_irq) → halt_if → id_in_ready_o → ctrl_fsm_ns

Consolidating halt_if to its simpler condition removes that arc. The change is pure combinational restructuring (no new register, no pipeline-stage change, no functional difference).

Measurements on a small-config Ibex (RV32IM, RV32MFast, RV32Zca, WB=1, BP=1, no PMP/icache/security) synthesised with Yosys + ltp:

  critical_path_gates     119 → 108   (-9.24%)
  fmax estimate @ 0.2ns   42.0 → 46.3 MHz (+10.2%)
  num_cells (generic)     17,846 → 17,840 (-6)
  CoreMark 100i cycles    37,362,985 (unchanged)
  CoreMark CRCs           0xe714 / 0x1fd7 / 0x8e3a (golden)
  return42 / hello / fwd_test  all pass with identical cycle counts

Single block edit; total +13 / -12 lines in ibex_controller.sv.

### DIFF
--- a/rtl/ibex_controller.sv
+++ b/rtl/ibex_controller.sv
@@ -607,9 +607,16 @@ module ibex_controller #(
           end
         end
 
-        // If entering debug mode or handling an IRQ the core needs to wait until any instruction in
-        // ID or WB has finished executing. Stall IF during that time.
-        if ((enter_debug_mode || handle_irq) && (stall || id_wb_pending)) begin
+        // cpu3/D9 T-008: hoist halt_if out from under special_req gating to
+        // shorten the LTP. The decode-time halt_if signal is consumed only by
+        // id_in_ready_o (= ~stall & ~halt_if & ~retain_id). When special_req
+        // is high retain_id is already 1, which forces id_in_ready_o low, so
+        // halt_if's exact value while special_req is high is observationally
+        // irrelevant. Decoupling halt_if from special_req here removes the
+        // special_req -> halt_if leg of the LTP, which arrives late because
+        // special_req depends on csr_pipe_flush -> csr_op_en -> instr_id_done.
+        // Combinational restructuring only; no new register, no new stage.
+        if (enter_debug_mode || handle_irq) begin
           halt_if = 1'b1;
         end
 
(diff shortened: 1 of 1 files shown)


elliotb-lowrisc:
Hi @adsurg, thanks for your contribution, 10% certainly looks like a handy speed-up! We'll hopefully find someone to take a look at it in detail soon.

SamuelRiedel:
Hi @adsurg, thank you for looking into Ibex. I checked the path you mentioned and your changes, and I have a few questions about this change:

- Since `id_in_ready_o` is computed using both `retain_id`, which depends on `special_req`, and `halt_if`, the logic leading into `id_in_ready_o` still includes `special_req`. It isn't clear to me how this restructuring shortens the true critical path.
- Concerning the synthesis, a frequency of around 40 MHz is incredibly low for Ibex. Could you provide the exact synthesis scripts and setup you used to generate these numbers so we can attempt to reproduce them? Which technology did you target, and do you have a detailed report of the critical path.

raylau1:
Closing this due to inactivity.

PR #2444, opened 2026-06-04
Add llvm-snippy random test generator flow

AnastasiyaChernikova:
This patch adds an experimental [llvm-snippy](https://github.com/LLVM-Snippy/llvm-snippy/releases/tag/snippy-3.0) integration to the Ibex DV flow.

The main motivation is to complement the existing riscv-dv tests with another generator that uses a different approach to program generation. riscv-dv remains the primary and well-integrated random instruction generator for Ibex, but llvm-snippy can be useful as an independent source of programs with a different generation profile, including different ways of controlling memory layout, call graph structure, register initialization, and instruction distribution. As a result, it can produce scenarios that are not necessarily well covered by the existing riscv-dv tests, or that would require additional dedicated riscv-dv configuration.

The practical value of this approach has already been demonstrated: during experiments with Snippy, an issue was found in the implementation of the Zcmp extension, described in issue [#2445](https://github.com/lowRISC/ibex/issues/2445). This shows that an independent generator can be useful not as a replacement for riscv-dv, but as a complementary source of test programs that increases the chance of finding corner cases in new or complex ISA extensions.

The integration is implemented as a separate Snippy flow and should not change the behavior of regular riscv-dv runs. Snippy scenarios are described by separate YAML files, can be run through the Makefile bridge, and reuse the existing Ibex infrastructure for build, simulation, and coverage. This makes it possible to use Snippy as an additional tool for targeted/random testing without polluting the main riscv-dv flow.

In its current form, this patch adds infrastructure for running Snippy tests in Ibex:
- CMake-based Snippy flow
- Makefile bridge for launching it from dv/uvm/core_ibex
- YAML-based description of Snippy scenarios
- support for selecting Snippy layouts and the YAML directory
- support for joint coverage collection from llvm-snippy and riscv-dv test runs

The goal of this patch is not to replace the current regression tests, but to give Ibex DV one more independent and configurable way to generate programs for RTL simulation and coverage-driven exploration.

Coverage results for unprivileged isa only with llvm-snippy tests:

```
Total Coverage Summary
+--------+-------+--------+--------+--------+--------+--------+
| GROUP  | SCORE | LINE   | TOGGLE | FSM    | BRANCH | ASSERT |
+--------+-------+--------+--------+--------+--------+--------+
| 100.00 | 78.51 | 82.82  | 86.09  | 36.49  | 74.23  | 91.42  |
+--------+-------+--------+--------+--------+--------+--------+
```


SamuelRiedel:
Hi @AnastasiyaChernikova, thanks for putting this PR together! It is really interesting to see an llvm-snippy integration working with Ibex. As I mentioned over on the issue, we can't promise yet that we will be able to merge this as officially integrating a new generator would be a significant ongoing maintenance commitment for us. 

That being said, we really appreciate the effort, and it's a cool concept. Could you elaborate a bit more on the specific advantages of having Snippy as a second generator? For instance, what are some concrete things or scenarios that Snippy can handle better, or more easily, than riscv-dv? 

Also, regarding the Zcmp issue (#2445) mentioned in the description: as I noted in the issue thread, those mismatches are very likely just due to the mainline cosim environment not being fully updated for the experimental Zc* extensions yet.

AnastasiyaChernikova:
@SamuelRiedel Thank you very much for the comments and for taking a look at the PR.

I fully understand the concern about maintenance cost. I am not proposing Snippy as a replacement for riscv-dv. Rather, I see it as an additional CI flow that could complement the existing riscv-dv-based regressions and provide a different instruction stimulus profile.

In my opinion, Snippy can bring value to Ibex in several ways.

1. Independent source of stimulus

Snippy is built on a different infrastructure and uses a different approach to program generation compared to riscv-dv. As a result, it can generate tests with a different instruction distribution, different program structure, and different corner cases. For CI, this is useful as an additional independent check: even if Snippy runs in the same Ibex testbench/cosimulation environment, it may activate scenarios that do not appear as frequently in the main riscv-dv flow.

2. Convenient support for ISA extensions through LLVM

Snippy is based on LLVM target descriptions. If an ISA extension is already supported in LLVM, in many cases it does not need to be implemented again inside the generator. For example, Snippy documentation describes support/configuration for extensions and features such as B/Zb*, C/Zc*, F/D, V, Zicbom/Zicboz, Zicond, scalar crypto extensions Zk*, vector crypto extensions Zvkn*/Zvks*. This can be useful for Ibex if new extensions are added in the future: the generator-side effort is usually lower than manually adding support for a new extension to a separate generator description.

3. More direct control over the instruction mix

Snippy allows instruction histograms and patterns to be specified explicitly in YAML configurations. This is convenient for targeted tests where we want to focus on a specific instruction subset, ISA extension, or operation class.

4. More flexible memory-oriented scenarios

One of Snippy’s strengths is the ability to control memory access patterns. It supports address ranges, stride/offset-based accesses, explicit addresses, burst-like memory sequences, and other memory schemes. This makes it possible to create denser and more directed stress on the load/store unit, the data memory path, and related pipeline scenarios.

5. More flexible control-flow generation

Snippy allows branches, loops, and call graphs to be configured separately, including loop ratio, nesting depth, branch distance, function calls, and the overall structure of the generated program. This is useful when we do not just want random branch/jump instructions, but want to control the shape of the control-flow graph and the dynamic behavior of the program. In particular, this can be useful for stress testing the BPU and fetch/control-flow logic, which could be valuable for Ibex as well.

6. Targeted dependency and hazard scenarios

Snippy is also convenient for dependency-oriented scenarios. For example, the available register set can be restricted to increase the probability of RAW/dependency cases, or dedicated configurations can be created for hazard-oriented testing. This helps generate denser dependency patterns than a fully general random instruction stream.

7. Corner cases and special patterns

Snippy is well suited for tests targeting rare operand values and special situations. For example, configurations can be built for specific operand values, sign-related cases, dense memory accesses, self-modifying code, and other scenarios that are difficult to obtain reliably by simply increasing the length of a random test.

8. Integration with the existing flow

Snippy generates regular executable artifacts that can be run in the existing Ibex DV environment, and the resulting coverage can be merged with coverage from riscv-dv and Snippy runs. The goal of the integration is not to rewrite the testbench or replace the riscv-dv infrastructure. The goal is to add another source of programs that go through the existing execution, cosimulation, and checking mechanisms.

For CI, I think the practical option would be to add Snippy as a second or optional job with a limited set of configurations. Such a job could run a compact set of targeted tests: arithmetic/bitmanip, memory patterns, branches/loops/calls, dependency/hazard scenarios, and a few corner-case configurations.

The generation time itself should not be a major CI bottleneck. Snippy is implemented in C++, and in our measurements it generated 50,000 instructions in under 2 seconds. Therefore, most of the additional runtime would come from cosimulation, not from test generation.

If the maintainers think this direction is useful, I would be happy to help bring this PR to an acceptable state, prepare a compact set of Snippy configurations for CI, help integrate this flow into CI, and support it afterwards. I can also help investigate possible failures, update configurations when the Ibex DV flow changes, and keep the integration aligned with changes in Snippy.

There is also a good upstream support story on the Snippy side: the project is actively developed, issues are usually handled quickly, and questions about the generator are answered promptly. Therefore, if generator-side problems appear, there is a realistic path to get help or fixes upstream.

AnastasiyaChernikova:
@SamuelRiedel, I wanted to gently follow up on this PR and ask whether any decision has been made regarding the possible integration of the llvm-snippy flow.

On my side, I have also run a small additional experiment to better evaluate the practical benefit of this flow for CI and coverage.

I compared the local runtime of the current riscv-dv test set and the selected llvm-snippy test set in single-threaded mode, without coverage collection. In my environment, a full riscv-dv run took around 29 hours on a single thread, while the selected llvm-snippy test set took around 2.5 hours within a same config.

At the same time, when running the selected llvm-snippy test set 3 times, it consistently reaches 100% coverage for the unprivileged ISA groups. As a result, when merging coverage from the riscv-dv and llvm-snippy runs, Snippy consistently covers 19 groups 
coverage_groups_comparison_table
that are not fully or not reliably covered by the current riscv-dv scenarios. Overall, this adds roughly 2% to the total groups coverage.

Therefore, I think adding llvm-snippy as an additional CI flow for unprivileged instructions could be useful not only as an independent source of stimulus, but also from a practical point of view: it could potentially reduce the total runtime needed to reach comparable coverage, while slightly improving the final coverage percentage.

AnastasiyaChernikova:
Hello! I’m happy to share an additional resource: a recording of my English [ youtube talk](https://youtu.be/fSsq5mm3I40?is=JuaY1BuVGIzB_QZA) about this work, including the llvm-snippy integration with Ibex, is now available.

I hope it provides helpful context for reviewing this patch and assessing whether this direction could be valuable for Ibex.

PR #2448, opened 2026-06-07
actions/checkout: v4 -> v6

hcallahan-lowrisc:
This fixes CI breakage due to the deprecation of NodeJS20 support in the default action runners.


SamuelRiedel (approved):
LGTM. Thanks for handling this.

PR #2449, opened 2026-06-08
[simple_system] Fix initial assignment of `clk_sys`

### LINKED ISSUE #2447: Fix ifdef VERILATOR breaking sv-tests ibex
## Observed Behavior

The Ibex test that runs on Verilator as part of sv-tests at https://chipsalliance.github.io/sv-tests-results/?v=verilator+ibex+ibex

fails with:

```
%Error-CONTASSINIT: build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/../src/lowrisc_ibex_ibex_simple_system_core_0/rtl/ibex_simple_system.sv:68:19: Continuous assignment to variable with initial value: 'clk_sys'
                                                                                                                                                                               : ... note: In instance 'ibex_simple_system'
                                                                                                                                                                            : ... Location of variable initialization
                    build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/../src/lowrisc_ibex_ibex_simple_system_core_0/rtl/ibex_simple_system.sv:132:12: ... Location of continuous assignment
                    ... For error description see https://verilator.org/warn/CONTASSINIT?v=5.049
```

## Expected Behavior

Does not fail.

See sv-tests for their command line.

## Notes

I believe this is because Verilator recently added/updated the continuous assignment error, previously this was (incorrectly) not an error.  I believe that the error is correct, and is due to the `ifdef VERILATOR code around here:

https://github.com/lowRISC/ibex/blob/2cf450db5133f6215c32d8463e65f8733e7323a3/examples/simple_system/rtl/ibex_simple_system.sv#L131

it should be updated to be an always or otherwise not create IEEE-illegal code when VERILATOR is defined. Thanks.

SamuelRiedel:
Remove the in-line initialization of the clock from the simple system, and handle the clock generation in a separate `initial` block instead.

Fixes https://github.com/lowRISC/ibex/issues/2447

### DIFF
--- a/examples/simple_system/rtl/ibex_simple_system.sv
+++ b/examples/simple_system/rtl/ibex_simple_system.sv
@@ -65,7 +65,7 @@ module ibex_simple_system (
   parameter bit                 BranchPredictor          = 1'b0;
   parameter                     SRAMInitFile             = "";
 
-  logic clk_sys = 1'b0, rst_sys_n;
+  logic clk_sys, rst_sys_n;
 
   typedef enum logic {
     CoreD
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Change looks good, thanks for fixing @SamuelRiedel :+1:

PR #2455, opened 2026-06-30
[cheriot,rtl] Add TRVK filter to Ibex sources

thommythomaso:
This PR introduces the temporal revocation (TRVK) filter to the Ibex sources without integrating it yet. The integration of the TRVK into Ibex is done in #2456.

The TRVK filter is constructed and integrated into Ibex according to the [OpenTitan CHERIoT MEmory Subsystem RFC](https://docs.google.com/document/d/17bFJLxabrlvvFqkowsxfCvOBtQeWylYygie5syd1Xoc/edit?usp=sharing). 

Ibex's documentation has yet to be updated by a follow-up PR. An [issue #2464](https://github.com/lowRISC/ibex/issues/2464) regarding this has been opened.

The verification of TRVK will be done together with the CHERIoT-capable Ibex core. We keep track of this in [issue #2466](https://github.com/lowRISC/ibex/issues/2466).


ibex_top.core
```diff
@@ -23,6 +23,7 @@ filesets:
       - rtl/ibex_register_file_ff.sv # generic FF-based
       - rtl/ibex_register_file_fpga.sv # FPGA
       - rtl/ibex_register_file_latch.sv # ASIC
+      - rtl/ibex_trvk.sv
```
andreaskurth, line 30: Adding `ibex_trvk.sv` to `files_rtl` here pulls in `stream_fork` / `stream_join_dynamic`, but `ibex_top.core` does not declare a dependency on the vendored common_cells core. The `files_rtl.depend` block above lists no `pulp-platform:common_cells:common_cells`, and no other `.core` file depends on it either — so fusesoc will not include those sources and `ibex_top` elaboration/lint will fail on unresolved `stream_fork`/`stream_join_dynamic`.

Please add it to `files_rtl.depend`:
```
      - pulp-platform:common_cells:common_cells
```

thommythomaso:

rtl/ibex_cheriot_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+package ibex_cheriot_pkg;
+
+  parameter int unsigned TOP_W     = 9;
+  parameter int unsigned BOT_W     = 9;
+  parameter int unsigned EXP_W     = 5;
+  parameter int unsigned OTYPE_W   = 3;
+  parameter int unsigned CPERMS_W  = 6;
+
+
+  // Obtain 32-bit representation of top
```
marnovandermaas: 33-bit

rtl/ibex_cheriot_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+  function automatic logic[32:0] get_bound33(logic [TOP_W-1:0] top, logic [1:0]  cor,
+                                             logic [EXP_W-1:0] exponent, logic [31:0] addr);
+    logic [32:0] t1, t2, mask, cor_val;
```
marnovandermaas, line 32: t1 and t2 should have more descriptive names.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+
+    return t1;
+  endfunction
+
+
+  // Update the top/base correction for a cap
+  function automatic logic [2:0] update_temp_fields(logic [TOP_W-1:0] top, logic [BOT_W-1:0] base,
+                                                    logic [BOT_W-1:0] addrmi);
+    logic top_hi, addr_hi;
+    logic [2:0] res3;
```
marnovandermaas, line 57: This should probably be called something like "correction_bits"

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+    .N_INP(32'd2)
+  ) u_stream_join_dynamic_d2h (
+    .inp_valid_i({revbm_rvalid_i,     downstream_rsp_out_valid}),
+    .inp_ready_o({revbm_rsp_ready,    downstream_rsp_out_ready}),
+    .sel_i      ({revbm_req_required, 1'b1                    }),
+    .oup_valid_o(upstream_rvalid_o),
+    .oup_ready_i(1'b1)
+  );
+
+  // Forward OBI payload between host and device
```
marnovandermaas: You use host and device terminology here but downstream and upstream terminology for the signals. Does it make sense to align these?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  assign downstream_we_o         = upstream_we_i;
+  assign downstream_be_o         = upstream_be_i;
+  assign downstream_addr_o       = upstream_addr_i;
+  assign downstream_wdata_o      = upstream_wdata_i;
+  assign downstream_wdata_intg_o = upstream_wdata_intg_i;
+  assign upstream_rdata_o        = downstream_rsp_out.data;
+  assign upstream_rdata_intg_o   = downstream_rsp_out.intg;
+  assign upstream_err_o          = downstream_rsp_out.err;
+
+  // Forward host to device CHERIoT tag w/o changes
```
marnovandermaas: Nit, probably best to spell out "without" here.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  assign downstream_tag_o = upstream_tag_i;
+
+  // Tag handling; always return tag except if we do a revocation bitmap lookup
+  assign upstream_tag_o = revbm_rvalid_i ? !revbm_revoked : downstream_rsp_out.tag;
```
marnovandermaas: Shouldn't this be?
```
assign upstream_tag_o = (revbm_rvalid_i ? !revbm_revoked : 1'b1) & downstream_rsp_out.tag;
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  ) u_prim_fifo_sync_align (
+    .clk_i,
+    .rst_ni,
+    .clr_i   (1'b0),
+    .wvalid_i(align_fork_valid),
+    .wready_o(align_fork_ready),
+    .wdata_i (upstream_addr_i[2]),
+    .rvalid_o(align_out_valid),
+    .rready_i(align_out_ready),
+    .rdata_o (align_out),
```
marnovandermaas: Maybe it's better to call this something like "misaligned_flag"

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+      end
+    end
+  end
+
+  // Pointer valid store
+  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_pointer_vaild_store
+    if(!rst_ni) begin
+      ptr_store_valid_q <= 1'b0;
+    end else begin
+      if (align_out_ready) begin
```
marnovandermaas: We don't have to check align_out_valid here?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+    .wdata_i (downstream_rsp_in),
+    .rvalid_o(downstream_rsp_out_valid),
+    .rready_i(downstream_rsp_out_ready),
+    .rdata_o (downstream_rsp_out),
+    .full_o  (),
+    .depth_o (),
+    .err_o   ()
+  );
+
+  assign downstream_rsp_in = '{
```
marnovandermaas: I think it makes a bit more sense putting this above u_prim_fifo_ds_rsp_store.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+
+  ////////////////////////////////
+  // Bitmap Address Calculation //
+  ////////////////////////////////
+
+  // Exponent is expanded (page 70, CHERIoT Architecture specification, Version 1.0)
+  assign cap_meta = '{
+    base:     downstream_rsp_out.data[8:0],
+    exponent: downstream_rsp_out.data[21:18] != 4'hf ? {1'b0, downstream_rsp_out.data[21:18]} :
+                                                       5'd24,
```
marnovandermaas: It probably makes sense to make this a package function to avoid replicating this logic throughout our design.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+    otype:    downstream_rsp_out.data[24:22],
+    cperms:   downstream_rsp_out.data[30:25]
+  };
+
+  // Not all fields are used
+  assign unused_cap_meta = ^{cap_meta.otype, cap_meta.cperms[5]};
+
+  // Check if cap is sealing cap
+  assign is_sealing_cap = (cap_meta.cperms[4:3] == 2'b00) && (|cap_meta.cperms[2:0]);
```
marnovandermaas: I think it makes sense to make this a package function.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+
+  parameter int unsigned TOP_W     = 9;
+  parameter int unsigned BOT_W     = 9;
+  parameter int unsigned EXP_W     = 5;
+  parameter int unsigned OTYPE_W   = 3;
+  parameter int unsigned CPERMS_W  = 6;
+
+
+  // Obtain 32-bit representation of top
+  function automatic logic[32:0] get_bound33(logic [TOP_W-1:0] top, logic [1:0]  cor,
```
marnovandermaas, line 30: I think it makes sense to call this "get_top33" right?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  assign unused_cap_correction = ^cap_correction[2:1];
+
+  // The MSB is unused in our case
+  assign {unused_cap_base_33, cap_base} = cap_base_33;
+
+  // Address in the revocation bitmap
+  assign revbm_cap_addr = cap_base - heap_base_addr_i;
+
+  // Bit address in the revocation bitmap (every bit corresponds to one 64-bit capability)
+  assign revbm_bit_addr = revbm_cap_addr >> 32'd3;
```
marnovandermaas: It probably makes sense to calculate the 3 using `$clog2(64/8)`

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  output logic        downstream_req_o,
+  input  logic        downstream_gnt_i,
+  input  logic        downstream_rvalid_i,
+  output logic        downstream_we_o,
+  output logic [3:0]  downstream_be_o,
+  output logic [31:0] downstream_addr_o,
+  output logic [31:0] downstream_wdata_o,
+  output logic [6:0]  downstream_wdata_intg_o,
+  input  logic [31:0] downstream_rdata_i,
+  input  logic [6:0]  downstream_rdata_intg_i,
```
marnovandermaas, line 50: It would be good to parameterise all of these.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  assign revbm_bit_addr = revbm_cap_addr >> 32'd3;
+
+  // Word address of the revocation bitmap
+  assign revbm_addr = revbm_bit_addr[RevBitmapAddrWidth+5-1:5];
+
+  // Bit select
+  assign revbm_bit_select = revbm_bit_addr[4:0];
+
+  // Capability base is outside of the bitmap range
+  assign revbm_out_of_range = |(revbm_bit_addr[31:RevBitmapAddrWidth+5]);
```
marnovandermaas: I think it probably makes sense to derive the 5 from the revocation bitmap width here so `$clog2(32)` and the 32 can be replaced by a parameter if you take the suggestion I make above.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+
+
+  //////////////////////
+  // Bitmap Interface //
+  //////////////////////
+
+  // We have loaded valid capability pointer, now we see valid metadata, not a sealing cap,
+  // and are pointing into the revocation bitmap
+  assign revbm_req_required = !is_sealing_cap          && // Not sealing cap
+                              ptr_store_valid_q        && // The base pointer stored is valid
```
marnovandermaas: This one really confused me, it would be much clearer if this signal was named "cap_load_valid_q"

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module ibex_trvk #(
```
marnovandermaas, line 9: I would prefer to rename this. One idea I had is "ibex_cap_load_filter".

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  // Element is consumed, if host handshakes response
+  assign align_out_ready = upstream_rvalid_o;
+
+
+  ///////////////////
+  // Pointer Store //
+  ///////////////////
+
+  // Pointer buffer is filled iff tag valid & 64-bit aligned
+  assign ptr_store_enable = downstream_rsp_out.tag && !align_out && align_out_valid &&
```
marnovandermaas: I would prefer to rename this "ptr_load_enable". I was very confused by this code thinking that this was about storing a pointer not loading one.

marnovandermaas: Also we should probably rename "ptr" to "cap".

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  //////////////////////
+  // Bitmap Interface //
+  //////////////////////
+
+  // We have loaded valid capability pointer, now we see valid metadata, not a sealing cap,
+  // and are pointing into the revocation bitmap
+  assign revbm_req_required = !is_sealing_cap          && // Not sealing cap
+                              ptr_store_valid_q        && // The base pointer stored is valid
+                              downstream_rsp_out.tag   && // We are looking at a capability
+                              downstream_rsp_out_valid && // The stored response is valid
```
marnovandermaas, line 355: Maybe the comment should say "The latched load response is valid"

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+                              align_out                && // We are on the second word of the cap
+                              align_out_valid          && // The alignment store is valid
```
marnovandermaas: Maybe this comment should say "The latched alignment bits are valid"

marnovandermaas (commented):
I've left a few comments on this, mainly to make the code more clear. Do we have any tests for this load filter? We should include at least a smoke test before merging this.

thommythomaso:
> I've left a few comments on this, mainly to make the code more clear. Do we have any tests for this load filter? We should include at least a smoke test before merging this.

I have a smoke test in OpenTitan of the entire CHERIoT subsystem. I would suggest to not add a dedicated test for just the filter here, especially as we are planning to merge the core soon which then will use and verify the filter.

rtl/ibex_cheriot_pkg.sv
thommythomaso: I would keep this aligned with the name originally used by the reference implementation of Microsoft. The plan is to merge the current `ibex_cheriot_pkg.sv` with the one that will be introduced with the core.

rtl/ibex_trvk.sv
thommythomaso: This is intentionally not done as Ibex does not support parameterized interface types.

rtl/ibex_trvk.sv
thommythomaso: I would keep this aligned with the nomenclature introduced by Microsoft.

rtl/ibex_cheriot_pkg.sv
thommythomaso: I would keep this aligned with the name originally used by the reference implementation of Microsoft. The plan is to merge the current ibex_cheriot_pkg.sv with the one that will be introduced with the core.

rtl/ibex_cheriot_pkg.sv
thommythomaso: Agreed, but we should align ourself with the nomenclature used in the core. @SamuelRiedel, will you rename these signals?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+      ptr_store_q <= '0;
+    end else begin
+      if (ptr_store_enable) begin
+        ptr_store_q <= downstream_rsp_out.data;
+      end
+    end
+  end
+
+  // Pointer valid store
+  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_pointer_vaild_store
```
SamuelRiedel: Nit:

```suggestion
  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_pointer_valid_store
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+module ibex_trvk #(
+  // The number of outstanding transaction the IP supports
+  parameter int unsigned NumOutstanding     = 32'd4,
+  // The width of the meta memory address space used to store the revocation bits
+  parameter int unsigned RevBitmapAddrWidth = 32'd9,
+  // The base address of the meta SRAM holding the revocation bitmap
+  parameter int unsigned RevBitmapBaseAddr  = 32'h0000_0000
+)(
+  input clk_i,
+  input rst_ni,
```
SamuelRiedel: Will this default to wire or logic?


```suggestion
  input  logic clk_i,
  input  logic rst_ni,
```

rtl/ibex_cheriot_pkg.sv
SamuelRiedel: I will change this to clean it up and make it more clear. But this will be part of a bigger refactoring, so I don't know whether it makes sense to already do this here, because we also want to be able to show where we took the original code from before aligning it with our styleguide.

SamuelRiedel (commented):
Thanks for this Thomas! Just a few very minor nits.

ibex_top.core
```diff
@@ -19,10 +19,12 @@ filesets:
       - lowrisc:prim:onehot_check
       - lowrisc:prim:onehot
       - lowrisc:prim:util
+      - pulp-platform:common_cells:common_cells
```
andreaskurth, line 25: I think dependencies for `prim_fifo_sync` and `prim_secded_inv_39_32_dec` are missing:

```suggestion
      - pulp-platform:common_cells:common_cells
      - lowrisc:prim:fifo
      - lowrisc:prim:secded
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+                              ptr_store_valid_q        && // The base pointer stored is valid
+                              downstream_rsp_out.tag   && // We are looking at a capability
+                              downstream_rsp_out_valid && // The stored response is valid
+                              align_out                && // We are on the second word of the cap
+                              align_out_valid          && // The alignment store is valid
+                              !revbm_out_of_range;        // We hit the heap range
+
+  // Assemble read-only request
+  assign revbm_req_o  = revbm_req_required && !revbm_outstanding_q;
+  assign revbm_addr_o = RevBitmapBaseAddr + {{32 - RevBitmapAddrWidth{1'b0}}, revbm_addr};
```
andreaskurth: `revbm_addr` is a *word* index (`revbm_bit_addr >> 5`, with `revbm_bit_addr[4:0]` used as `revbm_bit_select`), whereas `RevBitmapBaseAddr` is documented as "the base address of the meta SRAM", which in Figure 2 of the memory subsystem RFC is a byte address. Is this port intentionally word-addressed, unlike `downstream_addr_o`? If so, could the parameter comments say so; if it's byte-addressed, doesn't `revbm_addr` need a `<< 2` here?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+
+  // We have loaded valid capability pointer, now we see valid metadata, not a sealing cap,
+  // and are pointing into the revocation bitmap
+  assign revbm_req_required = !is_sealing_cap          && // Not sealing cap
+                              ptr_store_valid_q        && // The base pointer stored is valid
+                              downstream_rsp_out.tag   && // We are looking at a capability
+                              downstream_rsp_out_valid && // The stored response is valid
+                              align_out                && // We are on the second word of the cap
+                              align_out_valid          && // The alignment store is valid
+                              !revbm_out_of_range;        // We hit the heap range
```
andreaskurth, line 358: This encodes the assumption that a capability load always shows up as exactly two consecutive, uninterrupted responses with `addr[2]` 0 then 1. If a foreign response ever lands in between, `ptr_store_valid_q` is cleared and the tag is forwarded **unrevoked**. I can't construct such a case with today's single-outstanding, in-order LSU, but a second requester upstream (e.g., the TRBE) could lead to problems here, right?

Could the assumption be stated in the module header and backed by an assertion (see my comment near `endmodule`)? And would a safer default (e.g., clearing the tag when the check cannot be performed) be an option here?

Separately, the filter never looks at `we`, so capability-load detection rests entirely on `addr[2]` plus the response tag. Is `downstream_tag_i` defined to be 0 on write responses? If not, a store would drive this state machine too and trigger a spurious bitmap read.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+    .Pass(1'b1),
+    .Depth(NumOutstanding),
+    .NeverClears(1'b1),
+    .Secure(1'b0)
+  ) u_prim_fifo_ds_rsp_store (
+    .clk_i,
+    .rst_ni,
+    .clr_i   (1'b0),
+    .wvalid_i(downstream_rvalid_i),
+    .wready_o(), // Not connected as OBI is not configured to use rready
```
andreaskurth: With `wready_o` unconnected, an overflow of this FIFO would be silent. `stream_fork` lets the downstream output handshake before the input does, so while `u_prim_fifo_sync_align` is full a `NumOutstanding+1`-th request can still go out downstream, even though both FIFOs are only `NumOutstanding` deep. (The same window also lets `upstream_rvalid_o` fire for a transaction `upstream_gnt_o` hasn't acknowledged yet.) Again not reachable with the current LSU, but it makes `NumOutstanding` mean something different from what its comment says.

Would it be simpler to drop `stream_fork` and gate the request on FIFO space, so the two paths can't diverge in the first place?

```systemverilog
assign downstream_req_o = upstream_req_i  & ~align_full;
assign upstream_gnt_o   = downstream_gnt_i & ~align_full;
assign align_fork_valid = upstream_req_i  & downstream_gnt_i & ~align_full;
```

Otherwise `Depth(NumOutstanding + 1)` here would also do. Either way, could `wready_o` be connected and asserted so this can never fail silently?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+                              align_out                && // We are on the second word of the cap
+                              align_out_valid          && // The alignment store is valid
+                              !revbm_out_of_range;        // We hit the heap range
+
+  // Assemble read-only request
+  assign revbm_req_o  = revbm_req_required && !revbm_outstanding_q;
+  assign revbm_addr_o = RevBitmapBaseAddr + {{32 - RevBitmapAddrWidth{1'b0}}, revbm_addr};
+
+  // Is the current capability marked as revoked?
+  assign revbm_revoked = revbm_rdata_i[revbm_bit_select] || revbm_err_i;
```
andreaskurth: An ECC failure on the bitmap response raises `revbm_data_intg_error_o` but does not revoke, unlike `revbm_err_i`. By the same argument you made for `revbm_err_i` (invalidating an erroneous bitmap read is the safer direction), should the integrity error be folded in here as well? Clearing a tag can only ever remove authority, and the system is heading for an alert in that case anyway.

```suggestion
  assign revbm_revoked = revbm_rdata_i[revbm_bit_select] || revbm_err_i ||
                         (|revbm_rsp_data_intg_error);
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+  };
+
+  // Not all fields are used
+  assign unused_cap_meta = ^{cap_meta.otype, cap_meta.cperms[5]};
+
+  // Check if cap is sealing cap
+  assign is_sealing_cap = (cap_meta.cperms[4:3] == 2'b00) && (|cap_meta.cperms[2:0]);
+
+  // Extract the middle field from the pointer, bounds depend on exponent, width fixed
+  assign addr_mid = ptr_store_q[cap_meta.exponent +: ibex_cheriot_pkg::BOT_W];
```
andreaskurth: For the expanded exponent 24 this selects bits `[32:24]` of a 32-bit vector, so the top bit is out of range and reads as `x` in simulation. It happens not to change `cap_base` (the correction gets shifted out at that exponent), but a shift avoids it and matches the Sail formulation `a_mid = truncate(a >> E, cap_mantissa_width)`:

```suggestion
  assign addr_mid = cheriot_bot_t'(ptr_store_q >> cap_meta.exponent);
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+
+
+  //////////////////////////////////
+  // Downstream Response Latching //
+  //////////////////////////////////
+
+  // downstream response store
+  prim_fifo_sync #(
+    .Width($bits(ds_rsp_t)),
+    .Pass(1'b1),
```
andreaskurth, line 276: With pass-through enabled, the whole base decompression (two shifts by a variable exponent, a 33-bit add, the comparators) sits combinationally between `downstream_rdata_i` and `revbm_req_o`/`revbm_addr_o`, i.e. on the memory response path. Do we expect this path to be uncritical? The Ibex RFC commits to not further increasing the critical path of the overall OpenTitan design.

rtl/ibex_trvk.sv
andreaskurth: `ptr_store_enable` (l. 223) already contains `align_out_ready`, so the term seems redundant under this `if`. What matters here is that the flop is *held* between the two words of a capability rather than cleared, right? Could the two conditions be separated, or a short comment added?

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,383 @@
...
+    end else begin
+      if (revbm_rvalid_i && revbm_rsp_ready) begin
+        revbm_outstanding_q <= 1'b0;
+      end else if (revbm_req_o && revbm_gnt_i) begin
+        revbm_outstanding_q <= 1'b1;
+      end
+    end
+  end
+
+endmodule
```
andreaskurth, line 421: The module has no assertions at all yet. Suggest starting with the following:
```systemverilog
// Every delivered response has its alignment information available.
`ASSERT(AlignValidOnRsp_A, upstream_rvalid_o |-> align_out_valid)
// The downstream response store never overflows (needs `wready_o` connected).
`ASSERT(DsRspFifoNoOverflow_A, downstream_rvalid_i |-> ds_rsp_wready)
// No unsolicited bitmap response; this is what keeps `upstream_tag_o` safe.
`ASSERT(RevbmRspOnlyWhenOutstanding_A, revbm_rvalid_i |-> revbm_outstanding_q)
// OBI request stability on the bitmap port.
`ASSERT(RevbmReqStable_A, revbm_req_o && !revbm_gnt_i |=> revbm_req_o && $stable(revbm_addr_o))
`ASSERT_INIT(RevBitmapAddrWidthRange_A, RevBitmapAddrWidth inside {[1:27]})
`ASSERT(HeapBaseAligned_A, heap_base_addr_i[2:0] == 3'b0)
```

The first two would turn the two contract issues I flagged above into something machine-checkable.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+    logic [32:0] t1, t2, mask, cor_val;
+
+    if (cor[1])
+      // negative sign extension
+      cor_val = {33{cor[1]}};
+    else
+      cor_val = {32'h0, (~cor[1]) & cor[0]};
+
+    cor_val = (cor_val << exponent) << TOP_W;
+    mask    = (33'h1_ffff_ffff << exponent) << TOP_W;
```
andreaskurth, line 41: When this function is used to compute the *base* (as `ibex_trvk` does), the shift is conceptually `BOT_W`, not `TOP_W`. Same value today, but it reads as top-only. Together with Marno's point on l. 14, would it be worth making the header comment say that this returns a 33-bit base *or* top?

vendor/pulp_common_cells.vendor.hjson
```diff
@@ -0,0 +1,28 @@
...
+
+    mapping: [
+        {
+            from: "src/stream_fork.sv",
+            to: "rtl/stream_fork.sv",
+            patch_dir: "stream_fork_sv"},
+        {
+            from: "src/stream_fork_dynamic.sv",
+            to: "rtl/stream_fork_dynamic.sv",
+            patch_dir: "stream_fork_dynamic_sv"},
```
andreaskurth: `stream_fork_dynamic` isn't instantiated anywhere. Could this mapping, its patch directory and the `.core` entry be dropped until something actually needs it?

andreaskurth (commented):
Thanks @thommythomaso, nice to see the TRVK land!I checked the CHERIoT-specific parts against the v1.0 spec and the Sail model and they match.

My remaining comments are almost entirely around integration, verification and documentation rather than the logic itself. At the current stage, many items can be deferred for follow-up in tracking issues, but those issues need to be created now and referenced here.

Three items that don't attach to a line:

**Verification.** Nothing in CI currently compiles this module: it isn't instantiated anywhere, so Verilator only parses it and never elaborates it, and there is no testbench. I understand the smoke test lives in OpenTitan and that the core PR will exercise the filter, but the memory subsystem RFC commits to a block-level test for the TRVK specifically. Please create a tracking issue and reference it here.

**Documentation.** There is nothing under `doc/03_reference/`. It would help integrators to have the port protocol (OBI plus the `tag` sideband), the revocation bitmap addressing convention, the placement requirements (lockstep domain, arbitration on the bitmap port, `heap_base_addr_i` being static and 8-byte aligned), and the expected handling of `revbm_data_intg_error_o` / `revbm_device_error_o` written down somewhere. Please create a tracking issue and reference it here.

**PR description.** Could you please add one? Both commit bodies are empty as well. A pointer to the RFCs and a note that the module is deliberately not instantiated yet would make this much easier to understand later.

rtl/ibex_trvk.sv
thommythomaso: OBI does not allow back pressure on the response path, so it is here not required.

rtl/ibex_trvk.sv
thommythomaso: I named these signals explicitly ptr as it stores the lower part of the capability (which corresponds to the C pointer). I feel if we rename it cap_load_valid, people might get the impression, we are storing a full capability.

rtl/ibex_trvk.sv
thommythomaso: it would default to logic. but its better to be explicit. I must have done a copypasta.

rtl/ibex_trvk.sv
thommythomaso: You are right.

rtl/ibex_trvk.sv
thommythomaso: I would say this assumption is valid in our design. A TRBE would have its own TRVK module to not make the TRBE (a system component) dependent on the core.

rtl/ibex_trvk.sv
thommythomaso: Should it become critical, it is planned to cut the path at the revbm port.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,418 @@
...
+
+  // Every delivered response has its alignment information available.
+  `ASSERT(AlignValidOnRsp_A, upstream_rvalid_o |-> misalign_flag_out_valid)
+  // The downstream response store never overflows.
+  `ASSERT(DsRspFifoNoOverflow_A, downstream_rvalid_i |-> downstream_rsp_wready)
+  // No unsolicited bitmap response; this is what keeps `upstream_tag_o` safe.
+  `ASSERT(RevbmRspOnlyWhenOutstanding_A, revbm_rvalid_i |-> revbm_outstanding_q)
+  // OBI request stability on the bitmap port.
+  `ASSERT(RevbmReqStable_A, revbm_req_o && !revbm_gnt_i |=> revbm_req_o && $stable(revbm_addr_o))
+  `ASSERT_INIT(RevBitmapAddrWidthRange_A, RevBitmapAddrWidth inside {[1:27]})
```
andreaskurth: The upper bound here came from my earlier suggestion and it is off by one, sorry. `revbm_out_of_range` (l. 341) selects `revbm_bit_addr[31:RevBitmapAddrWidth+5]`, so `RevBitmapAddrWidth = 27` gives the reversed range `[31:32]`. The binding constraint is 26:

```suggestion
  `ASSERT_INIT(RevBitmapAddrWidthRange_A, RevBitmapAddrWidth inside {[1:26]})
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,418 @@
...
+  };
+
+  // Not all fields are used
+  assign unused_cap_meta = ^{cap_meta.otype, cap_meta.cperms[5]};
+
+  // Check if cap is sealing cap
+  assign is_sealing_cap = ibex_cheriot_pkg::is_sealing_cap(cap_meta.cperms[4:0]);
+
+  // Extract the middle field from the pointer, bounds depend on exponent, width fixed
+  assign addr_mid = ptr_storage_q >> cap_meta.exponent;
```
andreaskurth: The shift fixes the out-of-range select, thanks. It does implicitly truncate 32 bits down to the 9-bit `cheriot_bot_t` though, which could lead to problems with lint. Would an explicit cast be worth it?

```suggestion
  assign addr_mid = cheriot_bot_t'(ptr_storage_q >> cap_meta.exponent);
```

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,418 @@
...
+
+// Assumes a tagged capability load always arrives downstream as exactly two consecutive,
+// uninterrupted 32-bit responses from a single, in-order requester.
+module ibex_trvk #(
+  // The number of outstanding transaction the IP supports
+  parameter int unsigned NumOutstanding     = 32'd4,
+  // The width of the meta memory address space used to store the revocation bits
+  parameter int unsigned RevBitmapAddrWidth = 32'd9,
+  // The base address of the meta SRAM holding the revocation bitmap
+  parameter int unsigned RevBitmapBaseAddr  = 32'h0000_0000,
```
andreaskurth: Now that `revbm_addr_o` is byte-granular (l. 360), could these two comments make the convention explicit? `RevBitmapAddrWidth` is a *word*-address width (9 bits -> 512 words -> 2 KiB), which isn't obvious from "the width of the meta memory address space".

Relatedly, `RevBitmapBaseAddr` now has to be 4-byte aligned for the `+` at l. 360 to behave. Worth an `ASSERT_INIT` alongside the others?

andreaskurth (commented):
Thanks @thommythomaso! This addresses everything from the last round and the answers on the sequencing assumption and the revbm timing path make sense to me. Only three small follow-ups, two of which are consequences of my own earlier suggestions.

andreaskurth (approved):
LGTM, thanks for addressing my feedback @thommythomaso!

PR #2456, opened 2026-06-30
[cheriot,rtl] Integrate TRVK filter into Ibex Top

thommythomaso:
This PR depends on #2455 and it integrates the TRVK filter into Ibex.

Currently the TRVK filter is ***not*** operated in lockstep, an issue is opened here: https://github.com/lowRISC/ibex/issues/2467. 
The TRVK filter is not yet verified in the Ibex repo, see https://github.com/lowRISC/ibex/issues/2466.


rtl/ibex_top.sv
```diff
@@ -1148,14 +1181,109 @@ module ibex_top import ibex_pkg::*; #(
...
   logic icache_alert_major_internal;
   assign icache_alert_major_internal = (|icache_tag_alert) | (|icache_data_alert);
 
-  assign alert_major_internal_o = core_alert_major_internal |
+  assign alert_major_internal_o = core_alert_major_internal     |
                                   lockstep_alert_major_internal |
-                                  icache_alert_major_internal;
+                                  icache_alert_major_internal   |
+                                  trvk_revbm_data_intg_error    |
+                                  trvk_revbm_device_error;
```
andreaskurth: Shouldn't these feed into `alert_major_bus_o`? Potentially worth discussing with Pascal and/or Pirmin

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -149,10 +151,20 @@ module core_ibex_tb_top;
...
+    .data_tag_o                (                           ),
     .data_err_i                (data_mem_vif.error         ),
 
+    .trvk_revbm_req_o          (                           ),
+    .trvk_revbm_gnt_i          (1'b0                       ),
+    .trvk_revbm_rvalid_i       (1'b0                       ),
+    .trvk_revbm_addr_o         (                           ),
+    .trvk_revbm_rdata_i        ('b0                        ),
+    .trvk_revbm_rdata_intg_i   ('b0                        ),
+    .trvk_revbm_err_i          (                           ),
```
andreaskurth: Suggest tying off (only the last line)

rtl/ibex_top.sv
```diff
@@ -1148,14 +1181,109 @@ module ibex_top import ibex_pkg::*; #(
...
+    assign trvk_revbm_data_intg_error = 1'b0;
+    assign trvk_revbm_device_error    = 1'b0;
+    assign unused_trvk = ^{
+      trvk_heap_base_addr_i,
+      trvk_revbm_gnt_i,
+      trvk_revbm_rvalid_i,
+      trvk_revbm_rdata_i,
+      trvk_revbm_rdata_intg_i,
+      trvk_revbm_err_i,
+      trvk_rtag,
```
andreaskurth, line 1269: suggest tying off in this branch (only the last line) - otherwise it's both undriven and unused, which may cause lint problems

rtl/ibex_top.sv
```diff
@@ -21,6 +22,8 @@ module ibex_top import ibex_pkg::*; #(
   parameter ibex_pkg::pmp_cfg_t     PMPRstCfg[PMP_MAX_REGIONS]   = ibex_pkg::PmpCfgRst,
   parameter logic [PMP_ADDR_MSB:0]  PMPRstAddr[PMP_MAX_REGIONS]  = ibex_pkg::PmpAddrRst,
   parameter ibex_pkg::pmp_mseccfg_t PMPRstMsecCfg                = ibex_pkg::PmpMseccfgRst,
+  parameter int unsigned            CheriotRevBitmapAddrWidth    = 32'd9,
+  parameter int unsigned            CheriotRevBitmapBaseAddr     = 32'h0,
```
andreaskurth: Suggest documenting the new parameters in `integration.rst`. Not a blocker - if you prefer to do this later, please create an issue.

Similarly, I suggest describing the internal workings of this PR and #2455 in the reference guide - also not blocking but should be tracked in an issue if defered.

andreaskurth (commented):
Mostly LGTM, thanks @thommythomaso!

A few comments/suggestions below

rtl/ibex_top.sv
thommythomaso: As `integration.rst` is quite outdated in general, I suggest we complete this in a dedicated pass.

thommythomaso: #2464.

rtl/ibex_top.sv
```diff
@@ -1148,15 +1181,112 @@ module ibex_top import ibex_pkg::*; #(
...
+    assign trvk_revbm_data_intg_error = 1'b0;
+    assign trvk_revbm_device_error    = 1'b0;
+    assign unused_trvk = ^{
+      trvk_heap_base_addr_i,
+      trvk_revbm_gnt_i,
+      trvk_revbm_rvalid_i,
+      trvk_revbm_rdata_i,
+      trvk_revbm_rdata_intg_i,
+      trvk_revbm_err_i,
+      trvk_rtag,
```
SamuelRiedel, line 1269: I think this should be trvk_wtag. Initially, the rtag and wtag were mixed up here, the wtag is always driven by ibex, but unused without the trvk. The rtag is driven by the trvk (so needs to be tied off here), and always goes to the core, so never unused. After Andreas's feedback, the rtag is now properly tied off, but we should remove the tie off of wtag, otherwise it has multiple drivers, and we should replace the assignment here with wtag.


```suggestion
      trvk_wtag,
```

rtl/ibex_top.sv
```diff
@@ -1148,15 +1181,112 @@ module ibex_top import ibex_pkg::*; #(
...
+    assign trvk_wtag   = 1'b0;
+    assign unused_trvk = trvk_rtag;
+
+  end else begin : gen_no_cheriot_trvk
+
+    logic unused_trvk;
+
+    assign trvk_revbm_req_o           = '0;
+    assign trvk_revbm_addr_o          = '0;
+    assign trvk_wtag                  = 1'b0;
```
SamuelRiedel, line 1255:
```suggestion
```

SamuelRiedel (commented):
I rebased my work on top and things worked fine. I only had one minor issue.

rtl/ibex_top.sv
```diff
@@ -13,6 +13,7 @@
  * Top level module of the ibex RISC-V core
  */
 module ibex_top import ibex_pkg::*; #(
+  parameter ibex_pkg::base_isa_e    BaseIsa                      = ibex_pkg::BaseIsaRV32IorCHERIoT,
```
andreaskurth: Should this default to `BaseIsaRV32I` for now? As it stands the TRVK filter is instantiated in every integration that doesn't opt out (simple system, riscv-compliance, all `ibex_configs.yaml` configs, and the syn/lint/FPGA flows), while the two flows that actually verify Ibex -- the UVM regression and `dv/formal` -- are the ones selecting `BaseIsaRV32I`.

Related: `BaseIsa` isn't exposed as a FuseSoC parameter in `ibex_top.core` / `ibex_top_tracing.core`, and it isn't in `ibex_configs.yaml` (whose header requires every config to specify the same parameter set). So there's currently no way to select the RV32I variant from a FuseSoC build at all. Could it be added in both places?

rtl/ibex_top.sv
```diff
@@ -91,10 +96,21 @@ module ibex_top import ibex_pkg::*; #(
...
   input  logic                                                         data_err_i,
 
+  // TRVK revocation bitmap read interface
+  output logic                                                         trvk_revbm_req_o,
+  input  logic                                                         trvk_revbm_gnt_i,
+  input  logic                                                         trvk_revbm_rvalid_i,
+  output logic [31:0]                                                  trvk_revbm_addr_o,
+  input  logic [31:0]                                                  trvk_revbm_rdata_i,
+  input  logic [6:0]                                                   trvk_revbm_rdata_intg_i,
+  input  logic                                                         trvk_revbm_err_i,
```
andreaskurth, line 112: None of the ten new top-level ports has an `ASSERT_KNOWN`, and `IbexDataReqPayloadX` (line 1297) doesn't include `data_tag_o` -- the rest of the file X-checks every port. Fine as a follow-up rather than part of this PR; could you file an issue for it?

rtl/ibex_top.sv
```diff
@@ -806,15 +839,15 @@ module ibex_top import ibex_pkg::*; #(
...
       instr_rdata_core,
       instr_err_i,
-      data_req_o,
-      data_gnt_i,
-      data_rvalid_i,
-      data_we_o,
-      data_be_o,
-      data_addr_o,
-      data_wdata_o,
+      trvk_req,
```
andreaskurth, line 848: The TRVK should be in the lockstep domain, right? (Not necessarily in this PR, but would need a follow-up issue.)

rtl/ibex_top.sv
```diff
@@ -1148,15 +1181,112 @@ module ibex_top import ibex_pkg::*; #(
     assign unused_scan = scan_rst_ni;
   end
 
+
+  //////////
+  // TRVK //
+  //////////
+
+  if (BaseIsa == BaseIsaRV32IorCHERIoT) begin : gen_cheriot_trvk
+    ibex_trvk #(
```
andreaskurth, line 1196: `MemECC` isn't forwarded here, and `ibex_trvk` unconditionally SECDED-decodes `revbm_rdata_intg_i`. With `MemECC = 0` an integrator has no reason to drive integrity on the bitmap port, and `prim_secded_inv_39_32_dec` reports all-zeros as an error (the inverted code's zero-data codeword is `39'h2A00000000`), so every bitmap response would raise `alert_major_bus_o`.

It looks cheap to support: `ibex_trvk.sv:360-368` is the only ECC-dependent logic in the module -- everything else just forwards the integrity bits -- so a `parameter bit MemECC = 1'b1`, a generate-if around the decoder with `revbm_data_intg_error_o = 1'b0` in the else branch, and `.MemECC(MemECC)` here should cover it. The module-side part would belong in #2455. If you'd rather not do it now, could you document `MemECC = 0` as unsupported with CHERIoT and open a follow-up issue?

rtl/ibex_top.sv
```diff
@@ -1148,15 +1181,112 @@ module ibex_top import ibex_pkg::*; #(
...
+      .NumOutstanding(32'd2),
```
andreaskurth: This `32'd2` is the same fact as `MaxOutstandingDSideAccesses` at line 1326, but that localparam sits inside `` `ifdef INC_ASSERT `` so the two can silently diverge -- and undersizing `NumOutstanding` would let a downstream request be issued while the align FIFO is full, decoupling the request and alignment queues - right?

Would you consider moving `MaxOutstandingDSideAccesses` out of the `` `ifdef `` into a module-level localparam and using it in both places?

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
```diff
@@ -95,29 +95,30 @@ module core_ibex_tb_top;
...
-    .SecureIbex           (SecureIbex          ),
-    .LockstepOffset       (LockstepOffset      ),
-    .ICacheScramble       (ICacheScramble      ),
-    .BranchPredictor      (BranchPredictor     ),
-    .DbgTriggerEn         (DbgTriggerEn        ),
-    .DmBaseAddr           (DmBaseAddr          ),
-    .DmAddrMask           (DmAddrMask          ),
-    .DmHaltAddr           (DmHaltAddr          ),
-    .DmExceptionAddr      (DmExceptionAddr     )
+    .BaseIsa              (ibex_pkg::BaseIsaRV32I),
```
andreaskurth, line 98: With `BaseIsaRV32I` hardcoded here and in `dv/formal/check/top.sv`, nothing exercises the TRVK path. Could you open a tracking issue for the DV work so it isn't forgotten before CHERIoT gets enabled anywhere?

ibex_core.core
```diff
@@ -12,6 +12,7 @@ filesets:
       - lowrisc:prim:clock_gating
       - lowrisc:prim:lfsr
       - lowrisc:prim:mubi
+      - lowrisc:prim:fifo
```
andreaskurth: `prim_fifo_sync` is used by `ibex_trvk.sv`, which is in `ibex_top.core`'s fileset -- should this dependency move there? It currently only resolves via the transitive dependency on `ibex_core`.

andreaskurth (commented):
Thanks @thommythomaso! I have a couple more comments :-) Please see below.

One comment that doesn't attach inline: on `rtl/ibex_top.sv:1523`: The `MajorAlertOnDMemIntegrityErr` assertion pairs the downstream trigger `data_rvalid_i` with the core-side data `data_rdata_core`. When the TRVK response FIFO already holds an entry those are different words, so the assertion decodes the wrong one, no? Independently, the `##[0:5]` window *could* be exceeded while the TRVK stalls a response for a bitmap lookup (although unlikely in our current systems). Would triggering on `trvk_rvalid` be the right fix?

rtl/ibex_top.sv
thommythomaso: https://github.com/lowRISC/ibex/issues/2467

dv/uvm/core_ibex/tb/core_ibex_tb_top.sv
thommythomaso: Tracked in https://github.com/lowRISC/ibex/issues/2466.

rtl/ibex_trvk.sv
```diff
@@ -0,0 +1,418 @@
...
+    .inp_ready_o({revbm_rsp_ready,    downstream_rsp_out_ready}),
+    .sel_i      ({revbm_req_required, 1'b1                    }),
+    .oup_valid_o(upstream_rvalid_o),
+    .oup_ready_i(1'b1)
+  );
+
+  // Forward OBI payload between upstream and downstream
+  assign downstream_we_o         = upstream_we_i;
+  assign downstream_be_o         = upstream_be_i;
+  assign downstream_addr_o       = upstream_addr_i;
```
nasahlpa, line 188: Thanks @thommythomaso for this PR.

I had a look from a FI perspective and I am wondering a bit about the architecture.

Take for example the data port of Ibex. Ibex top exposes `data_*` and `data_*_shadow`. For now, only a single instance of `ibex_trvk` is instantiated that operates on the main core data port. From a functional perspective this works as the data output ports are just routed through without any logic (downstream = upstream).

From my understanding so far, from a security point of view the architecture is not ideal as the attacker could mess around with FIFOs (e.g., _u_prim_fifo_ds_rsp_store_) within this module or flops. Then, the faulty data would be distributed to the lockstep and the fault would not be detected.

Instead, I propose to instantiate the _ibex_trvk_ module twice and drive it with the main and the shadow core.

thommythomaso: Thanks @nasahlpa. Indeed, the current implementation is not yet protected against FI. The TRVK filter has to be run in lockstep as well, we opened an issue https://github.com/lowRISC/ibex/issues/2467 to keep track of this task.

nasahlpa: Ah thanks, I've missed that issue :-)

Would it make sense to have this issue in the RTL as well? But I am fine if not - we just need to make sure not to forget that.

rtl/ibex_top.sv
```diff
@@ -1148,15 +1187,114 @@ module ibex_top import ibex_pkg::*; #(
...
+      .upstream_rvalid_o      (trvk_rvalid),
+      .upstream_we_i          (trvk_we),
+      .upstream_be_i          (trvk_be),
+      .upstream_addr_i        (trvk_addr),
+      .upstream_wdata_i       (trvk_wdata),
+      .upstream_wdata_intg_i  (trvk_wdata_intg),
+      .upstream_rdata_o       (trvk_rdata),
+      .upstream_rdata_intg_o  (trvk_rdata_intg),
+      .upstream_err_o         (trvk_err),
+      .upstream_tag_i         (trvk_rtag),
```
andreaskurth: The `w`/`r` naming looks inverted after this swap. In `ibex_trvk`, `upstream_tag_i` feeds `downstream_tag_o` (core to memory, i.e. the write tag) and `upstream_tag_o` is derived from the response tag (memory to core, the read tag). So this makes `trvk_rtag` the store tag and `trvk_wtag` the load tag; also the opposite of what @SamuelRiedel described in the thread below ("the wtag is always driven by ibex ... the rtag is driven by the trvk ... and always goes to the core").

Should the connection go back to `.upstream_tag_i(trvk_wtag)` / `.upstream_tag_o(trvk_rtag)` and keep only the `unused_trvk` part of the fix? The tie-off comment above would then match again too -- as it stands, `trvk_rtag` is tied off but carries the direction away from the core.

andreaskurth (commented):
Thanks @thommythomaso, this addresses everything from the last round. Just one thing left, below.

rtl/ibex_top.sv
thommythomaso: The direction of the tags and the tie-offs are now consistent.

andreaskurth (approved):
LGTM, thanks @thommythomaso!

PR #2457, opened 2026-07-02
[rtl] Generalize SRAM configuration interfaces

thommythomaso:
This PR renames the SRAM cfg ports to align with https://github.com/lowRISC/opentitan/pull/30509.


SamuelRiedel (approved):
This looks good to me, thanks. If I understood correctly, we need the latest version of the RAM prims, but other updates in the prims currently break in Ibex. I suggest we create an issue to pull in the latest version of the prims and align Ibex to it in the future to minimize the patches again.

PR #2459, opened 2026-07-13
[rtl] Switch to ratified bitmanip extensions

thommythomaso:
This PR switches the RTL of Ibex from our currently unratified bitmanip extension to the officially ratified extensions. 

An overview of the changes can be found [here](https://docs.google.com/presentation/d/1Z2lPl5-F6TvsAc5hoF7DbrsuqBSoYQzXY0x-95SaBJo/edit?usp=sharing).

This PR only introduces the RTL changes and ensures the CI tests pass. Certain scalar crypto extensions are not yet fully verified; see #2468.

### DIFF
--- a/doc/01_overview/compliance.rst
+++ b/doc/01_overview/compliance.rst
@@ -8,7 +8,8 @@ It follows these specifications:
 * `RISC-V Instruction Set Manual, Volume II: Privileged Architecture, document version 20211203 (December 4, 2021) <https://github.com/riscv/riscv-isa-manual/releases/download/Priv-v1.12/riscv-privileged-20211203.pdf>`_.
   Ibex implements the Machine ISA version 1.12.
 * `RISC-V External Debug Support, version 0.13.2 <https://content.riscv.org/wp-content/uploads/2019/03/riscv-debug-release.pdf>`_
-* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ and `version 0.93 (draft from January 10, 2021) <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.93.pdf>`_
+* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ (Zba, Zbb, Zbc, Zbs sub-extensions)
+* `RISC-V Scalar Cryptography Extension, version 1.0.1 <https://github.com/riscv/riscv-crypto/releases/download/v1.0.1-scalar/riscv-crypto-spec-scalar-v1.0.1.pdf>`_ (Zbkb, Zbkx sub-extensions only)
 * `PMP Enhancements for memory access and execution prevention on Machine mode (Smepmp) version 1.0 <https://github.com/riscv/riscv-tee/blob/191b563b08b31cc2974d604a3b670d8666a2e093/Smepmp/Smepmp.pdf>`_
 
 Many features in the RISC-V specification are optional, and Ibex can be parametrized to enable or disable some of them.
@@ -35,8 +36,12 @@ In addition, the following instruction set extensions are available.
      - 2.0
      - optional
 
-   * - **B**: Standard Extension for Bit-Manipulation Instructions
-     - 1.0.0 + 0.93 [#B_draft]_
+   * - **B**: Standard Extension for Bit-Manipulation Instructions (Zba, Zbb, Zbc, Zbs)
+     - 1.0.0
+     - optional
+
+   * - **Zbkb, Zbkx**: Scalar Cryptography Bit-Manipulation Instructions
+     - 1.0.1
      - optional
 
    * - **Zicsr**: Control and Status Register Instructions
@@ -69,12 +74,3 @@ Ibex currently supports the following features according to the RISC-V Privilege
 * All CSRs listed in :ref:`cs-registers`
 * Performance counters as described in :ref:`performance-counters`
 * Vectorized trap handling as described at :ref:`exceptions-interrupts`
-
-.. rubric:: Footnotes
-
-.. [#B_draft] Ibex fully implements the ratified version 1.0.0 of the RISC-V Bit-Manipulation Extension including the Zba, Zbb, Zbc and Zbs sub-extensions.
-   In addition, Ibex also supports the remaining Zbe, Zbf, Zbp, Zbr and Zbt sub-extensions as defined in draft version 0.93 of the RISC-V Bit-Manipulation Extension.
-   Note that the latter sub-extensions may change before being ratified as a standard by the RISC-V Foundation.
-   Ibex will be updated to match future versions of the specification.
-   Prior to ratification this may involve backwards incompatible changes.
-   Additionally, neither GCC or Clang have committed to maintaining support upstream for unratified versions of the specification.
--- a/doc/03_reference/instruction_decode_execute.rst
+++ b/doc/03_reference/instruction_decode_execute.rst
@@ -65,43 +65,35 @@ Other blocks use the ALU for the following tasks:
 * The LSU uses it to increment addresses when performing two accesses to handle an unaligned access
 
 Bit-Manipulation Extension
-  Support for the `RISC-V Bit-Manipulation Extension version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ and `draft version 0.93 from January 10, 2021 <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.93.pdf>`_ is optional. [#B_draft]_
+  Support for the `RISC-V Bit-Manipulation Extension version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_, together with the Zbkb and Zbkx sub-extensions of the ratified Scalar Cryptography Extension, is optional.
   It can be enabled via the enumerated parameter ``RV32B`` defined in :file:`rtl/ibex_pkg.sv`.
   By default, this parameter is set to "ibex_pkg::RV32BNone" to disable the bit-manipulation extension.
 
-  There are three versions of the bit-manipulation extension available:
+  There are two versions of the bit-manipulation extension available:
   The balanced version comprises a set of sub-extensions aiming for good benefits at a reasonable area overhead.
   It can be selected by setting the ``RV32B`` parameter to "ibex_pkg::RV32BBalanced".
-  The OTEarlGrey version comprises all sub-extensions except for the Zbe.
-  This version can be selected by setting the ``RV32B`` parameter to "ibex_pkg::RV32BOTEarlGrey".
-  The full version comprises all sub-extensions and can be selected by setting the ``RV32B`` parameter to "ibex_pkg::RV32BFull".
+  The full version additionally comprises the Zbc, Zbkb and Zbkx sub-extensions and can be selected by setting the ``RV32B`` parameter to "ibex_pkg::RV32BFull".
   The following table gives an overview of which sub-extensions are implemented in each version and of which instructions are implemented as multi-cycle instructions.
   Multi-cycle instructions are completed in 2 cycles.
   All remaining instructions complete in a single cycle.
 
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Bit-Manipulation Sub-Extension | Spec.   | Balanced | OTEarlGrey | Full | Multi-Cycle Instr. |
-  +================================+=========+==========+============+======+====================+
-  | Zba (Address generation)       | v.1.0.0 |    X     |     X      |  X   | None               |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbb (Base)                     | v.1.0.0 |    X     |     X      |  X   | rol, ror[i]        |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbc (Carry-less multiply)      | v.1.0.0 |          |     X      |  X   | None               |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbs (Single-bit)               | v.1.0.0 |    X     |     X      |  X   | None               |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbe (Bit compress/decompress)  | v.0.93  |          |            |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbf (Bit-field place)          | v.0.93  |    X     |     X      |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbp (Permutation)              | v.0.93  |          |     X      |  X   | None               |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbr (CRC)                      | v.0.93  |          |     X      |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbt (Ternary)                  | v.0.93  |    X     |     X      |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
+  +------------------------------------+---------+----------+------+--------------------+
+  | Bit-Manipulation Sub-Extension     |  Spec.  | Balanced | Full | Multi-Cycle Instr. |
+  +====================================+=========+==========+======+====================+
+  | Zba (Address generation)           | v.1.0.0 |    X     |  X   |        None        |
+  +------------------------------------+---------+----------+------+--------------------+
+  | Zbb (Base)                         | v.1.0.0 |    X     |  X   |    rol, ror[i]     |
+  +------------------------------------+---------+----------+------+--------------------+
+  | Zbc (Carry-less multiply)          | v.1.0.0 |          |  X   |        None        |
+  +------------------------------------+---------+----------+------+--------------------+
+  | Zbs (Single-bit)                   | v.1.0.0 |    X     |  X   |        None        |
+  +------------------------------------+---------+----------+------+--------------------+
+  | Zbkb (Bit-manipulation for crypto) | v.1.0.0 |          |  X   |        None        |
+  +------------------------------------+---------+----------+------+--------------------+
+  | Zbkx (Crossbar permutation)        | v.1.0.0 |          |  X   |        None        |
+  +------------------------------------+---------+----------+------+--------------------+
 
-  The implementation of the Bit-Manipulation Extension comes with an area overhead of 2.7 kGE for the balanced version, 6.1 kGE for the OTEarlGrey version, and 7.5 kGE for the full version.
+  The implementation of the Bit-Manipulation Extension comes with an area overhead of 2.4 kGE for the balanced version and 5.9 kGE for the full version, relative to a configuration without any bit-manipulation support.
   These numbers were obtained by synthesizing the design with Yosys and relaxed timing constraints.
 
 
@@ -169,12 +161,3 @@ Source File: :file:`rtl/ibex_load_store_unit.sv`
 
 The Load-Store Unit (LSU) interfaces with main memory to perform load and store operations.
 See :ref:`load-store-unit` for more details.
-
-.. rubric:: Footnotes
-
-.. [#B_draft] Ibex fully implements the ratified version 1.0.0 of the RISC-V Bit-Manipulation Extension including the Zba, Zbb, Zbc and Zbs sub-extensions.
-   In addition, Ibex also supports the remaining Zbe, Zbf, Zbp, Zbr and Zbt sub-extensions as defined in draft version 0.93 of the RISC-V Bit-Manipulation Extension.
-   Note that the latter sub-extensions may change before being ratified as a standard by the RISC-V Foundation.
-   Ibex will be updated to match future versions of the specification.
-   Prior to ratification this may involve backwards incompatible changes.
-   Additionally, neither GCC or Clang have committed to maintaining support upstream for unratified versions of the specification.
--- a/dv/cosim/spike_cosim.cc
+++ b/dv/cosim/spike_cosim.cc
@@ -73,7 +73,8 @@ SpikeCosim::SpikeCosim(const std::string &isa_string, uint32_t start_pc,
   processor->set_ibex_flags(secure_ibex, icache_en);
   processor->set_debug_module_range(dm_start_addr, dm_end_addr);
 
-  initial_proc_setup(start_pc, start_mtvec, mhpm_counter_num);
+  bool rv32b_enabled = isa_string.find("Zb") != std::string::npos;
+  initial_proc_setup(start_pc, start_mtvec, mhpm_counter_num, rv32b_enabled);
 
   if (log) {
     processor->set_debug(true);
@@ -565,12 +566,20 @@ void SpikeCosim::set_cpuctrlsts_double_fault_seen() {
 }
 
 void SpikeCosim::initial_proc_setup(uint32_t start_pc, uint32_t start_mtvec,
-                                    uint32_t mhpm_counter_num) {
+                                    uint32_t mhpm_counter_num,
+                                    bool rv32b_enabled) {
   processor->get_state()->pc = start_pc;
   processor->get_state()->mtvec->write(start_mtvec);
 
   processor->get_state()->csrmap[CSR_MARCHID] =
       std::make_shared<const_csr_t>(processor.get(), CSR_MARCHID, IBEX_MARCHID);
+
+  uint32_t misa_val = processor->get_csr(CSR_MISA);
+  if (rv32b_enabled) {
+    misa_val |= 1 << 1;
+  }
+  processor->get_state()->csrmap[CSR_MISA] =
+      std::make_shared<const_csr_t>(processor.get(), CSR_MISA, misa_val);
 
   processor->set_mmu_capability(IMPL_MMU_SBARE);
 
@@ -821,16 +830,6 @@ void SpikeCosim::fixup_csr(int csr_num, uint32_t csr_val) {
       // For Ibex, mtvec.MODE is set to vectored and
       // mtvec.BASE must be 256-byte aligned
       reg_t new_val = (csr_val & mtvec_and_mask) | mtvec_or_mask;
-#ifdef OLD_SPIKE
-      processor->set_csr(csr_num, new_val);
-#else
-      processor->put_csr(csr_num, new_val);
-#endif
-      break;
-    }
-    case CSR_MISA: {
-      // For Ibex, misa is hardwired
-      reg_t new_val = 0x40901104;
 #ifdef OLD_SPIKE
       processor->set_csr(csr_num, new_val);
 #else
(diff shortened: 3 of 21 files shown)


rtl/ibex_decoder.sv
```diff
@@ -465,13 +465,21 @@ module ibex_decoder #(
             {7'b000_0101, 3'b110}, // max
             {7'b000_0101, 3'b101}, // minu
             {7'b000_0101, 3'b111}, // maxu
-            {7'b000_0100, 3'b100}, // pack
-            {7'b000_0100, 3'b111}, // packh
```
marnovandermaas, line 487: The zexth commit also remove pack and packh

rtl/ibex_decoder.sv
```diff
@@ -844,8 +844,9 @@ module ibex_decoder #(
                   5'b0_0000: alu_operator_o = ALU_SRL;   // Shift Right Logical by Immediate
                   5'b0_1000: alu_operator_o = ALU_SRA;   // Shift Right Arithmetically by Immediate
                   5'b0_1001: alu_operator_o = ALU_BEXT;  // Extract bit specified by immediate.
+                  // rori (Zbb, also imported by Zbkb): rotate right by immediate
```
marnovandermaas: Why does changing this comment complete Zbkb?

doc/01_overview/compliance.rst
```diff
@@ -8,7 +8,8 @@ It follows these specifications:
 * `RISC-V Instruction Set Manual, Volume II: Privileged Architecture, document version 20211203 (December 4, 2021) <https://github.com/riscv/riscv-isa-manual/releases/download/Priv-v1.12/riscv-privileged-20211203.pdf>`_.
   Ibex implements the Machine ISA version 1.12.
 * `RISC-V External Debug Support, version 0.13.2 <https://content.riscv.org/wp-content/uploads/2019/03/riscv-debug-release.pdf>`_
-* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ and `version 0.93 (draft from January 10, 2021) <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.93.pdf>`_
+* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_
+* `RISC-V Scalar Cryptography Extension, version 1.0.1 <https://github.com/riscv/riscv-crypto/releases/download/v1.0.1-scalar/riscv-crypto-spec-scalar-v1.0.1.pdf>`_ (Zbkb, Zbkx sub-extensions only)
```
marnovandermaas, line 12: The commit for this shouldn't be marked as RTL. Probably [doc,dv] would be better.

marnovandermaas (commented):
Initial comments from me. Do you think it is worth squashing the commits that just remove things? I think it may be excessive to have 9 separate commits for removing instructions.

thommythomaso:
> Initial comments from me. Do you think it is worth squashing the commits that just remove things? I think it may be excessive to have 9 separate commits for removing instructions.

I have them as individual commits to help people review the changes. I would have squashed all into one after the review process. What do you think?

doc/01_overview/compliance.rst
```diff
@@ -8,7 +8,8 @@ It follows these specifications:
 * `RISC-V Instruction Set Manual, Volume II: Privileged Architecture, document version 20211203 (December 4, 2021) <https://github.com/riscv/riscv-isa-manual/releases/download/Priv-v1.12/riscv-privileged-20211203.pdf>`_.
   Ibex implements the Machine ISA version 1.12.
 * `RISC-V External Debug Support, version 0.13.2 <https://content.riscv.org/wp-content/uploads/2019/03/riscv-debug-release.pdf>`_
-* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ and `version 0.93 (draft from January 10, 2021) <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.93.pdf>`_
+* `RISC-V Bit-Manipulation Extension, version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_
```
SamuelRiedel: Should we explicitly mention the subextensions? The "B" only stands for Zba, Zbb, and Zbs in the ISA string, but then the Zbc is also part of the bit manipulation extension. This is a bit confusing in the RISC-V spec already, but I would explicitly mention what we support already in this file here and in the list below.

doc/01_overview/compliance.rst
```diff
@@ -36,7 +37,11 @@ In addition, the following instruction set extensions are available.
      - optional
 
    * - **B**: Standard Extension for Bit-Manipulation Instructions
-     - 1.0.0 + 0.93 [#B_draft]_
+     - 1.0.0 [#B_draft]_
```
SamuelRiedel: This footnote doesn't really make sense anymore. We have the same footnote, named B_draft internally that describes what the fully ratified B extension entails and then also refer to the same footnote for the scalar crypto one. Let's have one footnote each or just leave them away, since the scalar crypto is already clear on its own and if we clarify the bitmanip above already, I also don't think we need a footnote specifying the subextensions anymore.

doc/03_reference/instruction_decode_execute.rst
```diff
@@ -65,44 +65,33 @@ Other blocks use the ALU for the following tasks:
 * The LSU uses it to increment addresses when performing two accesses to handle an unaligned access
 
 Bit-Manipulation Extension
-  Support for the `RISC-V Bit-Manipulation Extension version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_ and `draft version 0.93 from January 10, 2021 <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.93.pdf>`_ is optional. [#B_draft]_
+  Support for the `RISC-V Bit-Manipulation Extension version 1.0.0 <https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf>`_, together with the Zbkb and Zbkx sub-extensions of the ratified Scalar Cryptography Extension, is optional. [#B_draft]_
```
SamuelRiedel: Again, outdated footnote name

doc/03_reference/instruction_decode_execute.rst
```diff
@@ -65,44 +65,33 @@ Other blocks use the ALU for the following tasks:
...
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbp (Permutation)              | v.0.93  |          |     X      |  X   | None               |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbr (CRC)                      | v.0.93  |          |     X      |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-  | Zbt (Ternary)                  | v.0.93  |    X     |     X      |  X   | All                |
-  +--------------------------------+---------+----------+------------+------+--------------------+
-
-  The implementation of the Bit-Manipulation Extension comes with an area overhead of 2.7 kGE for the balanced version, 6.1 kGE for the OTEarlGrey version, and 7.5 kGE for the full version.
-  These numbers were obtained by synthesizing the design with Yosys and relaxed timing constraints.
```
SamuelRiedel, line 105: Do we have any new numbers for those? Even if we could simply estimate them?

dv/cosim/spike_cosim.cc
```diff
@@ -565,13 +566,21 @@ void SpikeCosim::set_cpuctrlsts_double_fault_seen() {
...
   processor->get_state()->csrmap[CSR_MARCHID] =
       std::make_shared<const_csr_t>(processor.get(), CSR_MARCHID, IBEX_MARCHID);
 
+  uint32_t misa_val = processor->get_csr(CSR_MISA);
+  if (rv32b_enabled) {
+    misa_val |= 1 << 1;
+  }
+  processor->get_state()->csrmap[CSR_MISA] =
+      std::make_shared<const_csr_t>(processor.get(), CSR_MISA, misa_val);
+
```
SamuelRiedel, line 583: Why is the B extension a special case here where we need to parse the ISA string ourselves and inject it into the MISA register? Shouldn't this be handled the same way as all the other extensions?

dv/uvm/core_ibex/riscv_dv_extension/ml_testlist.yaml
```diff
@@ -80,7 +80,7 @@
     +disable_compressed_instr=0
     +randomize_csr=0
     +enable_b_extension=1
-    +enable_bitmanip_groups=zbb,zb_tmp,zbt,zbs,zbp,zbf,zbe,zbc,zbr
+    +enable_bitmanip_groups=zbb,zb_tmp,zbs,zbp,zbc
```
SamuelRiedel: RISC-V DV will be outdated compared to the extensions Ibex implements now, right? So zb_tmp, and zbp will need to be restructured/removed at some point within riscv-dv as well. But currently, they are emitting supported instructions that are now in the the Zbb, the Zbkb, and Zbkx extensions, right?

dv/uvm/core_ibex/riscv_dv_extension/testlist.yaml
```diff
@@ -1153,7 +1136,6 @@
     +enable_zbb_extension=1
     +enable_zbs_extension=1
     +enable_b_extension=1
-    +enable_bitmanip_groups=zbf,zbt
```
SamuelRiedel, line 1156: This test will still generate lots of the unratified bitmanip instructions. The `enable_b_extension` flag will enable all extensions and the `enable_bitmanip_groups` flag specified which subgroups should really be active. By default, all of them, so by removing this line, we now have all the unratified old bitmanip extensions enabled here. We should remove the `enable_b_extension` as well.

examples/simple_system/ibex_simple_system.cc
```diff
@@ -53,11 +53,7 @@ std::string SimpleSystem::GetIsaString() const {
       extensions += "_Zba_Zbb_Zbs_XZbf_XZbt";
       break;
 
-    case 2:  // RV32BOTEarlGrey
-      extensions += "_Zba_Zbb_Zbc_Zbs_XZbf_XZbp_XZbr_XZbt";
-      break;
-
-    case 3:  // RV32BFull
+    case 2:  // RV32BFull
       extensions += "_Zba_Zbb_Zbc_Zbs_XZbe_XZbf_XZbp_XZbr_XZbt";
```
SamuelRiedel: Here we still have all the old extensions.

rtl/ibex_decoder.sv
```diff
@@ -144,32 +140,15 @@ module ibex_decoder #(
...
-    assign unused_clk = clk_i;
-    assign unused_rst_n = rst_ni;
-
-    // always zero
-    assign use_rs3_q = use_rs3_d;
-  end
+  logic unused_clk;
+  logic unused_rst_n;
+  assign unused_clk = clk_i;
+  assign unused_rst_n = rst_ni;
```
SamuelRiedel: Let's entirely remove the clock and reset if we have no more state in the decoder in any configuration.

rtl/ibex_decoder.sv
```diff
@@ -402,41 +371,34 @@ module ibex_decoder #(
...
             end else begin
               unique case (instr[31:27])
                 5'b0_0000,                                                             // srli
                 5'b0_1000: illegal_insn = (instr[26:25] == 2'b00) ? 1'b0 : 1'b1;       // srai
 
-                5'b0_0100: begin                                                       // sroi
-                  illegal_insn = (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) ? 1'b0 : 1'b1;
-                end
                 5'b0_1100,                                                             // rori
                 5'b0_1001: illegal_insn = (RV32B != RV32BNone) ? 1'b0 : 1'b1;          // bexti
```
SamuelRiedel: These instructions require `instr[25]` to be zero as well in RV32, which is never checked, right?

rtl/ibex_decoder.sv
```diff
@@ -365,15 +344,13 @@ module ibex_decoder #(
           3'b001: begin
             unique case (instr[31:27])
               5'b0_0000: illegal_insn = (instr[26:25] == 2'b00) ? 1'b0 : 1'b1;        // slli
-              5'b0_0100: begin                                                        // sloi
-                illegal_insn = (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) ? 1'b0 : 1'b1;
-              end
               5'b0_1001,                                                              // bclri
               5'b0_0101,                                                              // bseti
               5'b0_1101: illegal_insn = (RV32B != RV32BNone) ? 1'b0 : 1'b1;           // binvi
```
SamuelRiedel: These instructions require `instr[25]` to be zero as well in RV32, which is never checked, right?

rtl/ibex_decoder.sv
```diff
@@ -453,7 +415,7 @@ module ibex_decoder #(
         rf_ren_b_o      = 1'b1;
         rf_we           = 1'b1;
         if ({instr[26], instr[13:12]} == {1'b1, 2'b01}) begin
-          illegal_insn = (RV32B != RV32BNone) ? 1'b0 : 1'b1; // cmix / cmov / fsl / fsr
+          illegal_insn = 1'b1; // cmix / cmov / fsl / fsr removed
```
SamuelRiedel: Do we still need this? They will simply fall under the default case later, right? I think the comment will be a bit confusing if someone looks at this without the context of the bitmanip just being cleaned up.

rtl/ibex_decoder.sv
```diff
@@ -950,47 +865,7 @@ module ibex_decoder #(
...
-                if (instr_first_cycle_i) begin
-                  use_rs3_d = 1'b1;
-                end else begin
-                  use_rs3_d = 1'b0;
-                end
-              end
-              default: ;
-            endcase
-          end
+          // Ternary bitmanip ops (cmix/cmov/fsl/fsr) removed; these encodings are illegal.
```
SamuelRiedel: Here we again have an empty if block and the else only contains a switch case with legal encodings, so we could directly use the switch case only, no?

rtl/ibex_tracer_pkg.sv
```diff
@@ -124,174 +116,30 @@ package ibex_tracer_pkg;
   parameter logic [31:0] INSN_BEXT = { 7'b0100100, 10'h?, 3'b101, 5'h?, {OPCODE_OP} };
 
   // ZBP
```
SamuelRiedel: Mention of old extension

rtl/ibex_alu.sv
```diff
@@ -588,68 +503,65 @@ module ibex_alu #(
...
+    // rev8 / brev8 / orc.b        //
+    ////////////////////////////////
 
-    // Only a subset of the general reverse and or-combine instructions are implemented in the
-    // balanced version of the B extension. Currently rev8 (shift_amt = 5'b11000) and orc.b
-    // (shift_amt = 5'b00111) are supported in the base extension.
+    // The generalized-reverse / or-combine butterfly below implements the ratified
+    // subset only: rev8  and orc.b.
 
     logic [4:0] zbp_shift_amt;
```
SamuelRiedel: Mention of old extension in variable name

SamuelRiedel (commented):
Thanks for cleaning up the bitmanip @thommythomaso. I did a quite thorough review, but most of it should be easy to fix.

The only thing that is left is that we should update RISC-V DV to also support the latest ratified extensions. I saw that the ratified B extensions are supported already, but for the scalar crypto ones, we currently simply enable the Zbp still, right? I think this is good enough for now, but we should create an issue documenting the current state, like which instructions are emitted by riscv-dv and which aren't and what we would ideally want to then clean up the DV in a second step at least.

rtl/ibex_decoder.sv
thommythomaso: Will be resolved by squashing the commits.

rtl/ibex_decoder.sv
thommythomaso: You are right, the comment does not fit the commit message. This will be fixed with the squash.

doc/03_reference/instruction_decode_execute.rst
thommythomaso: I will update the estimates.

dv/cosim/spike_cosim.cc
thommythomaso: B is not that well defined, right?

dv/uvm/core_ibex/riscv_dv_extension/ml_testlist.yaml
thommythomaso: https://github.com/lowRISC/ibex/issues/2468

doc/01_overview/compliance.rst
```diff
@@ -35,8 +36,12 @@ In addition, the following instruction set extensions are available.
      - 2.0
      - optional
 
-   * - **B**: Standard Extension for Bit-Manipulation Instructions
-     - 1.0.0 + 0.93 [#B_draft]_
+   * - **B**: Standard Extension for Bit-Manipulation Instructions (Zba, Zbb, Zbc, Zbs)
+     - 1.0.0
+     - optional
```
vogelpi, line 41: Please note that **B** just comprises Zba, Zbb and Zbs. The collection comprising Zba, Zbb, Zbc and Zbs is called RISC-V bit manipulation extension v.1.0.0 (see https://www.ece.lsu.edu/ee4720/doc/riscv-bitmanip-1.0.0.pdf, it's the same doc you referenced above but this doesn't reference "B" anymore).

So, here we should probably say:
- **B** + Zbc because that's what we use our ISA string
- or then Zba, Zbb, Zbc, Zbs as above

rtl/ibex_alu.sv
```diff
@@ -254,64 +226,26 @@ module ibex_alu #(
...
   always_comb begin
-    if (bfp_op) begin
-      shift_amt[4:0] = bfp_off;  // length field of bfp control word
-    end else begin
-      shift_amt[4:0] = instr_first_cycle_i ?
-          (operand_b_i[5] && shift_funnel ? shift_amt_compl[4:0] : operand_b_i[4:0]) :
-          (operand_b_i[5] && shift_funnel ? operand_b_i[4:0] : shift_amt_compl[4:0]);
-    end
+    shift_amt[4:0] = instr_first_cycle_i ? operand_b_i[4:0] : shift_amt_compl[4:0];
   end
```
vogelpi, line 233: You now have an always_comb with a single ternary statement in it. Can you please just convert this to an assign?

rtl/ibex_alu.sv
```diff
@@ -320,27 +254,22 @@ module ibex_alu #(
...
     // select shifter input
-    // for bfp, sbmode and shift_left the corresponding bit-reversed input is chosen.
+    // for sbmode and shift_left the corresponding bit-reversed input is chosen.
     if (RV32B == RV32BNone) begin
       shift_operand = shift_left ? operand_a_rev : operand_a_i;
     end else begin
       unique case (1'b1)
-        bfp_op:       shift_operand = bfp_mask_rev;
         shift_sbmode: shift_operand = 32'h8000_0000;
         default:      shift_operand = shift_left ? operand_a_rev : operand_a_i;
```
vogelpi, line 267: You now have quite a few `unique case (1'b1)` statements with a single case plus the default in here. I think an if/else would make more sense but I also appreciate the diff now being smaller meaning the review is simpler.

rtl/ibex_alu.sv
```diff
@@ -588,68 +503,65 @@ module ibex_alu #(
       endcase
     end
 
-    ////////////////////////////////////
-    // General Reverse and Or-combine //
-    ////////////////////////////////////
+    ////////////////////////////////
+    // rev8 / brev8 / orc.b        //
+    ////////////////////////////////
```
vogelpi, line 508: As you anyway change this, is there a reason for not minimizing the width of the header? i.e.,
```
    //////////////////////////
    // rev8 / brev8 / orc.b //
    //////////////////////////
```

rtl/ibex_alu.sv
```diff
@@ -983,317 +800,46 @@ module ibex_alu #(
...
     end
 
     //////////////////////////////////////
     // Multicycle Bitmanip Instructions //
     //////////////////////////////////////
-    // Ternary instructions + Shift Rotations + Bit Compress/Decompress + CRC
-    // For ternary instructions (zbt), operand_a_i is tied to rs1 in the first cycle and rs3 in the
-    // second cycle. operand_b_i is always tied to rs2.
+    // Shift Rotations
+    // operand_a_i is tied to rs1, operand_b_i is tied to rs2.
```
vogelpi, line 813: The shift rotations are now the only multi-cycle instructions of the ALU and those are the most simple instructions of the original set of multi-cycle instructions. IIRC, we decided to make them multi-cycle as the register is now there anyway and it was better to use it compared to also making the shifter wider.

But now, we only keep the register and the control (+ verification complexity) to implement an instruction which could probably simply be implemented with a slightly different shifter. Have you considered this option?

rtl/ibex_decoder.sv
```diff
@@ -365,15 +331,14 @@ module ibex_decoder #(
...
             unique case (instr[31:27])
               5'b0_0000: illegal_insn = (instr[26:25] == 2'b00) ? 1'b0 : 1'b1;        // slli
-              5'b0_0100: begin                                                        // sloi
-                illegal_insn = (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) ? 1'b0 : 1'b1;
-              end
               5'b0_1001,                                                              // bclri
               5'b0_0101,                                                              // bseti
-              5'b0_1101: illegal_insn = (RV32B != RV32BNone) ? 1'b0 : 1'b1;           // binvi
+              5'b0_1101: illegal_insn = (instr[26:25] == 2'b00 && RV32B != RV32BNone) ?
+                                         1'b0 : 1'b1;                                  // binvi
```
vogelpi, line 337: Nit: the `// binvi` comment should be moved left by one space ;-)

rtl/ibex_decoder.sv
```diff
@@ -402,41 +359,35 @@ module ibex_decoder #(
...
                 5'b0_0000,                                                             // srli
                 5'b0_1000: illegal_insn = (instr[26:25] == 2'b00) ? 1'b0 : 1'b1;       // srai
 
-                5'b0_0100: begin                                                       // sroi
-                  illegal_insn = (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) ? 1'b0 : 1'b1;
-                end
                 5'b0_1100,                                                             // rori
-                5'b0_1001: illegal_insn = (RV32B != RV32BNone) ? 1'b0 : 1'b1;          // bexti
+                5'b0_1001: illegal_insn = (instr[26:25] == 2'b00 && RV32B != RV32BNone) ?
+                                           1'b0 : 1'b1;                                 // bexti
```
vogelpi, line 370: Nit: move comment `// bexti` left by 1 space.

rtl/ibex_decoder.sv
```diff
@@ -402,41 +359,35 @@ module ibex_decoder #(
...
                 5'b0_0101: begin
-                  if (RV32B == RV32BOTEarlGrey || RV32B == RV32BFull) begin
-                    illegal_insn = 1'b0;                                              // gorci
-                  end else if (instr[24:20] == 5'b00111) begin
-                    illegal_insn = (RV32B == RV32BBalanced) ? 1'b0 : 1'b1;            // orc.b
-                  end else begin
-                    illegal_insn = 1'b1;
-                  end
+                  // orc.b (Zbb): gorci restricted to shamt 0x07
+                  illegal_insn = (instr[25:20] == 6'b000111 && RV32B != RV32BNone) ?
```
vogelpi, line 384: Nit: 6'00_0111 would be more readable.

vogelpi (commented):
Thanks @thommythomaso for taking this effort on in the first place and for the PR. This is nice work!

Apart from a couple of nits, I think we should do the following things before merging this PR:
- We should consider implementing ror/rol in the shifter rather than making it a multicycle instruction. Because these instructions were never a driver for introducing multi-cycle instructions in the ALU, but once the support was there, it was good to re-use the infrastructure to avoid making the shifter more complex. Now, it would be possible to have everything single cycle and I believe moving the complexity to the shifter would not increase the critical path but reduce complexity and area.
- There are quite a few single-line (plus default) unique_case statements in the decoder now. While it's easier to review the changes as is. We should rewrite these statements IMO to increase readability.

In general, I wondering how confident we are that when adding DV support for the scalar crypto, we don't find out that stuff is missing or not properly implemented? Have @SamuelRiedel and @marnovandermaas done a thorough enough review such that we are reasonable confident? On my end, I mostly focused on ensuring we don't drop an instruction which is not meant to be dropped (according to your sheet @thommythomaso ).

SamuelRiedel:
> Thanks @thommythomaso for taking this effort on in the first place and for the PR. This is nice work!
> 
> Apart from a couple of nits, I think we should do the following things before merging this PR:
> 
> * We should consider implementing ror/rol in the shifter rather than making it a multicycle instruction. Because these instructions were never a driver for introducing multi-cycle instructions in the ALU, but once the support was there, it was good to re-use the infrastructure to avoid making the shifter more complex. Now, it would be possible to have everything single cycle and I believe moving the complexity to the shifter would not increase the critical path but reduce complexity and area.
> * There are quite a few single-line (plus default) unique_case statements in the decoder now. While it's easier to review the changes as is. We should rewrite these statements IMO to increase readability.
> 
> In general, I wondering how confident we are that when adding DV support for the scalar crypto, we don't find out that stuff is missing or not properly implemented? Have @SamuelRiedel and @marnovandermaas done a thorough enough review such that we are reasonable confident? On my end, I mostly focused on ensuring we don't drop an instruction which is not meant to be dropped (according to your sheet @thommythomaso ).

@vogelpi I reviewed the hardware thoroughly, including the scalar crypto extensions, and I am confident we aren't missing any instructions. As noted though, the DV still needs cleanup. While I'd prefer a clean DV run to ensure all corner cases are covered, this PR doesn't add any new instructions/logic and the old DV still runs. Therefore, IMO we match the current D1 state, can merge the RTL, and address the DV in the next step.

thommythomaso:
I rebased now on the current master, once the CI passes, I will address the feedback.

rtl/ibex_alu.sv
thommythomaso: I left the structure to minimize the diff. But I agree this can most likely be simplified. I am currently not sure which would be the lowest-risk option to proceed with as a change in this instruction could change the critical path negatively. I would opt for leaving it as is for now with an issue that this instruction could potentially be simplified.

doc/03_reference/instruction_decode_execute.rst
```diff
@@ -65,43 +65,33 @@ Other blocks use the ALU for the following tasks:
...
+  | Zbb (Base)                         | v.1.0.0 |    X     |  X   |
+  +------------------------------------+---------+----------+------+
+  | Zbc (Carry-less multiply)          | v.1.0.0 |          |  X   |
+  +------------------------------------+---------+----------+------+
+  | Zbs (Single-bit)                   | v.1.0.0 |    X     |  X   |
+  +------------------------------------+---------+----------+------+
+  | Zbkb (Bit-manipulation for crypto) | v.1.0.0 |          |  X   |
+  +------------------------------------+---------+----------+------+
+  | Zbkx (Crossbar permutation)        | v.1.0.0 |          |  X   |
+  +------------------------------------+---------+----------+------+
```
vogelpi: Nit: this should be v.1.0.1 I believe (for Zbkb and Zbkx).

vogelpi (approved):
Many thanks @thommythomaso , this looks great!

SamuelRiedel (approved):
Thanks a lot. Looks good to me now.

PR #2460, opened 2026-07-13
Contributors license agreement

marnovandermaas:
This pull request adds a contributors license agreement to the Ibex repository. It also adds automatic checks in CI.

There is a **private** repository (for privacy reasons) where the signatures are stored: https://github.com/lowRISC/cla-signatures

And a **private** repository that shows how the action works: https://github.com/lowRISC/cla-test


ci/lint-commits.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
elliotb-lowrisc: Is there benefit in using `#!/usr/bin/env bash` ?

CLA.md
```diff
@@ -0,0 +1,31 @@
...
+
+4. You represent that you are legally entitled to grant the above license. If your employer(s) has rights to intellectual property that you create that includes your Contributions, you represent that you have received permission to make Contributions on behalf of that employer, that your employer has waived
+such rights for your Contributions to the Project, or that your employer has executed a separate Corporate CLA with the Project.
```
elliotb-lowrisc: Style oddity - line break in a paragraph

CLA.md
```diff
@@ -0,0 +1,31 @@
...
+
+5. You represent that each of Your Contributions is Your original creation (see section 7 for submissions on behalf of others). You represent that Your Contribution submissions include complete details of any third-party license or other restriction (including, but not limited to, related patents and trademarks) of which you are personally aware and which are associated with any part of Your Contributions.
+
+6. You are not expected to provide support for Your Contributions, except to the extent You desire to provide support. You may provide support for free, for a fee, or not at all. Unless required by applicable law or agreed to in writing, You provide Your Contributions on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied, including, without
+limitation, any warranties or conditions of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A PARTICULAR PURPOSE.
```
elliotb-lowrisc: Style oddity - line break in a paragraph

elliotb-lowrisc (commented):
Looks fairly good to me. A few minor comments (see inline).

Also, is it intentional that the CLA.md Markdown file is not following the OpenTitan standard style of one sentence per line? I note that README.md change is using a different style again, but is consistent with the rest of that file.

ci/lint-commits.sh
marnovandermaas: Yes, definitely!

CLA.md
marnovandermaas: Thank you, should be fixed now.

CLA.md
marnovandermaas: Thanks! This should be fixed now.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,52 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # The below token should have repo scope and must be manually added by you in the repository's secret and is needed because the signatures live in a remote repository.
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.SIGNATURE_REPO_SECRET }}
+        with:
+          path-to-signatures: 'signatures/version1/cla.json'
+          path-to-document: '${{ env.CLA_URL }}' # e.g. a CLA or a DCO document
+          # branch should not be protected
+          branch: 'main'
+          allowlist: user1,bot*
```
SamuelRiedel: Why are we allowing those users? And is it potentially risky to wildcard and allow any user starting with _bot_ to bypass the CLA? That is what this does, right?

util/lint_commits.py
```diff
@@ -0,0 +1,189 @@
...
+
+from git import Repo
+
+error_msg_prefix = 'ERROR: '
+warning_msg_prefix = 'WARNING: '
+
+# Maximum length of the summary line in the commit message (the first line)
+# There is no hard limit, but a typical convention is to keep this line at or
+# below 50 characters, with occasional outliers.
+COMMIT_MSG_MAX_SUMMARY_LEN = 100
```
SamuelRiedel, line 18: Should we directly limit this to 50 then?

util/lint_commits.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Check for an empty line separating the summary line from the long
+    # description.
+    if len(lines) > 1 and lines[1] != "":
+        error(
+            "The second line of a commit message must be empty, as it "
+            "separates the summary from the long description.", commit)
+        success = False
+
+    # Check that the commit message contains at least one Signed-off-by line
```
SamuelRiedel, line 89: Are we doing both the sign-off line and the CLA bot? I think we have to specify this in the README and contributor's guide in that case, because currently, we don't

SamuelRiedel (commented):
Thanks for implementing this @marnovandermaas. I really appreciate it. Just a few minor questions. And finally, do we need the sign-off line and the CLA signed in the comment? If so, I think we should specify this in the README. In any case, I think it could help adding just a quick mention in the README that we will prompt new users to sign the CLA when they open their first PR. Otherwise, it might be a bit confusing to figure out how to sign the CLA.

.github/workflows/cla.yml
marnovandermaas: Thanks, I've removed these because we don't need this as we don't automatically write to our repo with bots.

util/lint_commits.py
marnovandermaas: Well this would disallows the "occasional outliers". I would prefer to keep this the same as OpenTitan from where I copied this file: https://github.com/lowRISC/opentitan/blob/master/util/lint_commits.py

util/lint_commits.py
marnovandermaas: Yes, I think both is what we are expecting here. I've added this to CONTRIBUTING.md

marnovandermaas:
I've now also added the "SIGNATURE_REPO_SECRET" to the repository.

util/lint_commits.py
SamuelRiedel: Ok fair.

SamuelRiedel (approved):

marnovandermaas:
I will merge this pull request. Then we can see the CLA checks running on future pull requests.

PR #2465, opened 2026-07-30
[dv] CSR testbench fixes

### LINKED ISSUE #2242: [dv] CS_registers golden-model error in ePMP extension
## Observed Behavior

DV test `lowrisc:ibex:tb_cs_registers` cannot pass given `+ntb_random_seed=111'
```
Error, transaction:
Register transaction:
Operation:  CSR Read
Address:    PMPCfg0
Read data:  6981400
Expected rdata: 6981484
- ../src/lowrisc_ibex_tb_cs_registers_0/tb/tb_cs_registers.sv:111: Verilog $finish
Received $finish() from Verilog, shutting down simulation.
[Reg driver] drove: 4970 register transactions

//-------------//
// TEST FAILED //
//-------------//
```

The provided golden-model in `base_register.cc` is wrong when handling CSRWrite to `pmpcfg`. It should first check mseccfg.MML then write, rather than deal with `// No reserved L/R/W/X values when MML Set`. This may due to SMEPMP extension's evolving.

```c++
uint32_t PmpCfgRegister::RegisterWrite(uint32_t newval) {
  uint32_t lock_mask = GetLockMask();
  uint32_t read_value = register_value_;

  register_value_ &= lock_mask;
  register_value_ |= (newval & ~lock_mask);
  register_value_ = HandleReservedVals(register_value_);

  return read_value;
}
```

## Expected Behavior

CS_Registers test should pass, RTL's dealing with mseccfg and pmpcfg is right.
At that point, mseccfg.MML is set, write to pmpcfg results in no change.
Image

## My Environment

**EDA tool and version:**

Verilator 4.210 2021-07-07 rev v4.210

**Operating system:**

Ubuntu Linux 22.04

8f4c75c5e43e61b994248307c814e01860303056

SamuelRiedel:
Fix tb_cs_registers golden model and testbench:

- Fix reset signal always being low under newer Verilator version
- Implement SMEPMP write suppression (rule 4b) in the golden model
- Fix incorrect PMP address register locking in the golden model

@ZhongYic00 and @brdr-life, does this resolve https://github.com/lowRISC/ibex/issues/2242 for you?

Closes #2242

### DIFF
--- a/dv/cs_registers/model/base_register.cc
+++ b/dv/cs_registers/model/base_register.cc
@@ -151,12 +151,34 @@ uint32_t PmpCfgRegister::GetLockMask() {
   return lock_mask;
 }
 
+// Returns a per-byte mask for bytes whose write must be suppressed under
+// SMEPMP rule 4b: MML=1, RLB=0, and the candidate byte has lock=1 with
+// {X,W,R} in {0x4(X), 0x2(W), 0x6(W+X), 0x5(X+R)}.
+uint32_t PmpCfgRegister::GetMmlSuppressMask(uint32_t candidate) {
+  BaseRegister *mseccfg = GetRegisterFromMap(kCSRMSeccfg);
+  assert(mseccfg);
+  uint32_t mseccfg_val = mseccfg->RegisterRead();
+  if (!(mseccfg_val & kMSeccfgMml) || (mseccfg_val & kMSeccfgRlb))
+    return 0;
+  uint32_t suppress_mask = 0;
+  for (int i = 0; i < 4; i++) {
+    uint8_t byte_val = (candidate >> (8 * i)) & 0xFF;
+    uint8_t xwr = byte_val & 0x7;
+    if ((byte_val & 0x80) &&
+        (xwr == 0x4 || xwr == 0x2 || xwr == 0x6 || xwr == 0x5)) {
+      suppress_mask |= (0xFFu << (8 * i));
+    }
+  }
+  return suppress_mask;
+}
+
 uint32_t PmpCfgRegister::RegisterWrite(uint32_t newval) {
   uint32_t lock_mask = GetLockMask();
   uint32_t read_value = register_value_;
+  uint32_t suppress_mask = GetMmlSuppressMask(newval);
 
-  register_value_ &= lock_mask;
-  register_value_ |= (newval & ~lock_mask);
+  register_value_ &= (lock_mask | suppress_mask);
+  register_value_ |= (newval & ~lock_mask & ~suppress_mask);
   register_value_ = HandleReservedVals(register_value_);
 
   return read_value;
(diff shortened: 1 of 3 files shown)


SamuelRiedel:

ZhongYic00:
LGTM 👏🏻

dv/cs_registers/tb/tb_cs_registers.sv
```diff
@@ -43,7 +43,13 @@ module tb_cs_registers #(
   //-----------------
   // Allow reset to be toggled by the top-level (in Verilator)
   // or a DPI call
+  // Note: under Verilator, in_rst_ni is an inout that eval_initial forces to 0,
+  // so we rely solely on the DPI reset driver (dpi_rst_ni).
```
elliotb-lowrisc, line 47: What is `eval_initial`?

dv/cs_registers/model/base_register.cc
```diff
@@ -151,12 +151,34 @@ uint32_t PmpCfgRegister::GetLockMask() {
   return lock_mask;
 }
 
+// Returns a per-byte mask for bytes whose write must be suppressed under
+// SMEPMP rule 4b: MML=1, RLB=0, and the candidate byte has lock=1 with
+// {X,W,R} in {0x4(X), 0x2(W), 0x6(W+X), 0x5(X+R)}.
```
elliotb-lowrisc, line 156: I suspect the latest official release of the RISC-V Priv spec (version 20260120) doesn't have "SMEPMP rule 4b" _as such_ anymore. Some content that may have been it seems to have been moved into a "Historical Rationale for Extensions" appendix (A.1.4.b), which doesn't seem as strong a thing to reference. Perhaps clarify or replace this reference?

elliotb-lowrisc: I found some other things confusing, but perhaps that was because I am had little prior knowledge of PMP.

1. I was confused that if this only restricted locked cfgs then it seemed to be redundant w.r.t. `PmpCfgRegister::GetLockMask()`, until I realised that this checked the value being _written_, rather than the existing value. Just me not catching the intended meaning of "candidate".
2. I was additionally confused for some time about why these particular combinations of permissions were included, until I read 3.1.19 "Adding a rule with __executable__ privileges that either is M-mode-only or a locked Shared-Region..." and the table in 6.2 together closely.

A larger explanatory comment here would be nice to avoid any future misunderstanding, but ultimately anyone making changes here should have a good understanding of the spec (which I apparently didn't) already.

dv/cs_registers/model/base_register.cc
```diff
@@ -175,8 +199,11 @@ uint32_t PmpCfgRegister::RegisterSet(uint32_t newval) {
 uint32_t PmpCfgRegister::RegisterClear(uint32_t newval) {
   uint32_t lock_mask = GetLockMask();
   uint32_t read_value = register_value_;
+  uint32_t candidate = register_value_ & (~newval | lock_mask);
+  uint32_t suppress_mask = GetMmlSuppressMask(candidate);
 
-  register_value_ &= (~newval | lock_mask);
+  register_value_ = (register_value_ & (lock_mask | suppress_mask)) |
+                    (candidate & ~lock_mask & ~suppress_mask);
```
elliotb-lowrisc, line 206: I don't think this is quite right. If both masks are all zeros (allow any bit to be cleared), then the existing register value will be entirely lost.

How about:
```
register_value_ &= (~newval | lock_mask | suppress_mask);
```

elliotb-lowrisc (changes requested):

PR #2470, opened 2026-08-10
[rtl] Make fetch FIFO clear dominate updates

Srinidhi-Magesh:
Local synthesis identifies fetch FIFO payload/error updates on the taken-branch redirect path. Qualifying their enables with !clear_i improves arrival time from 2.89 ns to 2.77 ns at a measured 0.44% area cost.
The change has passed directed clear/push/pop testing, Ibex hello_test, and a standalone SymbiYosys/Z3 proof.
Would this timing/area tradeoff and implementation-level payload stability behavior be considered appropriate for an upstream RTL change?

### DIFF
--- a/rtl/ibex_fetch_fifo.sv
+++ b/rtl/ibex_fetch_fifo.sv
@@ -239,14 +239,14 @@ module ibex_fetch_fifo #(
         if (!rst_ni) begin
           rdata_q[i] <= '0;
           err_q[i]   <= '0;
-        end else if (entry_en[i]) begin
+        end else if (entry_en[i] && !clear_i) begin
           rdata_q[i] <= rdata_d[i];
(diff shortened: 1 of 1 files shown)


Srinidhi-Magesh:
recheck

thommythomaso:
Hi @Srinidhi-Magesh, thank you for looking into Ibex. I checked your changes, and I would like to know a bit more on how you arrived at the synthesis results you have presented in your message.

* What synthesis setup (tool, version, PDK, std-cell library, corner) did you use for your analysis?
* How exactly did you measure the arrival time? Does it impact the critical path?
* To what area does the 0.44% relate to? If it is measured compared to Ibex an increase of 0.44% seems too significant for just adding two additional gates.

PR #2471, opened 2026-08-11
[ci] Skip CLA check in the merge queue

andreaskurth:
Prior to this commit, the `CLA Assistant` workflow was not triggered
for merge groups.  The required `CLAAssistant` check therefore never
reported a result for a merge group, which blocked the merge queue
until the check timed out.

The CLA Assistant action cannot check a merge group, because it needs
a pull request to determine the authors and to post its comment on.

This commit adds a `merge_group` trigger to the workflow.  The
condition on the only step of the workflow is false for that event, so
the job completes without running any step and the `CLAAssistant`
check reports success.  The CLA is thus only checked on a pull
request, before that pull request enters the merge queue.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -2,6 +2,12 @@ name: "CLA Assistant"
 on:
   issue_comment:
     types: [created]
+  # The CLA Assistant action cannot check a merge group, so the job below runs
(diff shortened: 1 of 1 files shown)


andreaskurth:

thommythomaso (approved):

marnovandermaas:
Thanks for making this change!

PR #2475, opened 2026-08-16
[rtl,cheriot] Implement CHERIoT ISA 

SamuelRiedel:
## Implement CHERIoT ISA

This PR integrates the full [CHERIoT ISA specification v1.0](https://github.com/microsoft/cheriot-sail/releases/) into Ibex, building on the rebased [microsoft/cheriot-ibex](https://github.com/microsoft/cheriot-ibex) code (the first commit in this series).

### Main features

This PR adds complete CHERIoT ISA support to Ibex. The key design decisions include:

*   **Merged integration**: All CHERIoT logic (decode, ID/EX control, CSRs, LSU, and register file) is merged directly into the existing Ibex pipeline modules. Only the capability execution stage (`ibex_cheriot_ex`) is a standalone new module.
*   **Spec-aligned types** (`ibex_cheriot_pkg`): Capability types are packed structs whose lower bits match the spec encoding 1:1, so their vector forms match the format in memory. Two forms are used throughout the design: `cap_t` (compressed capabilities with correction factors, which live in the register file and on load/store ports) and `decoded_cap_t` (uncompressed bounds and permissions, which are used inside the core). The correction factors (2 bits) are stored in the register file to prevent the need for expensive comparators on every capability read.
*   **No revocation engine inside the pipeline**: The TRVK filter (`ibex_trvk`) is connected after the LSU in `ibex_top` to keep the pipeline clean.
*   **No stack zeroing and no TBRE engine**: The `cheri_stkz` and `cheri_tbre` modules from the Microsoft fork are not ported. The scope of this PR is limited strictly to the ISA.
*   **Capability register file**: The lower 16 integer registers carry a `cap_t` metadata word. The upper 16 registers are integer-only. All three register file variants (FF, FPGA, and latch) are updated consistently.
*   All modules are renamed from `cheri_*` to `ibex_cheriot_*` or `ibex_*`, and legacy instruction formats are dropped.

### Parametrization

*   **`BaseIsa`** (`ibex_pkg::base_isa_e`): This is a static build-time switch. `BaseIsaRV32I` elaborates away all CHERIoT hardware (no change through this PR). `BaseIsaRV32IorCHERIoT` includes both RV32I and CHERIoT modes.
*   **`cheriot_enable_i`** (`ibex_mubi_t`): This is a dynamic runtime switch. When CHERIoT hardware is present and this signal is deasserted, the core behaves as a plain RV32I core. This signal has to be held constant throughout execution with one exception. Ibex can switch from Off (RV32I mode) to On (CHERIoT mode) exactly once at runtime. Ibex can never switch from CHERIoT mode to RV32I mode without a reset.

### Verification

*   RV32I regressions pass with `BaseIsaRV32I`.
*   `BaseIsaRV32IorCHERIoT` with `cheriot_enable_i` turned off runs all existing RV32I tests without regressions.
*   CHERIoT mode (`cheriot_enable_i` turned on) is verified with basic smoketests.
*   Full CHERIoT DV (both directed and constrained-random) is coming in a follow-up PR.
*   OpenTitan now has `BaseIsaRV32IorCHERIoT` enabled, meaning we have both ISAs available in this configuration. However, all current testbenches (cosim, simple systems, formal) still disable CHERIoT using the `cheriot_enable_i` signal.

### To-Dos for follow up PRs

- [x] Implement register file register sharing.
- [ ] CSR register sharing.
- [ ] DV PRs for smoketest
- [ ] DV PR for cosim with cheriot sail


rtl/ibex_register_file_ff.sv
```diff
@@ -67,6 +75,24 @@ module ibex_register_file_ff #(
     end
 
     assign rf_reg[i] = rf_reg_q;
+
+    if (BaseIsa == BaseIsaRV32IorCHERIoT) begin : g_rf_cap_flops
```
andreaskurth: This gives 32 entries of 32-bit data plus a 35-bit capability, whereas the approved RFC proposes a single 16x65-bit file where each entry serves two 32-bit GPRs in base-ISA mode, right?

As implemented, the capability flops of entries 16-31 can never hold a non-zero value: CHERIoT masks register addresses to 4 bits in `ibex_decoder`, and base-ISA mode always writes `NULL_CAP`. So that's ~560 permanently-zero flops, and synthesis probably will have a hard time optimizing them away because `rf_wcap_a_i` isn't a constant.

Could we implement the shared 16x65 arrangement here, or is there a reason it doesn't work that we should write down? I'd rather resolve this one in this PR than defer it.

rtl/ibex_compressed_decoder.sv
```diff
@@ -354,7 +390,12 @@ module ibex_compressed_decoder #(
             // (c.lui hints are translated into a lui hint)
             instr_o = {{15 {instr_i[12]}}, instr_i[6:2], instr_i[11:7], {OPCODE_LUI}};
 
-            if (instr_i[11:7] == 5'h02) begin
+            // c.incaddr16csp -> cincoffsetimm csp, csp, nzimm
```
andreaskurth, line 390: This remap (and `c.incaddr4cspn` above) is needed because a plain `addi` writes a NULL capability in CHERIoT mode. Doesn't Zcmp need the same treatment?

`cm.push`/`cm.pop` expand via `cm_sp_addi()` to `addi x2, x2, imm`, which would clear `csp`'s tag and bounds; `cm_push_store_reg()`/`cm_pop_load_reg()` use `sw`/`lw`, so `cra`'s backward sentry is lost in memory and the `jalr x0, 0(x1)` from `cm_ret_ra()` then takes a tag violation (cd = cnull, cs1 = cra requires a tagged backward sentry). `cm.mvsa01`/`cm.mva01s` similarly clear tags on argument registers. The `opentitan` config enables `RV32ZcaZcbZcmp` together with `BaseIsaRV32IorCHERIoT`, so this is reachable.

Could we either expand these to CHERIoT equivalents? Or is Zcmp incompatible with CHERIoT? Follow-up issue is fine with me, not a blocker for this PR.

rtl/ibex_cs_registers.sv
```diff
@@ -1747,4 +1979,260 @@ module ibex_cs_registers import ibex_pkg::*; #(
...
+                             (cheriot_csr_addr_i == CHERIOT_SCR_MEPCC) &&
+                             (cheriot_csr_op_i == CHERIOT_CSR_RW);
+
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        mepc_cap <= ROOT_CAP_TX;
+      end else if ((cheriot_enable_i == IbexMuBiOn) && csr_save_cause_i && ~debug_csr_save_i &&
+                   ~debug_mode_i) begin
+        mepc_cap <= pcc_exc_cap;
+      end else if ((cheriot_enable_i == IbexMuBiOn) && mepc_en) begin  // legacy cssrw; NMI recover
```
andreaskurth: On `mret` out of an NMI, `mepc_en` is asserted from the `nmi_mode_i` restore path, so this sets `mepc_cap` to `NULL_CAP`. `mstack` only saves the 32-bit `mepc`, not the capability.

Doesn't that make the interrupted handler's own `mret` install an untagged PCC and take a CHERI tag violation, i.e. turn a recoverable NMI into an unrecoverable one in CHERIoT mode? Would stacking the capability alongside `mstack_epc_q` be the fix?

Follow-up issue is fine, not a blocker.

rtl/ibex_cheriot_ex.sv
```diff
@@ -0,0 +1,1025 @@
...
+                                    && chk_cs1_otype_45) ||
+                                   ((rf_waddr_i == 0) && (rf_raddr_a_i != 5'h1)
+                                    && (chk_cs1_otype_0 || chk_cs1_otype_1)) ||
+                                   ((rf_waddr_i == 5'h1) && (chk_cs1_otype_0 | chk_cs1_otype_23)) ||
+                                   ((rf_waddr_i != 0) && (chk_cs1_otype_0 | chk_cs1_otype_1)));
+
+      perm_vio_vec[PVIO_EX]    = ~rf_fullcap_a.perms.EX;
+    end else if (cheriot_operator_i.CCSR_RW) begin
+      perm_vio_vec[PVIO_ASR]   = ~pcc_cap_i.perms.SR;
+      illegal_scr_addr         = ~debug_mode_i & (csr_addr_o < 27);
```
andreaskurth: CSpecialRW's Sail treats only SCRs 28-31 as existing and calls `handle_illegal()` for everything else. Two deviations here:

- `< 27` still admits SCR 27 (ZTOPC). Since `ztop_rdata_i`/`ztop_rcap_i` are tied to zero in `ibex_core.sv` and there is no write path, `cspecialrw cd, 27, cs1` silently reads 0 and drops the write instead of trapping.
- The `~debug_mode_i` gate makes SCRs 0-23 legal in debug mode too, where they also read 0 via the `default` arm of the SCR read mux.

Should this become `< 28`, with an explicit allowance for the debug SCRs 24-26 only? Follow-up issue is fine, not a blocker.

rtl/ibex_cs_registers.sv
```diff
@@ -666,7 +780,9 @@ module ibex_cs_registers import ibex_pkg::*; #(
         CSR_MSCRATCH: mscratch_en = 1'b1;
 
         // mepc: exception program counter
-        CSR_MEPC: mepc_en = 1'b1;
+        // disabled for pure cap mode (only allow cap writes)
+        CSR_MEPC: mepc_en = ~(BaseIsa == BaseIsaRV32IorCHERIoT)
```
andreaskurth, line 796: §7.10 says that accessing the replaced RISC-V CSRs (`mtvec`, `mepc`) via the CSR* instructions should raise a Reserved Instruction exception in CHERIoT mode.

Here, and for `CSR_MTVEC` a few lines below, the write is silently dropped instead, and the corresponding reads in the read mux still succeed. So software can `csrw mtvec, x` and get no indication that it had no effect. Should these two be forced into `illegal_csr` when `cheriot_enable_i == IbexMuBiOn`?

Follow-up issue is fine, not a blocker.

rtl/ibex_decoder.sv
```diff
@@ -636,10 +783,129 @@ module ibex_decoder #(
             default: csr_illegal = 1'b1;
           endcase
 
+          // always allow access to the following CSRs even without ASR permission
+          //   -- 0xC01-0xC9F (unpriviledged counters)
+          //   -- 0xB01-0xB9F (m-mode counters).
+          //      note 0xb01 is undefined per rvi spec. CSR register logic will handle it.
+          csr_cheriot_always_ok_o = (BaseIsa == BaseIsaRV32IorCHERIoT) &
```
andreaskurth, line 778: Table 7.1 lists only `cycle(h)`, `time(h)`, `instret(h)` and `hpmcounter(h)` (all read-only) as accessible without ASR. This condition also covers 0xB00-0xB9F, so `mcycle`/`minstret`/`mhpmcounter*` can be read without ASR permission.

Writes are separately blocked by the ASR gate on `csr_we_int`, so the effect is read access rather than tampering, but per spec those accesses should trap. Worth narrowing to the 0xC00 range? Follow-up issue is fine.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -3,69 +3,965 @@
...
+  // Implicit permission masks for each compressed permission format (spec v1.0, chapter 7.13.1)
+  parameter perms_t PERM_MRW_IMSK = '{LD:1, MC:1, SD:1, default:0}; // Memory cap-read-write
+  parameter perms_t PERM_MRO_IMSK = '{LD:1, MC:1, default:0};       // Memory cap-read-only
+  parameter perms_t PERM_MWO_IMSK = '{SD:1, MC:1, default:0};       // Memory cap-write-only
+  parameter perms_t PERM_MDO_IMSK = '{default:0};                   // Memory data-only
+  parameter perms_t PERM_EXE_IMSK = '{EX:1, MC:1, LD:1, default:0}; // Executable
+  parameter perms_t PERM_SEA_IMSK = '{default:0};                   // Sealing
+
+  // Decode the 6-bit compressed permission encoding to the full 12-bit permissions field.
+  function automatic perms_t cheriot_expand_perms(cperms_t cperms);
```
andreaskurth, line 203: `perms` is only assigned inside the if/else-if chain, with no final `else`. The six branches do cover all 64 `cperms` encodings today, so this is functionally correct, but an unconditional `perms = '0;` (or a `default`) up front would keep it X-free if the chain is ever edited. Follow-up is fine.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -3,69 +3,965 @@
...
+  function automatic logic [1:0] cheriot_get_base_correction(cap_cor_t cap_cor);
+    logic unused_top_cor_bit;
+    unused_top_cor_bit = cap_cor[1];
+    return {2{cap_cor[0]}};
+  endfunction
+
+  // Update a capability's address and recompute correction fields. Invalidate the tag if not
+  // representable.
+  // chktop/chkbase enable explicit top33/base32 bounds checks used for PCC updates.
+  function automatic decoded_cap_t cheriot_set_address(decoded_cap_t in_cap, logic [31:0] newptr,
```
andreaskurth: `chktop`/`chkbase` are 0 at all three call sites (`set_address_comb` twice and `cheriot_pcc_to_mepc`), so `ptr_below_top` and the two extra terms in the condition below are dead. Drop the two arguments?

rtl/ibex_cheriot_ex.sv
```diff
@@ -0,0 +1,1025 @@
...
+    if (cheriot_setaddr_sel_i == SETADDR_PCC_PCNXT) begin
+      tfcap1 = pcc_cap_i;
+      taddr1 = pc_id_nxt;
+    end else if (cheriot_setaddr_sel_i == SETADDR_PCC_ARITH) begin
+      tfcap1 = pcc_cap_i;
+      taddr1 = addr_result;
+    end else if (cheriot_setaddr_sel_i == SETADDR_RFA_ARITH) begin
+      tfcap1 = rf_fullcap_a;
+      taddr1 = addr_result;
+    end else if ((cheriot_setaddr_sel_i == SETADDR_SCR) && scr_legalization) begin
```
andreaskurth, line 633: This branch looks redundant. MTCC/MEPCC legalization only clears low address bits, and the `rf_rdata_a[1:0] != 2'b00` / `rf_rdata_a[0] != 1'b0` checks in `main_ex` already clear the tag in exactly the cases where the address changes, so `setaddr1_outcap`'s representability result is always overridden by `trcap.valid = rf_fullcap_a.valid`, and the recomputed `cap_cor` is identical to the input.

Can `SETADDR_SCR` and `scr_legalization` be dropped entirely?

rtl/ibex_if_stage.sv
```diff
@@ -199,8 +219,13 @@ module ibex_if_stage import ibex_pkg::*; #(
     end
 
     unique case (exc_pc_mux_i)
-      EXC_PC_EXC:     exc_pc = { csr_mtvec_i[31:8], 8'h00                };
-      EXC_PC_IRQ:     exc_pc = { csr_mtvec_i[31:8], 1'b0, irq_vec, 2'b00 };
+      EXC_PC_EXC:     exc_pc = ((BaseIsa == BaseIsaRV32IorCHERIoT)
```
andreaskurth, line 222: `csr_mtvec_i[0]` is initialised from `csr_mtvec_init_i` at boot, so a core that boots in base-ISA mode has it set. After a runtime switch to CHERIoT mode, exceptions would then be vectored here until software writes MTCC, even though CHERIoT supports direct mode only.

Should the direct-mode selection depend on `cheriot_enable_i` alone rather than also on `~csr_mtvec_i[0]`? Follow-up issue is fine.

rtl/ibex_cheriot_pkg.sv
```diff
@@ -3,69 +3,965 @@
...
+    logic LD;  // [5]  load
+    logic SL;  // [4]  store local capability
+    logic LM;  // [3]  load mutable
+    logic SD;  // [2]  store
+    logic LG;  // [1]  load global
+    logic GL;  // [0]  global
+  } perms_t;
+
+  // Sealing types (spec v1.0, chapter 7.13.2)
+  parameter otype_t OTYPE_UNSEALED       = 3'd0;
```
andreaskurth: otype 1 (interrupt-inheriting forward sentry) isn't named here, so `ibex_cheriot_ex` ends up comparing against the bare literal `3'h1` in the CJALR otype checks. Add an `OTYPE_SENTRY_INHERIT = 3'd1` for symmetry with the other four?

andreaskurth (commented):
Thanks @SamuelRiedel for this work! As agreed, the first commit imports the Microsoft's CHERIoT implementation, which is based on an earlier fork of Ibex, and rebases it onto the current Ibex, and the second commit reimplements the CHERIoT ISA in a way that fits Ibex's architecture and design very neatly. This two-commit approach makes code provenance and differences explicit, which is important.

I checked your CHERIoT ISA implementation against the v1.0 spec (chapter 9 Sail listings, the §8 encoding tables, and §7.13) and the approved OpenTitan RFC and am convinced that many/most aspects are correct - nice!

One item needs changing in this PR, I think: The register file doesn't seem to implement the RFC's shared 16x65-bit arrangement.

The remaining comments are functional and conformance gaps, and some of them I think are major, but none of them blocks this PR -> follow-up issues.

rtl/ibex_top_tracing.sv
```diff
@@ -52,6 +55,7 @@ module ibex_top_tracing import ibex_pkg::*; #(
 
   input  logic [31:0]                                                  trvk_heap_base_addr_i,
 
+  input  ibex_mubi_t                                                   cheriot_enable_i,
```
andreaskurth, line 58: This pin needs to be connected for the instance at `dv/riscv_compliance/rtl/ibex_riscv_compliance.sv:168` (hence CI currently fails).

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
```diff
@@ -35,8 +36,13 @@ class ibex_mem_intf_response_agent extends uvm_agent;
     if (!uvm_config_db#(bit)::get(null, "", "SecureIbex", secure_ibex)) begin
       secure_ibex = 1'b0;
     end
+    if (!uvm_config_db#(bit)::get(null, "", "CHERIoT", cheriot)) begin
+      cheriot = 1'b0;
+    end
 
-    cfg.fixed_data_write_response = secure_ibex;
+    // Write response data fields are X when fixed_data_write_response is unset. CHERIoT's TRVK
```
thommythomaso, line 43: Did we not have such a `DataKnown_A` assertion in Ibex before?

rtl/ibex_register_file_ff.sv
SamuelRiedel: Yes absolutely, I already started working on that but then decided to go the simple route first. I will quickly finish this and add it.

rtl/ibex_compressed_decoder.sv
SamuelRiedel: You're right. Expanding them to CHERIoT equivalents should not be an issue. I can fix this quickly.

rtl/ibex_cs_registers.sv
SamuelRiedel: I'll have a look and fix it if it's quick, or will create an issue to track this.

rtl/ibex_cheriot_ex.sv
SamuelRiedel: Yes, done.

rtl/ibex_cs_registers.sv
SamuelRiedel: Good catch. I fixed this quickly.

rtl/ibex_decoder.sv
SamuelRiedel: Done.

rtl/ibex_cheriot_pkg.sv
SamuelRiedel: This was also quicker to fix than create an issue.

rtl/ibex_cheriot_pkg.sv
SamuelRiedel: Yes, I had this on my cleanup plan, we can drop them and I did it now.

rtl/ibex_compressed_decoder.sv
SamuelRiedel: I was too quick here. Implementing it is easy, but the extensions are not really compatible IMO. Because of the `ra` (and other capabilities) we might push to the stack. The Zcmp increments the `sp` by 4 bytes for each register, but if we want to allow pushing capabilities we need 8 bytes per word/capability. I didn't see anything regarding this with CHERIoT, but CHERI mentions that [Zcmp is incompatible](https://cpp-lang.org/riscv-cheri-full.html#_rv64_rv64y_rvc_loadstore_mapping_summary), and they are creating their [own version of it](https://cpp-lang.org/riscv-cheri-full.html#ZCMP_CHERI) which is not ratified. Let's mark them as incompatible for now and defer this to later. I will keep a copy of the RTL that just uses 8-bytes per stack element.

rtl/ibex_compressed_decoder.sv
```diff
@@ -13,23 +14,23 @@
...
+  parameter ibex_pkg::rv32zc_e   RV32ZC   = ibex_pkg::RV32ZcaZcbZcmp,
+  parameter bit                  ResetAll = 1'b0,
+  parameter ibex_pkg::base_isa_e BaseIsa  = ibex_pkg::BaseIsaRV32IorCHERIoT
 ) (
   input  logic                 clk_i,
   input  logic                 rst_ni,
   input  logic                 valid_i,
   input  logic                 id_in_ready_i,
   input  logic [31:0]          instr_i,
+  input  ibex_mubi_t           cheriot_enable_i,
```
thommythomaso, line 27: There is a secure adaptations between them, right? The surrounding CHERIoT system uses OT mubi.

rtl/ibex_top.sv
```diff
@@ -157,6 +162,9 @@ module ibex_top import ibex_pkg::*; #(
   output logic [ 3:0]                                                 rvfi_mem_wmask,
   output logic [31:0]                                                 rvfi_mem_rdata,
   output logic [31:0]                                                 rvfi_mem_wdata,
+  output logic                                                        rvfi_mem_is_cap,
```
thommythomaso, line 165: the CHERIoT subsystem currently uses the same bit for write capabilities and hinting. keeping them separate in Ibex is fine though. When implementing the security countermeasures, we need to ensure that the capabilities between the two lockstep domains are implemented using mubi.

thommythomaso (commented):
I did a quick review focusing mainly on the interfaces to the rest of the CHERIoT system.

rtl/ibex_cheriot_ex.sv
SamuelRiedel: Yes, I dropped it.

rtl/ibex_if_stage.sv
SamuelRiedel: Yes, correct. I fixed this quickly.

rtl/ibex_cheriot_pkg.sv
SamuelRiedel: Right. There were also other hardcoded comparisons missing the parameter. Fixed it all.

rtl/ibex_top_tracing.sv
SamuelRiedel: Done.

dv/uvm/core_ibex/common/ibex_mem_intf_agent/ibex_mem_intf_response_agent.sv
SamuelRiedel: No, I was also surprised, but I think the DV on purpose drives X in the non-secure mode.

rtl/ibex_compressed_decoder.sv
SamuelRiedel: Yes, we had the same issue with `mcounteren_writable_i`

rtl/ibex_top.sv
SamuelRiedel: True, I can combine them.

rtl/ibex_register_file_ff.sv
```diff
@@ -67,6 +74,24 @@ module ibex_register_file_ff #(
     end
 
     assign rf_reg[i] = rf_reg_q;
+
+    if ((BaseIsa == BaseIsaRV32IorCHERIoT) && (i < 16)) begin : g_rf_cap_flops
```
andreaskurth: Following up on the earlier thread on this line, where you said you'd finish the shared file: the `i < 16` limit does remove the ~560 permanently-zero capability flops, so the dead-flop half is fixed. The RFC's actual proposal (one 16x65-bit file where each entry also serves two 32-bit GPRs in base-ISA mode) isn't there yet, so this is still 32x32 data + 16x35 cap = 1584 bits against the RFC's 1040. Are you still planning to land that here?

`ibex_register_file_latch.sv` still instantiates capability latches for all 31 entries, 16 of which can never be non-zero, so the same limit applies there. (The FPGA variant looks fine as-is; RAM inference wants the full depth.)

rtl/ibex_cs_registers.sv
```diff
@@ -1747,4 +1991,260 @@ module ibex_cs_registers import ibex_pkg::*; #(
...
+                             (cheriot_csr_op_i == CHERIOT_CSR_RW);
+
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        mepc_cap <= ROOT_CAP_TX;
+      end else if ((cheriot_enable_i == IbexMuBiOn) && csr_save_cause_i && ~debug_csr_save_i &&
+                   ~debug_mode_i) begin
+        mepc_cap <= pcc_exc_cap;
+      end else if ((cheriot_enable_i == IbexMuBiOn) && mepc_en) begin  // legacy cssrw; NMI recover
+        mepc_cap <= NULL_CAP;
```
andreaskurth: Following up on the earlier thread on this branch: still open, so an issue to track it would be good.

One thing that changed around it: now that `164c9945` makes a legacy `csrw mepc` trap in CHERIoT mode, `illegal_csr` feeds `illegal_csr_insn_o` which gates `csr_we_int`, so the `CSR_MEPC` arm of the write case can no longer assert `mepc_en` in CHERIoT mode. The NMI restore path is therefore the *only* remaining way into this branch, which both sharpens the bug and makes the `// legacy cssrw; NMI recover` comment stale. Worth updating the comment either way.

rtl/ibex_cheriot_ex.sv
```diff
@@ -0,0 +1,1010 @@
...
+
+          csr_access_o  = ~instr_fault;
+          csr_op_o      = CHERIOT_CSR_RW;
+          csr_op_en_raw = ~instr_fault && is_write && ~is_ztop;
+          csr_addr_o    = cheriot_cs2_dec_i;
+
+          if (cheriot_cs2_dec_i == CHERIOT_SCR_MTCC) begin
+            // MTVEC/MTCC legalization (clear tag if checking fails)
+            csr_wdata_o = {rf_rdata_a[31:2], 2'b00};
+            trcap = rf_rcap_a;
```
andreaskurth: My earlier request to drop the `SETADDR_SCR` path was too broad, sorry! It also removed the `cap_cor` recompute, and I think that part wasn't redundant.

`trcap = rf_rcap_a` carries the correction factors computed for the *original* address, but `csr_wdata_o` stores the legalized one. On the path where legalization actually changes the address the tag is cleared, but `cap_cor` is still architecturally observable: read MTCC back with `CSpecialRW` and apply `CGetBase`/`CGetTop` to the untagged result. With `E=0`, `B=0x002`, `T=0x008` and an address ending in `0x3`, the original gives `addr_mid=3 -> addr_hi=0 -> cap_cor=2'b00` and the legalized address gives `addr_mid=0 -> addr_hi=1 -> cap_cor=2'b11`, so the decoded base differs by 512. The old comment ("so that RTL behavior would match sail") was pointing at exactly this case.

Would you mind either recomputing `cap_cor` for the legalized address here, or checking whether the Sail co-simulation compares the metadata of untagged capabilities at all? If it doesn't, this is harmless and worth a short comment saying so. Same applies to the MEPCC branch below.

andreaskurth (commented):
Thanks for turning these around so quickly, @SamuelRiedel; nice update!

I re-reviewed only the delta. Three things left: the register file is still only half-done, the NMI/MEPCC item is still open, and one of my earlier requests turned out to be too broad and has introduced a small divergence (details inline).

rtl/ibex_cheriot_ex.sv
SamuelRiedel: You're right. The correction will be off and won't match Sail because Sail doesn't store the correction factor anywhere and always recomputes it. The capability will be invalid anyway, so there won't be a mismatch on this instruction, but as you mentioned, there will be one if we read and recompute the bounds back later. Here is the reference [sail code](https://github.com/CHERIoT-Platform/cheriot-sail/blob/main/src/cheri_insts.sail#L405).

Sail never compares anything on its own, but our co-sim does, so we have some control over this. It doesn't compare the corrections (since they don't exist in Sail), so the write to MTCC is fine for now and the comparison will only mismatch once we read back the bounds like you described. We can easily revert the change I made earlier and address how we want to handle this in the Sail co-sim later.

rtl/ibex_register_file_ff.sv
SamuelRiedel: The register files are all updated to share the registers as defined in the RFC. Please have a look. I ran the more extensive regressions locally with the FF register file and it passed. I also ran the CHERIoT smoketests locally and extended them to write/read each register with all three configurations.

rtl/ibex_cs_registers.sv
SamuelRiedel: I took the time to look into this in detail now, and your are right. I went ahead and fixed it by adding a capability register (mstack_epc_cap_q) to properly preserve the capability alongside mstack_epc_q, and updated the stale comment.

I verified this with a test that triggers an NMI and asserts that the MEPCC tag survives the round-trip instead of dropping to NULL_CAP.

rtl/ibex_top.sv
SamuelRiedel: I double-checked this again. The CHERIoT subsystem should already be aligned because `data_tag_o` is already set by the LSU for both capability loads and stores. The rvfi signal here is only used for the cosim and the testrig framework, which currentli still need this. So the code should already hint the subsystem correctly, right?

The only thing remaining is switching to mubi once we have the lockstep domain.

rtl/ibex_register_file_ff.sv
```diff
@@ -4,17 +4,56 @@
...
+ *    as `rf_shared` storage to hold CHERIoT capability metadata. Therefore, `CapWidth` must be
+ *    >= `DataWidth` so that in non-CHERIoT mode, standard data writes to upper registers can be
+ *    zero-extended into `rf_shared`.
+ *
+ * These configuration options result in the following physical register allocations:
+ * +---------+-------+-------+--------------------+--------------------+-----------------------+
+ * | CHERIoT | Dummy | RV32E | rf_data Flops      | rf_shared Flops    | Total Storage (Flops) |
+ * +---------+-------+-------+--------------------+--------------------+-----------------------+
+ * |    1    |   1   |   0   | 16 x DataWidth     | 16 x CapWidth      | 16 x Data + 16 x Cap  |
+ * |    1    |   1   |   1   | 16 x DataWidth     | 16 x CapWidth      | 16 x Data + 16 x Cap  |
```
andreaskurth, line 37: RV32E is *don't care* if CHERIoT & Dummy, right?

On another note, that table gives a nice overview :+1: Suggest putting it into the docs and referencing it from here and the other reg file implementations.

Totally fine as follow-up, since this PR doesn't extend the doc yet.

rtl/ibex_lockstep.sv
```diff
@@ -580,68 +635,89 @@ module ibex_lockstep import ibex_pkg::*; #(
...
 
       .raddr_a_i        (shadow_rf_raddr_a),
       .rdata_a_o        (shadow_rf_rdata_a_intg),
+      .rcap_a_o         (shadow_rf_rcap_a_ecc),
       .raddr_b_i        (shadow_rf_raddr_b),
       .rdata_b_o        (shadow_rf_rdata_b_intg),
+      .rcap_b_o         (shadow_rf_rcap_b_ecc),
       .waddr_a_i        (shadow_rf_waddr_wb),
       .wdata_a_i        (shadow_rf_wdata_wb_ecc[RegFileDataEccWidth-1:RegFileDataWidth]),
+      .wcap_a_i         (shadow_rf_wcap_ecc_wb),
```
nasahlpa: I think this is wrong - it should be similar to FF register file (line 662).

rtl/ibex_lockstep.sv
```diff
@@ -580,68 +635,89 @@ module ibex_lockstep import ibex_pkg::*; #(
...
```
nasahlpa: I think this is wrong - it should be similar to FF register file (line 662).

rtl/ibex_top.sv
```diff
@@ -73,6 +73,8 @@ module ibex_top import ibex_pkg::*; #(
   input  prim_ram_1p_pkg::ram_1p_cfg_req_t [ibex_pkg::IC_NUM_WAYS-1:0] ram_cfg_icache_data_i,
   output prim_ram_1p_pkg::ram_1p_cfg_rsp_t [ibex_pkg::IC_NUM_WAYS-1:0] ram_cfg_icache_data_o,
 
+  input  ibex_mubi_t                                                   cheriot_enable_i,
```
nasahlpa, line 76: Good to have that MuBi encoded. Would it make sense to add a MuBi encoding checker at some place and connect it with an alert? Or are you planning to do this in OpenTitan where you instantiate Ibex?

rtl/ibex_cs_registers.sv
SamuelRiedel: This was fixed in the latest commit

rtl/ibex_cheriot_ex.sv
SamuelRiedel: This was reverted after a followup discussion

rtl/ibex_cs_registers.sv
```diff
@@ -216,31 +249,41 @@ module ibex_cs_registers import ibex_pkg::*; #(
...
+  logic         mscratch_en;
+  logic [31:0]  mepc_q, mepc_d;
+  logic         mepc_en;
+  exc_cause_t   mcause_q, mcause_d;
+  cap_t         mepc_cap;
+  logic         mcause_en;
+  logic [31:0]  mtval_q, mtval_d;
+  logic         mtval_en;
+  logic [31:0]  mtvec_q, mtvec_d;
+  cap_t         mtvec_cap;
```
nasahlpa, line 268: Some existing registers in this module (e.g., mstatus) are protected via shadow registers. Can you think of any new registers that would benefit from this protection as well?

rtl/ibex_register_file_fpga.sv
```diff
@@ -25,59 +48,162 @@ module ibex_register_file_fpga #(
...
+    // Data: raddr[4]=0 → rf_data (x0-x15)
+    //       raddr[4]=1 → rf_shared lower DataWidth bits (x16-x31), non-CHERIoT only.
+    // The bank-select is AND-gated with !cheriot_enabled so a spurious raddr[4]=1 in CHERIoT
+    // mode reads from rf_data rather than returning capability bits as integer data.
+    assign rdata_a_o = (raddr_a_i[4] && !cheriot_enabled) ?
+      DataWidth'(rf_shared[raddr_a_i[3:0]]) : rf_data[raddr_a_i[3:0]];
+    assign rdata_b_o = (raddr_b_i[4] && !cheriot_enabled) ?
+      DataWidth'(rf_shared[raddr_b_i[3:0]]) : rf_data[raddr_b_i[3:0]];
+
+    // Cap: gated to CapWordZeroVal in non-CHERIoT mode; x0 always returns CapWordZeroVal.
```
nasahlpa, line 141: Is it guaranteed that `CapWordZeroVal` is a valid ECC code word? Maybe it would make sense to add an assertion like: `WordZeroVal == CapWordZeroVal`

nasahlpa (commented):
Thanks Sam for your work!

I looked into the code from a fault injection perspective. I can confirm that in non-CHERI mode the lockstep comparison is identical to before. In CHERI mode, all relevant new signals are compared.

I also can confirm that the register file split (main and shadow core RF instances) look secure in the RF variant we use.

Regarding functionality - have you tested if switching from non-CHERI mode to CHERI mode works with the register file? Or do we need to clear the register file before doing the switch?

rtl/ibex_decoder.sv
```diff
@@ -169,24 +195,65 @@ module ibex_decoder #(
...
+                                (instr_rs3[4] & use_rs3_d & rf_ren_a_o) |
+                                (rf_waddr_o[4]   & rf_we_or_load));
   end else begin : gen_rv32e_reg_check_inactive
     assign illegal_reg_rv32e = 1'b0;
   end
 
+  if (CheriLimit16Regs) begin : gen_cheriot_reg_check_active
+    // Trap only on CHERIoT instructions that use an out-of-range capability register (>= c16).
+    // Standard RV32I instructions may freely use x0-x31.
+    assign illegal_reg_cheriot = (cheriot_enable_i == IbexMuBiOn) & instr_is_cheriot_o &
```
andreaskurth: Now that `516ec4e3` shares the storage for x16-x31 with the capability metadata, I think this check has to become unconditional in CHERIoT mode. Spec 7.3 says CHERIoT has 16 registers ("The 16 32-bit integer registers from RV32 are extended into 65-bit capabilities [...] `$x0..$x15` referring to their address parts"), but this only traps when `instr_is_cheriot_o` is set, and the address masking at lines 212-218 is gated the same way, so a plain RV32I instruction executing in CHERIoT mode still reaches the register file with `raddr[4]`/`waddr[4]` set.

That used to degrade gracefully because 32 real data registers existed. With the shared file it doesn't: `add x20, x21, x22` in CHERIoT mode reads `rf_data[5]`/`rf_data[6]` (i.e. x5/x6), because the upper-bank select is AND-gated with `!cheriot_enabled`, and the write to x20 is dropped entirely, because `rf_data[i]` requires `!waddr_a_i[4]` and `rf_shared[i]` requires `!cheriot_enabled`. The new register files' own `CheriotWaddrMSBClear` / `CheriotRaddrAMSBClear` / `CheriotRaddrBMSBClear` assertions fire on exactly this case, so the RTL contradicts itself right now.

Would dropping the `instr_is_cheriot_o` term here (and the corresponding gating on the masking above) work, so that CHERIoT mode simply implies RV32E register addressing? The two comments would need updating as well. I'd like this one resolved in this PR, since the sharing commit turns it into a silent-corruption path rather than a conformance gap.

rtl/ibex_register_file_ff.sv
```diff
@@ -24,85 +63,264 @@ module ibex_register_file_ff #(
...
+  input  logic [CapWidth-1:0]  wcap_a_i,
   input  logic                 we_a_i
 );
 
-  localparam int unsigned ADDR_WIDTH = RV32E ? 4 : 5;
-  localparam int unsigned NUM_WORDS  = 2**ADDR_WIDTH;
+  if (BaseIsa == BaseIsaRV32IorCHERIoT) begin : g_cheriot_rf
+
+    // CapWidth must be larger than DataWidth to ensure no truncation in the non-CHERIoT mode
+    `ASSERT_INIT(CapWidthGTEDataWidth, CapWidth >= DataWidth)
```
andreaskurth, line 91: This file, `ibex_register_file_fpga.sv:73`, and `ibex_register_file_latch.sv:73` now use `` `ASSERT_INIT ``/`` `ASSERT ``, but none of the three has `` `include "prim_assert.sv" ``, unlike every other Ibex file that uses those macros. All three are the first entries in `ibex_top.core`'s fileset, so they currently compile only because a dependency fileset has already defined the macros in the same compilation unit.

rtl/ibex_cheriot_ex.sv
```diff
@@ -0,0 +1,1028 @@
...
+          instr_fault   = perm_vio | illegal_scr_addr;
+
+          csr_access_o  = ~instr_fault;
+          csr_op_o      = CHERIOT_CSR_RW;
+          csr_op_en_raw = ~instr_fault && is_write && ~is_ztop;
+          csr_addr_o    = cheriot_cs2_dec_i;
+
+          if (cheriot_cs2_dec_i == CHERIOT_SCR_MTCC) begin
+            // MTVEC/MTCC legalization (clear tag if checking fails)
+            // note we don't reall need cheriot_set_address checks here - it's only used to update
```
andreaskurth: Now that this path is back, could the comment be rewritten to say why it is needed? As written ("we don't reall need cheriot_set_address checks here") it reads like dead code, which is what led me to ask for its removal last round.

Something like: the `cheriot_set_address` call is required to recompute `cap_cor` for the legalized address, so the resulting (untagged) capability matches Sail in the case where legalization actually changes the address. Also `reall` -> `really`.

rtl/ibex_lockstep.sv
```diff
@@ -41,6 +42,7 @@ module ibex_lockstep import ibex_pkg::*; #(
   parameter bit                     RegFileECC                  = 1'b0,
   parameter int unsigned            RegFileDataWidth            = 32,
   parameter int unsigned            RegFileDataEccWidth         = 39,
+  parameter int unsigned            RegFileCapEccWidth          = REGCAP_W + 7,
```
andreaskurth, line 45: `ibex_top` passes `RegFileCapEccWidth` explicitly to the main core (`ibex_top.sv:399`) but relies on this default for the lockstep instance, whereas the analogous `RegFileDataEccWidth` is passed explicitly at `ibex_top.sv:1146`. The value works out (42 is right for the shadow, since it stores the 7 ECC bits on top of the 35 cap bits) but the asymmetry between the two is easy to trip over. Pass it explicitly from `ibex_top` to match the data-width parameter?

andreaskurth (commented):
Thanks @SamuelRiedel! Everything from the last round is resolved, nice :+1:

A few more changes: the register-file sharing has surfaced a disagreement with the decoder that I think has to be fixed alongside it, plus three smaller items. Details inline.

rtl/ibex_lockstep.sv
SamuelRiedel: Thank you, good catch. I fixed this.

rtl/ibex_top.sv
SamuelRiedel: I think it makes sense to do this in Ibex. Ideally in the duplicated core, right? I will add this.

rtl/ibex_cs_registers.sv
SamuelRiedel: Good point. I think MTCC (MTVEC's capability) should get the same since MTVEC is protected. Otherwise, none of the protected CSRs get a capability part. And of the CHERIoT-only SCRs, I'm not sure if the root capabilities (MTDC, MScratchC) could be protected as well. They are anyway all duplicated in the shadow core, so any fault should show up in the lockstep comparison before they can leave the core.

Are you fine with creating an issue for this, and when we optimize the CSRs, and make them use the `ibex_csr` module as well, we directly address adding shadow copies as well?

rtl/ibex_register_file_fpga.sv
SamuelRiedel: I think neither is guaranteed to be a valid ECC to the parameter in the main register file. We would need an assertion that checks whether ECC(ibex_top.regfile.WordZeroVal) == ibex_top.lockstep.regfile.WordZeroVal, right? 

But the word and cap don't need to match. While neither ISA defines register values at reset, we initialize them mainly to establish the initial ECC. And now we set the lower half to "word" values, upper half with "capability" values mainly because the width is different. Both are currently zero. After reset, any writes to the register files will keep their values and ECC permanently synchronized, regardless of how the registers are used.

rtl/ibex_register_file_ff.sv
SamuelRiedel: > RV32E is don't care if CHERIoT & Dummy, right?

Concerning the number of registers instantiated, yes. But the logic is a bit different, because in the RV32E mode, we don't share the capability registers and block access to the shared registers in the non-CHERIoT mode. But if CHERIoT and Dummy, enabling RV32E makes no sense, because it doesn't save HW.

Yes, good point. I will move it to the doc in the doc PR.

rtl/ibex_decoder.sv
SamuelRiedel: Yes, you're right. I fixed this with the changes you suggested. In the future, we could clean this up a bit more neatly.

rtl/ibex_register_file_ff.sv
SamuelRiedel: Fixed

rtl/ibex_cheriot_ex.sv
SamuelRiedel: Done

rtl/ibex_lockstep.sv
SamuelRiedel: Done.

rtl/ibex_cs_registers.sv
nasahlpa: I agree with the MTCC register. Creating an issue sounds good to me, thanks!

rtl/ibex_register_file_fpga.sv
nasahlpa: > I think neither is guaranteed to be a valid ECC to the parameter in the main register file. We would need an assertion that checks whether ECC(ibex_top.regfile.WordZeroVal) == ibex_top.lockstep.regfile.WordZeroVal, right?

Ah yes, right.
> After reset, any writes to the register files will keep their values and ECC permanently synchronized, regardless of how the registers are used.

Agreed, but what happens if we are switching from non- CHERI mode to CHERI mode? If we are switching the mode, aren't the values computed in the old mode checked with the new mode code?

rtl/ibex_top.sv
SamuelRiedel: I added this: https://github.com/lowRISC/ibex/pull/2475/changes/0cc88fb6d4001cfe342226d32d74413ade099a44

I ran into an issue that the check needs to be gated somehow, otherwise we have combinatorial alerts at reset. Especially the shadow core triggers alerts initially because we reset the shadow inputs to zero, so during reset, the mubi is invalid. Do you think we should directly reset the mubi there to Off?

Instead of using the reset as a gating signal now, I gated with the fetch_enable, which should indicate whether we're actually executing instructions. @nasahlpa do you think that is a good solution?

rtl/ibex_register_file_fpga.sv
SamuelRiedel: If we switch mode, currently we don't change any registers. So all of them will have valid value+ECC pairs. And we always compare the full width of the register, even if the upper bits are never accessed in the non-CHERIoT mode. So for the ECC checker, nothing changes depending on mode. And a switch should be treated like a fresh boot, so the registers will have to be initialized by sw.

rtl/ibex_register_file_ff.sv
andreaskurth: Tracked in https://github.com/lowRISC/ibex/issues/2479

rtl/ibex_cs_registers.sv
andreaskurth: Tracked in https://github.com/lowRISC/ibex/issues/2478

rtl/ibex_compressed_decoder.sv
andreaskurth: Tracked in https://github.com/lowRISC/ibex/issues/2477.

andreaskurth (approved):
LGTM, many thanks @SamuelRiedel for all this work!

SamuelRiedel:
Thanks for all the quick and helpful feedback! I've just squashed the fixes into their respective commits as a final cleanup.

rtl/ibex_register_file_fpga.sv
nasahlpa: I was thinking of this output of the register file:
https://github.com/SamuelRiedel/ibex/blob/a4c00bf4065746b166794621a4a1f68daca1c804/rtl/ibex_register_file_ff.sv#L227-L228
When we are switching from non-cheri mode to cheri mode we could have old data in rf_shared. And then,  `rcap_a_o` gets checked by:
https://github.com/SamuelRiedel/ibex/blob/a4c00bf4065746b166794621a4a1f68daca1c804/rtl/ibex_core.sv#L1263
here we have a different decoder than for the plain data:
`rdata_o` gets checked by:
https://github.com/SamuelRiedel/ibex/blob/a4c00bf4065746b166794621a4a1f68daca1c804/rtl/ibex_core.sv#L1227
So SW would basically need to be initialized by SW as you mentioned.

SamuelRiedel: You're right. The fact that the ZeroECC was the same independent of the decoder led me to believe that zero extending a 32-bit word on the wider ECC path would result in the same ECC value because the zeros don't add to the sum. But for this to work, we would have to use the same generator matrix, right? Would this be possible?

We also have to be careful about x16 because it's a real register in the rv32i mode, but always reads zero as the x0 capability part in CHERIoT. Currently, this should be fine, because in CHERIoT, we hardcode the output to ZeroCap and ZeroECC.

I see a few options here:

- Force the software to zero out the registerfile before the switch. The ZeroECC will match and the x16 corner case is covered.
- Add hardware to reset the registers when the switch happens. At the switch, we have to flush the pipeline anyway and could also add some logic to hardware enforce that the regsiters are reset to the values we want for the capabilities.
- Always route out the full registerwidth into the ECC. So even in non-cheriot mode, we could have the upper bits routed to the ECC path, even though they will always be zero in this mode. Then we always use the narrow ECC path on the lower registers and the wider ECC path for the upper ones and nothing changes from the ECC's perspective between switches. But we have to handle the width truncation outside of the registerfile, which is not very clean.

What do you prefer? I will also open a dedicated issue for this discussion.

SamuelRiedel: Tracked here: https://github.com/lowRISC/ibex/issues/2480

PR #2481, opened 2026-08-27
dv: enable read-only CSRs in csr_description.yaml

Talha-Dmr:
Fixes first item of #1337

CSR test generation now correctly handles read-only fields via riscv_csr_instr allow_ro_write handling (csr[11:10]==2'b11 check). Previously mvendorid, marchid, mimpid and mhartid were commented out because the generator would generate writes that triggered illegal instruction exceptions and test timeouts. Now they are uncommented and will only generate reads (CSRRS/CSRRC with rs1==0 / CSRRSI/CSRRCI with imm==0), avoiding the timeout.

Test: verified csr_description.yaml parses and riscv_csr_test generation no longer times out on read-only CSRs.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
+++ b/dv/uvm/core_ibex/riscv_dv_extension/csr_description.yaml
@@ -63,80 +63,81 @@
 #      msb: 25
 #      lsb: 0
 
-# CSR test generation cannot deal with read-only fields. Leaving these here
-# commented out so they can be used once the read-only issue is fixed.
-# https://github.com/lowRISC/ibex/issues/1337 tracks the required improvements
-#- csr: mvendorid
-#  description: >
-#    Machine Vendor ID Register providing JEDEC manufacturer ID
-#  address: 0xF11
-#  privilege_mode: M
-#  rv32:
-#    - field_name: Bank
-#      description: >
-#        Number of continuation codes in JEDEC ID
-#      type: R
-#      reset_val: 0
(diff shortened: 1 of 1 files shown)


Talha-Dmr:

elliotb-lowrisc:
@Talha-Dmr, thanks for taking an interest in the Ibex DV.

What makes you say that read-only fields are correctly handled now? What changed?

Talha-Dmr:
Hi @elliotb-lowrisc,

Thanks for pointing this out. After tracing the test execution flow in `dv/uvm/core_ibex/scripts/run_instr_gen.py` (lines 82–86), I realized my assumption was incorrect: `riscv_csr_test` does not run the SystemVerilog `riscv_csr_instr.sv` generator (which has `allow_ro_write` constraints). Instead, it runs `vendor/google_riscv-dv/scripts/gen_csr_test.py`.

In `gen_csr_test.py`, the generator loops through all `csr_ops` (`csrrw`, `csrrs`, `csrrc`, `csrrwi`, etc.) for every CSR in `csr_description.yaml` without filtering write operations for read-only registers. Consequently, simply uncommenting these CSRs in YAML will still cause `csrrw`/`csrrwi` writes to be emitted, raising illegal instruction exceptions on Ibex hardware and causing timeouts without an appropriate handler.

Since this needs to be addressed either in the generator script or with an explicit test trap handler rather than just uncommenting the YAML entries, I am closing this PR to avoid any confusion. Apologies for the noise!

PR #2482, opened 2026-08-27
dv: handle illegal CSR access via test_fail

Talha-Dmr:
Fixes second item of #1337

When riscv_csr_test generates an illegal CSR access that was not intended (enable_illegal_csr_instruction==0), the previous illegal instruction handler just skipped the instruction (mepc+4; mret) which could cause a test timeout if the test expected the CSR write to succeed. Now it jumps to test_fail instead, so unexpected illegal accesses are reported as a test failure rather than a timeout. When enable_illegal_csr_instruction==1 the original skip behavior is kept for intentional illegal CSR tests.

Depends on #2481 (first item of #1337) which enables read-only CSRs. This PR should be merged after #2481.

### DIFF
--- a/dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
+++ b/dv/uvm/core_ibex/riscv_dv_extension/ibex_asm_program_gen.sv
@@ -79,6 +79,17 @@ class ibex_asm_program_gen extends riscv_asm_program_gen;
     gen_section(get_label("ecall_handler", hart), instr);
   endfunction
 
+  // Illegal instruction handler - for #1337 second item, keep default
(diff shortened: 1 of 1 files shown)


elliotb-lowrisc:
Hi @Talha-Dmr, how does this depend on https://github.com/lowRISC/ibex/pull/2481?

Talha-Dmr:
Hi @elliotb-lowrisc,

You are completely right — there is no dependency on #2481. This PR was an incomplete draft that only delegated to `super.gen_illegal_instr_handler`, and furthermore `gen_csr_test.py` generates standalone assembly rather than utilizing this class.

I am closing this PR as well to keep the tracker clean. Thank you for the review and guidance!

PR #2483, opened 2026-08-28
[rtl] Add `ASSERT_KNOWN` to all ibex_top outputs

SamuelRiedel:
Add assertions to check all outputs are known at any time. The only exception is `crash_dump_o`, which is by definition not defined when `ResetAll` is not set. That is, when not all registers are resettable. This is only the case for non-secure implementations of Ibex.


gautschimi (approved):
Looks good to me. thanks!

rtl/ibex_top.sv
```diff
@@ -1407,6 +1417,11 @@ module ibex_top import ibex_pkg::*; import ibex_cheriot_pkg::*; #(
   `ASSERT_KNOWN_IF(IbexTrvkRevbmRPayloadX,
     {trvk_revbm_rdata_i, trvk_revbm_rdata_intg_i, trvk_revbm_err_i}, trvk_revbm_rvalid_i)
 
+  // Crash dump depends on FFs that are only defined at reset if the ResetAll flag is set
+  if (ResetAll) begin : g_reset_all_ibex_assert_known
+    `ASSERT_KNOWN(IbexCrashDumpX, crash_dump_o)
```
gautschimi: You could only disable the assertions for the fields that can be X (`current_pc`, `next_pc`). But I think it's fine.

nasahlpa (approved):

rtl/ibex_top.sv
SamuelRiedel: I added this quickly. Thanks :)

gautschimi (approved):

PR #2484, opened 2026-08-28
[doc] Update documentation to align with recent RTL changes

SamuelRiedel:
Some new changes to Ibex were not reflected in the documentation. This PR updates it:

- add new `ibex_top` parameters
- add cross-refs throughout the parameter list to explain what they do without duplication
- add missing ports and their description
- udpate instantiation snippet updated to match current RTL
- add missing CSRs
- add pipeline details for pipeline flush on CSR writes
- expand explanation on Zcmp behavior concerning interrupts

The CHERIoT documentation is deferred to a separate PR


doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheScrNumPrinceRoundsHalf`` | int (1..5)          | 2                       | Number of PRINCE half-rounds used for ICache address scrambling                              |
+|                                  |                     |                         | (only relevant when ICacheScramble == 1)                                                     |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheTweakInfection``         | bit                 | SecureIbex              | Enable :ref:`icache-tweak-infection`                                                         |
+|                                  |                     |                         | (only relevant when ICacheScramble == 1)                                                     |
```
nasahlpa: Is that correct? As far as I remember, it is only useful when having SecureIbex enabled (because then ECC errors are flagged by a wrong un-XOR operation) and when we  have an icache available.

nasahlpa (commented):
LGTM apart from one question.

doc/02_user/integration.rst
SamuelRiedel: You're right. That was a copy-paste error.

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``PMPGranularity``               | int (0..31)         | 0                       | Minimum granularity of PMP address matching                                                  |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``PMPNumRegions``                | int (1..16)         | 4                       | Number implemented PMP regions (ignored if PMPEnable == 0)                                   |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``MHPMCounterNum``               | int (0..10)         | 0                       | Number of :ref:`performance-counters`                                                        |
```
andreaskurth, line 261: Following this new `:ref:` lands on `performance_counters.rst:85-90`, which uses `WidthMHPMCounters`/`NumMHPMCounters` (RTL names them `MHPMCounterWidth`/`MHPMCounterNum`) and caps the useful count at 8 (`mhpmcounter3` - `mhpmcounter10`). The RTL defines events up to `mhpmcounter12` (`ibex_cs_registers.sv:1583-1597`), i.e. 10, which matches this row and that page's own event table. Worth fixing there too while we're at it?

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICache``                       | bit                 | 0                       | Enable :ref:`icache` instead of prefetch buffer                                              |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheECC``                    | bit                 | 0                       | Enable :ref:`icache-ecc` (if ICache == 1)                                                    |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``BranchPredictor``              | bit                 | 0                       | *EXPERIMENTAL* Enable :ref:`branch-prediction`                                               |
```
andreaskurth, line 305: The `:ref:` target still calls the parameter `BranchPrediction` (`instruction_fetch.rst:34`). Rename it there too so the cross-reference doesn't contradict the name in this row?

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
+| ``DbgHwBreakNum``                | int (>= 1)          | 1                       | Number of hardware breakpoints (debug triggers) supported                                    |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``SecureIbex``                   | bit                 | 0                       | Enable various :ref:`security`. Note: SecureIbex == 1'b1 and                                 |
+|                                  |                     |                         | RV32M == ibex_pkg::RV32MNone is an illegal combination.                                      |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``LockstepOffset``               | int (>= 1)          | 1                       | Number of cycles to delay the :ref:`lockstep` shadow core                                    |
```
andreaskurth, line 314: Add "(only relevant when SecureIbex == 1)" here, like the neighbouring rows carry? `Lockstep = SecureIbex`, so the offset has no effect otherwise.

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``MemECC``                       | bit                 | SecureIbex              | Enable SECDED ECC for :ref:`bus-integrity`                                                   |
```
andreaskurth, line 316: Following this new `:ref:` lands on `security.rst:93`, which still says "When the SecureIbex parameter is set, incoming data will be checked ..." (same in `load_store_unit.rst:68`). Now that `MemECC` is documented as an independent parameter, should those two name `MemECC` instead?

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``MemDataWidth``                 | int (>= 32)         | MemECC ? 32 + 7 : 32    | Data width of memory interface including integrity bits                                      |
```
andreaskurth, line 318: `int (>= 32)` reads as if this were freely choosable, but it is derived from `MemECC` and effectively fixed at 32 or 39: `ibex_top.sv:864` slices `data_wdata_core[MemDataWidth-1:32]` into a `prim_buf #(.Width(7))`, so any other value breaks elaboration. Say "derived" or "32 or 39" in the range column?

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheScramble``               | bit                 | 0                       | Replace ICache tag/data RAMs with                                                            |
+|                                  |                     |                         | :ref:`scrambling RAM <icache-scramble-key>` primitives                                       |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheScrNumPrinceRoundsHalf`` | int (1..5)          | 2                       | Number of PRINCE half-rounds used for ICache address scrambling                              |
```
andreaskurth, line 323: Should this be *data* scrambling? `ICacheScrNumPrinceRoundsHalf` feeds `NumPrinceRoundsHalf` of `prim_ram_1p_scr`, which sizes the PRINCE cipher generating the data keystream. Address scrambling is governed by `NumAddrScrRounds`, hardcoded to 2 in `ibex_top.sv:264`.

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
+|                                  |                     |                         | (only relevant when ICacheScramble == 1)                                                     |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``ICacheTweakInfection``         | bit                 | SecureIbex              | Enable :ref:`icache-tweak-infection`                                                         |
+|                                  |                     |                         | (only relevant when ICache == 1 and SecureIbex == 1)                                         |
```
andreaskurth, line 327: Following up on @nasahlpa's thread: isn't `ICacheECC == 1` the actual requirement rather than `SecureIbex == 1`? `TweakInfection` is passed straight to `ibex_icache` gated only on `ICache` (`ibex_if_stage.sv:298-306`), and a wrong un-XOR is only detectable if the ECC is checked. `SecureIbex` is just the default for both.

doc/02_user/integration.rst
```diff
@@ -89,175 +86,272 @@ Instantiation Template
...
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``DmHaltAddr``                   | int                 | 0x1A110800              | Address to jump to when entering Debug Mode                                                  |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``DmExceptionAddr``              | int                 | 0x1A110808              | Address to jump to when an exception occurs while in Debug Mode                              |
++----------------------------------+---------------------+-------------------------+----------------------------------------------------------------------------------------------+
+| ``RndCnstIbexKey``               | logic[]             | RndCnstIbexKeyDefault   | Scrambling key for ICache data (only relevant when ICacheScramble                            |
```
andreaskurth, line 343: These are the *reset* values of the scrambling key/nonce, used until `scramble_key_valid_i` delivers a real one (`ibex_top.sv:634`), and `scramble_key_q` feeds the tag RAMs as well as the data RAMs. "Reset value of the ICache scrambling key/nonce" would be more precise for both rows?

doc/02_user/integration.rst
```diff
@@ -274,84 +368,112 @@ This is well supported by most tools but some care is needed when overriding the
...
+| ``scramble_*``               | Scrambling key interface, see :ref:`icache`                            |
++------------------------------+------------------------------------------------------------------------+
+| ``debug_*``                  | Debug interface, see :ref:`debug-support`                              |
++------------------------------+------------------------------------------------------------------------+
+| ``crash_dump_o``             | A set of signals that can be captured on reset to aid crash debugging. |
++------------------------------+------------------------------------------------------------------------+
+| ``double_fault_seen_o``      | A double fault was observed, see :ref:`double-fault-detect`            |
++------------------------------+-------------------------+-----+----------------------------------------+
+| ``fetch_enable_i``           | ibex_mubi_t             | in  | Allow the core to fetch instructions.  |
+|                              |                         |     | If set to IbexMuBiOff, the core will   |
```
andreaskurth, line 418: With `SecureIbex`, *any* value other than `IbexMuBiOn` pauses fetching, not only `IbexMuBiOff`; without it only bit 0 is examined (`ibex_core.sv:648-656`). Phrase it as "If not set to `IbexMuBiOn`"?

doc/03_reference/cs_registers.rst
```diff
@@ -189,6 +212,35 @@ Reset Value: ``0x0000_0001``
...
+|        |      | Bits above ``MHPMCounterNum + 2`` always read as zero.                              |
++--------+------+-------------------------------------------------------------------------------------+
+| 2      | RW   | **IR:** Enable U-mode access to ``instret`` / ``instreth``.                         |
++--------+------+-------------------------------------------------------------------------------------+
+| 1      | R    | **TM:** Always reads as zero. The ``time`` CSR is not implemented.                  |
++--------+------+-------------------------------------------------------------------------------------+
+| 0      | RW   | **CY:** Enable U-mode access to ``cycle`` / ``cycleh``.                             |
++--------+------+-------------------------------------------------------------------------------------+
+
+Writes to ``mcounteren`` are only accepted when the ``mcounteren_writable_i`` input is set to ``IbexMuBiOn``.
```
andreaskurth, line 239: `performance_counters.rst:77` describes the same mechanism but calls the signal `mcounteren_writeable`. Fix the name there and cross-reference here, so the behaviour is only spelled out once?

doc/03_reference/pipeline_details.rst
```diff
@@ -43,7 +43,11 @@ Read the description for more information.
 |                       |                                      | RISCV-V RV32I Base Integer Instruction Set.                 |
 +-----------------------+--------------------------------------+-------------------------------------------------------------+
 | CSR Access            | 0                                    | CSR Access Instruction are defined in 'Zicsr' of the        |
-|                       |                                      | RISC-V specification.                                       |
+|                       |                                      | RISC-V specification. Note that any CSR *write* flushes the |
```
andreaskurth, line 46: "any CSR *write*" is slightly too strong: `mscratch` and `mepc` are exempt via `no_flush_csr_addr` (`ibex_id_stage.sv:593`), and the RTL comment says "on most CSR modification". Mention the two exceptions?

doc/03_reference/pipeline_details.rst
```diff
@@ -103,3 +107,16 @@ Read the description for more information.
 |                       |                                      | 'Zcmp' of the RISC-V specification. Internally, they are    |
 |                       |                                      | implemented as two `addi rd, rs1, 0` instructions.          |
 +-----------------------+--------------------------------------+-------------------------------------------------------------+
+
+.. note::
+
+   **Zcmp interrupt and exception behaviour**
```
andreaskurth, line 113: Could the note also cover debug-mode entry? It is blocked over a wider window than interrupts: `INSTR_EXPANDED` *and* `INSTR_EXPANDED_COMMIT` (`ibex_controller.sv:474-477`), versus `INSTR_EXPANDED_COMMIT` only for `handle_irq` (`ibex_controller.sv:499`).

doc/03_reference/pipeline_details.rst
```diff
@@ -103,3 +107,16 @@ Read the description for more information.
...
+
+   * **Interrupt masking:** Interrupts are suppressed for the duration of a Zcmp push/pop commit sequence.
```
andreaskurth, line 115: `cm.push` never reaches `INSTR_EXPANDED_COMMIT` (it goes `INSTR_EXPANDED` -> `INSTR_EXPANDED_LAST`), so interrupts are not suppressed during it. The sentence two lines down already says as much, so dropping "push/" here would remove the contradiction.

doc/03_reference/pipeline_details.rst
```diff
@@ -103,3 +107,16 @@ Read the description for more information.
...
+     The sequence therefore appears atomic to the interrupt controller.
+     That is, from the stack update to the return jump of the ``cm.pop*`` instructions and the entire ``cm.mv*`` instructions.
+
+   * **minstret counting:** An entire Zcmp push/pop sequence (regardless of how many register load/store micro-operations it expands to) is counted as a single retired instruction in ``minstret``.
+
+   * **Exception/interrupt during expansion:** If an exception or interrupt is signalled before the commit sequence has started, the Zcmp state machine is flushed and the partial sequence is abandoned.
+     The core takes the trap as if the Zcmp instruction had not yet started, so software can safely restart it on return from the handler.
```
andreaskurth, line 122: "as if the Zcmp instruction had not yet started" seems a bit strong: a partial `cm.push` has already written stack memory (negative offsets from `sp`, see `cm_push_store_reg`) and a partial `cm.pop` has already overwritten registers. The reason restarting is safe is that `sp` is adjusted last, so the micro-ops are idempotent. Say that instead?

andreaskurth (commented):
Thanks @SamuelRiedel for updating the Ibex documentation with many recent RTL changes (not just CHERIoT)!

This mostly LGTM, but I have see a few points that should be changed: one factual mix-up on `ICacheScrNumPrinceRoundsHalf`, a few over- or under-stated qualifiers, and some stale text sitting at the far end of the new cross-references.

doc/02_user/integration.rst
SamuelRiedel: Yes. Done.

doc/02_user/integration.rst
SamuelRiedel: Yes, this is also fixed.

doc/02_user/integration.rst
SamuelRiedel: fixed

doc/02_user/integration.rst
SamuelRiedel: done

doc/02_user/integration.rst
SamuelRiedel: fixed

doc/02_user/integration.rst
SamuelRiedel: you're right. this is fixed as well

doc/02_user/integration.rst
SamuelRiedel: fixed

doc/02_user/integration.rst
SamuelRiedel: I changed it accordingly

doc/02_user/integration.rst
SamuelRiedel: done

doc/03_reference/cs_registers.rst
SamuelRiedel: done, and I added a reference back to the performance counters

doc/03_reference/pipeline_details.rst
SamuelRiedel: added them

doc/03_reference/pipeline_details.rst
SamuelRiedel: yes, I added it

doc/03_reference/pipeline_details.rst
SamuelRiedel: That's right. I clarified this

doc/03_reference/pipeline_details.rst
SamuelRiedel: Yes, I changed this sentence

SamuelRiedel:
@nasahlpa @andreaskurth, thank you both for your feedback. I addressed all of it and squashed it into the corresponding commit. Let me know if you have more feedback. The formal CI pipeline failure is unrelated to this, since we only change the doc here.

nasahlpa (approved):

PR #2488, opened 2026-09-02
sim: display ISA configuration string at simulation startup

### LINKED ISSUE #1364: Display configuration parameters in Verilator simulation
To help users identify how their Verilator simulation of Ibex was built it would be nice to display the parameters (or name of the configuration) that were chosen when running the simulation -- either by default, or as part of a config parameter (e.g. `--version`).

The parameter values should be available from Verilator at runtime, so we probably don't need any compile-time magic to pass them in.

devtyagi3909:
Fixes #1364.

`GetIsaString()` was already implemented in `SimpleSystem` to return the Spike-compatible ISA string derived from the elaborated Verilator parameters (e.g. `rv32imc`, `rv32imcb`). However it was never called or displayed, so users had no way to confirm which ISA parameters their simulation was actually built with without cross-referencing build flags.

This adds a single line to the startup banner in `SimpleSystem::Run()`:

```
Simulation of Ibex
==================
ISA: rv32imc
```

No new logic introduced — this strictly wires the existing function into the existing output path.


devtyagi3909:

PR #2492, opened 2026-09-04
[simple_system] Add BaseIsa parameter

mingiiiiiiii:
`util/ibex_config.py` emits `--BaseIsa` for every configuration, but
`examples/simple_system/ibex_simple_system.core` does not declare it. The build
command documented in `examples/simple_system/README.md` therefore fails during
argument parsing:

fusesoc --cores-root=. run --target=sim --setup --build
    lowrisc:ibex:ibex_simple_system $(util/ibex_config.py small fusesoc_opts)

fusesoc run lowrisc_ibex_ibex_simple_system_0: error: unrecognized arguments:
--BaseIsa=ibex_pkg::BaseIsaRV32I

`BaseIsa` was added to `ibex_top.core` and `ibex_top_tracing.core` when CHERIoT
support landed, but the simple system example was not updated.

This declares it in the same way, using the same datatype, default and
description, and adds it to the target parameter list.

### Testing

Against `upstream/master` (34b0705):

- Before: the command above fails with the error shown.
- After: setup completes, and the generated `.vc` file contains
  `-DBaseIsa=ibex_pkg::BaseIsaRV32I`.

The default is unchanged, so the elaborated design is identical for existing
builds.


mingiiiiiiii:

PR #2494, opened 2026-09-07
[nix] Fix OSS formal nix env

SamuelRiedel:
Fix the failing OSS Formal flow in CI:

- Remove the `mkshell-minimal` package since it was moved away from GitHub to [Codeberg](https://codeberg.org/viperML/mkshell-minimal). Since we already include another minimal shell, reuse that one for the formal flow as well.
- Update the nixpkgs version to fix the issue where requests from crates.io were blocked: https://github.com/rust-lang/crates.io/issues/13482


hcallahan-lowrisc (approved):
Thanks Sam! Assuming the CI job now passes, LGTM.

Issue #2461, opened 2026-07-16
[DV] Add regression tests for Zcmp instruction traps and atomicity


SamuelRiedel:
In PR https://github.com/lowRISC/ibex/pull/2374, we implemented several RTL fixes for Zcmp expanded instructions. While the RTL changes are merged, they have only been tested manually. We need to implement dedicated regression tests in our DV environment to prevent regressions in the future.

### The following fixes require DV coverage:
* State Machine Reset: The Zcmp state machine now correctly resets when an interrupt or exception aborts the instruction.
* Atomic Commits: Interrupts are deferred during the committing sequence of Zcmp instructions (e.g., entire `cm.mv*` instructions, and the final sequence of `cm.pop*`).
* Debug Requests Deferred: Debug requests are entirely deferred until completion. To a debugger single-stepping, a multi-cycle Zcmp instruction appears as a single instruction.
* Performance Counters: Zcmp expanded instructions now correctly count as only a single instruction in the `instret` counter.

### Manual Verification Completed
These changes were validated manually using the following methodology:
* Generated a `riscv-dv` test with the following seed:
  ```bash
  make TEST=riscv_zcmp_directed_test SEED=12503
  ```
* Ran a TCL simulator script that sweeps a phase-shifting window of interrupts and debug requests across all Zcmp instructions. 
* Verified the core aborts when allowed, continues otherwise, atomically commits the final instructions, and completely flushes its state upon abort.
* Verified `instret` counting with a dedicated assembly program.

<details>
<summary>Click to expand: TCL Interrupt Injection Script used for manual verification</summary>

```tcl
# --- Configuration Parameters ---
set report_dir      "irq_external_i/master"    ;# Folder to save reports
set clk_period      20         ;# Clock period in ns (time added per iteration)
set half_clk        [expr {$clk_period / 2}] ;# Phase shift in ns
set post_force_time 1000   ;# Runtime after forcing the interrupt
set log_file        "out/run/tests/riscv_zcmp_directed_test.12503/trace_core_00000000.log"

# --- Define Events to Sweep ---
# Format: { {base_time_1 iterations_1} {base_time_2 iterations_2} ... }
    # # cm.mva01s
    # {45012  7}
    # # cm.popretz
    # {86292  25}
    # # cm.mvsa01
    # {88912  7}
    # # cm.pop
    # {91292  25}
    # # 2x cm.mv
    # {249492  10}
    # # cm.popret
    # {281652  25}
set sweep_events {
[...]

PR #2497, opened 2026-09-14
[dv] Fix directed-test configuration and scheduling

kulan-pal:
Three small fixes to the directed-test flow in `dv/uvm/core_ibex`. None changes which
tests run on the baseline; they are prerequisites for directed tests that need a
plusarg of their own (Zcmp and dummy-instruction tests in follow-up PRs, see #2461).

### `sim_opts` is silently dropped from directed-test entries

`scripts/run_rtl.py` reads `sim_opts` from the test options for both riscv-dv and
directed tests, but the pydantic schema for `directed_testlist.yaml`
(`scripts/directed_test_schema.py`) has no such field, so pydantic's default
`extra = "ignore"` discarded it. A directed test could not pass a plusarg through its
own entry. The field is added to `DConfig` as `Optional[str] = None`: existing
entries stay valid, and a config default can still be overridden per test.

### A test with N `rtl_params` entries is scheduled N times

In `scripts/ibex_cmd.py::filter_tests_by_config` the accepting `append` sat inside
the loop over `rtl_params`, so a test listing N parameters was appended N times:
`make` warned about duplicate targets and `regr.log` counted every seed N times.
Every existing entry lists one parameter, which hid it. The append moves to the
loop's `else:` clause, which runs once and only when no parameter caused a `break`.

### `directed_testlist.yaml` regenerates differently per locale

`directed_tests/gen_testlist.py` listed the vendored test directories with `ls`,
whose order follows the locale's collation, so the committed YAML could not be
checked against its generator. The listings now come from `sorted(os.listdir())`,
which is code-point order on every machine. The regenerated YAML is included.

### Verification

Baseline `34b07057`.

* Regeneration is byte-identical after this change. The 944 entries and their
  contents are unchanged; only the order of a few vendored tests differs. With the
  unmodified generator, running under the C locale produces a different file from the
  committed one.
* Python only: `filter_tests_by_config` on the `opentitan` config returns a test with
  two `rtl_params` entries twice before this change and once after; one-parameter,
  no-parameter and rejected tests behave the same before and after.
* End to end (VCS U-2023.03-SP2, Spike `ibex_cosim`, `IBEX_CONFIG=opentitan`): a
  directed test with two `rtl_params` entries and `sim_opts: +disable_cosim=1` is
  scheduled once and the plusarg appears on the simulator command line. Existing
  directed tests and `riscv_arithmetic_basic_test` pass unchanged.
* `util/lint_commits.py --no-merges` clean.

### Disclosure

Per CLA section 9: the analysis and code in this PR were produced with the
assistance of an AI coding tool (Claude Code) and independently reviewed with a
second tool (OpenAI Codex). I have reviewed and understood every change and take
full responsibility for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Rn6C7Eyq1LchFeP9w6up9P


kulan-pal:

PR #2502, opened 2026-09-14
[rtl,CHERIoT] Properly track outstanding memory transactions assertions

SamuelRiedel:
CHERIoT uses its own LSU request signals which were missing from some assertions tracking different LSU properties. This aligns the CHERIoT capability load/stores, which are two memory transactions, with the unaligned load/stores, which do the same. Simultaneously, we also fix a few edgecases, where the NoMemResponseWithoutPendingAccess could give false positives.


rtl/ibex_core.sv
```diff
@@ -1361,20 +1361,27 @@ module ibex_core import ibex_pkg::*; import ibex_cheriot_pkg::*; #(
...
+                                (id_stage_i.lsu_req_dec | id_stage_i.cheriot_lsu_req_dec) &
                                 ~id_stage_i.lsu_we;
-  assign outstanding_store_id = id_stage_i.instr_executing & id_stage_i.lsu_req_dec &
+  assign outstanding_store_id = id_stage_i.instr_executing &
+                                (id_stage_i.lsu_req_dec | id_stage_i.cheriot_lsu_req_dec) &
                                 id_stage_i.lsu_we;
 
+  // lsu_busy covers all non-IDLE states, but in WAIT_GNT the second grant hasn't been issued yet
+  // so no outstandng response can arrive there.
+  assign lsu_expecting_resp = lsu_busy & (load_store_unit_i.ls_fsm_cs != WAIT_GNT);
```
gautschimi, line 1375: `lsu_expecting_resp` is declared/assigned outside if (WritebackStage) but only ever read inside it. WritebackStage defaults to 0, so this looks like it'll be an unused-signal lint warning in that config unless CI only lints WritebackStage=1. Might want to move the declaration/assign inside the if (WritebackStage) block.

rtl/ibex_core.sv
```diff
@@ -1361,20 +1361,27 @@ module ibex_core import ibex_pkg::*; import ibex_cheriot_pkg::*; #(
...
+
+  assign outstanding_load_id  = id_stage_i.instr_executing &
+                                (id_stage_i.lsu_req_dec | id_stage_i.cheriot_lsu_req_dec) &
                                 ~id_stage_i.lsu_we;
-  assign outstanding_store_id = id_stage_i.instr_executing & id_stage_i.lsu_req_dec &
+  assign outstanding_store_id = id_stage_i.instr_executing &
+                                (id_stage_i.lsu_req_dec | id_stage_i.cheriot_lsu_req_dec) &
                                 id_stage_i.lsu_we;
 
+  // lsu_busy covers all non-IDLE states, but in WAIT_GNT the second grant hasn't been issued yet
```
gautschimi, line 1373: Not quite sure I understand this comment

rtl/ibex_core.sv
gautschimi, line 1375: also load_store_unit_i.ls_fsm_cs is a hierarchical assignment. please avoid. lets route this condition out of the LSU

SamuelRiedel: Good catch. I moved it inside the WritebackStage block.

rtl/ibex_core.sv
SamuelRiedel: I expanded the comment a bit. Essentially, we have to check the LSU's state machine to see whether it is expecting a response to come back. The `lsu_busy` comes close to be that signal, but there is one state which is an exception, which is when we still wait for a grant. Is is clearer now?

rtl/ibex_core.sv
SamuelRiedel: Fair point. I kept it that way, because it previously already used hierarchical signals and other assertions do the same. This code here is only used for assertions and if `INC_ASSERT` is defined. So I can easily route out a signal, but it will have a bit of overhead, because it will be an unused port whenever we don't have the WB Stage or don't compile assertions. What do you think given this? Should I still route it out or leave as is?

gautschimi: ok I missed this. If it is only for an assertion, it is fine

gautschimi (approved):

PR #2504, opened 2026-09-17
[rtl] Remove ztop related dead code

SamuelRiedel:
This was used to indicate the revocation memory location, which we implement differently


gautschimi (approved):
looks good

PR #2506, opened 2026-09-17
[dv] Include CHERIoT Load/Store/Jump signals in fcov assertions

SamuelRiedel:
Those fcov assertions were not properly updated when adding CHERIoT. Because those CHERIoT instructions are decoded with different control signals, we have to look out for both.


elliotb-lowrisc:
Just to check, does `InstrCategoryBranchCorrect` not also need updating anymore? It did at one point, but I could believe some implementation refinement has changed that and I can't see it firing when I run a test locally.

elliotb-lowrisc:
Actually, looking back at my original notes I can't see `InstrCategoryBranchCorrect` mentioned, so perhaps I made a copy-paste or assumption error. Apologies

elliotb-lowrisc (approved):
All three assertions changed here go from firing to not firing in my testing, so all good!

PR #2507, opened 2026-09-18
[nix] Fix python in nix

SamuelRiedel:


elliotb-lowrisc (approved):
I'm not sure how, but this seems to restore my ability to build the eda_shell

nix/pythonEnv/default.nix
```diff
@@ -17,18 +17,22 @@
...
           inputs.pyproject-build-systems.overlays.default
           (inputs.uv2nix_hammer_overrides.overrides pkgs)
+          uvLockedOverlay
+          (final: prev: {
+            # PyYAML 6.0 uses setuptools as its build backend without declaring
+            # it in its package metadata.
+            pyyaml = prev.pyyaml.overrideAttrs (old: {
+              nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ final.setuptools ];
+            });
+          })
```
hcallahan-lowrisc, line 35:
```suggestion
          uvLockedOverlay
          (inputs.uv2nix_hammer_overrides.overrides pkgs)

```

hcallahan-lowrisc: You can actually just do this (I got this the wrong way around when I originally wrote this code, but the uv2nix_hammer provides all the fixes we need, including the pyyaml one).

hcallahan-lowrisc (approved):
Thanks @SamuelRiedel, LGTM! One small suggestion to simplify slightly.

